PR #43, opened 2014-11-05
Iterator for Poller

Divius:
While switching EventLoop to iterator instead of indexing, I spotted a
potencial problem: io_process expects list of events not to change during it's
execution, but self is passed as &mut to the handler, making it possible (at
least in theory) to break it. After switch to an iterator, this issue is caught
by borrow checker actually. I switched the code to collect()ing all events into
a Vec as a workaround for now and looking forward to better ideas.

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -10,6 +10,11 @@ pub struct Poll {
     events: os::Events
 }
 
+pub struct EventsIterator<'a> {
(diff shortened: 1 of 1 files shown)


carllerche:
Sorry for the delay. I ended up merging this at abbe04d3f1d9d51cf49a27a7b958dc385c066d1a

PR #86, opened 2015-01-02
Publicize the io and os modules for downstream users.

### LINKED ISSUE #85: `os` should be public, so that other libraries using mio can implement new mio-compatible IoHandle's
The bindings and utility functions are valuable to those writing other extensions, like support for reading/writing from files.

reem:
The contents of these modules are useful to consumers of mio.

Closes #85


carllerche:
Closing since `IoDesc` is already exposed.

PR #87, opened 2015-01-02
Export TimerResult and TimerError, as requested in #61

### LINKED ISSUE #61: Some types should be exported in the public API
Some types are not exported in the public API of mio, but are used as return type of some functions of the public API.
They are visible in the generated documentation, but are not links.

So far I found:

- `EventLoopError`, used in `EventLoopResult`
- `TimerResult`, which is the return type of `EventLoop::timeout`
- the trait `IoHandle`, not sure about this one. It appears in various generics in the lib, but leaving it private effectively forbids using them with non-mio structures.

reem:
EventLoopError and EventLoopResult, as well as IoHandle, have all already
been made public.

Fixes #61

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -133,6 +133,8 @@ pub use event_loop::{
 };
 pub use timer::{
     Timeout,
+    TimerError,
+    TimerResult
 };
 pub use os::token::{
     Token,


carllerche:
Merged c3a4d21de2b252e1f59dbda9387502841cf18bf5

PR #88, opened 2015-01-03
Update to rust master: fixed array length fixes + lib stabilizations

vhbit:


carllerche:
Thanks for the PR. I ended up fixing this w/ the newer breaking changes as well 00198f1488dfcfd9a60abfe0cb0ce0b437b69d78

PR #95, opened 2015-01-05
Update to new memory ordering enum

awerush:
The memory ordering moved from std::sync::atomic into the 
std::sync::atomic::Ordering enum. This commit fixes the unresolve-issues.

rustc 0.13.0-nightly (ad9e75938 2015-01-05 00:26:28 +0000)

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -370,7 +370,8 @@ impl<H> EventLoopError<H> {
 mod tests {
     use std::str;
     use std::sync::Arc;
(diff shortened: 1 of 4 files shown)


carllerche:
Hey there. Thanks for the PR. I think this has already been handled by #94.

awerush:
Greetings. That #94 PR was my second try at not messing up with git/GitHub. ;-)
I closed it because of the unnecessary merge I created. This one does the same thing, it's just couple of commits smaller.

PR #98, opened 2015-01-06
Added getsockname for TcpSocket. 

aleksandrpak:
Also changed moving to borrowing in TcpSocket::getpeername.

I had problems building it with latest nix-rust:
`/Users/aleksandrpak/Projects/nix-rust/src/sched.rs:65:22: 65:23 error: obsolete syntax: ``|uint| -> bool`` closure type syntax
/Users/aleksandrpak/Projects/nix-rust/src/sched.rs:65 pub type CloneCb<'a> = ||:'a -> int;
note: use unboxed closures instead, no type annotation needed`

I have not found usage of this type so I just removed and it was OK for building mio but don't know about other projects. If you want I can send PR for this too.

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -191,9 +191,13 @@ pub mod tcp {
             try!(os::bind(&self.desc, addr));
             Ok(TcpListener { desc: self.desc })
         }
-        
-        pub fn getpeername(self) -> MioResult<SockAddr> {
+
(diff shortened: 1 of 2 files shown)


aleksandrpak:
nix-rust is already fixed so CI should pass now

carllerche:
Sorry for the delay. I pushed this here 022c9f6f1a46a725c2f3da6f3eca237b30ebfc35

PR #101, opened 2015-01-09
big update for rust nightly: uint/usize, fmt! changes

rrichardson:


carllerche:
I updated mio to the alpha but without doing the `int` -> `isize` change yet, so that is still up for grabs. 9a5c27f92c7dbf268ea1dad1fa0d68233c46918c

carllerche:
Handled here: 1773d6126bb67210167dc1baebccc23ff68519d7

PR #103, opened 2015-01-10
Make Slab<T> Send if T is.

### LINKED ISSUE #90: Eventloop is not longuer Send
Following last rust changes, EventLoop is not longer considered `Send` by the compiler. It effectively prevents initialising an EventLoop and then spawing a new thread to make it run.

The fix might be as simple as adding
```
unsafe impl<T: 'static, M: Send> Send for EventLoop<T, M> {}
```
but it can probably be handled better on the various subtypes of the library.

vberger:
Nothing prevents `Slab<T>` to be `Send` if `T` is `Send`, and it is the only obstacle to `EventLoop` being `Send`.

Closes #90 .

### DIFF
--- a/src/util/slab.rs
+++ b/src/util/slab.rs
@@ -26,6 +26,8 @@ const MAX: uint = int::MAX as uint;
 // When Entry.nxt is set to this, the entry is in use
 const IN_USE: int = -1;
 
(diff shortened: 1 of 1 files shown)


carllerche:
<3 thanks for the PR

PR #105, opened 2015-01-15
Refactoring notify

rrichardson:
I ran into some issues when I tried to push more than 10,000 msgs/sec out of an eventloop.   This scheme simplifies notification, removes the queue count and moves to a simpler boolean hint as to whether or not it needs to kick the awakener.  I have tested this approach with well over 30,000 messages per second with positive results. (26 microsecond round trip times in aggregate for messages)

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -1,5 +1,6 @@
 use std::default::Default;
 use std::time::duration::Duration;
+use std::sync::mpsc::SendError;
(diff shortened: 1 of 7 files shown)


carllerche:
Since the original issue that triggered this has been fixed, I'm going to close this. If there is anything else in the PR that should still be merged, could you open a new PR? :heart:

PR #109, opened 2015-01-20
Explicitly depend on bitflags, since it was moved.

reem:


carllerche:
This has been handled by #106

Cargo.toml
```diff
@@ -7,6 +7,7 @@ authors = ["Carl Lerche <me@carllerche.com>"]
 [dependencies]
 time = "*"
 log = "*"
+bitflags = "*"
```
fantix: Hmm, comparing to #106 , maybe this one should be used?

PR #111, opened 2015-01-24
Make ByteBuf Sync and Send

fhartwig:
Implement Sync and Send for ByteBuf. I believe this is safe, since `ByteBuf` is basically just a fancy `Vec<u8>` and has no internal mutability or similar problems.

### DIFF
--- a/src/buf/byte.rs
+++ b/src/buf/byte.rs
@@ -112,6 +112,9 @@ impl Writer for ByteBuf {
     }
 }
 
+unsafe impl Send for ByteBuf { }
+unsafe impl Sync for ByteBuf { }
+
 #[cfg(test)]
 mod test {
     use buf::*;


carllerche:
ByteBuf can be mutated. I don't think it can be `Sync`. Being `Send` is fine though.

fhartwig:
Unless I misunderstand the docs for [Sync](http://doc.rust-lang.org/std/marker/trait.Sync.html), mutability is not a problem, only interior mutability is (e.g. structs containing `RefCell`s). `ByteBuf` has simple inherited mutability. A `Vec` can be mutated, but it's still `Sync` (as long as its elements are `Sync`).

I don't actually need the `Sync` instance, though, so I can remove it if you prefer.

carllerche:
I'm trying to get it done, but I've been hitting issues installing the rust nightly.

carllerche:
I pulled in the `Send` here: 79b8c90508e8e64c14491b001ff61ec8091de88e

Thanks for the PR.

PR #112, opened 2015-01-28
Get mio to build with rust nightly

fhartwig:
The `io` module is now called `old_io`, the `Writer`'s `write` method has been renamed to `write_all`.


carllerche:
Thanks :heart: committed here: 977ec309fb5accbb069351301067840648c6ab9e

PR #113, opened 2015-02-01
Fix rustc warnings

fhartwig:
This is pretty trivial:
  - replace `fmt::Show` by `fmt::Debug`
  - use slicing syntax instead of `slice` and relatives
  - explicitly allow specific features instead of `#[allow(unstable)]`


carllerche:
Thanks for the PR :heart:

PR #124, opened 2015-02-21
Preliminary UDP v6 support

vhbit:


vhbit:
Note: requires https://github.com/carllerche/nix-rust/pull/81 to land before

carllerche:
Looking good. I am currently working on a large change to mio. See #127. I will have to think about how to apply this change.

carllerche:
So, for the first part of my work on mio reform, I think that I'm going to attempt to update mio (and nix) to use `std::io::{SocketAddr, IpAddr}` etc... Part of this involves moving code that should be in nix out of mio. I think the sock addr conversions etc.. should be moved to nix, and i started doing some of the work. Could you move the relevant ipv6 stuff to nix?

I will attempt to have mio master reflect the nix change tonight.

vhbit:
> Could you move the relevant ipv6 stuff to nix?

 Sure, tomorrow or day after tomorrow.

vhbit:
@carllerche ohh, it seems by the time I've got a chance to work on it - it's already builtin. Thanks a lot!

carllerche:
Yeah, I just pulled in in while I was in there :smile: I'm going to close this now.

PR #127, opened 2015-02-22
MIO Reform

carllerche:
**Currently a work in progress**

I am posting a PR here to hopefully get some feedback as the work happens.

## Goals

1. Update mio to track the latest changes in Rust std.
2. Get rid of `MioResult` and `MioError` in favor of `std::io::Result` and `std::io::Error`.
3. Get rid of the `IoAcceptor` trait
4. Remove re-implementations of the various `net` types that already exist in `std`.
5. Freshen up the APIs to match more recent Rust idioms.
6. Be friendlier to using blocking sockets with mio.
7. Remove "windows support" and any related abstractions
8. Misc cleanup

#### Rename `IoReader` to `TryRead` and `IoWriter`

Now that `std::io::Read` and `std::io::Write` are light wrappers over their respective sys calls, it should be fine to use them with mio, especially if we want to be friendlier to using blocking sockets as well. However, there still should be an API optimized for non-blocking operations. This will be handled by `TryRead` and `TryWrite`.

#### Get rid of `MioResult` and `MioError`

`std::io::Error` is now a wrapper around errnos, it should be possible to just use that vs. having a separate type in mio.

#### Get rid of `IoAcceptor`

Rust std got rid of it, and I don't think we need it either.

#### Remove re-implementations of the various `net` types that already exist in `std`.

Now that all of the various `std::net` types are light wrappers around system calls, it should be possible for mio to use them directly. mio will be able to implement `TryRead` and `TryWrite` for them as well, making them friendlier for the non-blocking case.

Mio will keep `TcpSocket` (and the other raw socket types) around to enable more advanced socket setup / configuration, but the socket will need to be converted to either a `TcpListener` or a `TcpStream` before accepting, reading, or writing.

The `Socket` trait will remain to allow configuring socket options.

#### Be friendlier to using blocking sockets with mio.

Polling isn't just for non-blocking sockets (though it probably is the majority case). I do want to make sure there are no assumptions that all sockets are non-blocking. `TcpSocket::v4()` currently returns a non-blocking socket w/ CLOEXEC set. I would propose keeping this, but also adding `TcpSocket::v4_opt(SockOpt::nonblocking() | SockOpt::cloexec())` or something like that. `TcpSocket::v4_opt(...)` would not set any flags by default.

Now, there should probably be more functions on `TcpSocket` as well, for example: `TcpSocket::connect(...) -> TcpStream` would create a non-blocking socket by default.

#### Remove "windows support" and any related abstractions

Originally, I was thinking that mio would support windows as well. However, I am starting to lean against this. I am currently leaning to have mio be for systems that support the readiness model. Windows supports the completion model (via IOCP). I'm thinking there should be a higher level IO library that supports both windows & posix systems.

As such, we can get rid of `IoDesc` in favor of just `Fd`. I also would like to rename `IoHandle` to `Listenable`.

I am interested in feedback. cc @wycats  @rrichardson @reem @vhbit @rozaliev @fhartwig @vberger @vbuslov


carllerche:
Also, the current state of the PR is an initial deep dive. I may start over and try to do it more incrementally now that I have a better sense of the dependencies.

tupshin:
" I'm thinking there should be a higher level IO library that supports both windows & posix systems."

This is exactly right. Let the lower levels expose their platform distinctions, which will temporarily reduce portability of code. But then people will build multiple(and that's a good thing)  abstractions on top that unify the two.

ayosec:
> I would propose keeping this, but also adding `TcpSocket::v4_opt(SockOpt::nonblocking() | SockOpt::cloexec())` or something like that. `TcpSocket::v4_opt(...)` would not set any flags by default.

Another option is to add methods on the `TcpSocket` to configure the socket.

```rust
let socket = TcpSocket::v4().nonblocking().cloexec();
```

See [pseudo/untested implementation](https://gist.github.com/ayosec/0125b27fd68df7f5d075).

carllerche:
@ayosec I initially thought of adding a builder style API, however there needs to be a way to atomically create the socket and set the flags. So, either there is an explicit socket builder type that has a "build" call, or there would need to be fns that can create the sockets w/ the flags in one go.

dcsommer:
+1 for these changes. I'd vote towards keeping the APIs streamlined/defaulted for non-blocking, since that is the niche mio is filling nicely right now.

vberger:
A big +1 for the unification with std types and traits.

Other than that, as a mere user not knowing a lot about epoll and others... I can't really say more. It overall looks really nice.

bjadamson:
+1 on this!! Personal opinion, builder APIs with an explicit build() method are really great for separating mutation and object construction from object use

vhbit:
LGTM.

I can see a pain point with multicast sockets on ipv6 (std lib has the same issue), but that's definitely should be addressed separately.

teotwaki:
Starting to work with Mio, one aspect that slightly bothers me is having to do something like the following...

```rust
extern crate mio;

use self::mio::*;
use self::mio::net::udp::*;
use self::mio::buf::{RingBuf, SliceBuf};
```

... without really knowing what's going on. All of a sudden, the namespace is cluttered with a bunch of fairly nondescript (`Token` is hardly going to be unique). I may be biased because even in C++ I usually never use things like `using namespace std` or what have you, and in Python I never use `from foo import *`, so this could be chalked up to personal preference, but as @carllerche offered on IRC:

> teotwaki | I have to admit, I'm not a massive fan of use foo::*;
> ~carllerche | teotwaki: even if it's foo::prelude::*?

Obviously, using `foo::prelude::*` wouldn't give access to everything inside `udp` or `tcp` or whatever new APIs are exposed, but it should provide enough traits/structs with names that have little chance of collision with user-code to allow use of specific items in the rest of the Mio API.

vbuslov:
+1 for unification with Rust std types.
Dropping potential Windows support is definitely necessary to keep implementation simple and fast.

reem:
After reading the whole changeset, I have some thoughts.

I am particularly averse to the disappearance of the `IoDesc` type, whose logic I'd much rather stayed in one place rather than being copied to every type which holds an `Fd`. It was already possible to get an `Fd` out of an `IoDesc`, so I don't really see a need to drop the type entirely.

I also dislike `Listenable`, since it contains the `able` suffix which is traditionally undesirable for traits and we already have an io action `listen` that this could be easily confused with. I'd much rather `Io` or another simple name, which simply indicates that this thing is an io object and can produce an `Fd` or `IoDesc`, and as a result can be registered on the event loop.

The elimination of windows support as a goal is unfortunate in the short term (as there is no replacement), but probably an unavoidable step for the project and healthy for long term goals. IOCP/Windows support can go in another library and a higher-level IO lib can select between the two.

I'm a big fan of most other cuts and changes, like renaming to `TryRead` and `TryWrite`, dropping `MioError`, `MioResult`, and `IoAcceptor` as well as many of the io types, and all the other stuff.

ayosec:
> IOCP/Windows support can go in another library

There is a [rust-iocp create](https://github.com/cyderize/rust-iocp). I guess that it still needs a lot of work.

carllerche:
@reem re: naming, I am open to alternatives.

reem:
Ideas for the old `IoHandle`: `Io`, `IoDesc`, `Desc`, `AsDesc`, `AsIoDesc,` `AsFd`, `AsRawFd`, `IoHandle`

carllerche:
Also, another big reason for getting rid of it was the unification with std types. Mio will do away with a significant chunk of net types. For the others, it makes sense to me to keep them similar to impls in std.

carllerche:
I am no a huge fan of Io as well since not all types that can be registered with the Mio event loop represent IO. There could be signal abstractions, timers, notifications of changed to shared mem, etc...

Another option I was considering was Evented

reem:
I'm actually surprised there is no `IoDesc`-like abstraction used in the implementation of `std::{io, net}`.

`Evented` is pretty good. I have a trait in rust-aio called `Evented` as well, which is basically a rolling combo of `Handler` and `IoHandle` in the old/current mio.

carllerche:
These goals are mostly completed by https://github.com/carllerche/mio/commit/06bf140e6928a4505df1520c0805905909f7b68a (and prior commits). However, there are still changes pending on figuring out #131

PR #129, opened 2015-02-28
Fix compilation on Linux.

vberger:
It was apparently forgotten in 147e3d3ac19ebf303042952df630c5a41abb9d95.


carllerche:
Thanks. I did indeed mess that up

PR #130, opened 2015-03-02
Support broadcast on UDP sockets

vhbit:
Requires https://github.com/carllerche/nix-rust/pull/86 to land.

### DIFF
--- a/src/net/mod.rs
+++ b/src/net/mod.rs
@@ -117,6 +117,14 @@ pub trait UnconnectedSocket {
     fn recv_from<B: MutBuf>(&mut self, buf: &mut B) -> MioResult<NonBlock<SocketAddr>>;
 }
 
+pub trait BroadcastSocket: Socket {
+    fn set_broadcast(&self, val: bool) -> MioResult<()> {
+        nix::setsockopt(self.desc().fd, nix::SockLevel::Socket, nix::sockopt::Broadcast, val)
+            .map_err(MioError::from_nix_error)
+    }
+}
+
+
 /*
  *
  * ===== Implementation =====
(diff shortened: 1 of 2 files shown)


carllerche:
I'm actually in the process of getting rid of mio's UDP socket impl in favor of the one in std (which already has broadcast), so this won't be needed anymore.

carllerche:
Ok, I pushed UDP support update. It now uses std::net types: https://github.com/carllerche/mio/commit/ec77c8fcc84d7c2fc3bb9b4a5e9bbc03badf732a

I'm definitely interested in feedback though since I personally haven't used UDP all that much. Closing this for now, but as always feel free to open new tickets.

vhbit:
I've tried to migrate, but was immediately blocked by missing support of `set_reuse_port` and `set_reuse_addr` in `std::net` udp socket, so for now I have no feedback either as have to stick with old version.

rrichardson:
That seems like pretty good feedback.

carllerche:
@vhbit You should still be able to access these functions. They are implemented on `Socket` which is implemented on `std::net::UdpSocket` (https://github.com/carllerche/mio/blob/master/src/net/udp.rs#L20)

Does that work for you?

vhbit:
The problem is that should be used before bind, which is impossible in
std::net as it combines socket creation and binding.

carllerche:
Ok, I should have a fix today

carllerche:
This should be handled in 29c5e32fe96e70b762e65fe7d59a605d367d9597

`udp::v4() -> NonBlock<UdpSocket>`

vhbit:
@carllerche thanks!

vhbit:
btw, have you discussed with the core team a way to create `std::net` sockets from existing fd without transmuting?

carllerche:
@vhbit I have, it should be coming pre 1.0

PR #133, opened 2015-03-03
Handler callback for when EINTR occurs from polling

jbschlosser:
Currently, an EINTR occurrence when polling will cause the event loop to exit. This fix handles EINTR in a more elegant way- calling a handler callback and continuing the event loop.


carllerche:
Looks good to me. I will try to get it merged today. @rrichardson any feedback?

rrichardson:
I think calling an interrupted handler is overkill, but it ultimately does
the right thing, so I'm OK with it.

carllerche:
Thanks. Merged b5501184b533f92d39fca165593a3220547e6127

PR #134, opened 2015-03-03
Add iterators for Slab

fhartwig:
I need the `iter_mut` method on `Slab` in one of my projects.
Possible issues:
- Iteration may need to look at every `Entry` in the slab (i,e. needs `Slab.cap` instead of `Slab.len` operations) in certain cases. I don't think this is a big problem.
- There is some `unsafe` code in the `next` method of `IterMut` (which does the same thing as `mut_entry`, but `mut_entry`'s result has the wrong lifetime). Using unsafe code for Iterators returning `&mut`s apparently can't be avoided, though.

### DIFF
--- a/src/util/slab.rs
+++ b/src/util/slab.rs
@@ -12,7 +12,7 @@ pub struct Slab<T> {
     len: isize,
     // The total number of elements that the slab can hold
     cap: isize,
-    // THe token offset
+    // The token offset
     off: usize,
     // Offset of the next available slot in the slab. Set to the slab's
     // capacity when the slab is full.
@@ -191,6 +191,23 @@ impl<T> Slab<T> {
         }
     }
 
+    pub fn iter(&self) -> SlabIter<T> {
+        SlabIter {
+            slab: self,
+            cur_idx: 0,
+            yielded: 0
+        }
+    }
+
+    pub fn iter_mut(&mut self) -> SlabMutIter<T> {
+        SlabMutIter {
+            slab: self,
+            cur_idx: 0,
+            yielded: 0
+        }
+    }
+
+
     #[inline]
     fn entry(&self, idx: isize) -> &Entry<T> {
         unsafe { &*self.mem.offset(idx) }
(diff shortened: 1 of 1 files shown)


carllerche:
Could you tell me more about why you need to iterate a slab? It wasn't a case that I originally considered, so I would like to know more.

fhartwig:
@carllerche Sure. I'm writing a bittorrent client, and I'm keeping a slab of connected peers (i.e. a TcpStream + some associated state). I sometimes need to send a message to all connected peers, so I need a way to iterate over them.

carllerche:
Sorry for letting this sit. I was mid mio reform and wanted to get through that. I _think_ this is fine, but I'm going to want to look through it again when I am fresh tomorrow morning. I'm quite tired now.

fhartwig:
Not a problem, makes perfect sense. If you decide that this isn't a good idea, I can always work around it in my application (by keeping a vector of active tokens or something).

fhartwig:
Just in case you have already pulled this branch: I just rebased it on the current master (and force-pushed).

src/util/slab.rs
```diff
@@ -320,6 +337,51 @@ impl<T> Entry<T> {
...
+    slab: &'a Slab<T>,
+    cur_idx: isize,
+    yielded: isize
+}
+
+impl <'a, T> Iterator for SlabIter<'a, T> {
+    type Item = &'a T;
+
+    fn next(&mut self) -> Option<&'a T> {
+        while self.cur_idx < self.slab.cap && self.yielded < self.slab.len {
```
carllerche: The first check is redundant no? As long as less items have been yielded than are contained in the slab, iteration should continue?

carllerche:
Just one bit of feedback now that I am fresh, besides that it is good to merge!

src/util/slab.rs
fhartwig: Yeah, you're right. I originally just had the first condition and then added the second as an optimisation. I'll fix this soon.

fhartwig:
Fixed :-)

carllerche:
Thanks!

PR #138, opened 2015-03-06
Allow error events to be handled in Handlers

CraZySacX:
Currently error events just spit out the error message.  This allows the handler to react to error events outside of readable and writable.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -310,7 +310,7 @@ impl<H: Handler> EventLoop<H> {
         }
 
         if evt.is_error() {
-            println!(" + ERROR");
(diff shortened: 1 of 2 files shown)


carllerche:
Ah yes... I completely forgot about this. I've been meaning to fix it :P

I'm wondering if it would be better to pass errors to the handler via `readable` and set the `ReadHint` -> error. The issue is that, I do not believe that every platform reliably detects errors without reading, so this may be a good case of just considering an error a "readable" event.

CraZySacX:
Yeah that sounds better.  I'll close this pull request and submit another.

PR #139, opened 2015-03-09
Error events are sent to readable in handler along with error in the readhint

CraZySacX:
Updated to send error event plus readhint to the readable method in handlers.  This should allow for the handlers to properly respond to error events.


carllerche:
Thanks! Merged: https://github.com/carllerche/mio/commit/4c724d7b8f4b6778145a5bf0d6cfac1c06f1798a

PR #140, opened 2015-03-21
Fix error type for latest rust nightly

fhartwig:
I'm guessing that this is the intended error type now. Apologies if I guessed wrong :-)


carllerche:
I ended up pulling #141 first. Thanks for the PR though!

PR #142, opened 2015-03-29
Fix compilation on Linux.

vberger:
These were apparently forgotten in last update to follow rust evolution.


carllerche:
GAH I failed! Thanks for the PR.

PR #150, opened 2015-04-12
Implement IntoInterator instances for Slab

fhartwig:
Since we have `IntoIterator` now, why not use it?
I'm not sure if these should replace the `iter` and `iter_mut` methods. What do you think? The `std::collections` types still have explicit `iter` methods as well as `IntoIterator` implementations, but that may just be because they wanted to avoid breaking too much existing code.


carllerche:
It seems fine to leave them. I'll merge now, thanks.

PR #153, opened 2015-04-19
Rustup

fhartwig:
This mostly fixes the build on the latest nightly, and it also fixes a lifetime bug in `SlabMutIter`'s `Iterator` instance.
For this to build, the nix package still needs to be fixed and the version bound bumped.


carllerche:
Thanks for the PR. I already pulled some of the commits (and tweaked another) before seeing this PR. Master should be working.

PR #156, opened 2015-04-21
Fix bug in Slab::get

fhartwig:
`Slab::get` currently panics when called with a non-existent token. I'm assuming it should return `None` instead. I've also added some basic tests for `get` and `get_mut`.


carllerche:
Merged thanks! fad8ffdfe551dfa983b9d6f73322ce122bf951b8

PR #167, opened 2015-05-14
Implement Read+Write under TryRead+TryWrite for Io types where applicable, and rename TryRead+TryWrite methods

jnicholls:
…, TcpStream and UnixStream.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -412,7 +412,7 @@ mod tests {
         let wcount = Arc::new(AtomicIsize::new(0));
         let mut handler = Funtimes::new(rcount.clone(), wcount.clone());
 
-        writer.write(&mut buf::SliceBuf::wrap("hello".as_bytes())).unwrap();
+        writer.try_write_buf(&mut buf::SliceBuf::wrap("hello".as_bytes())).unwrap();
         event_loop.register(&reader, Token(10)).unwrap();
 
         let _ = event_loop.run_once(&mut handler);
@@ -420,7 +420,7 @@ mod tests {
 
         assert_eq!((*rcount).load(SeqCst), 1);
 
-        reader.read(&mut b).unwrap();
+        reader.try_read_buf(&mut b).unwrap();
 
         assert_eq!(str::from_utf8(b.flip().bytes()).unwrap(), "hello");
     }
--- a/src/io.rs
+++ b/src/io.rs
@@ -17,13 +17,13 @@ pub trait Evented {
 }
 
 pub trait TryRead {
-    fn read<B: MutBuf>(&mut self, buf: &mut B) -> Result<Option<usize>> {
+    fn try_read_buf<B: MutBuf>(&mut self, buf: &mut B) -> Result<Option<usize>> {
         // Reads the length of the slice supplied by buf.mut_bytes into the buffer
         // This is not guaranteed to consume an entire datagram or segment.
         // If your protocol is msg based (instead of continuous stream) you should
         // ensure that your buffer is large enough to hold an entire segment (1532 bytes if not jumbo
         // frames)
-        let res = self.read_slice(buf.mut_bytes());
+        let res = self.try_read(buf.mut_bytes());
 
         if let Ok(Some(cnt)) = res {
             buf.advance(cnt);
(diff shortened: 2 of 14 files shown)


carllerche:
How is the ambiguity between `TryRead::read` and `Read::read` handled?

rrichardson:
This is a fairly minor nit, but I think it would be more general and extensible (and less code) to impl  TryRead for all socket type things like Io and UnixStreams, and impl Read for all implementors of TryRead. ( T : TryRead).  
This way it is less work for anyone who creates a new type (in the Mio create or otherwise) and wishes to make it (try)readable. It also makes more sense to me from a theoretical standpoint, since impls of traits move from the specific to the general,  and TryRead is more specific than Read.

jnicholls:
Funny enough I was considering doing this on Friday, except the opposite: `impl<T: Read> TryRead for T` and `impl<T: Write> TryWrite for T`.  I forgot to code that up.

I agree with your sentiment about TryRead -> Read, however pragmatically it means that we are taking the general result (Read), wrapping it into a specialized result (TryRead) and then unwrapping it again to impl the general trait (Read).  This seems clunky to me.  Read naturally comes before TryRead from an implementation standpoint: you take the result from Read/Write, check it, and create the TryRead/TryWrite result out of it.

rrichardson:
It does seem weird,  I guess it comes down to a design choice.   I we wanted mio to speak only with nonblocking sockets/fds, then I would say that TryRead would be the fundamental type.  However, since epoll and friends work just fine with blocking sockets as well, I guess this may not be the case, so having read/try_read as siblings makes a bit of sense.

jnicholls:
I guess if you did not want to enable external Read/Write implementations to take advantage of TryRead/TryWrite APIs in Mio, then your suggestion of Try -> Non-Try does make the most sense.  But if you did want blocking versions to automatically fit into non-blocking APIs, then Non-Try -> Try makes more sense.  I'm not sure what the future of Mio holds, but that's the consideration.

And you beat me to it, haha.  Spot on.

jnicholls:
With the impending Windows IOCP support, I think your suggestion of TryRead -> Read is the safest thing to do for now.  It can always be changed to open the doors to blocking types, rather than closing the doors on those types and potentially break code.

jnicholls:
As I am going through TryRead -> Read, I lose context of the underlying OS error code during the unwrapping of the TryRead result.  I'm adding a FIXME comment to reinstate that error code for unix and windows, when Windows support rolls in.  It's not necessarily critical, but it is a small loss-in-translation.

jnicholls:
Ah man, since Read/Write are not defined in the mio crate I can't implement them generically on type T even though type T must implement a trait defined in the mio crate (TryRead/TryWrite).  Bummer.

jnicholls:
I've updated the PR to implement TryRead/TryWrite on top of Read/Write, and implement only Read/Write for the Mio types.  This goes against what we said earlier, but check it out and consider it.  It's definitely clean.

I might even retract my previous statement about Mio APIs.  I imagine you guys will utilize the Evented trait to scope the types that can pass into APIs more or less.

rrichardson:
Arg!  I forgot about that..  Best laid plans I reckon.

jnicholls:
I rebased these changes on top of the code reorganization.  Please review, thanks.

carllerche:
Hey, this looks good. I have one minor thought.

What if `try_read` was renamed -> `try_read_buf` and `try_read_slice` -> `try_read`. I'm wondering if that would be better as it would more closely mirror the `Read` trait. (Same for `Write`).

Of course, I could merge this PR and do this change as a separate change.

jnicholls:
I agree, that makes a lot of sense.  I'll modify the commit accordingly.

On Thursday, May 21, 2015, Carl Lerche <notifications@github.com> wrote:

> Hey, this looks good. I have one minor thought.
>
> What if try_read was renamed -> try_read_buf and try_read_slice ->
> try_read. I'm wondering if that would be better as it would more closely
> mirror the Read trait. (Same for Write).
>
> Of course, I could merge this PR and do this change as a separate change.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/carllerche/mio/pull/167#issuecomment-104453881>.
>


-- 
Sent from Gmail Mobile

jnicholls:
Commit has been updated, thanks for the feedback.

jnicholls:
Bump

carllerche:
There was no internet [where I was](http://en.wikipedia.org/wiki/Crater_Lake) over the long weekend. Going to look at this now.

jnicholls:
Thanks Carl, hope you had a great weekend!

I noticed you put in the commit directly rather than merge the PR.  You might want to close the PRs afterwards just to keep things tidy.

PR #171, opened 2015-05-17
Remove the from_nix_error() hack

jnicholls:
Utilize the Error::from_raw_os_error() method on std::io::Error.  This is compatible with Rust 1.0.0+.


carllerche:
Hey, I had this pending on a branch that I should have merged. It is currently on master:

https://github.com/carllerche/mio/blob/master/src/sys/unix/mod.rs#L38-L40

Thanks for the PR though :heart:

PR #177, opened 2015-05-21
Add shutdown function

hoxnox:
Provided by nix::socket::shutdown.

### DIFF
--- a/src/net/tcp.rs
+++ b/src/net/tcp.rs
@@ -92,7 +92,15 @@ pub struct TcpStream {
     sys: sys::TcpSocket,
 }
 
-pub use sys::Shutdown;
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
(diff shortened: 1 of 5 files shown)


carllerche:
Looks good. One minor nit, I would redefine the `Shutdown` struct in mio. I think it would go best in [net/mod.rs](https://github.com/carllerche/mio/blob/master/src/net/mod.rs). Generated docs will be better and one day, there will be windows support which won't have access to nix :smile:

hoxnox:
Yah, i thought about it. It would be better to define Shutdown in mio.

carllerche:
Cool, I'll keep an eye open for the change :+1:

hoxnox:
I redefined Shutdown in net::tcp. Not in net because TcpStream, which defines shutdown function, is in net::tcp.

carllerche:
Good call! Looks good to me.

PR #180, opened 2015-05-26
Use trace level logging for logging in tick

thijsc:
The mio debug logging output is very verbose, which makes it hard to have a readable debug log a program that's using mio. I've therefore moved the most often occurring debug log calls to the trace level.

This change should allow somebody working on mio to have very detailed logging output, while not making a host processes's debug logs hard to read. Let me know what you think.


carllerche:
Good idea, thanks. Merged: 285b1fb5b53296fa7a7e9cd192c3442f61495654

PR #181, opened 2015-05-27
Change 'register' functions to use Evented Trait Object

rrichardson:
When attempting to make an event-loop wrapper which can conveniently manage multiple protocol handlers, the one problem I ran into with mio is being able to provide Sized Evented objects for consumption.  In some cases I was able to know the exact Evented impl and Any::downcast down to the type, in other cases I was not so lucky.  

I have a branch which simply changes deregister/register/register_opt in EventLoop and Poll to take an &Evented rather than  E : Evented.  This requires no code change for customers as a ref will already coerce. All tests pass.  

Also,  I have added a Slab::replace which works like std::mem::replace. Very handy for moving values in and out of a Slab without giving up a token.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -180,17 +180,17 @@ impl<H: Handler> EventLoop<H> {
     }
 
     /// Registers an IO handle with the event loop.
-    pub fn register<E: Evented>(&mut self, io: &E, token: Token) -> io::Result<()> {
+    pub fn register(&mut self, io: &Evented, token: Token) -> io::Result<()> {
         self.poll.register(io, token, Interest::all(), PollOpt::level())
     }
 
     /// Registers an IO handle with the event loop.
-    pub fn register_opt<E: Evented>(&mut self, io: &E, token: Token, interest: Interest, opt: PollOpt) -> io::Result<()> {
+    pub fn register_opt(&mut self, io: &Evented, token: Token, interest: Interest, opt: PollOpt) -> io::Result<()> {
         self.poll.register(io, token, interest, opt)
     }
 
     /// Re-Registers an IO handle with the event loop.
-    pub fn reregister<E: Evented>(&mut self, io: &E, token: Token, interest: Interest, opt: PollOpt) -> io::Result<()> {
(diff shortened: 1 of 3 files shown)


dwrensha:
Would it work instead to `?Sized` bound, as I do here: https://github.com/dwrensha/mio/commit/2f4c2d42d878958e3d2aea751b0fc5328d4b83ff? (I've sat on this change because I ended up not needing it.)

rrichardson:
For reasons unbeknownst to me,  that change forces me to annotate every invocation of a register function with ::<TcpStream> to get the tests to pass, even when I clearly annotate the type of the socket in its instantiation.    I really have no idea why that would happen, ?Sized does add ambiguity, but only in allowing unsized items, it doesn't allow a broader range of concrete types.

rrichardson:
e.g. 
```rust
test/test_battery.rs:240:16: 240:103 error: type annotations required: cannot resolve `_ : mio::io::Evented` [E0283]
test/test_battery.rs:240     event_loop.register_opt(&sock, CLIENT, Interest::writable(), PollOpt::edge() | PollOpt::oneshot()).unwrap();
```

carllerche:
hmm... @dwrensha's patch _should_ work... @rrichardson do you have full code somewhere that shows the problem?

@dwrensha I would take your diff as PR :smile:

carllerche:
@rrichardson I don't think that switching to trait objects here is going to be a great idea. I'd rather figure out why you are having problems getting trait objects working in the first place...

I would like the `replace` fn on `Slab` though. A separate PR would be appreciated!

rrichardson:
Try @dwrensha's diff. The tests won't compile. The compiler needs
additional type annotation for register for the exact socket type.

This theoretically *should* work I think.
It might be a bug in rustc, not sure.

My issue is that, in order to manage arbitrary protocols with arbitrary
connection types, I am storing contexts as trait objects. There isn't
another way i can think of to do it.

I can fathom situations in which the consumer of the API may not have the
type information handy to downcast the context in order to get a statically
sized object in order to call register.

rrichardson:
I just tested @dwrensha's branch from #182 . It works.  I'll close this and make a new PR for Slab::replace

rrichardson:
added #183 based on the #182 branch

PR #183, opened 2015-05-27
Added 'replace' to Slab which works like mem::replace

rrichardson:


carllerche:
Committed: 7e7a561dcfa09572696c37fc9db969aec5d08bd2

Issue #178, opened 2015-05-24
Unintuitive behaviour when sending notifications to a dropped EventLoop


Vaelden:
Right now, when a notification is sent to an EventLoop that has been dropped (explicitly or its thread panicked), no error is returned and the notification stays "stuck" in the queue, and is dropped only when the last associated Sender is dropped.
I find that a little bit unintuitive (completely different from channels in the standard library) it and makes difficult to recover from a panic in an EventLoop.

So I would suggest two ideas to improve that:

* Introduce a new special state in Notify (e.g. `DROPPED`) that would be atomically set when the EventLoop is dropped. When sending a message, Senders would check the state (as usual) but return an error containing the message if the event loop was dropped (like standard channels).
* For the drop strategy, I would suggest one of the following strategies:
  * When panicking / dropping and after setting the `DROPPED` state, the event loop would grab all remaining messages and drop them. This would ensure that they don't stay in an "unreachable" state without being dropped for too long (e.g. if a Sender is still in scope somewhere)
  * Keep the current plan of the last Sender is responsible for dropping everything, but allow a Sender to retrieve the queued messages if the EventLoop has panicked (might allows to recover without losing too many messages)
  *  Keep it as is

Do you have any views on the feasibility of this?

carllerche:
Thanks for opening this issue. What you are describing is something that had slipped my mind. Your description of the proposed fix is spot on. For the drop strategy, I would opt for the first option. So, to summarize, the plan would be:

> Introduce a new special state in Notify (e.g. `DROPPED`) that would be atomically set when the EventLoop is dropped. When sending a message, Senders would check the state (as usual) but return an error containing the message if the event loop was dropped (like standard channels).

> When panicking / dropping and after setting the DROPPED state, the event loop would grab all remaining messages and drop them. This would ensure that they don't stay in an "unreachable" state without being dropped for too long (e.g. if a Sender is still in scope somewhere)

I would happily accept a PR that implements this :smile: (hopefully with tests!)

carllerche:
This is resolved by #189

PR #189, opened 2015-05-30
Drop notifications

Vaelden:
Here is the fix for #178

I have made a couple of implementation choices:

* Named the new state `CLOSED` instead of `DROPPED`, as potentially you could want to expose the `close()` function
* As currently the `Notify` structure is only used in mpsc mode and `close()` is not exposed, I do not check for the `CLOSED` state when reading the channel in `check()`, and only inserted a `debug_assert!()`
* To ensure that all messages are dropped from the queue, a `send()` will in very special cases have to drop a notification. That could potentially be a problem when the drop is expensive, and with panicking destructors (which are already a bad idea in Rust)

Potential enhancements:

* Expose the `close()` function if you think it could be useful. This will require to handle the `CLOSED` case in the `check()` function
    * You could then think about a `reopen()` function, but it could be tricky to implement, and I am not sure there would be use cases

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -335,6 +335,12 @@ impl<H: Handler> EventLoop<H> {
 
 unsafe impl<H: Handler> Sync for EventLoop<H> { }
 
+impl <H: Handler> Drop for EventLoop<H> {
+    fn drop(&mut self) {
+        self.notify.close();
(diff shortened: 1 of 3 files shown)


carllerche:
Looks good to me! Thanks!

PR #190, opened 2015-06-03
slab: Add `Slab::create_and_insert`.

dpc:
Hi,

I find current `insert` limiting, so I come up with this. Comments?

### DIFF
--- a/src/util/slab.rs
+++ b/src/util/slab.rs
@@ -109,7 +109,7 @@ impl<T> Slab<T> {
         if idx == self.entries.capacity() {
             Err(val)
         } else {
-            match self.insert_with(move |_| { val }) {
+            match self.insert_with(move |_| val ) {
                 None => panic!("Slab::insert_with() should"),
                 Some(token) => Ok(token)
             }
--- a/src/util/slab.rs
+++ b/src/util/slab.rs
@@ -129,6 +129,38 @@ impl<T> Slab<T> {
         Ok(self.idx_to_token(idx))
     }
 
+    /// Like `insert` but for objects that require newly allocated
+    /// Token in their constructor.
+    pub fn create_and_insert<F>(&mut self, f : F) -> Option<Token>
+    where F : FnOnce(Token) -> T {
+        let idx = self.nxt;
+
+        if idx == self.entries.len() {
+            // Using an uninitialized entry
+            if idx == self.entries.capacity() {
+                // No more capacity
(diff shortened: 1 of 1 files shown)


carllerche:
Looks good. I agree, I have hit cases where this would be useful.

Minor nitpicks. What about naming this `insert_with` and refactoring `insert()` to use this function?

dpc:
OK. If this is mergeable idea than I will work on re-factoring.

carllerche:
I think it is. Just to make sure that I understood it, the function reserves a token, then calls a builder function w/ that token allowing creating the value w/ the token in one go (vs. having to store Option<Token>).

However, there is one thing to consider: How to handle panics. Aka, what happens if the builder function panics? I believe, in this case, the token will leak. Is that correct?

dpc:
The token is passed to builder before anything is done with the slab, so in case of panic in the builder the token should not leak. Maybe I should write a test for this.

I've refactored insert, it's a bit tricker than ideal, because I can't see a way to both pass `val` to the builder function and potentially pass in `Err`. I think optimizer should have no problem with removing the additional check.

dpc:
BTW. The slab allocator seems like something that could be usable as a small separate library.

src/util/slab.rs
```diff
@@ -104,15 +104,33 @@ impl<T> Slab<T> {
 
     pub fn insert(&mut self, val: T) -> Result<Token, T> {
         let idx = self.nxt;
+        // check fail condition before val gets moved by insert_with,
+        // so `Err(val)` can be returned
+        if idx == self.entries.capacity() {
+            Err(val)
+        } else {
+            match self.insert_with(move |_| { val }) {
```
carllerche: You shouldn't need the brackets here.

carllerche:
Looking good. The main reason that this isn't a separate lib is that it is bound to `Token` (and so is the rest of mio).

It may still be a good idea to split it out though.

dpc:
I don't know how to test leaks on panic. I need to use threads for it, and then if I need to wrap shared slab in `Mutex` to make it `Sync`, which will leads to problem if builder panics when lock is held anyway. In mio the handler is working in a single thread anyway, right?

carllerche:
Thanks! I merged this here: https://github.com/carllerche/slab/commit/ac96f1a8ab99af4e1a2b3eea0ce86371806da78e

Issue #184, opened 2015-05-28
Confusing behavior when multiple events for a single socket are fired in a single event loop iteration


carllerche:
Issue representing the confusion experienced in #172 and #179.

My current understanding of the confusion is, given epoll & a socket that received a `HUP` and is writable, mio will first call the `Handler::readable` w/ `ReadHint` set to `Hup` followed by the `Handler::writable` event. The confusion arises when the socket is closed during the `Handler::readable` call, or perhaps the token associated with that token is reused. The next call to `Handler::writable` for the "finalized" socket is unexpected.

Part of the original API design consideration is how to unify epoll w/ kqueue. With kqueue, the readable & writable notifications are provided separately. As far as I am aware, there is no guarantee as to which order (readable or writable) the events will be seen.

I don't believe that either solutions described by @rrichardson in #179 would be possible to implement (efficiently) for kqueue. The reason why hup / error are provided as a `ReadHint` is that it is not 100% portable for mio to detect them. However, the `hup` / `error` state can be discovered by the mio by reading from the socket, which is why they are considered `readable` events. In other words, to correctly use mio, hup / error states **must** be discovered by listening for `Handler::readable`, reading from the socket, and seeing an error. As such, doing anything with the `ReadHint` argument passed to `Handler::readable` is simply an optimization. I also don't see a way to implement passing `ReadHint` (renamed to `EventHint`) to `Handler::writable` callback in a portable fashion.

rrichardson:
In this case, the confusion arose when the user got a hup and then a writable event for the same token. In isolation, the obvious thing to do when you get a hup event is to unregister the socket and remove it and the token from the slab.  However, immediately, the writable event is called for the token, regardless of what one does with the token.  This results in unexpected behavior. 

Given the current design, one would have to update the state in their connection's state machine indicating a disconnection, so that they can ignore the writable callback. 

An approach that would allow them to not have to manage a potential disconnection in their state machine is to simply pass the ReadHint information to both readable and writable, then the convention would be to check the  Hint for err/hup before touching the socket.  (more specifically, on a read event, one would likely read then deregister, and on a write event, they would just deregister) 

If they're on OSX, then they'll just have to understand that they'll never get formal hup/err hints (afaik) so nothing really changes.

carllerche:
The current behavior / API is also problematic when dealing with the remote end shutting down part of the connection....

rrichardson:
Yes. That scenario applies to my above comment.   

ping @dpc and @hoxnox

[...]

PR #194, opened 2015-06-06
consolidated the behavior of kqueue and epoll into IoEvent iterator (for #184)

rrichardson:
~~*Note*: I am only able to test this on linux, so I am not sure if the kqueue impl works correctly. If someone on a bsd machine could test that would be fabulous.~~  Thanks, Travis

Major attractions:

* No more ReadHint, everything is included into Interest which is passed to ready()
* EventsIterator is now defined inside sys/unix (this is because epoll and kqueue use different data structures for their IoEvent iterator) 
* event_loop.poll now uses EventsIterator 
* All tests use the new ready() handler and all tests pass. 

Two potential issues: 

* Interest should probably be renamed, since it reflects both inbound and outbound event bitmasks. 
* In order to iterate through self.poll.events iterator from event_loop and pass &mut self into the handler, I had to move the events struct out poll temporarily during the poll() and then return it. I'm not totally stoked about this.  The other potential option is to use a loop {} until events.iter().next() returns None, and copy the contents out of each cell.  The 2nd option might be a tad faster, but is uglier :) One involves a guaranteed memcpy of the Events struct into and out of Poll.  The 2nd copies Token and Interest out of each event.

### DIFF
--- a/src/sys/unix/epoll.rs
+++ b/src/sys/unix/epoll.rs
@@ -3,6 +3,7 @@ use event::IoEvent;
 use nix::sys::epoll::*;
 use nix::unistd::close;
 use std::os::unix::io::RawFd;
+use std::slice::Iter;
 
 #[derive(Debug)]
 pub struct Selector {
@@ -37,6 +38,8 @@ impl Selector {
                            .map_err(super::from_nix_error));
 
         unsafe { evts.events.set_len(cnt); }
+
+        evts.coalesce();
 
         Ok(())
     }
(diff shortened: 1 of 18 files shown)


src/sys/unix/epoll.rs
```diff
@@ -116,46 +119,59 @@ impl Drop for Selector {
...
-        }
+            if e.events.contains(EPOLLOUT) {
+                kind = kind | Interest::writable();
+            }
 
-        // EPOLLHUP - Usually means a socket error happened
-        if epoll.contains(EPOLLERR) {
-            kind = kind | Interest::error();
-        }
+            // EPOLLHUP - Usually means a socket error happened
```
danburkert: Is this comment still relevant?  It seems to be outdated.

danburkert:
I looked into the `test_readable` failure on OS X.  It appears that the `Funtimes::ready` handler is getting called with an Interest of `Writable | Hinted`.  That suggests that somehow the readable/writeable interests are getting mixed up in kqueue specific code, but I could not find anything obviously wrong in the patch.

carllerche:
Nice. I'm going to have to take some time for this... It does look like CI fails on OS X right now though :)

rrichardson:
Yea. Dan and I spent an hour yesterday trying to debug that failing test
case.  So far it appears that kevent is simply not suppling a readable
event for that fd, even though we are successfully registering read
interest.  This is obviously not a satisfactory answer, but it is very hard
for me to test this behavior from Linux :)

danburkert:
I tracked the issue down to some undefined behavior with how KEvent structs were being used.  Previously a Vec<KEvent> was allocated, and the length was unsafely incremented, and then the resulting slot was initialized with values.  In my understanding this is UB behavior in Rust, since it requires treating uninitialized memory as a mut reference.  I removed the unsafe usage, and the tests pass.  The commit is [here](https://github.com/danburkert/mio/commit/13a3555f9d99e2984d6e6952acbc5188c4ab72b1), and I've opened a PR on @rrichardson's branch which this PR is based on.

dpc:
I like this. Any ETA when this will get merged? I could remove some workarounds from here and there with it.

carllerche:
@dpc I'm going to work on this next! It's a big change so I am going to try to break it down a bit.

carllerche:
Closing in favor of #207

PR #201, opened 2015-06-26
Use `FromRawFd` from std

fhartwig:
Now that the `FromRawFd` trait can be used from rust stable, I've removed the mio-specific `FromRawFd`. I've also updated the .travis.yml file to use rust 1.1, since this won't build on 1.0 anymore (incidentally: would it be a good idea to use `stable` instead of hard-coded version numbers?).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: rust
 sudo: false
 rust:
   - nightly
-  - 1.0.0
+  - 1.1.0
 
 os:
   - linux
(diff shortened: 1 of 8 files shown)


carllerche:
Thanks for the PR.

Regarding `stable` vs. hardcoded version numbers, it seems like a good idea to run CI on the lowest supported version of Rust. Switching to std for `FromRawFd` increases the rust version dependency (which is OK in this case). I think it is good to explicitly catch it.

PR #206, opened 2015-07-07
pending does not need to be mutable

fkr:


carllerche:
Woops, I think I did this change without looking at the PR. Sorry.

PR #208, opened 2015-07-08
Move inserting and unlinking timeout log to trace level

thijsc:
Found two more log statements that occur very often as part of the normal loop operation.


carllerche:
Makes sense to me

carllerche:
Merged. Thanks: 1b2c5ead21f62041bb789195edae8396a57782bf

Issue #204, opened 2015-07-04
Some additional traits inheriting from Evented for outside libraries to use.


dpc:
Could `mio` define some additional traits inheriting from `Evented` trait?

In mioco I'm planing to use dynamic dispatch, so the coroutine needs to do everything through trait functions.

It seems to me that mio defined the following clases of `Evented` structs:

* Readers
 * try_read
 * close (inbound)
* Writers
 * try_write
 * close (outbound)
* Listeners
 * accept
 * close (stop listening)
* Connectors
 * connect
* anything else? timers, signals etc?

Could this traits be added to mio? Is this a good idea?

carllerche:
Probably a good idea :smile: I will keep this issue open and come back around to it soon.

carllerche:
Could you say more about why you need to separate close for both read and write. I am assuming that it is to expose socket shutdown, in which case could it make sense to have a `Shutdown` trait that has `shutdown_read()` and `shutdown_write()` functions?

For Listeners, is there a reason to separate out `close()` from dropping the listener?

dpc:
The only case that I'm aware of where inbound and outbound close makes a difference is TCP. For everything else it could be aliased or something similiar. In `mioco` I pass to coroutine uniform array of trait objects, so all the possible trait methods needs to be in that trait and it's up to coroutine to use them properly, but `mio` could group the Evented into any set of distinct traits, and `mioco` could provide default implementation for them just fine. So I'm open to any design of such traits. All I want to avoid is to implement such common trait one-by-one for every `Evented` struct supported by `mio`.

dpc:
Now when I think about it, if `mio` could group `Evented`, then maybe I could use static dispatch and array of enums.

carllerche:
What do you mean by group `Evented`?

carllerche:
I guess, I'm wondering why your lib `mioco` couldn't define it's own trait that extends the set of mio traits that it cares about.

dpc:
"group `Evented`" - define types of `Evented`-implementing structs, that behave the same way. 

mioco does already implement a trait that it uses for everything:

https://github.com/dpc/mioco/blob/master/src/lib.rs#L87

But then I have to implement it for every `struct` from `mio`: https://github.com/dpc/mioco/blob/master/src/lib.rs#L111

If `mio` had a common traits that behave the same than I could do:

```
impl<T> Evented for T where T : mio::Stream {

}
```

```
impl<T> Evented for T where T : mio::Listener {

}
```

[...]

PR #215, opened 2015-07-19
Introduce `Listener` trait.

dpc:
As per discussion in #204.

Abstract `accept()` operation, so that `mioco` can have a generic wrapper over them.

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -69,6 +69,12 @@ impl<T: Write> TryWrite for T {
     }
 }
 
+pub trait TryAccept {
+    type Output;
+
+    fn accept(&self) -> Result<Option<Self::Output>>;
+}
+
 /*
  *
  * ===== Helpers =====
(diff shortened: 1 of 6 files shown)


dpc:
This is for review only. I believe it might break existing software, as they need to import a `Listener` trait to use `accept`.

Some questions:

Is `Output` OK, or should implementing types include the `Option` themselves?

carllerche:
Thinking about it, `TryAccept` seems better than `Listener` since there is only one fn, and it is consistent with `TryRead` and `TryWrite`.

Also, a backwards compatible way to add a trait would be to leave `accept` on the original structs and have the trait impl call the struct impl.

dpc:
OK.

dpc:
All done.

carllerche:
Thanks. Merged: 478cef3f66716bd8584620b09838a177028edda9

PR #217, opened 2015-07-21
Fix assumption about queue order. Fixes #216 .

### LINKED ISSUE #216: Memory sync issue under multithreaded concurrency
It seems that something along the `notify.rs` and `util/mpmc_bounded_queue.rs` codepath having memory coherence issues under heavy multithreaded loads that utilize `notify()`.

Running the following program in `--release` mode will quickly demonstrate the issue (< 30s on my linux laptop): https://gist.github.com/jamwt/b9df06a87411ea2846aa

Typical Output:

```
$ time cargo run --release
     Running `target/release/mio`
@0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
@100000
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
the sum is 0
thread '<main>' panicked at '[BUG] at this point there should always be a message', ../src/libcore/option.rs:330
An unknown error occurred

To learn more, run the command again with --verbose.

real	0m3.473s
user	0m0.320s
sys	0m5.940s

```

jamwt:
The mpmc queue can fill slots out of order if a context switch occurs
between "reserving" a slot and marking that slot as filled.

So, insertion into slot 2 can look successful (and cause mio to
increment its own queue length counts) before an insertion into slot 1
is fully committed.

So, then when mio actually reads the queue, it gets None, because
slot 1, the "next" slot, isn't completely done yet.  Yielding the
scheduler to other threads will allow the partially-committed slot
to finish committing, and the queue will eventually be seen as having
both elements, slots 1 and 2.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -314,6 +314,11 @@ impl<H: Handler> EventLoop<H> {
                     handler.notify(self, msg);
                     cnt -= 1;
                 },
+                // If we expect messages, but the queue seems empty, a context
+                // switch has occurred in the queue's push() method between
+                // reserving a slot and marking that slot; let's spin for
+                // what should be a very brief period of time until the push
+                // is done.
                 None => thread::yield_now(),
             }
         }
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -3,7 +3,7 @@ use event::{IoEvent, EventSet, PollOpt};
 use notify::Notify;
 use timer::{Timer, Timeout, TimerResult};
 use std::default::Default;
-use std::{io, fmt, usize};
+use std::{io, fmt, thread, usize};
 
 /// Configure EventLoop runtime details
 #[derive(Copy, Clone, Debug)]
@@ -309,11 +309,13 @@ impl<H: Handler> EventLoop<H> {
 
     fn notify(&mut self, handler: &mut H, mut cnt: usize) {
         while cnt > 0 {
-            let msg = self.notify.poll()
-                .expect("[BUG] at this point there should always be a message");
-
-            handler.notify(self, msg);
-            cnt -= 1;
+            match self.notify.poll() {
+                Some(msg) => {
+                    handler.notify(self, msg);
+                    cnt -= 1;
+                },
+                None => thread::yield_now(),
+            }
         }
     }
 


carllerche:
Could you elaborate a bit more? Which part of the `push` function causes this?

jamwt:
Focusing on:

https://github.com/carllerche/mio/blob/master/src/util/mpmc_bounded_queue.rs#L102-L108

There are two pushers, thread A and thread B, and reactor thread, thread R.

 1. Thread A executes the CAS on line 102, and successfully reserves slot 0
 2.   *context switch to B*
 3. Thread B executes the CAS on line 102, and successfully reserves slot 1
 4. Thread B executes the sequence store on line 106, and marks the slot reserved
 5. Thread B exits `push()`, and mio increments `self.state` no the `NotifyInner` object. ( https://github.com/carllerche/mio/blob/master/src/notify.rs#L160 )
 6. *context switch to R*
 7. R does `.check()` on the notify, which returns one item ready, which represents thread B's push.
 8. R attempts to process one item, and it calls `.pop()` on the queue
 9. Pop gets the next slot to process from the atomic, which is slot 0 ( in this block https://github.com/carllerche/mio/blob/master/src/util/mpmc_bounded_queue.rs#L123-L127 ), and compares it to the sequence value modulo the ring size
 10. The sequence value is still slot - ring_size, since thread A was context switched out before executing the sequence store at line 106.  So `diff < 0`, and None is returned.
 11. The `EventLoop` panics on the empty queue in thread R.

Basically, a bad context switch in the wrong spot can have a "queue size" of 1, but the front of the queue be null, until thread that owns slot 0 gets scheduled back in so it can mark the slot and set the value.  We just need to yield the scheduler for a very short period of time so the OS can schedule that thread and populate the value.

(Validated all this with the repro script and tracing.)

carllerche:
Yep, this makes total sense now. Thanks.

Also, how did you do tracing?

carllerche:
Merged: f67357b1bf9f01a2fa74573796e5e26108562b11

PR #221, opened 2015-07-26
Link to `mioco` in README.

### LINKED ISSUE #191: Introducing coroutine-based mio connection handlers.
Hi,

Just to let you know, I got the basics of `mioco` working: https://github.com/dpc/mioco/

It narrows down to being able to handle each connection with a function running in per-connection coroutines, that block and unblock automatically and are scheduled quickly in userspace.

Example function:

https://github.com/dpc/mioco/blob/master/examples/echo.rs#L70

What's nice is that the performance of the simple tcp echo seem roughly the same as [non-coroutine example that I played with before](https://github.com/dpc/mioecho), and similar to libev-based tcpecho server.

dpc:
Can `mioco` get a link from `mio`'s README? :)

Fixes #191.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,13 +53,19 @@ or higher level libraries.
 Currently, MIO only supports Linux and Darwin. The goal is to support
 all platforms that support Rust and the readiness IO model.
 
-## Reading
+## Resources
 
 Please submit PRs containing links to MIO resources.
+
+### Reading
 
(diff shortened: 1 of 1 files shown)


carllerche:
Merged: 94113ff13d44055cfead673ff58ddda5883e9200

dpc:
Thanks!

PR #222, opened 2015-07-28
WIP: Getting started guide & example

carllerche:
Not actually ready to merge yet... but any comments are welcome. I will keep updating the PR as I continue working on the document.
## Things to document
- [ ] If registering both `Readable` & `Writable` w/ edge + oneshot, what happens when receiving `ready` with only one.

[Rendered](https://github.com/carllerche/mio/blob/getting-started/doc/getting-started.md)

### DIFF
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -11,7 +11,8 @@ Of course, you will need Rust installed. If you haven't already, get it
 here: [rust-lang.org](https://www.rust-lang.org).
 
 The complete echo server can be found
-[here](../examples/ping_pong/src/server.rs).
+[here](../examples/ping_pong/src/server.rs). A client example is
+[here](../examples/ping_pong/src/client.rs).
 
 > **Note:** As of the time of writing, Mio does not support Windows
 > (though support is currently in progress).
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -1,0 +1,259 @@
+# Getting Started
+
+This section will serve as an introductory Mio tutorial. It assumes that
+you have familiarity with the [Rust](http://www.rust-lang.org/)
+programming language and the [Cargo](https://crates.io) tool. It will
+start from generating a new [Rust](http://www.rust-lang.org/) project
+using [Cargo](https://crates.io) up to writing a simple TCP echo server
+and client.
+
+Of course, you will need Rust installed. If you haven't already, get it
+here: [rust-lang.org](https://www.rust-lang.org).
+
+> **Note:** As of the time of writing, Mio does not support Windows
+> (though support is currently in progress).
+
+## Setting up the project
+
+The first step is getting a new Cargo project setuup. In a new
+directory, run the following:
+
+```not_rust
+cargo new pingpong --bin
+cd pingpong
+```
+
+Now, open the directory in your favorite text editor. You should see the
+following files:
+
+* src/main.rs
+* Cargo.toml
+
+If you are not already familiar with Cargo, you can learn more about it
+[here](http://doc.crates.io/).
+
+Open `Cargo.toml` and add a dependency on Mio by putting the following a
+the bottom of the file:
+
+```not_rust
+[dependencies.mio]
+git = "https://github.com/carllerche/mio"
+```
+
+> Since the guide is currently written assuming some APIs that are not
+> available on crates.io yet, we depend on the master branch of Mio on
+> GitHub.
+
+Save the file, then compile and run the project using the following
+command:
+
+```not_rust
+cargo run
+```
+
+You will see some Cargo related output followed by `Hello, world!`. We
+haven't written any code yet and this is the default behavior of a
+freshly generated Cargo project.
+
+## Writing the Echo Server
+
+Let's start by writing a very simple server that accepts connections and
+does nothing with them. The connections will be accepted and shutdown
+immediately after.
+
+Here is the entire code, we'll step through it in a bit.
+
+```
+extern crate mio;
+
+use mio::tcp::*;
+
+const SERVER: mio::Token = mio::Token(0);
+
+struct Pong {
+    server: TcpListener,
+}
+
+impl mio::Handler for Pong {
+    type Timeout = ();
+    type Message = ();
+
+    fn readable(&mut self, event_loop: &mut mio::EventLoop<Pong>, token: mio::Token, hint: mio::ReadHint) {
+        match token {
+            SERVER => {
+                println!("the server socket is ready to accept a connection");
+                match self.server.accept() {
+                    Ok(Some(connection)) => {
+                        println!("accepted a socket, exiting program");
+                        event_loop.shutdown();
+                    }
+                    Ok(None) => {
+                        println!("the server socket wasn't actually ready");
+                    }
+                    Err(e) => {
+                        println!("listener.accept() errored: {}", e);
+                        event_loop.shutdown();
+                    }
+                }
+            }
+            _ => panic!("Received unknown token"),
+        }
+    }
+}
+
+fn main() {
+    let address = "0.0.0.0:6567".parse().unwrap();
+    let server = TcpListener::bind(&address).unwrap();
+
+    let mut event_loop = mio::EventLoop::new().unwrap();
+    event_loop.register(&server, SERVER);
+
+    println!("running pingpong server");
+    event_loop.run(&mut Pong { server: server });
+}
+```
+
+Let's break it down. The first step (at the beginning of the `main`
+function), is to create a TCP listener. This will create the socket,
+bind to the specified address, and start listening for inbound
+connections.
+
+The next step is to register the socket with the event loop.
+
+### The Event Loop
+
+The Mio event loop is able to monitor many sockets and notify the
+application when the state of a socket changes. The application
+registers sockets with the event loop. This is done by supplying a
+`Token` with the socket, associating the two. When the event loop is
+started, the application passes in a custom event handler. Whenever the
+state of any socket changes, the event loop will notify the event
+handler, calling the appropriate event function on the handler and
+passing in the originally supplied `Token`.
+
+In our case, the event handler is the `Pong` struct as it implements the
+`mio::Handler` trait. We only define the `readable` function, but the
+`mio::Handler` trait has [other
+functions](http://rustdoc.s3-website-us-east-1.amazonaws.com/mio/master/mio/trait.Handler.html)
+that can be defined to handle other event types.
+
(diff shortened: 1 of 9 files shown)


doc/getting-started.md
```diff
@@ -0,0 +1,583 @@
...
+                }
+            }
+            _ => panic!("Received unknown token"),
+        }
+    }
+}
+
+fn main() {
+    let address = "0.0.0.0:6567".parse().unwrap();
+    let server = TcpListener::bind(&address).unwrap();
```
allan-simon: I found in other project this syntax of error handling to be quite elegant rather than just an `unwrap` 

```rust
 let sock = TcpListener::bind(&addr).ok().expect("Failed to bind address");
```

doc/getting-started.md
```diff
@@ -0,0 +1,583 @@
...
+immediately after.
+
+Here is the entire code, we'll step through it in a bit.
+
+```rust
+extern crate mio;
+
+use mio::tcp::*;
+
+const SERVER: mio::Token = mio::Token(0);
```
allan-simon: hermanradtke.com/2015/07/22/creating-a-multi-echo-server-using-rust-and-mio.html
quote from one of the comment in the code from this blog post 

```rust
            // I don't use Token(0) because kqueue will send stuff to Token(0)
            // by default causing really strange behavior. This way, if I see
            // something as Token(0), I know there are kqueue shenanigans
            // going on.
            token: Token(1),
```

doc/getting-started.md
```diff
@@ -0,0 +1,584 @@
...
+
+```not_rust
+cargo new pingpong --bin
+cd pingpong
+```
+
+Now, open the directory in your favorite text editor. You should see the
+following files:
+
+* src/server.rs
```
tanriol: `src/main.rs`?

seanmonstar:
What does `hup` mean exactly? That the socket closed in some violent manner?

From a client's perspective, if a server closes the connection after writing out it's response, will that trigger `hup`? In that case, it'd be proper to just read on the socket til EOF.

du5t:
The first thing I did when trying to understand mio is write a block diagram to see how my application's data source/sink, mio's socket(s), handler(s), and event loop worked together. It would be really helpful to have one of those.

The next thing I did was write a series of steps that seem to be embedded in the getting started guide. I'm still getting them sorted, but it looks something like:

1. Create+configure Event Loop
2. Create Handler(s) for Sockets
3. Create+configure Sockets
4. Register Sockets with Event Loop, passing appropriate Token(s) to Socket.
5. Run Event Loop (with a plan for handling any other app tasks in other threads)

A list of things like that to check off to ensure you've got everything covered would also be very excellent.

007pig:
I have a question: In Ping Pong example, why register_opt() uses edge + oneshot but reregister() uses only oneshot as its opts? Thanks!

tshepang:
The type inference in this line caused me to pause:

    let address = std::net::SocketAddr::from_str("0.0.0.0:6568").unwrap();

I would rather we have:

    let address = std::net::SocketAddr::from_str("0.0.0.0.1:6568").unwrap();

It is less clever, but also less surprising.

doc/getting-started.md
```diff
@@ -0,0 +1,584 @@
...
+> [`State`](../examples/ping_pong/src/server.rs#L249). The full source of
+> the echo server in its final form is
+> [here](../examples/ping_pong/src/server.rs).
+
+```rust
+fn ready(&mut self, event_loop: &mut mio::EventLoop<Pong>, events: mio::EventSet) {
+    match self.state {
+        State::Reading(ref mut buf) => {
+            assert!(events.is_readable(), "unexpected events; events={:?}", events);
+            self.read(event_loop);
```
matklad: Should it be `self.read(event_loop, &mut buf)`? The `buf` variable is unused.

cseghrce:
how can i register client socket and sever socket in same event_loop please?
i wanna 
A -> B -> C 
A act as client , B act as server for A but client for C and C is a Main server. please help me.

doc/getting-started.md
reem: The downside here is that you lose the information contained in the `io::Error`. Fortunately there is now an `expect` on `Result` as well which will log both the error and your message on error.

You could do: `let server = TcpListener::bind(&address).expect("Failed to bind listener.");` to get nicer messages without losing info.

doc/getting-started.md
```diff
@@ -0,0 +1,584 @@
...
+Let's start by writing a very simple server that accepts connections and
+does nothing with them. The connections will be accepted and shutdown
+immediately after.
+
+Here is the entire code, we'll step through it in a bit.
+
+```rust
+extern crate mio;
+
+use mio::tcp::*;
```
reem: I'd avoid glob imports both here and in example code, though it can make the final code a bit briefer, it makes it much harder to follow since you don't know where symbols came from.

doc/getting-started.md
reem: I'd stick a quick comment here like you have in the source (https://github.com/carllerche/mio/blob/getting-started/examples/ping_pong/src/server.rs#L12-L14) + a note that "we use tokens to identify the different sockets we are interested in" to make things a bit clearer. As is, I would be confused about tokens all the way until the tokens section, which is distracting from the rest of the code.

doc/getting-started.md
```diff
@@ -0,0 +1,584 @@
...
+    server: TcpListener,
+}
+
+impl mio::Handler for Pong {
+    type Timeout = ();
+    type Message = ();
+
+    fn ready(&mut self, event_loop: &mut mio::EventLoop<Pong>, token: mio::Token, events: mio::EventSet) {
+        match token {
+            SERVER => {
```
reem: I'd also add a note here that just says "We received an event on the SERVER token, so we know it's relevant to the server socket."

doc/getting-started.md
```diff
@@ -0,0 +1,584 @@
...
+        }
+    }
+}
+
+fn main() {
+    let address = "0.0.0.0:6567".parse().unwrap();
+    let server = TcpListener::bind(&address).unwrap();
+
+    let mut event_loop = mio::EventLoop::new().unwrap();
+    event_loop.register(&server, SERVER);
```
reem: and then here, saying "Here we register our interest in the server socket, and associate it with our SERVER token."

doc/getting-started.md
```diff
@@ -0,0 +1,584 @@
...
+    fn ready(&mut self, event_loop: &mut mio::EventLoop<Pong>, token: mio::Token, events: mio::EventSet) {
+        match token {
+            SERVER => {
+                // Only receive readable events
+                assert!(events.is_readable());
+
+                println!("the server socket is ready to accept a connection");
+                match self.server.accept() {
+                    Ok(Some(socket)) => {
+                        println!("accepted a socket, exiting program");
```
reem: Say "shutting down the event loop" here, then add another print at the end of main saying "exiting program". This makes the whole flow of starting up, running, then shutting down and exiting the event loop clearer.

doc/getting-started.md
```diff
@@ -0,0 +1,584 @@
...
+Our `Pong` struct looks like this now:
+
+```rust
+struct Pong {
+    server: TcpListener,
+    connections: Slab<Connection>,
+}
+```
+
+#### Tokens
```
reem: some of my token comments could be mitigated by moving this up higher, as is I would just be confused for quite some time.

reem:
Looking at the annotated source, I think many of the comments could be copied into the source snippets here - even though they're explained after, having comments in the code makes things much clearer.

aochagavia:
The note about not supporting Windows is obsolete now :wink:

nixpulvis:
Maybe it's unavoidable, but the complexity of the basic example seems a bit much. New users of this library might want to see the minimal code to read from STDIN for example. Avoiding making a bunch of structs and abstractions for a simpler example might better illustrate how MIO actually works, then a longer example for some advice on how to make it work for larger projects.

tshepang:
me agrees with @nixpulvis

carllerche:
@nixpulvis Perhaps, the example is more of a tutorial that touches on all the concepts. I can think more about the simplest "hello world" case, but that would most likely be with the `Poll` API, so not *that* much overlap.

STDIN in a portable way (including windows) is going to be a bit tricky...

mdg:
Where it lists File Operations as a non-goal, would it be possible to link to an explanation for why that is? It seems like not that different from doing network stream i/o.

doc/getting-started.md
```diff
@@ -0,0 +1,584 @@
...
+
+Of course, you will need Rust installed. If you haven't already, get it
+here: [rust-lang.org](https://www.rust-lang.org).
+
+The complete echo server can be found
+[here](../examples/ping_pong/src/server.rs). A client example is
+[here](../examples/ping_pong/src/client.rs).
+
+> **Note:** As of the time of writing, Mio does not support Windows
+> (though support is currently in progress).
```
remram44: This disagrees with the [README](https://github.com/carllerche/mio/commit/168fcb5938897eadc9f57e5bc52ae148108e1778#diff-04c6e90faac2675aa89e2176d2eec7d8R59)?

oconnor663:
Initial experiences with the guide:

- noticed the version in the Cargo.toml file is out of date, I'll try the latest one
- the nix build produces warnings, but builds fine
- oops, using the latest 0.5.1 version seems to produce a build error, switching back to 0.4.1 like the tutorial recommended
- that works but I see more warnings

Crazy-Owl:
Please note that in 0.5.1 `event_loop.register()` takes 4 arguments.

beatgammit:
Is anyone working on this? I recently got a non-trivial example with `mio` working, so I think I'm ready to contribute to a getting started guide. I don't want to duplicate any work though, so let me know if you're working on something.

doc/getting-started.md
```diff
@@ -0,0 +1,584 @@
...
+
+If you are not already familiar with Cargo, you can learn more about it
+[here](http://doc.crates.io/).
+
+Open `Cargo.toml` and add a dependency on Mio by putting the following a
+the bottom of the file:
+
+```toml
+[dependencies]
+mio = "0.4.1"
```
sateffen: We're at mio 0.6.0 already :)

allan-simon:
@beatgammit  i'm not working on it

vi:
> Getting started guide
> Currently a work in progress: Getting Started.

Links to old v0.4 guide telling Windows is not supported? Shall it be marked as obsolete and not be the first link in the README.md?

carllerche:

PR #227, opened 2015-08-11
commit -> rev

tomjakubowski:
The right key to use to pin a Cargo git dependency is `rev`, not `commit`; this was causing mio to depend on the latest master of `"https://github.com/carllerche/bytes"`. Unfortunately Cargo doesn't seem to warn about the unused key in this case!


carllerche:
Thanks for pointing this out. I ended up updating the commit at the same time: 77775a151d5fb9cedcb6b298f06198aec2c8e62a

PR #231, opened 2015-08-20
Implement Handler::tick()

### LINKED ISSUE #219: Does deregister clear events already pending delivery? Should it?
Let's say I have two event sources registered, both of them fired at the same time and landed in pending events list for current iteration.

While handling the first event, I do `deregister` on the other one. Will I still get it? It seems to me that I will, which is very inconvenient, as it makes hard (impossible?) to safely deregister events other than the one that just fired.

A spin of the above case is: What if I do `deregister` and right after used `register_opt` on different event source, using the same token.

Another spinoff is: What if I do `reregister` using different interest.

If I am not missing anything, it seems to me that either: the end of each event delivery iteration should have a notification, during which calls like `deregister` and `reregister` can be safely used, or `deregister` and `reregister` should make sure to remove any events that are not to be delivered anymore.

hjr3:
The `Handler::tick()` method is called at the beginning of each event
loop tick.

Fixes #219

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -263,6 +263,7 @@ impl<H: Handler> EventLoop<H> {
         self.io_process(handler, events);
         self.notify(handler, messages);
         self.timer_process(handler);
+        handler.tick(self);
         Ok(())
     }
 
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -29,4 +29,8 @@ pub trait Handler: Sized {
     /// Invoked when `EventLoop` has been interrupted by a signal interrupt.
     fn interrupted(&mut self, event_loop: &mut EventLoop<Self>) {
     }
+
+    /// Invoked at the end of an event loop tick.
+    fn tick(&mut self, event_loop: &mut EventLoop<Self>) {
+    }
 }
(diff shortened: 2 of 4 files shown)


src/event_loop.rs
```diff
@@ -263,6 +263,7 @@ impl<H: Handler> EventLoop<H> {
         self.io_process(handler, events);
         self.notify(handler, messages);
         self.timer_process(handler);
+        handler.tick(self);
```
hjr3: I wonder if I should move this above the call to `timer_process`. Here is how the logs look currently:

```
TRACE:mio::event_loop: event loop tick
TRACE:mio::timer: tick_to; now=40; tick=39
TRACE:mio::timer: ticking; curr=Token(18446744073709551615)
TRACE:mio::timer: ticking; curr=Token(18446744073709551615)
TRACE:mob_server::server: Handling end of tick
TRACE:mio::event_loop: event loop tick
TRACE:mio::timer: tick_to; now=42; tick=41
TRACE:mio::timer: ticking; curr=Token(18446744073709551615)
TRACE:mio::timer: ticking; curr=Token(18446744073709551615)
TRACE:mob_server::server: Handling end of tick
```

carllerche: Why are you wondering this? Not following. Everything looks good to me, let me know if you want it merged :+1:

carllerche:
Thanks! Merged

dpc:
Nice!

PR #232, opened 2015-08-20
Remove UnsafeCell from Unix Awakener

alexcrichton:
Add some impls of Read/Write for `&T` (like the stdlib does) to express that a
mutable borrow isn't needed, allowing removal of the unsafe cells in the
awakener.

### DIFF
--- a/src/net/unix.rs
+++ b/src/net/unix.rs
@@ -197,6 +197,12 @@ impl Read for PipeReader {
     }
 }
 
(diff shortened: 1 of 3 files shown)


carllerche:
:+1:

carllerche:
Merged. Thanks: 4dadf21d381d29c02820bdc62d8200ec1b50f109

PR #233, opened 2015-08-20
Remove Selector/Io from public API

alexcrichton:
(discussed on IRC)


alexcrichton:
er, looks like this is needed as public for the `Evented` trait currently, will have to figure out something else later.

PR #234, opened 2015-08-20
Add missing trait annotations for rust nightly

alexcrichton:
Warnings about these started appearning on nightly and they'll eventually turn
into hard errors, so may as well fix them now!


carllerche:
Merged. Thanks: 736d7320c90403969c7b5ef02c03e829061444b6

PR #236, opened 2015-08-21
Test on to Rust 'stable' instead of Rust '1.1.0'

frewsxcv:
Would prevent having to upgrade the version number after every Rust release

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: rust
 sudo: false
 rust:
   - nightly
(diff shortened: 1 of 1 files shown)


hjr3:
:+1:

carllerche:
I plan on explicitly testing using the oldest version of Rust that MIO supports. I want a change here to indicate that MIO has increased it's minimum Rust version.

frewsxcv:
Fair enough :+1:

PR #239, opened 2015-08-25
Preliminary Windows TCP/UDP support

alexcrichton:
These commits add preliminary support for the TCP/UDP API of mio, built on top of [IOCP](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365198%28v=vs.85%29.aspx) using some [raw Rust bindings](https://github.com/alexcrichton/wio) plus some [networking extensions](https://github.com/alexcrichton/net2-rs) as the foundational support. This support is definitely **still experimental** as there are likely to be a number of bugs and kinks to work out.

I haven't yet done much benchmarking as there are still a number of places I would like to improve the implementation in terms of performance. I've also been focusing on getting "hello world" and the in-tree tests working ASAP to start getting some broader usage and feedback. High level docs are available in the `src/sys/windows/mod.rs` file and the TCP/UDP implementations are quite similar in terms of how they're implemented.

Not many new tests were added, but all tests (other than those using unix sockets) are passing on Windows and an `appveyor.yml` file was also added to enable AppVeyor CI support to ensure this doesn't regress.

cc #155

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,18 +22,24 @@ log   = "0.3.1"
 nix   = "0.3.9"
 libc  = "0.1.8"
 slab  = "0.1.0"
-winapi = "0.1.23"
 clock_ticks = "0.0.5"
-
-[dependencies.bytes]
-git = "https://github.com/carllerche/bytes"
-rev = "7edb577d0a"
+bytes = { git = "https://github.com/carllerche/bytes", rev = "7edb577d0a" }
+winapi = "0.2.1"
+wio = { git = "https://github.com/alexcrichton/wio" }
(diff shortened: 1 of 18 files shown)


alexcrichton:
I should also point out that this does not close #155 as I believe mio will probably want some deeper API changes to improve Windows/Unix support. Additionally, there two primary aspects which are not currently implemented:

* The `PollOpt::level()` option will always return an error on Windows right now.
* Moving a handle between event loops or adding it to multiple event loops will not work very well right now.

There's still lots of legwork to accomplish, but this should get a number of initial use cases up and running!

PR #253, opened 2015-08-28
windows: Remove Selector::io

### LINKED ISSUE #243: windows: Remove Selector::io
This is a field in the Windows implementation of `Selector` which keeps track of all pending I/O. There are two problems with this:

* Hash map lookups are slammed quite a bit as every I/O operation must be registered
* Callbacks are all `Box<T>` which means that each operation has at least one allocation

The map could possibly store `Arc<T>` instead of `Box<T>` (and all I/O handles internally have an `Arc`), which would solve the allocation problem but not the hash map problem. To kill two birds with one stone we could possibly do:

```rust
#[repr(C)]
pub struct MioOverlapped {
    real_overlapped: wio::Overlapped,
    finished: fn(...),
}
```

This way whenever I/O completes the event loop can cast the `OVERLAPPED` pointer to a `MioOverlapped` and call the `finished` function with the relevant contextual information. The callbacks can do some unsafe casting to get back to the original I/O object. Quite unsafe, but less allocations!

alexcrichton:
This commit removes the internal hash map from Windows' Selector implementation
keeping track of what to do after any I/O completes. The "completion callback"
is now stored right after the OVERLAPPED structure in memory, and unsafe casting
is done to get back to the original `Arc<T>` and process the I/O event.

Currently this involves vendoring a small copy of `Arc<T>` to ensure that the
memory layout is guaranteed, but this guarantee could in theory be added to the
standard library one day.

Closes #243

### DIFF
--- a/src/sys/windows/from_raw_arc.rs
+++ b/src/sys/windows/from_raw_arc.rs
@@ -1,0 +1,116 @@
+//! A "Manual Arc" which allows manually frobbing the reference count
+//!
+//! This module contains a copy of the `Arc` found in the standard library,
+//! stripped down to the bare bones of what we actually need. The reason this is
+//! done is for the ability to concretely know the memory layout of the `Inner`
+//! structure of the arc pointer itself (e.g. `ArcInner` in the standard
+//! library).
+//!
+//! We do some unsafe casting from `*mut OVERLAPPED` to a `FromRawArc<T>` to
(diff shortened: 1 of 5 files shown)


carllerche:
:+1:

alexcrichton:
Landed in https://github.com/carllerche/mio/commit/98f128177ba889d100e38f295db12a7d7260a13e

PR #254, opened 2015-08-28
windows: Pool buffers on the selector

### LINKED ISSUE #245: windows: Pool buffers for reads/writes
Right now all reads/writes allocate a buffer whenever they happen. This can be quite expensive and it'd probably be more useful to pool buffers between operations. For example the event loop could have a cache of buffers which are handed out on reads/writes.

Probably want to benchmark the impact before/after as well.

alexcrichton:
This adds a cache of buffers to be stored locally on the selector for new
requests to use for both reads and writes. The cache has a maximum size of the
number of buffers it can store and new buffers are allocated if the cache is
empty at the time.

Closes #245

### DIFF
--- a/src/sys/windows/buffer_pool.rs
+++ b/src/sys/windows/buffer_pool.rs
@@ -1,0 +1,20 @@
+pub struct BufferPool {
+    pool: Vec<Vec<u8>>,
+}
+
+impl BufferPool {
+    pub fn new(cap: usize) -> BufferPool {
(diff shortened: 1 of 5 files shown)


carllerche:
:+1:

alexcrichton:
Landed in https://github.com/carllerche/mio/commit/13d5fc9e11bbe96f1e4b9593645766dcf8e5a4d4

PR #257, opened 2015-08-31
Replace EventLoop::register with register_opt

### LINKED ISSUE #248: Remove EventLoop::register, rename register_opt to register
The `register` API probably doesn't have the right default for most use cases, and most implementations end up calling `register_opt` anyway. This should probably be reflected by removing `register` and renaming `register_opt` to take its place.

alexcrichton:
Closes #248


carllerche:
:+1:

alexcrichton:
Merged as c07deda

PR #259, opened 2015-09-03
Move EventLoopConfig to the builder pattern

### LINKED ISSUE #247: Move EventLoopConfig to be a builder
A builder can be more extensible in terms of future additions (no risk of breaking existing clients), and it also provides a vector to provide platform-specific extensions via traits.

alexcrichton:
This allows it to be extensible in the future without breaking backwards
compatibility, as well being idiomatic in terms of configuration with a builder.
The `Copy` trait has also been removed as it one day may not contain `Copy`
configuration values (e.g. strings/vectors).

Closes #247

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 * Don't re-export bytes types
 * Preliminary Windows support
 * Renamed `EventLoop::register_opt` to `EventLoop::register`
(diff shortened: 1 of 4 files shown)


carllerche:
:+1:

alexcrichton:
Merged as https://github.com/carllerche/mio/commit/8b84f3307a94ddf9e5e2aca6a39663e00705f3af

PR #260, opened 2015-09-03
Don't take Buf/MutBuf for UDP

### LINKED ISSUE #255: mio::udp::UdpSocket.recv_from() API differs from std::net::UdpSocket API
`std::net::UdpSocket::recv_from()` returns `Result<(usize, SocketAddr)>`. `mio::udp::UdpSocket.recv_from()` returns `Result<Option<SocketAddr>>`.

`02:18 <~ carllerche> if you could file an issue w/ API inconsistencies that you are aware of... would be good`

Unaware of more, but this could be an issue for taking notes of such things.

alexcrichton:
Bring these methods in line with the `std::net` equivalents by having the same
signatures, but returning `Option<T>` instead of `T` to represent the "would
block" case via `None`.

Closes #255

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
   is also no longer `Copy`. (#259)
 * `TcpSocket` is no longer exported in the public API (#262)
 * `TcpListener` now returns the remote peer address from `accept` as well (#275)
+* The `UdpSocket::{send_to, recv_from}` methods are no longer generic over `Buf`
+  or `MutBuf` but instead take slices directly. The return types have also been
+  updated to return the number of bytes transferred. (#260)
 
 # 0.4.1 (July 21)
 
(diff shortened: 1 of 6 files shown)


carllerche:
Should there be a `_buf` variant?

alexcrichton:
It depends, I would personally think that the `bytes` crate should be totally independent of mio, in which case there wouldn't be a `_buf` variant. On the other hand if the two are tied together closely then it probably makes sense. Regardless you can build the `_buf` variant on this version externally either way.

nixpulvis:
I agree that there should be a way to deal with non buffered UDP messages, but I'm also curious if the interface for buffered responses can be more ergonomic. Having a `Result<Option<T>, E>` is a bit cumbersome.

carllerche:
I think this would be fine to merge, but it has conflicts.

carllerche:
:+1: Let's bring back `_buf` functions before the next release as part of #286

PR #261, opened 2015-09-04
Various Windows optimizations

alexcrichton:
* Hold onto mutexes instead of releasing and then immediately re-locking
* Don't move in and out of state so often in `read_done` for TCP


carllerche:
:+1:

PR #262, opened 2015-09-04
Remove TcpSocket from the public API

### LINKED ISSUE #249: Migrate TCP sockets to std construction patterns
The standard library has an extension library, [net2](https://github.com/alexcrichton/net2-rs) for fine-grained control over the construction of sockets. This pattern isn't quite followed in mio just yet, and it would perhaps be nice to mirror.

This issue would also enable mirroring the socket extensions in `net2` by not providing socket options on `TcpListener` that only make sense for `TcpStream` (and vice versa).

Additionally, the precise API of the socket options may wish to align with `net2` as I believe they currently mirror what the standard library used to do, which is a bit outdated in a few locations.

alexcrichton:
This commit removes `mio::tcp::TcpSocket` as a public type, only exposing the
`TcpStream` and `TcpListener` types. This change is motivated for a few reasons:

1. This mirrors the design of `std::net` more closely by having a static
   distinction between types and statically ruling out cases like calling
   `peer_addr` on a `TcpListener`.
2. This allows re-using standard extension methods on the `std::net` types,
   especially in the Windows implementation (which is now greatly simplified).

The primary reason to have `TcpSocket`, however, is to have fine-grained control
over socket creation and the ability to set various options. This is
implemented, however, in the `net2` extension crate which provides the
`TcpBuilder` type to configure a socket before it's bound/listened/etc. Methods
are provided on mio's `TcpStream` and `TcpListener` to be constructed from these
bare types so they can be passed in.

Closes #249

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
 # 0.5.0 (unreleased)
 
 * Don't re-export bytes types
-* Preliminary Windows support
-* Renamed `EventLoop::register_opt` to `EventLoop::register`
+* Preliminary Windows support (#239)
+* Renamed `EventLoop::register_opt` to `EventLoop::register` (#257)
 * `EventLoopConfig` is now a builder instead of having public struct fields. It
-  is also no longer `Copy`.
+  is also no longer `Copy`. (#259)
+* `TcpSocket` is no longer exported in the public API (#262)
 
 # 0.4.1 (July 21)
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ clock_ticks = "0.0.5"
 bytes = { git = "https://github.com/carllerche/bytes", rev = "7edb577d0a" }
 winapi = "0.2.1"
 wio = { git = "https://github.com/alexcrichton/wio" }
-net2 = "0.2.7"
+net2 = "0.2.9"
 
 [dev-dependencies]
 env_logger = "0.3.0"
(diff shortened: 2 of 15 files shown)


alexcrichton:
Another slightly different option would be to purge mentions of `TcpBuilder` entirely from the public API and instead just talk `std::net` types. The builder has methods to extract `TcpStream` or `TcpListener` at any time, so for example `mio::tcp::TcpStream::connect_builder(&builder, &addr)` could instead be `connect_builder(stream, &addr)` (taking ownership of `std::net::TcpStream`).

I think I prefer this API (e.g. not exporting types of other crates) a little more, WDYT?

rrichardson:
This looks fine on one condition: There has to be a way to call getsockopt and setsockopt on a socket after the socket has been bound/connected.  It is the only way to get certain bits of information from the socket in some cases.

alexcrichton:
Ah yeah you can certainly still do that, the `AsRawFd` trait implementation the minimum escape hatch needed, and beyond that I suspect new methods can certainly be added to the mio networking types (e.g. to mirror those on `net2::TcpStreamExt`)

carllerche:
I am OK w/ this as long as no functionality is lost.

/cc @reem @jamwt @dpc @hjr3 and whoever else!

hjr3:
:+1: i think this is a good change

jamwt:
All of our stuff only uses `TcpListener` and `TcpStream`, so +1 here.

dpc:
Fine with me.

alexcrichton:
Merged as https://github.com/carllerche/mio/commit/06b359a188d802162a69020f0624a4884848e7c3!

PR #270, opened 2015-09-13
flush changelist more frequently

### LINKED ISSUE #267: kqueue: spurious events after register() and immediate close()
With mio v.0.4.3 on OSX, if I run the below program and run `nc 127.0.0.1 12345` in another terminal, it prints "stream event. THIS SHOULD NEVER HAPPEN".

```rust
extern crate mio;

use std::net::ToSocketAddrs;

pub struct Handler {
    listener_is_ready: bool,
}

impl Handler {
    pub fn new() -> Handler {
        Handler { listener_is_ready: false }
    }
    pub fn reset(&mut self) {
        self.listener_is_ready = false;
    }
}

type EventLoop = ::mio::EventLoop<Handler>;

const LISTENER_TOKEN : ::mio::Token = ::mio::Token(0);
const STREAM_TOKEN : ::mio::Token = ::mio::Token(1);

impl ::mio::Handler for Handler {
    type Timeout = ();
    type Message = ();
    fn ready(&mut self, _event_loop: &mut EventLoop, token : ::mio::Token, _events: ::mio::EventSet) {
        if token == LISTENER_TOKEN {
            println!("listener is ready!");
            self.listener_is_ready = true;
        } else if token == STREAM_TOKEN {
            println!("stream event. THIS SHOULD NEVER HAPPEN");
        }
    }
}

pub fn main() {
    let mut event_loop: EventLoop = ::mio::EventLoop::new().unwrap();
    let addr = "127.0.0.1:12345".to_socket_addrs().unwrap().next().expect("could not parse address");
    let listener = ::mio::tcp::TcpListener::bind(&addr).unwrap();
    assert!(listener.accept().unwrap().is_none());
    event_loop.register_opt(&listener,LISTENER_TOKEN,
                            ::mio::EventSet::readable(), ::mio::PollOpt::edge()).unwrap();

    let mut handler = Handler::new();
    loop {
        println!("tick");
        event_loop.run_once(&mut handler).unwrap();
        if handler.listener_is_ready {
            match listener.accept().unwrap() {
                None => println!("accept failed!?"),
                Some(s) => {
                    event_loop.register_opt(&s, STREAM_TOKEN,
                                            ::mio::EventSet::writable() |
                                            ::mio::EventSet::readable(),
                                            ::mio::PollOpt::edge()).unwrap();

                    // Immediately deregister. No events should come in now.
                    event_loop.deregister(&s).unwrap();

                    // Note: I also expect that if I omit this deregister, the stream should still not
                    // receive any events because it will be dropped and therefore removed from the EventLoop.
                }
            }
        }
        handler.reset()
    }
}
```

dwrensha:
This fixes #267.

### DIFF
--- a/src/sys/unix/kqueue.rs
+++ b/src/sys/unix/kqueue.rs
@@ -38,10 +38,10 @@ impl Selector {
     pub fn register(&mut self, fd: RawFd, token: Token, interests: EventSet, opts: PollOpt) -> io::Result<()> {
         trace!("registering; token={:?}; interests={:?}", token, interests);
 
-        try!(self.ev_register(fd, token.as_usize(), EventFilter::EVFILT_READ, interests.contains(EventSet::readable()), opts));
-        try!(self.ev_register(fd, token.as_usize(), EventFilter::EVFILT_WRITE, interests.contains(EventSet::writable()), opts));
+        self.ev_register(fd, token.as_usize(), EventFilter::EVFILT_READ, interests.contains(EventSet::readable()), opts);
+        self.ev_register(fd, token.as_usize(), EventFilter::EVFILT_WRITE, interests.contains(EventSet::writable()), opts);
 
-        Ok(())
+        self.flush_changes()
     }
 
     pub fn reregister(&mut self, fd: RawFd, token: Token, interests: EventSet, opts: PollOpt) -> io::Result<()> {
@@ -51,13 +51,13 @@ impl Selector {
     }
 
     pub fn deregister(&mut self, fd: RawFd) -> io::Result<()> {
-        try!(self.ev_push(fd, 0, EventFilter::EVFILT_READ, EV_DELETE));
-        try!(self.ev_push(fd, 0, EventFilter::EVFILT_WRITE, EV_DELETE));
+        self.ev_push(fd, 0, EventFilter::EVFILT_READ, EV_DELETE);
+        self.ev_push(fd, 0, EventFilter::EVFILT_WRITE, EV_DELETE);
 
-        Ok(())
+        self.flush_changes()
     }
 
-    fn ev_register(&mut self, fd: RawFd, token: usize, filter: EventFilter, enable: bool, opts: PollOpt) -> io::Result<()> {
+    fn ev_register(&mut self, fd: RawFd, token: usize, filter: EventFilter, enable: bool, opts: PollOpt) {
         let mut flags = EV_ADD;
 
         if enable {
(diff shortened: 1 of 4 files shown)


dwrensha:
`kevent()` allows the user to register interest in events (via the `changelist` argument) in the same system call as requesting triggered events (via the `eventlist` argument). mio attempts to take advantage of this by batching `kevent()` calls; if there are no more than 1024 registrations, they are deferred until the `kevent()` call on the next `EventLoop::run_once()` call. This approach is enticing because it reduces the number of system calls. However, it has problematic semantics. What if an mio client registers interest on a file descriptor and then closes the descriptor before the next `run_once()` call? What ought to happen is that the client should receive no notifications about that descriptor. What happens in mio's current implementation is that the client gets notified that the (now invalid!) file descriptor is ready. I think  `kevent()` actually reports an error (via `EV_ERROR`) for these bad descriptors and mio ignores it --- that's another bug, and somewhat orthogonal.

carllerche:
Thanks. This change looks very reasonable to me.

I would :heart: :heart: if you could take the snippet in #267 and make it a test. This will be extra helpful in making sure windows supports the same semantics.

dwrensha:
> I would :heart: :heart: if you could take the snippet in #267 and make it a test.

OK. I'll do that and add it to this pull request.

dwrensha:
I've added a test that fails before this change and succeeds afterwards. Unfortunately, it fails on Windows too. :( https://ci.appveyor.com/project/carllerche/mio/build/1.0.42/job/yvvu3sspy95wd176

alexcrichton:
@dwrensha could you try [applying this patch](https://gist.github.com/anonymous/fdfe95117fb79bef31f1) to get the test to pass on Windows? I haven't run it locally but I think that'll do the trick.

alexcrichton:
Er, actually just verified that it passed on Windows! If you wanna just throw that on this branch then once CI is green this should be good to go @carllerche?

dwrensha:
@alexcrichton: Thanks! I've applied your patch. However, I'm worried that it doesn't cover all cases.  Does Windows reuse socket handles? Could it happen that a different socket is using the socket handle  by the time `run_once()` gets called, causing it to slip past your `INVALID_SOCKET` test?

alexcrichton:
So I wouldn't put it past Windows to reuse `SOCKET` handles, but could you elaborate on where you think the badness may arise? 

To give some background on the current implementation, however: currently each `SOCKET` is uniquely owned by the corresponding mio type, and when that type is dropped the socket is closed to ensure that all pending I/O is canceled and returns immediately. To do this we replace the current I/O object with `FromRawSocket::from_raw_socket(INVALID_HANDLE)`, which means that the transfer of ownership via the replacement closes the original object, and all future I/O will return an error as the socket is `INVALID_HANDLE`. There are still allocated resources for the socket, however, (e.g. buffers) which are only free'd once all the pending I/O returns and completes.

So the intention of these checks is to basically never generate an event if the socket has previously known to have been closed. In that sense I'm not quite sure where the reuse of an invalid handle would happen (you're either using your uniquely owned one or an always-invalid one), but I'm curious what you're thinking as well!

dwrensha:
Oh, I didn't realize that the `INVALID_SOCKET` value is coming from your implementation rather than from Windows. I am less worried now. :)

Elaboration of what I was thinking: one somewhat-plausible-looking way to deal with #267 in the kqueue() case would be to keep the current batching and to just ignore any "Bad file descriptor" errors. However, the fact that file descriptors and mio tokens can both be reused makes that approach untenable, I think, because you could end up matching events to the wrong I/O objects. I just wanted to make sure that we weren't going to hit similar problems here.

alexcrichton:
Ah ok cool, thanks for the clarification!

Merged as https://github.com/carllerche/mio/commit/2c3435c248df53324b955b34aa841c17253ff4f2

PR #274, opened 2015-09-14
Add a link to an article about "rotor" library

tailhook:
I've set link text to something different from article title because I feel the original title is too generic to be useful in this list. Feel free to change this.

[Review on Reviewable](https://reviewable.io/reviews/carllerche/mio/274)


carllerche:
Merged, thanks 8c87253824f306bfe8d11028d88ab885818c2eff

PR #275, opened 2015-09-15
Return remote peer address in TcpListener::accept

### LINKED ISSUE #258: TcpListener.accept() should return SocketAddr as well

alexcrichton:
This brings the API more in line with the `std::net::TcpListener`'s signature by
returning a pair of both the accepted connection and the remote socket address
if successful.

Closes #258

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * `EventLoopConfig` is now a builder instead of having public struct fields. It
   is also no longer `Copy`. (#259)
 * `TcpSocket` is no longer exported in the public API (#262)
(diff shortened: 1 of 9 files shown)


tailhook:
Well, I used to think that TryAccept should work for both Unix sockets and TCP sockets. I think this breaks it because SocketAddr doesn't work for Unix sockets, right?

carllerche:
This is true. It would be possible for `TcpListener::accept` to vary from `TryAccept::accept` in this case.

alexcrichton:
Updated to not change the signature on the `TryAccept` implementation

carllerche:
:+1:

PR #276, opened 2015-09-15
kqueue: fix bug where an error on registration prevented the changelist from getting flushed.

dwrensha:
The new test case fails in the current implementation.


carllerche:
Thanks! Merged d249d850b5b6957f7d6ad7fcb541e55e3063b897

PR #280, opened 2015-09-20
mio: don't wake up every second/timer tick when run() is called

luca-barbieri:
Currently run() results in the process waking up every second or every timer
tick if set lower because a timeout is passed to the OS wait function.

This changes the code so that run_once() still uses the timeout, but
run() will instruct the OS to wait forever when there are no timers.

The notify mechanism already uses a pipe for awakening the event loop,
so there should be no need to use a timeout for that.

This saves CPU time and battery, and is generally the correct behavior.

Miscellaneous notes:
- We also fix epoll because the maximum timeout supported by the Linux kernel is i32::MAX, not isize::MAX.
- We fix next_tick_ms() being truncated from u64 to usize
- It requires carllerche/nix-rust#192 to allow specifying no timeout to kevent()
- The public API of Poll::poll is changed, but that can be avoided if desired.

The choice of usize-types milliseconds for the timeout is not ideal, but should be good enough for now: epoll_wait() doesn't support any higher resolution, so kernel high-resolution timers would be required for better resolution on Linux and a dedicated API might be needed anyway for those.

NOTE: this introduces a next_timer_ms() function that currently uses the next tick if any timer is present. This is obviously not good and instead it should return the time until the first timer, but I'm not sure if the current timer implementation copes with that. The timer implementation itself seems to have serious problems anyway like no handling for times far in the future and no way to get the first timer in O(1).

Only lightly tested, please review.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -7,8 +7,6 @@ use std::{io, fmt, thread, usize};
 /// Configure EventLoop runtime details
 #[derive(Clone, Debug)]
 pub struct EventLoopConfig {
-    io_poll_timeout_ms: Option<usize>,
-
     // == Notifications ==
     notify_capacity: usize,
     messages_per_tick: usize,
@@ -24,27 +22,12 @@ impl EventLoopConfig {
     /// specified.
     pub fn new() -> EventLoopConfig {
         EventLoopConfig {
-            io_poll_timeout_ms: None,
             notify_capacity: 4_096,
             messages_per_tick: 256,
             timer_tick_ms: 100,
             timer_wheel_size: 1_024,
             timer_capacity: 65_536,
         }
-    }
-
-    /// Sets the default amount of time that a thread will be blocked in I/O
-    /// before timing out.
-    ///
-    /// If the event loop receives no I/O events during the specified timeout
-    /// then it will use this timeout to return control back to the original
-    /// program and run one more tick of the event loop.
-    ///
-    /// The default value for this is None, which means that the event loop
-    /// will sleep until I/O, a notification or a timer expires
-    pub fn io_poll_timeout_ms(&mut self, timeout: Option<usize>) -> &mut Self {
-        self.io_poll_timeout_ms = timeout;
-        self
     }
 
     /// Sets the maximum number of messages that can be buffered on the event
@@ -89,6 +72,7 @@ pub struct EventLoop<H: Handler> {
     timer: Timer<H::Timeout>,
     notify: Notify<H::Message>,
     config: EventLoopConfig,
+    pub tick_ms: Option<usize>
 }
 
 // Token used to represent notifications
@@ -127,6 +111,7 @@ impl<H: Handler> EventLoop<H> {
             timer: timer,
             notify: notify,
             config: config,
+            tick_ms: None
         })
     }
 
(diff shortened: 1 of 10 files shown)


dwrensha:
As an alternative to these proposed changes, would it make sense to change `EventLoopConfig::io_poll_timout_ms` to be an `Option<usize>` rather than just a `usize`, as it is now? Then you could specify "no timeout" as a configuration option when you construct an event loop.

luca-barbieri:
@dwrensha I thought that the timeout was not observable in the run() case, but there is in fact a tick() callback that is called unconditionally, so indeed it needs to be done as you say.

I pushed a new version that does that. It incompatibly changes both the public API and the default tick, with the rationale that a finite tick should not be used if possible and if it's used the application needs to pick the duration anyway.

carllerche:
Sorry for the delay. I pushed the nix change to master. Would you mind changing `Cargo.toml` to point to the correct nix rev? If you don't get around to it, I can do it as well.

luca-barbieri:
I changed it and rebased.

Note that while this prevents waking up when there are no timers, it still wakes up every timer tick when there are timers and more work would be required to fix that, which might involve some design decisions on how to evolve the timer code (in particular whether to adopt a cascading timer wheel so that buckets don't have timers for multiple wheel periods).

src/event_loop.rs
```diff
@@ -232,6 +218,21 @@ impl<H: Handler> EventLoop<H> {
...
 
+    pub fn tick_ms(&self) -> Option<usize> {
+        self.tick_ms
+    }
+
+    pub fn set_tick_ms(&mut self, tick_ms: Option<usize>) {
+        self.tick_ms = tick_ms
+    }
+
+    pub fn shorten_tick_ms(&mut self, tick_ms: usize) {
```
carllerche: Could you explain why this fn is needed?

src/event_loop.rs
```diff
@@ -33,19 +30,6 @@ impl EventLoopConfig {
...
-    ///
-    /// If the event loop receives no I/O events during the specified timeout
-    /// then it will use this timeout to return control back to the original
-    /// program and run one more tick of the event loop.
-    ///
-    /// The default value for this is 1000.
-    pub fn io_poll_timeout_ms(&mut self, timeout: usize) -> &mut Self {
-        self.io_poll_timeout_ms = timeout;
-        self
-    }
```
alexcrichton: Perhaps the config could provide the ability to customize the `tick_ms` option as well?

carllerche:
So, out of curiosity. Is there value in having a tick len vs. just pausing indefinitely?

dwrensha:
> So, out of curiosity. Is there value in having a tick len vs. just pausing indefinitely?

Now that I think about it, I can't come up with any good reason why the tick length should be configurable. I think pausing indefinitely is the right thing to do.

carllerche:
@dwrensha can confirm.

carllerche:
I'm going to close this in favor of #314 (based on this code, thanks!).

Regarding the timer included with Mio, it is explicitly a very coarse grained timer. The trade off is *very* efficient timeout registration & cancelation. The implementation is very simple to achieve those goals. There is a little bit of overhead due to the naive wakeup every 100ms & walk until the next tick. However, those are reasonable tradeoffs given the benefits IMO.

PR #285, opened 2015-09-24
test/test_*echo_server.rs: Handle Ok(None)

geofft:
The Getting Started guide admonishes us, "Even if we just received a ready notification, there is no guarantee that a read from the socket will succeed and not return `Ok(None)` so we must handle that case as well." Fix these two test cases to heed that advice and reregister the socket instead of panicking.


carllerche:
Thanks! e3ac615bc4dce781a73d58345e934c8d046eb41a

PR #291, opened 2015-10-08
Support sending/receiving FDs over UNIX sockets

geofft:
Add `UnixStream::read_recv_fd` and `write_send_fd`, which can read and write a `RawFd` that is passed over the socket. The file descriptor is duplicated as if with `dup()`, except that it can be passed across processes.

Also add `try_read_recv_fd`, `try_read_buf_recv_fd`, `try_write_send_fd`, and `try_write_buf_send_fd`, analogous to the functions on the `TryRead` and `TryWrite` traits.

Since this functionality is unique to UNIX domain sockets, the new functions are directly on `UnixStream` and not part of a trait.

---

Depends on a nix with carllerche/nix-rust#197 (cmsg, merged earlier today) and carllerche/nix-rust#198 (fix sockaddr_un, just proposed for merge). Also I have a tiny refactoring of `to_non_block` which captures a common pattern in a more readable way IMO, but if you don't like it I can drop it.

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -55,17 +55,13 @@ pub trait TryWrite {
 
 impl<T: Read> TryRead for T {
     fn try_read(&mut self, dst: &mut [u8]) -> Result<Option<usize>> {
-        self.read(dst)
-            .map(|cnt| Some(cnt))
-            .or_else(to_non_block)
+        self.read(dst).map_non_block()
     }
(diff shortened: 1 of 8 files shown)


carllerche:
This was merged! f4b3c3570c54472ff79127d1290f4762c1849f99

PR #293, opened 2015-10-13
Add module 'fd' that exposes file handle operations

dflemstr:
I need to interact with basic file handles such as `stdin` and `stdout` and I didn't find a way of doing that with the existing `0.5.x` API.  This PR adds support for the basic file handles and also an entry point for wrapping arbitrary file descriptors.

### DIFF
--- a/src/fd.rs
+++ b/src/fd.rs
@@ -1,0 +1,53 @@
+use {io, Evented, EventSet, Io, PollOpt, Selector, Token};
+use std::io::{Read, Write};
+use std::os::unix::io::RawFd;
+
+pub struct FileStream(Io);
+
+pub fn stdin() -> FileStream {
+    FileStream(Io::from_raw_fd(0))
+}
+
+pub fn stdout() -> FileStream {
+    FileStream(Io::from_raw_fd(1))
+}
+
+pub fn stderr() -> FileStream {
+    FileStream(Io::from_raw_fd(2))
+}
+
+impl FileStream {
+    pub unsafe fn from_raw_fd(fd: RawFd) -> Self {
+        FileStream(Io::from_raw_fd(fd))
+    }
+}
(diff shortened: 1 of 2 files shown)


dflemstr:
The tests are failing on `nightly` for reasons unrelated to this PR, I hope that's okay! :-)

dimbleby:
Overlaps with `EventedFd` somewhat.

dflemstr:
@dimbleby maybe we should rather expose `EventedFd` and constructors for the standard file descriptors? I had the impression  that `EventedFd` is not really part of the cross platform since it's exported in an ad hoc manner and Unix specific; it would be nice to have something that's eventually supported on all platforms.

dimbleby:
`FileStream` also looks pretty platform-specific: per its use of `std::os::unix::io::RawFd` and the failing AppVeyor build.

But realIy I just wanted to point out the overlap.  I guess I do think that if mio ends up with both `EventedFd` and the proposed `FileStream` then that would look odd, but I don't have any particular view on how best to rationalise this.

dflemstr:
I will of course try to fix the builds on other platforms, but before it would be good to know what the preferred action is: keeping this interface and maybe also supporting it on Windows, or extending `EventedFd` so that it also supports `Read` and `Write`?

dflemstr:
I have abandoned this approach and gone back to using `Io` in `mio` version `0.4.x`, but I will leave the PR open for `0.5.x`.

carllerche:
Thanks for the PR!

The implementation (using `Io`) would also imply that when the handle goes out of scope, it will get closed.

It is pretty *nix specific which would require some thought. I think that I will stick with the `EventedFd` bit for now.

If you think there is a different strategy or still think that this is worth while, I would open an issue for discussion.

PR #294, opened 2015-10-18
Coerce kqueue filter to always listen for hup

### LINKED ISSUE #240: only EventSet::hup() doesn't generate events on OSX
I wrote this test recently and it passed on Linux but deadlocked on OSX. It looks like the kqueue implementation doesn't like listening with just `EventSet::hup()`?

```rust
extern crate mio;

use mio::{EventLoop, Handler, Token, EventSet, PollOpt};
use mio::tcp::{TcpListener, TcpStream};

fn main() {
    struct H { listener: TcpListener }

    impl Handler for H {
        type Timeout = ();
        type Message = ();

        fn ready(&mut self, event_loop: &mut EventLoop<Self>, token: Token,
                 _events: EventSet) {
            if token == Token(1) {
                let s = self.listener.accept().unwrap().unwrap();
                event_loop.register_opt(&s, Token(3), EventSet::all(),
                                        PollOpt::edge()).unwrap();
                drop(s);
            } else if token == Token(2) {
                event_loop.shutdown();
            }
        }
    }

    let mut e = EventLoop::new().unwrap();
    let l = TcpListener::bind(&"127.0.0.1:0".parse().unwrap()).unwrap();
    let s = TcpStream::connect(&l.local_addr().unwrap()).unwrap();

    e.register_opt(&l, Token(1), EventSet::readable(), PollOpt::edge()).unwrap();
    e.register_opt(&s, Token(2), EventSet::hup(), PollOpt::edge()).unwrap();

    let mut h = H { listener: l };
    e.run(&mut h).unwrap();
}
```

hjr3:
Changes the kqueue registration code to check if both the read and write
filters are disabled. If they are and we want to listen for hup, it
forces the read filter to be enabled.

Fixes #240

### DIFF
--- a/src/sys/unix/kqueue.rs
+++ b/src/sys/unix/kqueue.rs
@@ -38,7 +38,16 @@ impl Selector {
     pub fn register(&mut self, fd: RawFd, token: Token, interests: EventSet, opts: PollOpt) -> io::Result<()> {
         trace!("registering; token={:?}; interests={:?}", token, interests);
 
-        self.ev_register(fd, token.as_usize(), EventFilter::EVFILT_READ, interests.contains(EventSet::readable()), opts);
+        // coerce the read filter
+        // see https://github.com/carllerche/mio/issues/240
+        let enable_read = if !interests.contains(EventSet::writable())
+                        && interests.contains(EventSet::hup()) {
+            true
+        } else {
+            interests.contains(EventSet::readable())
+        };
+
+        self.ev_register(fd, token.as_usize(), EventFilter::EVFILT_READ, enable_read, opts);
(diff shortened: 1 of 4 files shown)


hjr3:
I am not sure if this the best solution. Alternatively, we could always force the read filter to be on and then check during the coalesce phase if there was interest in read.

arthurprs:
Doesn't this forces the event_loop to receive unwanted read events?

hjr3:
@arthurprs hmm it could. looks like i will have to make those changes to coalesce anyways.

carllerche:
I commented on #240. I think the correct solution will be to limit the cross platform API to be kqueue compatible in this case. I don't think that the ability to register `HUP` interest w/o readable is a strong requirement. If it is, it can be exposed as a linux specific extension.

Feel free to re-open if you disagree.

PR #295, opened 2015-10-19
epoll: Should also work with Android.

overminder:
Compiled the tests in rustc-1.4-beta with gcc targeting Android API 9.
All the tests (mio, test, smoke and tcp) passed in an ARM Android 4.2.2.


carllerche:
Thanks! Merged: c5088aeaae7e4be0f5690013a68467599995e929

Issue #298, opened 2015-11-04
Failed to build a project on Mac after 'cargo update'


workanator:
Hello,

I did `cargo update` in my project directory and a new version of `libc` was added.

```
me$ cargo update  
    Updating registry `https://github.com/rust-lang/crates.io-index`  
      Adding libc v0.2.0
```

Then I tried to build my project and got that result.

```
me$ cargo build  
 Downloading libc v0.2.0  
   Compiling libc v0.2.0  
   Compiling clock_ticks v0.0.5  
   Compiling winapi v0.1.23  
.../.cargo/registry/src/github.com-0a35038f75765ae4/clock_ticks-0.0.5/src/lib.rs:31:42: 31:60 error: import `mach_timebase_info` conflicts with value in this module [E0255]  
...//.cargo/registry/src/github.com-0a35038f75765ae4/clock_ticks-0.0.5/src/lib.rs:31     use libc::{timeval, timezone, c_int, mach_timebase_info};  
                                                                                                                                      ^~~~~~~~~~~~~~~~~~  
.../.cargo/registry/src/github.com-0a35038f75765ae4/clock_ticks-0.0.5/src/lib.rs:31:42: 31:60 help: run `rustc --explain E0255` to see a detailed explanation  
.../.cargo/registry/src/github.com-0a35038f75765ae4/clock_ticks-0.0.5/src/lib.rs:37:13: 37:75 note: conflicting value here  
..../.cargo/registry/src/github.com-0a35038f75765ae4/clock_ticks-0.0.5/src/lib.rs:37         pub fn mach_timebase_info(info: *mut mach_timebase_info) -> c_int;  
                                                                                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  
error: aborting due to previous error  
Build failed, waiting for other jobs to finish...  
Could not compile `clock_ticks`.
```
I used "*" as the mio version in my .toml but I believe it's of version 0.4.3. I suggest you should avoid using `clock_ticks` crate because it's deprecated now and switch to `time` crate.

Thank you.

workanator:
I tried to clone mio and replace `clock_ticks` with `time` and the compile. The compilation was successful but I cannot test now because my project is in deep development.

rrichardson:
Yea, we should switch back to the time crate (Why was this ever included?)

Even the clock_ticks repo agrees: https://github.com/tomaka/clock_ticks

[...]

PR #299, opened 2015-11-04
Use clock_ticks 0.1.0

dimbleby:
clock_ticks 0.0.5 had a wildcard dependency on libc, which is now pulling in 0.2.0, which breaks the build on Windows.

clock_ticks 0.1.0 specifies libc 0.1, which should fix things again.

I've taken the liberty of bumping the version number to 0.4.4 - it'd be great if you could publish this so we can all have fixed builds again.

Thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name          = "mio"
-version       = "0.4.3"
+version       = "0.4.4"
 license       = "MIT"
 authors       = ["Carl Lerche <me@carllerche.com>"]
 description   = "Lightweight non-blocking IO"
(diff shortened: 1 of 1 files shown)


dimbleby:
Hm - I see that the builds are all broken, in ways beyond the scope of this pull request:

- nightly builds are broken as per the fix you've made to master at 92ad1ae5.  Perhaps you'll want to apply that to this branch too?
- Rust 1.1 builds are broken because clock_ticks 0.1.0 is doing something that presumably wasn't supported at 1.1

If you're determined to support older versions of stable Rust then the clock_ticks thing puts you in a bind:
- older versions are broken per this pull request
- newer versions don't support older Rust

dimbleby:
And I've just seen #298, which says that you're planning on losing clock_ticks anyway.

I expect that'll make this pull request redundant then.

PR #306, opened 2015-11-22
NetBSD tweaks

polyfractal:
Two minor tweaks to get mio running under netbsd.  The conditional compilation for `ev_push` is because netbsd uses an i64 for `data`, instead of usize like the other OS's.

This depends on https://github.com/rust-lang-nursery/net2-rs/pull/20 and https://github.com/carllerche/nix-rust/pull/207 (and subsequently bumping the revision of nix-rust in `Cargo.toml` once/if merged)

Verified by cross-compiling the Getting Started ping_pong demo for a rumprun unikernel (which uses the netbsd drivers). There may well be other problems or something that I didn't tweak correctly, since the example only touches a small portion of the mio codebase...but it should at least provide a base to work from.

Obligatory, fun screenshot:

mio_rumprun_qemu

### DIFF
--- a/src/sys/unix/kqueue.rs
+++ b/src/sys/unix/kqueue.rs
@@ -77,6 +77,7 @@ impl Selector {
         self.ev_push(fd, token, filter, flags);
     }
 
+    #[cfg(not(target_os = "netbsd"))]
     fn ev_push(&mut self, fd: RawFd, token: usize, filter: EventFilter, flags: EventFlag) {
         self.changes.sys_events.push(
             KEvent {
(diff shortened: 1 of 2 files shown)


alexcrichton:
:confetti_ball: 

Nice!

carllerche:
Sorry for the delay. I'm going to look at this before I get v0.5 out!

carllerche:
Thanks! Merged: 99ee9f229bcefa7862775d3389e0516924d55d35

polyfractal:
Thanks @carllerche!

PR #309, opened 2015-11-26
Associate sockets with a `Selector`

### LINKED ISSUE #308: Consider linking Mio's net types w/ an EventLoop
**tl;dr** When a Mio network type is first registered with an `EventLoop` instance, permanently associate the socket with the `EventLoop`.

## Context

On *nix systems, `EventLoop` instances and FDs are fully decoupled. The same FD can be watched from any number of `EventLoop` instances on any number of threads. However, this isn't true on Windows.

On windows, IOCP handles get associated with threads and sockets get associated with IOCP handles. This means that, in order to get the full *nix behavior, a global IOCP handle has to be created and shared across all `EventLoop` instances.

## Proposal

Mio already provides it's own cross platform network types. I propose permanently linking each value with the first `EventLoop` that it gets registered with. This will follow the windows semantics. If an already linked network type is registered with a different `EventLoop`, the register call will return with an `Error`.

If a user of Mio wants to register the same FD with multiple `EventLoop` instances on *nix platforms, this is possible by registering with `EventedFd`.

carllerche:
Closes #308

Windows requires that sockets are used exclusively with a single IOCP handle.
Unix platforms do not have this requirement. In order to provide a cross
platform API, Mio's network types artificially impose this limitations on Unix
platforms.

If a user of Mio would like to use a single socket with multiple event loops
on Unix platforms, this is possible by using `std::net` types and the
`EventedFd` wrapper.

/cc @alexcrichton

### DIFF
--- a/src/sys/unix/tcp.rs
+++ b/src/sys/unix/tcp.rs
@@ -1,4 +1,5 @@
 use io::MapNonBlock;
+use std::cell::Cell;
 use std::io::{Read, Write};
 use std::net::{self, SocketAddr};
 use std::os::unix::io::{RawFd, FromRawFd, AsRawFd};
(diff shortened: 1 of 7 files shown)


src/sys/unix/tcp.rs
```diff
@@ -14,11 +15,13 @@ use sys::unix::eventedfd::EventedFd;
 #[derive(Debug)]
 pub struct TcpStream {
     inner: net::TcpStream,
+    selector_id: Cell<Option<usize>>,
```
alexcrichton: This has the consequence of making the types not `Sync`, which perhaps isn't desired? Concurrently reading/writing with a `TcpStream` may be useful in some cases.

alexcrichton:
lgtm

PR #310, opened 2015-11-26
Mark socket impl types on Windows as Send

### LINKED ISSUE #292: Failed to build on Windows
I am the author of `coio-rs`. I just tried to make `coio-rs` to be build under windows, but I just encountered this weird build error: https://ci.appveyor.com/project/zonyitoo/coio-rs

```
     Running `rustc examples\tcp-echo-server.rs --crate-name tcp_echo_server --crate-type bin -g --out-dir C:\projects\coio-rs\target\debug\examples --emit=dep-info,link -L dependency=C:\projects\coio-rs\target\debug -L dependency=C:\projects\coio-rs\target\debug\deps --extern mio=C:\projects\coio-rs\target\debug\deps\libmio-6fe12d503368ab8d.rlib --extern chrono=C:\projects\coio-rs\target\debug\deps\libchrono-5bb9dfc5de7b8e03.rlib --extern log=C:\projects\coio-rs\target\debug\deps\liblog-6e27e2da5513596d.rlib --extern clap=C:\projects\coio-rs\target\debug\deps\libclap-f423c1c9ad813993.rlib --extern context=C:\projects\coio-rs\target\debug\deps\libcontext-15e9a9895fb6d966.rlib --extern libc=C:\projects\coio-rs\target\debug\deps\liblibc-165b5479aec0459c.rlib --extern rand=C:\projects\coio-rs\target\debug\deps\librand-a5a71c849766362c.rlib --extern num_cpus=C:\projects\coio-rs\target\debug\deps\libnum_cpus-14d331afbbde5712.rlib --extern deque=C:\projects\coio-rs\target\debug\deps\libdeque-174fff4ca43959ba.rlib --extern env_logger=C:\projects\coio-rs\target\debug\deps\libenv_logger-bbd82f3547b4db71.rlib --extern coio=C:\projects\coio-rs\target\debug\libcoio.rlib -L native=C:\projects\coio-rs\target\debug\build\context-15e9a9895fb6d966\out -L native=C:\projects\coio-rs\target\debug\build\context-15e9a9895fb6d966\out`
examples\tcp-echo-server.rs:48:17: 48:33 error: the trait `core::marker::Sync` is not implemented for the type `core::cell::UnsafeCell<wio::overlapped::Overlapped>` [E0277]
examples\tcp-echo-server.rs:48                 Scheduler::spawn(move || {
                                               ^~~~~~~~~~~~~~~~
examples\tcp-echo-server.rs:48:17: 48:33 help: run `rustc --explain E0277` to see a detailed explanation
examples\tcp-echo-server.rs:48:17: 48:33 note: `core::cell::UnsafeCell<wio::overlapped::Overlapped>` cannot be shared between threads safely
examples\tcp-echo-server.rs:48:17: 48:33 note: required because it appears within the type `mio::sys::windows::selector::Overlapped`
examples\tcp-echo-server.rs:48:17: 48:33 note: required because it appears within the type `mio::sys::windows::tcp::StreamIo`
examples\tcp-echo-server.rs:48:17: 48:33 note: required because it appears within the type `mio::sys::windows::tcp::StreamImp`
examples\tcp-echo-server.rs:48:17: 48:33 note: required because it appears within the type `mio::sys::windows::tcp::TcpStream` 
examples\tcp-echo-server.rs:48:17: 48:33 note: required because it appears within the type `mio::net::tcp::TcpStream`
examples\tcp-echo-server.rs:48:17: 48:33 note: required because it appears within the type `coio::net::tcp::TcpStream`
examples\tcp-echo-server.rs:48:17: 48:33 note: required because it appears within the type `[closure@examples\tcp-echo-server.rs:48:34: 69:18 stream:coio::net::tcp::TcpStream, addr:std::net::addr::SocketAddr]`
examples\tcp-echo-server.rs:48:17: 48:33 note: required by `coio::scheduler::Scheduler::spawn`
```

Ref: https://github.com/alexcrichton/wio/issues/1

carllerche:
Closes #292

The implementations are thread safe however since they use `UnsafeCell`, they
need to explicitly implement `Send`.

I visually inspected the code to make sure that the types are sendable. I don't know if there is a good way to test it.

cc/ @alexcrichton


alexcrichton:
In theory I'd think these should be automatically inferred? I personally tend to prefer avoiding `unsafe impl Send` wherever possible, as the types are probably not `Send` for a good reason if they're not auto-tagged as such.

That being said having assertions that they're sendable seems fine

carllerche:
Well, they aren't `Send` because of the `UnsafeCell` used in `Overlapped` no?

alexcrichton:
Ah that may be true, yeah, although in that case the `unsafe impl Send` may be more appropriate on `Overlapped` rather than the streams as a whole as it's much more targeted and easy to verify

carllerche:
The reason why i didn't put it on that type was because that type itself isn't safe for Send. The implementation of each socket type and how it uses the Overlapped is what makes the socket type Send

test/tcp.rs
```diff
@@ -247,3 +247,15 @@ fn listen_then_close() {
     let mut h = H;
     e.run_once(&mut h).unwrap();
 }
+
+#[test]
+fn test_tcp_sockets_are_send() {
+    fn foo<T: Send>(t: T) {
```
reem: you could write this test without having to actually create TcpListener's and TcpStreams by just doing:

```rust
fn assert_send<T: Send>() {}
assert_send::<TcpListener>();
assert_send::<TcpStream>();
```

alexcrichton:
@carllerche hm that's a good point, although I would perhaps prefer to make that interface `unsafe` in that case while still allowing the type to be `Send`.

There's quite a bit of logic going on in `TcpStream` and I can see accidentally adding something non-`Sync` in the future which ends up getting papered over by the `unsafe` impl

carllerche:
>  I would perhaps prefer to make that interface `unsafe`

Not following this. Which interface?

alexcrichton:
Ah sorry to clarify I mean making all the methods on `Overlapped` unsafe while making the type itself `Send`. Then part of the "unsafe contract" of using the type correctly will be to ensure it's used in an appropriate fashion.

Issue #307, opened 2015-11-23
Define Mio's behavior


carllerche:
Exact mio behavior should be documented and consistent across platforms.

A few cases to document:
- [ ] A `TcpStream` is registered with an `EventLoop`, the `EventLoop` is closed, a new one is created and the socket is registered with the new one.
- [ ] A `TcpStream` is registered, then deregistered, then `registered` with a different `EventLoop`
- [ ] A socket is registered with multiple `EventLoop` values.
- [ ] what happens if you register or reregister a socket with `EventSet::none()`
- [ ] what happens if you close a socket without deregistering it
  - [ ] more events after the current tick finishes?
- [ ] Require readable or writable interest to be present in the `EventSet`
- [ ] Should `TryRead` & `TryWrite` take `&mut self` or `&self`?
- [ ] Should `Poll` be public and what should the API be. Can `IoEvent` be renamed to `Event`.
- [ ] What happens if a socket is registered with both edge & level (answer is probably: Error)
- [ ] What happens if you use the same token with multiple sockets.
  - [ ] look at `kqueue` implementation
  - [ ] Windows may also be a hazard here
- [ ] Deregistering a socket with an `EventLoop` that does not own it.
- [ ] Reregister before register #416
- [ ] Use `Duration` for timeouts?
- [ ] What happens when an IOCP handle is closed when there is active I/O?
- [ ] All types should be `Sync` and `Send`
  - [ ] use `AtomicUsize` on Unix with compare and swap + relaxed

reem:
Other things:
  - [ ] what happens if you register or reregister a socket with `EventSet::none()`
  - [ ] what happens if you close a socket without deregistering it

reem:
Also, given the recent changes for tracking the event loop a socket is registered on, do we care if mio's io objects are `Sync`?

carllerche:
We probably do want them to be `Sync`. It'll be worth revisiting.

carllerche:
Idea: type encode socket registration state: https://www.reddit.com/r/rust/comments/3vdz26/mio_v05_released_now_with_windows_support/cxn8p8t

reem:
Do we care about defining the behavior of mio under `fork`? I know this is a place where kqueue and epoll differ rather significantly.

danburkert:
Another thing:
* What happens when you read or write to a socket before the first readable or writable interest is fired?

[...]

PR #313, opened 2015-12-02
Implement level triggering for windows

### LINKED ISSUE #241: windows: Implement PollOpt::level()
Right now this polling option just returns an error if it's ever seen on Windows. This... isn't good! The `Selector` on Windows should probably manage what's readable and what's writable so after a level option comes in it knows to keep the timeout at 0ms and to keep returning I/O events for the readable/writable-ness.

I think the hard part here may be figuring out how to do this without having to iterate over all the I/O handles in the world? I haven't given this too much thought just yet, however.

carllerche:
The windows selector uses a linked list of currently active level-triggered events. When an event arrives and the socket is registered with level-triggered, the event is tracked in this linked list. Every call to select events will include all elements in this list. When an operation is performed on the socket such that the socket is no longer ready, the event is removed from the linked list.

Closes #241

## Remaining

- [x] Move calls to `unset_readiness` into `schedule_foo` functions.
- [x] Possibly combine `Registration::deregister` & `deregister2`
- [x] Remove `LinkedList`
- [x] UDP support
- [x] Uncomment prior tests using level

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -116,6 +116,7 @@ mod token;
 pub use event::{
     PollOpt,
     EventSet,
+    IoEvent,
 };
 pub use event_loop::{
     EventLoop,
@@ -157,7 +158,8 @@ pub use notify::{
     NotifyError,
 };
 pub use poll::{
-    Poll
+    Poll,
+    Events,
 };
 pub use timer::{
     Timeout,
--- a/src/notify.rs
+++ b/src/notify.rs
@@ -189,15 +189,16 @@ impl<M: Send> NotifyInner<M> {
 
 impl<M: Send> Evented for Notify<M> {
     fn register(&self, selector: &mut Selector, token: Token, interest: EventSet, opts: PollOpt) -> io::Result<()> {
+        assert!(opts.is_edge(), "awakener can only be registered using edge-triggered events");
         self.inner.awaken.register(selector, token, interest, opts)
     }
 
-    fn reregister(&self, selector: &mut Selector, token: Token, interest: EventSet, opts: PollOpt) -> io::Result<()> {
-        self.inner.awaken.reregister(selector, token, interest, opts)
+    fn reregister(&self, _: &mut Selector, _: Token, _: EventSet, _: PollOpt) -> io::Result<()> {
+        panic!("awakener is never reregistered");
     }
 
-    fn deregister(&self, selector: &mut Selector) -> io::Result<()> {
-        self.inner.awaken.deregister(selector)
+    fn deregister(&self, _: &mut Selector) -> io::Result<()> {
+        panic!("awakener is never deregistered");
     }
 }
 
--- a/src/sys/windows/selector.rs
+++ b/src/sys/windows/selector.rs
@@ -3,6 +3,7 @@ use std::io;
 use std::mem;
 use std::os::windows::prelude::*;
 use std::sync::{Arc, Mutex};
+use std::collections::hash_map::{Entry, HashMap};
 
 use slab::Index;
 use winapi::*;
@@ -29,7 +30,6 @@ struct SelectorInner {
     /// The actual completion port that's used to manage all I/O
     port: CompletionPort,
 
-
     /// A list of deferred events to be generated on the next call to `select`.
     ///
     /// Events can sometimes be generated without an associated I/O operation
@@ -42,13 +42,9 @@ struct SelectorInner {
     /// Primitives will take buffers from this pool to perform I/O operations,
     /// and once complete they'll be put back in.
     buffers: Mutex<BufferPool>,
-}
 
-pub struct Registration {
-    selector: Option<Arc<SelectorInner>>,
-    token: Token,
-    opts: PollOpt,
-    interest: EventSet,
+    /// A list of registered level triggered `IoEvent`s
+    level_triggered: Mutex<HashMap<usize, IoEvent>>,
 }
 
 impl Selector {
@@ -59,6 +55,7 @@ impl Selector {
                     port: cp,
                     defers: Mutex::new(Vec::new()),
                     buffers: Mutex::new(BufferPool::new(256)),
+                    level_triggered: Mutex::new(HashMap::new()),
                 }),
             }
         })
@@ -69,10 +66,10 @@ impl Selector {
                   timeout_ms: usize) -> io::Result<()> {
         // If we have some deferred events then we only want to poll for I/O
         // events, so clamp the timeout to 0 in that case.
-        let timeout = if self.inner.defers.lock().unwrap().len() > 0 {
-            0
-        } else {
+        let timeout = if self.should_block() {
             timeout_ms as u32
+        } else {
+            0
         };
 
         // Clear out the previous list of I/O events and get some more!
@@ -93,6 +90,7 @@ impl Selector {
         // then it means it was our awakener, so just generate a readable
         // notification for it and carry on.
         let dst = &mut events.events;
+
         for status in events.statuses[..n].iter_mut() {
             if status.overlapped() as usize == 0 {
                 dst.push(IoEvent::new(EventSet::readable(),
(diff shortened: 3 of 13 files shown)


src/lib.rs
```diff
@@ -116,6 +116,7 @@ mod token;
 pub use event::{
     PollOpt,
     EventSet,
+    IoEvent,
```
alexcrichton: How come this ended up being exported? Shouldn't this be an internal implementation detail?

src/sys/windows/selector.rs
```diff
@@ -166,16 +186,61 @@ impl Registration {
...
+            let event = IoEvent::new(set, self.token);
+
+            if self.opts.is_edge() {
+                events.push(event);
+
+                if self.opts.is_oneshot() {
+                    trace!("deregistering because of oneshot");
+                    self.interest = EventSet::none();
+                }
+            } else {
```
alexcrichton: I'm not quite sure if there are well defined semantics for both edge and level, but perhaps this should be `if opts.is_edge()` followed by `if opts.is_level()` instead of using `else` to make them disjoint?

src/sys/windows/tcp.rs
```diff
@@ -323,6 +328,7 @@ impl Read for TcpStream {
                 // Once the entire buffer is written we need to schedule the
                 // next read operation.
                 if cursor.position() as usize == cursor.get_ref().len() {
+                    me.iocp.unset_readiness(EventSet::readable());
```
alexcrichton: Perhaps these calls to `unset` can happen in the `schedule_foo` functions? I think that may be a bit more central of a location and it'll also handle the case where the operation completes immediately (e.g. the readiness doesn't need to be unset)

src/sys/windows/tcp.rs
```diff
@@ -412,9 +425,12 @@ impl Drop for TcpStream {
         // This is achieved by replacing our socket with an invalid one, so all
         // further operations will return an error (but no further operations
         // should be done anyway).
-        self.inner().socket = unsafe {
+        inner.socket = unsafe {
             net::TcpStream::from_raw_socket(INVALID_SOCKET)
         };
+
+        // Then run any finalization code including level notifications
+        inner.iocp.deregister2();
```
alexcrichton: Perhaps this could just be `let _ = inner.iocp.deregister()`? (instead of adding a new function)

src/sys/windows/selector.rs
```diff
@@ -42,13 +42,9 @@ struct SelectorInner {
...
     buffers: Mutex<BufferPool>,
-}
 
-pub struct Registration {
-    selector: Option<Arc<SelectorInner>>,
-    token: Token,
-    opts: PollOpt,
-    interest: EventSet,
+    /// A list of registered level triggered `IoEvent`s
+    level_triggered: Mutex<LinkedList<IoEvent>>,
```
alexcrichton: It might be interesting to move this mutex inside `LinkedList` eventually as in theory `list.get(entry)` shouldn't require much synchronization (or there could be atomic ops in play).

That being said I would want to think through that much more thoroughly and this seems totally fine for now (especially as `level` is less common than `edge`)

src/sys/windows/ll.rs
```diff
@@ -0,0 +1,219 @@
...
+            self.curr = &curr.next;
+            &curr.value
+        })
+    }
+}
+
+unsafe impl<T> Send for Entry<T> {}
+unsafe impl<T> Sync for Entry<T> {}
+
+type Link<T> = Option<Box<Node<T>>>;
```
alexcrichton: Man I always find linked lists somewhat difficult to read, we've got `Link`, `Rawlink`, `Node`, `Box`, ...

src/sys/windows/ll.rs
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    /// Convert the `Rawlink` into an Option value
+    ///
+    /// **unsafe** because:
+    ///
+    /// - Dereference of raw pointer.
+    /// - Returns reference of arbitrary lifetime.
+    unsafe fn resolve_mut<'a>(&mut self) -> Option<&'a mut T> {
+        mem::transmute(self.p)
```
alexcrichton: Perhaps this could use `&mut *self.p`, and oh dear this is unsafe...

src/sys/windows/ll.rs
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    fn ensure_same_ll(&self, entry: &Entry<T>) {
+        assert!(entry.ll == self as *const LinkedList<T> as *mut LinkedList<T>, "entry belongs to a different LinkedList");
+    }
+}
+
+unsafe impl<T> Send for LinkedList<T> {}
+
+pub struct Entry<T> {
```
alexcrichton: The lack of a lifetime connection between `Entry` and `LinkedList` makes me somewhat uneasy here, and the blind `unsafe impl<T> Send for LinkedList<T>` is also somewhat worrying (e.g. no bounds on `T`). I wonder if it would perhaps be better to just start out with a `HashMap` instead of a linked list? That should in theory still have `O(1)` lookup, addition, and removal (just with some higher constants out in front), but the perf for the level implementation could always be improved later on.

src/lib.rs
carllerche: `Poll` is a public API that provides polling on only sockets (no timer / notifications). It returns `IoEvent`. I'm confused how Mio compiled before without having this exported.

However, I will track this as part of the api discussion

src/sys/windows/ll.rs
carllerche: Technically, `Entry` is only used as a token supplied to `LinkedList` for ops and `LinkedList` does a verification step to ensure that `Entry` is correct... however thinking a bit more about it, there is an edge case... I can switch to a `HashMap` for now and punt.

src/sys/windows/selector.rs
carllerche: The options are disjoint in this case. You should not be able to register a socket as both edge & level. I will track this as part of #307.

src/sys/windows/tcp.rs
carllerche: I don't think that this is possible. I am renaming the functions though. One function has to check that the supplied `Selector` matches the one associated with the socket, the other function just does the deregistering.

carllerche:
@alexcrichton I have made the changes you suggest. Look good? TCP support should be done, but I still have some more outstanding items (tracked in the PR description).

src/sys/windows/selector.rs
alexcrichton: Ah ok cool, sounds good to me

src/sys/windows/tcp.rs
```diff
@@ -494,6 +518,9 @@ impl ListenerImp {
             State::Empty => {}
             _ => return
         }
+
+        me.iocp.unset_readiness(EventSet::readable());
```
alexcrichton: I think this may want to be deferred until the actual `mem::forget` below. If the `me.iocp.defer` branch is taken the readiness should still be applicable.

alexcrichton: Oh wait scratch that, it looks like `defer` will end up calling `push_event` which ends up doing the right thing for level triggering, so this should be fine.

src/sys/windows/selector.rs
```diff
@@ -221,14 +286,36 @@ impl Registration {
...
+    }
+
+    pub fn checked_deregister(&mut self, selector: &Selector) -> io::Result<()> {
+        match self.selector {
+            Some(ref s) => {
+                let inner1: &SelectorInner = &*selector.inner;
+                let inner2: &SelectorInner = &*s;
+
+                if inner1 as *const SelectorInner != inner2 as *const SelectorInner {
+                    return Err(other("socket registered with other selector"));
```
alexcrichton: May be good to add a test for this as well? Although that can also probably be tracked on the issue as well.

carllerche: I tracked this on the Mio behavior issue.

src/notify.rs
```diff
@@ -189,15 +189,16 @@ impl<M: Send> NotifyInner<M> {
 
 impl<M: Send> Evented for Notify<M> {
     fn register(&self, selector: &mut Selector, token: Token, interest: EventSet, opts: PollOpt) -> io::Result<()> {
+        assert!(opts.is_edge(), "awakener can only be registered using edge-triggered events");
```
alexcrichton: :+1:

src/sys/windows/selector.rs
```diff
@@ -194,9 +257,15 @@ impl Registration {
         if self.selector.is_some() {
             return Err(other("socket already registered"))
         }
+
         try!(Registration::validate_opts(opts));
         try!(selector.inner.port.add_socket(self.token.as_usize(), socket));
         self.associate(selector, token);
+
+        if opts.is_level() {
+            self.key = Some(socket.as_raw_socket() as usize);
```
alexcrichton: Perhaps this could happen unconditionally? Just helps follow the code at least (either way shouldn't matter)

carllerche:
Ok, this is not a candidate for merging (I will squash before merging)

cc/ @alexcrichton

src/sys/windows/selector.rs
```diff
@@ -166,16 +186,59 @@ impl Registration {
...
+                    trace!("deregistering because of oneshot");
+                    self.interest = EventSet::none();
+                }
+            } else {
+                let selector = self.selector.as_ref()
+                    .expect("expected a selector");
+
+                let mut level = selector.level_triggered.lock().unwrap();
+
+                match level.entry(self.key.expect("expected registration key")) {
```
alexcrichton: You could also do this match via `or_insert(event)` and then or-in the kind, it at least avoids the `Entry` import (but works either way)

alexcrichton:
I think there may also want to be some writable level  handling for UdpSocket, but otherwise lgtm r+

src/lib.rs
reem: rustc's privacy tracking is not so thorough - if something is declared with `pub` rustc considers it public and allows it in exported signatures, it need not be actually exported at the crate level

PR #314, opened 2015-12-03
Reduce unnecessary poll wakeups

### LINKED ISSUE #289: Unnecessary polling
This code which connects to and reads from a unix socket at `/tmp/socket` contains nothing that would require periodic wakeups, but strace shows it passing a timeout of 199ms or so to repeated calls to `epoll_wait`:
```
extern crate mio;

use std::path::Path;
use mio::{EventLoop, Handler, Token, EventSet};

struct E;

impl Handler for E {
	type Timeout = ();
	type Message = ();
}

fn main() {
	let mut e = EventLoop::<E>::new().unwrap();

	let addr = Path::new("/tmp/socket");

	let server = mio::unix::UnixStream::connect(addr).unwrap();
	e.register(&server, Token(2), EventSet::all(), mio::PollOpt::edge()).unwrap();
	e.run(&mut E).unwrap();
}
```
(To test, run `rm -f /tmp/socket; ncat -l -U /tmp/socket` in a terminal first.)

It would be much better for battery life if mio passed the maximum possible timeout to epoll_wait based on registered timers and actual timeouts requested by operations. In general mio should get this right for users because avoiding polling is one of the primary reasons to use an event-loop lib.

carllerche:
Based on #280 by luca-barbieri

Currently run() results in the process waking up every second or every timer
tick if set lower because a timeout is passed to the OS wait function.

This changes the code so that run_once() still uses the timeout, but
run() will instruct the OS to wait forever when there are no timers.

The notify mechanism already uses a pipe for awakening the event loop,
so there should be no need to use a timeout for that.

This saves CPU time and battery, and is generally the correct behavior.

Miscellaneous notes:

 * We also fix epoll because the maximum timeout supported by the Linux kernel is
   i32::MAX, not isize::MAX.
 * We fix next_tick_ms() being truncated from u64 to usize
 * It requires carllerche/nix-rust#192 to allow specifying no timeout to kevent()
 * The public API of Poll::poll is changed, but that can be avoided if desired.

Closes #289

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -2,14 +2,12 @@ use {Handler, Evented, Poll, NotifyError, Token};
 use event::{IoEvent, EventSet, PollOpt};
 use notify::Notify;
 use timer::{Timer, Timeout, TimerResult};
-use std::{io, fmt, thread, usize};
+use std::{cmp, io, fmt, thread, usize};
 use std::default::Default;
 
 /// Configure EventLoop runtime details
 #[derive(Clone, Debug)]
 pub struct EventLoopConfig {
-    io_poll_timeout_ms: usize,
-
     // == Notifications ==
     notify_capacity: usize,
     messages_per_tick: usize,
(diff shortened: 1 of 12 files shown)


alexcrichton:
Nice! :yen:

src/event_loop.rs
```diff
@@ -274,7 +258,7 @@ impl<H: Handler> EventLoop<H> {
     /// Spin the event loop once, with a timeout of one second, and notify the
     /// handler if any of the registered handles become ready during that
     /// time.
-    pub fn run_once(&mut self, handler: &mut H) -> io::Result<()> {
+    pub fn run_once(&mut self, handler: &mut H, mut timeout_ms: Option<usize>) -> io::Result<()> {
```
reem: since `std::time::Duration` is stable, is there a reason not to use it?

carllerche: I will track this in the API ticket. To date Mio has used *_ms

Issue #316, opened 2015-12-05
`tick()` not being called after update to 0.5


dpc:

dpc:
My current test is wrong, and after fixing it, I'm having troubles reproducing the same situation as in my code. Investigating.

dpc:
Funny enough, the "wrong" test was exactly the problem.

In case of lack of events, `EventLoop::run()` will just run, as no timeout is being used.

dpc:
I've updated the PR according to previous irc chat.

lcy03406:
I don't think it's a problem. If you want to tick the loop periodically, you can register a timer on it.
But the name 'tick()' is somewhat misleading, consider rename it to 'turn()' or something?

dpc:
The workaround (use own loop calling run_once) worked for me.

PR #317, opened 2015-12-05
Introduce option for steady ticks.

dpc:
This test current hangs, attesting issue #316.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -12,6 +12,8 @@ pub struct EventLoopConfig {
     notify_capacity: usize,
     messages_per_tick: usize,
 
+    tick_ms: Option<usize>,
+
     // == Timer ==
     timer_tick_ms: u64,
     timer_wheel_size: usize,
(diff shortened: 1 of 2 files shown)


dpc:
I don't know why Windows test fails. Please help. :)

dpc:
Ping. :)

carllerche:
Sorry for the delay. I actually pulled this in as part of a larger refactor I was working on that hasn't landed yet. This will be in soon though.

carllerche:
Ah sorry I thought this was a different PR that. I don't think that I'm going to pull this. It would be fairly simple to get this functionality as Mio is currently. You would just have to is implement [your own `run` fn](https://github.com/carllerche/mio/blob/master/src/event_loop.rs#L254-L263) with the functionality that you want. It's really short right now.

dpc:
So you're saying I should call run_once() in loop in my own function? I guess that's OK for me. Thanks.

carllerche:
Exactly. It also has the advantage of you being able to add further logic around blocking besides steady tick.

PR #324, opened 2015-12-10
Relax the Send bound on notify messages.

reem:
With the relaxed bound, all types can be used as notify messages, but only if
the notify message type is Send will Senders be Send or Sync (and then notify
messages can actually travel between threads).

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -381,25 +381,23 @@ impl <H: Handler> Drop for EventLoop<H> {
 }
 
 /// Sends messages to the EventLoop from other threads.
-pub struct Sender<M: Send> {
+pub struct Sender<M> {
     notify: Notify<M>
 }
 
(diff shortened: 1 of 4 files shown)


reem:
The travis failure on rust 1.1 is interesting, I guess the dead_code lint changed since then. The appveyor failure looks unrelated.

Is there a reason travis is on 1.1 instead of 1.4?

carllerche:
The travis version is pinned to the oldest supported version of Rust. I believe that we are currently planning on changing it to Rust 1.3 for Mio 1.0

reem:
Any update here? The travis builds are now passing, the appveyor failure still looks unrelated.

carllerche:
Thanks! I will merge this.

carllerche:
Merged: 504b376cccf2d23c3ceb31a11b0a5bb12e3b719f

PR #326, opened 2015-12-10
Remove all warnings during cargo test.

reem:
Primarily replace uses of the deprecated ::std::thread::sleep_ms with the util
test function sleep_ms.

Also did some small cleanup of imports.

### DIFF
--- a/test/test.rs
+++ b/test/test.rs
@@ -54,7 +54,8 @@ mod ports {
     }
 }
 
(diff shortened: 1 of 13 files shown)


reem:
Urgh, Duration is stable in 1.4 but not in 1.1.

carllerche:
Closing this. I pulled in some of it and then tweaked it to get rid of warnings on 1.3. Thanks!

PR #335, opened 2015-12-26
Implement error::Error for TimerError

mnussbaum:
I'm currently trying to wrap Mio for a coroutine library, and this will help me simplify my error handling. I think it is also a plus for Mio to implement the standard error traits as widely as possible.

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -313,10 +313,6 @@ impl error::Error for TimerError {
     fn description(&self) -> &str {
         self.desc
     }
-
-    fn cause(&self) -> Option<&error::Error> {
-        None
-    }
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


src/timer.rs
```diff
@@ -303,6 +309,16 @@ impl TimerError {
     }
 }
 
+impl error::Error for TimerError {
+    fn description(&self) -> &str {
+        self.desc
+    }
+
+    fn cause(&self) -> Option<&error::Error> {
```
reem: This method has a default impl, so you don't have to re-implement it.

src/timer.rs
```diff
@@ -294,6 +294,12 @@ pub struct TimerError {
     desc: &'static str,
 }
 
+impl fmt::Display for TimerError {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        write!(fmt, "{}", self.desc)
```
reem: It would be nice to include the `TimerErrorKind` in this output as well, not just the description.

mnussbaum:
Thanks for the fast feedback! Removed the redundant `cause` and added the kind into the display.

mnussbaum:
Closing as I believe this was merged in cf8a9a2e5cd5f4c5359caf147ec175c254372b15

Issue #186, opened 2015-05-28
Expose API to set SO_REUSEADDR / SO_REUSEPORT


sdroege:
Maybe as part of bind()/bound(), a boolean allow_reuse parameter. See this note for some things to keep in mind when using these sockopts: https://git.gnome.org/browse/glib/tree/gio/gsocket.c#n1898

I could prepare a patch if such an API would make sense to you

carllerche:
Currently, `SO_REUSEADDR` (as defined by *nix) is exposed via `TcpSocket` (See [here](https://github.com/carllerche/mio/blob/master/src/net/tcp.rs#L60-L62)). `SO_REUSEADDR` is also set by default when calling `TcpListener::bind()`

I would say that `SO_REUSEPORT` could be exposed via `set_reuseport(val: bool)`. Thoughts?

Also, which types are you referring to?

sdroege:
See the link I provided for some comments about the two different sockopts :) What they actually mean is not platform independent, so GLib does different things on different platforms to try to get the closest behaviour to what you usually want and only provides an "allow_reuse" boolean.

sdroege:
I'm referring mostly to UDP sockets btw. It might also make sense for datagram Unix domain sockets, but not sure if it makes sure for stream sockets (TCP / stream Unix domain sockets)

rrichardson:
Also Reliable UDP and DTLS!  ;)  (a boy can dream)

sdroege:
Not sure it makes sense of R-UDP or DTLS, probably only for connection-less sockets :) Otherwise you have different sockets confusing each others connection state machinery

pythonesque:
> I would say that SO_REUSEPORT could be exposed via set_reuseport(val: bool). Thoughts?

This would be very useful.

bluejekyll:
I'll jump on the request for SO_REUSEPORT. This option will allow for process oriented zero-downtime restarts. Simplifies deployments and configuration reloads.

tailhook:
> See the link I provided for some comments about the two different sockopts :) What they actually mean is not platform independent, so GLib does different things on different platforms to try to get the closest behaviour to what you usually want and only provides an "allow_reuse" boolean.

It's strange. I believe Glib is wrong, at least on linux. SO_REUSEPORT changes semantics of the socket in a way that many (I believe most) application don't want.

> I'll jump on the request for SO_REUSEPORT. This option will allow for process oriented zero-downtime restarts.

[...]

PR #338, opened 2016-01-01
Added set_reuse_address() and as_raw_socket() APIs

lhecker:
Using the set_reuse_address() API that is being introduced in this commit for ~~TcpListener and~~ UdpSocket sets SO_REUSEADDR on all platforms.
Furthermore it also sets SO_REUSEPORT on BSD platforms for UdpSockets, to implement a near-cross-platform behaviour, since most other platforms treat SO_REUSEADDR as if both options would have been set on BSD.

----

IMHO This PR is a good solution for #186. Since mio should be a abstraction of the differences between all platforms, it's necessary to provide a `reuse_address` API that hides those differences too and the solution in this commit implements the biggest common ground between those.
If someone depends on some platform specific behavior I recommend the usage of `as_raw_fd()` and the newly introduced `as_raw_socket()` functions instead (combined with `cfg!()` and the usual C APIs).
Furthermore this PR adds `ws2_32-sys` as a dependency but I don't think that this poses a problem since `miow` also depends on it.

### DIFF
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -74,6 +74,10 @@ impl UdpSocket {
     pub fn set_multicast_time_to_live(&self, ttl: i32) -> io::Result<()> {
         self.sys.set_multicast_time_to_live(ttl)
     }
+
+    pub fn set_reuse_address(&self, reuse_address: bool) -> io::Result<()> {
+        self.sys.set_reuse_address(reuse_address)
+    }
 }
 
 impl Evented for UdpSocket {
--- a/src/sys/unix/udp.rs
+++ b/src/sys/unix/udp.rs
@@ -130,6 +130,26 @@ impl UdpSocket {
             .map_err(super::from_nix_error)
     }
 
+    pub fn set_reuse_address(&self, reuse_address: bool) -> io::Result<()> {
+        try!(nix::setsockopt(self.as_raw_fd(), nix::sockopt::ReuseAddr, &reuse_address)
+                 .map_err(super::from_nix_error));
+
+        if cfg!(any(target_os = "bitrig",
+                    target_os = "dragonfly",
+                    target_os = "freebsd",
+                    target_os = "ios",
+                    target_os = "macos",
+                    target_os = "netbsd",
+                    target_os = "openbsd")) {
+            // To implement portable behaviour for SO_REUSEADDR with UDP sockets
+            // we need to also set SO_REUSEPORT on BSD-based platforms.
+            try!(nix::setsockopt(self.as_raw_fd(), nix::sockopt::ReusePort, &reuse_address)
+                     .map_err(super::from_nix_error));
+        }
+
+        Ok(())
+    }
+
     fn associate_selector(&self, selector: &Selector) -> io::Result<()> {
         let selector_id = self.selector_id.get();
 
--- a/src/sys/windows/tcp.rs
+++ b/src/sys/windows/tcp.rs
@@ -508,6 +508,12 @@ impl TcpListener {
     }
 }
 
+impl AsRawSocket for TcpListener {
+    fn as_raw_socket(&self) -> RawSocket {
+        self.inner().socket.as_raw_socket()
+    }
+}
+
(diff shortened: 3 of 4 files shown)


lhecker:
### Note:

I force-pushed some changes just now to my branch. This **removes** the `TcpListener::set_reuse_address()` methods. Reason being that TcpListener only features a `TcpListener::bind()` method but not a `v4()` and `v6()` method (which `UdpSocket` features though). This makes modifying the SO_REUSEADDR value impossible (since that needs to happen before the socket is bound).
The addition of a enum that encapsulates a `Empty`, `Builder` and `Bound` state like windows' `UdpSocket` classes have, are necessary here in the future, but should be solved in a seperate issue IMHO. After this adding a `set_reuse_address()` method should be simple since the `TcpBuilder` already abstracts this away.

alexcrichton:
This is also possible through net2's [`UdpBuilder::reuse_address`](http://doc.rust-lang.org/net2-rs/net2/struct.UdpBuilder.html#method.reuse_address) FWIW

lhecker:
@alexcrichton I'm using `UdpBuilder::reuse_address` for the windows implementation, but the unix one doesn't use `UdpBuilder` at all, which is why I'm manually using `nix`. So... Is there something I've missed?

@carllerche Are there still some blockers left for this PR which I should fix? Otherwise it would be awesome if you could merge this, because having `SO_REUSEADDR` for UDP is IMHO quite important and I could _really_ need it right now... :smile:

lhecker:
Oh and I think I should also highlight something I've already mentioned in my second comment:
> `TcpListener::set_reuse_address()` methods [are not possible]. Reason being that TcpListener only features a TcpListener::bind() method but not a v4() and v6() method (which UdpSocket features though). This makes modifying the SO_REUSEADDR value impossible (since that needs to happen before the socket is bound).

alexcrichton:
Ah yes, right, the `UdpSocket` APIs haven't caught up to the TCP ones which are intended to mirror the standard library, so the vector for adding this option may not always be there once they mirror

lhecker:
How do you plan on providing `SO_REUSEADDR` options then? The only way this could work is if you would provide UdpBuilder as well as TcpBuilder structs, but none of them exist in mio yet.
I guess this is going to be a future expansion, huh?
If that's the case I could change the PR to reflect that, but I would need some instructions about how you guys would like it to look like... It would also be a much larger change than this one.

alexcrichton:
Ah the idea is that `net2` is used for custom creation of sockets which are then passed to mio

carllerche:
At this point, I think I agree w/ @alexcrichton. If we can keep socket configuration in line w/ the story for std, that would be ideal. Right now, you can use net2 to configure your socket and then use the socket with mio and that is good for v1.0 and we can see how things evolve with the std::net sync io types.

lhecker:
Do you have an example on how to convert a net2 socket to a mio one? I don't see how to do this…

Also: Will there ever be a Builder Pattern in mio (like the one in net2)? The built-in configurability for socket creation in mio is quite cumbersome compared to net2…

carllerche:
@alexcrichton do you have a quick snippet? I will definitely add this to the docs.

alexcrichton:
The basic idea is that `std::net` is how mio and net2 can communicate, so you'd use `net2`'s builder types, then you'd create a `std::net` type, and then it'd be handed back to mio. For example with TCP you'd use [`TcpListener::from_listener`](http://rustdoc.s3-website-us-east-1.amazonaws.com/mio/master/mio/tcp/struct.TcpListener.html#method.from_listener).

Unfortunately the UDP bindings in mio haven't seen the same update that the TCP ones have, so the same API isn't available there just yet. Instead you'll have to combine `IntoRawFd` and `FromRawFd` for now I believe.

lhecker:
Thanks for your example! I understand where you guys are coming from, especially since net2 is a Rust RFC...
I just hope that there will be some intermediate solution for now since UDP sockets without REUSEADDR is kinda useless. :sweat_smile:

PR #347, opened 2016-01-25
WIP: Fix oneshot on windows

carllerche:
*Work in progress - do not merge*

Oneshot on windows does not currently behave correctly when the socket being reregistered has unread data.

### DIFF
--- a/src/sys/windows/tcp.rs
+++ b/src/sys/windows/tcp.rs
@@ -189,6 +189,11 @@ impl StreamImp {
     fn schedule_read(&self, me: &mut StreamInner) {
         match me.read {
             State::Empty => {}
+            State::Ready(_) | State::Error(_) => {
+                let mut set = EventSet::readable();
+                me.iocp.defer(set);
+                return;
+            }
             _ => return,
         }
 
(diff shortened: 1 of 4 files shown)


src/sys/windows/tcp.rs
```diff
@@ -189,6 +189,11 @@ impl StreamImp {
     fn schedule_read(&self, me: &mut StreamInner) {
         match me.read {
             State::Empty => {}
+            State::Ready(_) | State::Error(_) => {
+                let mut set = EventSet::readable();
+                me.iocp.defer(set);
+                return;
+            }
             _ => return,
```
alexcrichton: Might be worth explicitly matching `Pending` here as it's the only other one.

alexcrichton:
lgtm

carllerche:
This is going to take some more (somewhat significant work).

The issue right now is that `reregister` unconditionally queues the token into the list of pending events, even if it as already been queued. A token should only be added to the queued list a single time.

In order to avoid scanning the current list of queued tokens before each push, a somewhat significant internal refactor will have to happen....

alexcrichton:
Hm, could you describe a little more in detail what's going on here? I'm afraid much of this is out of cache for me at least

carllerche:
Closing in favor of #356

PR #354, opened 2016-02-10
Getting started

ereichert:
Updated the example for master.  Rewrote large parts of the getting started guide and updated all the links.

Hope this helps you out.

Should also mention that I merged master into getting-started.


carllerche:

PR #356, opened 2016-02-13
Fix windows bugs

carllerche:
WIP. Created the PR for review.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -169,6 +169,11 @@ impl EventSet {
     }
 
     #[inline]
+    pub fn is_none(&self) -> bool {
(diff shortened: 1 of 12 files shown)


carllerche:
cc @alexcrichton

alexcrichton:
To be honest I don't really have an idea what's going on here. What are the bugs being fixed again?

I'm mostly pretty uneasy about this `Arc<UnsafeCell<EventInner>>` where it's sort of optionally locked here and there. There isn't really any documentation about what's going on there, when the locks are necessary, or what the access patterns are. Seems like it may be easy to mix up in later refactorings?

PR #364, opened 2016-02-25
impl IntoRawFd for types that impl AsRawFd

canndrew:


canndrew:
Fixes https://github.com/carllerche/mio/issues/363

canndrew:
Why are the travis builds using stable 1.3.0 instead of 1.6.0? `IntoRawFd` has been stable since 1.4.0

carllerche:
As of now, Mio currently officially supports 1.3.0. @alexcrichton do you have thoughts about changing this dep?

alexcrichton:
I've yet to hear of *anyone* specifically requesting support for an older version of Rust, so I would personally think it'd be fine to switch to just supporting stable/beta/nightly

carllerche:
I merged this here: b47873d2cca59a4410e5755e89b5d5bc9364c1d6. I also bumped the min supported Rust version to 1.6.

Thanks.

PR #366, opened 2016-02-25
OpenBSD make

polezaivsani:


polezaivsani:
I wonder if there's something wrong about these patches or time and patience alone would cut it?

carllerche:
Merged, thanks 60d54772f19199c8888876df1194b75eb72c6c53

PR #369, opened 2016-03-01
Fix Clippy warnings

mcarton:
Nothing really bad here.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -481,7 +481,7 @@ mod tests {
         let wcount = Arc::new(AtomicIsize::new(0));
         let mut handler = Funtimes::new(rcount.clone(), wcount.clone());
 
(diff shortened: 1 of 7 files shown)


mcarton:
Oh, why do you test against *nightly* and *1.3.0*? The current release is *1.6.0* (and you can just ask travis to test against *stable*, it will pick the last release).

carllerche:
Hi @mcarton, we test explicitly against Rust 1.3 because that is the oldest version of Rust officially supported by Mio. When we decide to stop supporting a version of Rust, we change it in travis :)

mcarton:
@carllerche ok, I fixed the thing that was not compatible with *1.3*, but you should still consider testing on both *1.3* and *stable* in that case :smile:.

carllerche:
Thanks! Merged: 7d4778f226e3f2479e802ccb0934e1e8db3b31a1

Issue #360, opened 2016-02-23
Proposal: Unify Sockets, Timers, and Channels


carllerche:
# Unify Sockets, Timers, and Channels

Currently, there are two runtime APIs: `Poll` and `EventLoop`. `Poll` is
the abstraction handling readiness for sockets and waiting for events on
these sockets. `EventLoop` is a wrapper around `Poll` providing a
timeout API and a cross thread notification API as well as the loop /
reactive (via `Handler`) abstraction.

I am proposing to extract the timer and the notification channel
features into standalone types that implement `Evented` and thus can be
used directly with `Poll`. For example:

```rust
let timer = mio::Timer::new();
let poll = mio::Poll::new();

poll.register(&timer, Token(0), EventSet::readable(), PollOpt::edge());

timer.timeout("hello", Duration::from_millis(1_000));

poll.poll();
```

The insight is that a lot of the code that is currently windows specific
is useful to all platforms. Stabilizing the API and providing it to all
platforms allows implementing `Evented` for arbitrary types.

## Advantages

- Mio would have a unified API.
- No change in overhead for any platform.
- Move (and improve) some of the currently windows-only code to all platforms.
- The notification channel backend could be configurable:

```rust
let (tx, rx) = mio::Channel::new(mpsc::channel());
poll.register(&rx, Token(0), EventSet::readable(), PollOpt::edge());
```

## Disadvantages

* Unsafe code
* More code (lock-free algorithms)

The primary disadvantage that I can think of is that the code path
around timers & the notification channel become slightly more
complicated. I don't believe that the change would have a meaningful
performance impact.

There is also additional code complexity for all platforms. However,
this code complexity already exists for Windows.

## Behavior

An `Evented` would mirror the behavior of a socket registered with
epoll. Specifically, in a single threaded environment:

* A value registered will trigger at most one notification per call to `poll`.
* A value registered with readable interest & edge triggers a notification once when it becomes readable.
* A value registered with readable interest & level triggers a notification every call to poll as long as the value is still readable.
* A value registered (or reregistered) with readable interest triggers a notification immediately if it is currently readable.
* If a value is registered with readable interest only and already has a pending writable notification, the event is discarded
* If a value has any pending notifications and is deregistered, the pending notifications are cleared.
* When a value is dropped, it will no longer trigger any further notifications.
* `Poll` is permitted to fire of spurious readiness events *except* if the value has been dropped.

In the presence of concurrency, specifically readiness being modified on
[...]

PR #371, opened 2016-03-03
WIP - Custom readiness queue

carllerche:
WIP, not ready for merging. This PR will be the first step towards #360


src/poll.rs
```diff
@@ -42,14 +48,133 @@ use std::time::Duration;
...
+
+    // The Token used to register the `Evented` with `Poll`. This can change,
+    // but only by calling `Poll` functions, so there will be no concurrency.
+    token: Token,
+
+    // Both interest and opts can be mutated
+    interest: Cell<EventSet>,
+
+    // Poll opts
+    opts: Cell<PollOpt>,
```
alexcrichton: I'm personally pretty nervous about the direction this abstraction is headed in. As is it seems like it's incredibly difficult to get right in terms of preventing race conditions and other undefined behavior, and it's barely leveraging Rust types for managing this unsafety. Specifically this structure is conglomeration of:

* Raw unsafe pointers
* non-thread-safe `Cell` types
* threadsafe atomics

This, combined with the `fn inner_mut(&self) -> &mut ReadinessQueueInner` signature below, make this very difficult to determine whether it's actually safe or not.

I think this would benefit quite a bit from extracting some of the unsafe components here to standalone and isolated features with safe interfaces, or very very clear unsafe interfaces with documentation as to what contracts need to be upheld. For example `next_all_notes` and `prev_all_nodes` looks a lot like a doubly linked list? Does the standard library's `LinkedList` not suffice here because atomics are necessary? Some other reason?

Similarly if there's a set of fields that are only accessed by a `Poll` on each node, then this may benefit from a more concrete interface to indicate such. For example I would expect something like:

```rust
struct PollFields { ... }

struct ReadinessNode {
    poll_fields: UnsafeCell<PollFields>,
}

impl ReadinessNode {
    fn poll_fields<'a>(&'a self, _proof: &'a Poll) -> &'a PollFields { /* ... */ }
}
```

Basically something like that where the API itself is very difficult to get wrong if properly used. That should help anyone else working around here or perhaps any future refactorings as well.

src/poll.rs
```diff
@@ -145,3 +270,339 @@ pub fn selector(poll: &Poll) -> &sys::Selector {
...
+        // If the node is currently ready, re-queue?
+        if !event::is_empty(self.readiness()) {
+            self.queue_for_processing(None);
+        }
+
+        Ok(())
+    }
+
+    pub fn readiness(&self) -> EventSet {
+        event::from_usize(self.node().events.load(Ordering::Relaxed))
```
alexcrichton: I would personally **very strongly recommend** using `SeqCst` everywhere. It's far easier to selectively relax orderings, and profiling can be used to figure out which ones are on the hot paths.

If everything is Relaxed or Acquire/Release it's incredibly difficult to reason about whether the orderings are actually correct. They're all basically related to global decisions, so you have to understand basically everything to reason about any local ordering as well. Thankfully if most orderings are `SeqCst` you can safely ignore most of the program as they'll work "as expected".

src/poll.rs
```diff
@@ -145,3 +270,339 @@ pub fn selector(poll: &Poll) -> &sys::Selector {
...
+        let queue = poll.readiness_queue.clone();
+        let node = queue.new_readiness_node(token, interest, opts);
+
+        Registration {
+            node: node,
+            queue: queue,
+        }
+    }
+
+    pub fn update(&self, poll: &Poll, interest: EventSet, opts: PollOpt) -> io::Result<()> {
```
alexcrichton: Since this `Registration` is already inherently tied to the original `Poll`, is there a benefit from taking it as an extra parameter here? It'd perhaps be nicer to not have to pass it?

src/poll.rs
```diff
@@ -145,3 +270,339 @@ pub fn selector(poll: &Poll) -> &sys::Selector {
...
+        if !self.queue.identical(&poll.readiness_queue) {
+            return Err(io::Error::new(io::ErrorKind::Other, "nope"));
+        }
+
+        self.node().interest.set(interest);
```
alexcrichton: This is kinda surprising to me in terms of what's going on. We're modifying a `Cell` which is explicitly documented as only accessible from `Poll`, which sounds like a data race waiting to happen?

This is sort of what I was alluding to above with where the unsafety here may be a little too much to reason about. It's very unclear which operations are unsafe and which aren't because the `unsafe` blocks are so wildly powerful in this module.

carllerche:
Looks like CI is green. I'm hoping to merge this!

src/channel.rs
```diff
@@ -0,0 +1,284 @@
...
+    (tx, rx)
+}
+
+impl SenderCtl {
+    /// Call to track that a message has been sent
+    pub fn inc(&self) -> io::Result<()> {
+        if 0 == self.inner.pending.fetch_add(1, Ordering::Acquire) {
+            // Toggle readiness to readable
+            if let Some(set_readiness) = self.inner.set_readiness.as_ref() {
+                // TODO: Don't ignore result
```
alexcrichton: This TODO is handled, right?

src/channel.rs
```diff
@@ -0,0 +1,284 @@
...
+    let rx = ReceiverCtl {
+        registration: Lazy::new(),
+        inner: inner,
+    };
+
+    (tx, rx)
+}
+
+impl SenderCtl {
+    /// Call to track that a message has been sent
```
alexcrichton: Maybe worth documenting what can be done in the case of an error? E.g. the readiness is not set, but the increment still happened

src/channel.rs
```diff
@@ -0,0 +1,284 @@
...
+        match *self {
+            StdSender::Bounded(ref v) => StdSender::Bounded(v.clone()),
+            StdSender::Unbounded(ref v) => StdSender::Unbounded(v.clone()),
+        }
+    }
+}
+
+impl<T> Receiver<T> {
+    pub fn try_recv(&self) -> Result<T, mpsc::TryRecvError> {
+        // TODO: When there is no value to recv, unset readiness
```
alexcrichton: This is done through `dec`, right?

src/lazy.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+impl<T> Lazy<T> {
+    pub fn new() -> Lazy<T> {
+        Lazy { val: UnsafeCell::new(None) }
+    }
+
+    pub fn is_some(&self) -> bool {
+        self.as_ref().is_some()
+    }
```
alexcrichton: I don't think this is a safe method to expose because concurrent calls to `Lazy::is_some` and `Lazy::set` is a data race

alexcrichton: er nvmd, I'm getting confused between these two

alexcrichton: er actually, this `Lazy` type seems like it's not used?

carllerche: `Lazy` is used here https://github.com/carllerche/mio/blob/custom-readiness-queue/src/channel.rs#L96

src/channel.rs
carllerche: Yes, `dec` will unset readiness. I think I was suggesting something else... I also think the TODO here is out of date. I will remove.

rrichardson:
LGTM

src/channel.rs
```diff
@@ -0,0 +1,282 @@
...
+            tx: self.tx.clone(),
+            ctl: self.ctl.clone(),
+        }
+    }
+}
+
+impl<T> StdSender<T> {
+    pub fn send(&self, t: T) -> Result<(), SendError<T>> {
+        match *self {
+            StdSender::Bounded(ref tx) => tx.send(t).map_err(SendError::from),
```
hjr3: Shouldn't this be using `try_send`?

PR #382, opened 2016-04-24
Separate Windows and Unix deps | remove dependency on deprecated time crate

rrichardson:
This might be a bit controversial wrt the separation of unix and windows deps. I'm not entirely sure what the intended behaviors are supposed to be.  But the net effect of this commit is that when you compile on Linux, it doesn't pull down or compile any of the windows based crates, this speeds things up and avoids potential snafus.

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -1,12 +1,13 @@
 use token::Token;
-use time::precise_time_ns;
+use std::time::{Instant, Duration};
 use std::{error, fmt, usize, iter};
-use std::cmp::max;
 
 use self::TimerErrorKind::TimerOverflow;
-
 const EMPTY: Token = Token(usize::MAX);
-const NS_PER_MS: u64 = 1_000_000;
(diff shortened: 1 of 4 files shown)


rrichardson:
To fully realize the dependency splitting, this also relies on https://github.com/rust-lang-nursery/net2-rs/pull/33

src/timer.rs
```diff
@@ -158,7 +156,7 @@ impl<T> Timer<T> {
         // Update the head slot
         self.wheel[slot] = token;
 
-        trace!("inserted timout; slot={}; token={:?}", slot, token);
+        println!("inserted timout; slot={}; token={:?}", slot, token);
```
hjr3: Did you intend to convert all the `trace!` macros to `println!`?

rrichardson: Crap, I forgot to change those back. Nice catch.

carllerche:
Merged. Thanks.

PR #389, opened 2016-05-14
Document deregister behavior

### LINKED ISSUE #351: EventLoop::deregister(TcpStream) falls to 'No such file or directory (os error 2)'
Hello,  
after a TCP socket is shutdown by remote host call to `EventLoop::deregister(TcpStream)` fails with error *No such file or directory (os error 2)*. MIO v0.4.x lacks of that behavior so I was able to deregister active or shutdown sockets without getting the error.  
In my project sockets can be closed by my application or by remote hosts and I need to deregister them properly. Is there a way I can test if socket is active before call to `EventLoop::deregister`?  
I run on Mac OS X El Captain, Rust 1.6, branch `master`.

hjr3:
- explain that closing the socket is usually good enough
- warn against using deregister with kqueue edge+oneshot

Fixes #351

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -265,6 +265,13 @@ impl<H: Handler> EventLoop<H> {
     }
 
     /// Deregisters an IO handle with the event loop.
+    ///
+    /// Both kqueue and epoll will automatically clear any pending events when closing a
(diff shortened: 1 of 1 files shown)


carllerche:
Thanks, merged here 34ac135025acb95c62516aa91148ffe64c86ce0c

PR #396, opened 2016-05-28
Update dependencies

alexbool:


carllerche:
Hi, thanks for submitting the PR. I think that the current versioning strategy works for Mio. It's important to note that Mio is not pinned to exact versions of dependencies, but is allowed to pull in any newer version than specified that satisfies semver.

PR #407, opened 2016-06-13
fix/windows-crash: fix crash due to unwrap on Windows

ustulation:
We run into cases every now and then when the remote address returned by miow crate is None. This usually happens during severing of connection and hence we are no longer interested in the remote address anyway. However mio crate tries to unwrap the returned option crashing in several of our usages (esp. when stressed). This fix aims to prevent that crash.

### DIFF
--- a/src/sys/unix/tcp.rs
+++ b/src/sys/unix/tcp.rs
@@ -5,11 +5,11 @@ use std::net::{self, SocketAddr};
 use std::os::unix::io::{RawFd, FromRawFd, AsRawFd};
 
 use libc;
-use net2::{TcpStreamExt, TcpListenerExt};
+use net2::{TcpStreamExt};
 use nix::fcntl::FcntlArg::F_SETFL;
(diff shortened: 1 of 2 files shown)


ustulation:
eh, this was a windows only fix and travis fails for nightly - the base branch seems to have trouble with with nightly travis build already ?

carllerche:
Yeah :( I'm trying to get this fixed, but it seems like a bunch of dependencies are having trouble.

alexcrichton:
Looks like this was merged in https://github.com/carllerche/mio/commit/d1e3a78a51265291101353b91ddf59d1e1fb70b9, so closing.

ustulation:
Ah yes i had raised to master and but it was required to raise a similar one for back-port according to https://github.com/carllerche/mio/issues/406#issuecomment-225561974 hence this PR. But i think things are all gearing up for 0.6.0 so this PR isn't required anymore.

PR #414, opened 2016-06-24
deps: use nix 0.6.0 from crates.io

posborne:
After a bit of a drought, a new release of nix has been published
to crates.io.  Minor changes were required in order to get things
working again related to some small API changes to properly model
flags to calls like `recvmsg`.

Signed-off-by: Paul Osborne <osbpau@gmail.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name          = "mio"
-version       = "0.5.0"
+version       = "0.5.2-dev"
 license       = "MIT"
 authors       = ["Carl Lerche <me@carllerche.com>"]
 description   = "Lightweight non-blocking IO"
@@ -19,7 +19,7 @@ exclude       = [
 
 [dependencies]
 log    = "0.3.1"
-nix    = { git = "https://github.com/carllerche/nix-rust", rev = "c4257f8a76" }
+nix    = "0.6.0"
 libc   = "0.2.4"
 slab   = "0.1.0"
 time   = "0.1.33"
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ rust:
   - nightly
   - stable
   # The following Rust version represents the oldest supported version of Mio
-  - 1.6.0
+  - 1.8.0
 
 os:
   - linux
--- a/src/sys/unix/uds.rs
+++ b/src/sys/unix/uds.rs
@@ -4,6 +4,7 @@ use sys::unix::{net, nix, Socket};
 use std::io::{Read, Write};
 use std::path::Path;
 use std::os::unix::io::{RawFd, IntoRawFd, AsRawFd, FromRawFd};
+use nix::sys::socket::MsgFlags;
 pub use nix::sys::socket::Shutdown;
 
 #[derive(Debug)]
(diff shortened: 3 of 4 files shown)


posborne:
`test test_timer::test_edge_without_polling` seems to fail intermittent.  Has anyone seen this on other branches or could this be specific to the nix change?

posborne:
> test test_timer::test_edge_without_polling seems to fail intermittent. Has anyone seen this on other branches or could this be specific to the nix change?

I looked into this and have included a fix although it is not directly related to the subject matter of this PR.  If you would like I can split.  Read the commit comments for more details: https://github.com/carllerche/mio/pull/414/commits/89489d07d39d474d3a112c773a60d6817d0560dd

posborne:
@alexcrichton @carllerche It looks like I might be bumping into https://github.com/rust-lang/cargo/issues/2090 with 1.6.0 (or at least something with similar symptoms).  Suggestions on how to fix?  Another option would be to move to only supporting rustc 1.8.0+ (which also solves net2 compat issues); I don't love that idea as I am currently using 1.7.0 (via https://github.com/meta-rust/meta-rust).

Cargo.toml
```diff
@@ -19,14 +19,14 @@ exclude       = [
...
-nix    = { git = "https://github.com/carllerche/nix-rust", rev = "c4257f8a76" }
+nix    = "0.6.0"
 libc   = "0.2.4"
 slab   = "0.1.0"
 time   = "0.1.33"
 bytes  = "0.3.0"
 winapi = "0.2.1"
 miow   = "0.1.1"
-net2   = { version = "0.2.19", default-features = false }
+net2   = { version = ">=0.2.19,<=0.2.23", default-features = false }
```
alexcrichton: I think this is why historical Cargo is hanging, and in general this is not idiomatic in Cargo and can cause lots of resolution problems. How come 0.2.24+ is undesirable?

posborne: 0.2.24 uses the new `cfg` syntax in Cargo.toml which is not compatible with the version of Cargo shipping with rust prior to 1.8.0.

alexcrichton: Is that just the " vs the ' in Cargo.toml? Older Cargos should just ignore the `cfg` business and there should be sections for compatibility there otherwise.

posborne: > Is that just the " vs the ' in Cargo.toml? Older Cargos should just ignore the cfg business and there should be sections for compatibility there otherwise.

Here's the failure without limiting to preventing going to .24: https://travis-ci.org/carllerche/mio/builds/140069153.  Are you saying that 1.6.0 should be able to parse the `cfg` if the syntax in net2 was different (using double quotes)?  If those blocks were ignored, I do not see how net2 would have all required dependencies: https://github.com/rust-lang-nursery/net2-rs/blob/master/Cargo.toml

alexcrichton: Ah yeah if the single quotes are changed to double quotes it'll parse successfully and Cargo will ignore it basically. We can add back in a few explicit sections (e.g. `target.${target_name}.dependencies`) for necessary platforms. All that being said though the case for older compatibility is pretty thin, so it may not be worth it.

In any case though I would not recommend merging this PR until this dependency ceiling is lifted as it's very likely to wreak havoc in the large ecosystem that depends on mio

carllerche: For better or for worse, the version of Cargo included w/ each version of Rust also needs to be supported on the same semver point release. If you break compatibility w/ Cargo you need to bump the semver..

I think net2 was fixed though

carllerche:
This is in now.. Thanks

PR #417, opened 2016-06-27
Clarify the meaning of None in run_once

aidanhs:


mayhewj:
The comment inside the function should probably be updated too.

carllerche:
Thanks. Merged here: 49b8e3f0fc12ddd7f95d68cfce8c70dc3effa804

PR #422, opened 2016-07-06
Use AtomicUsize to get TCP/UDP streams Sync

alexcrichton:
Also add impls of Read/Write for &TcpStream and such to ensure the methods an be
called concurrently as well.


carllerche:
Merged into `dev` 5c651b1d4ecbcbbd2c946b92f1ba8eb291f03efb

PR #425, opened 2016-07-09
Extract Events from Poll

alexcrichton:
This will eventually allow one Poll to be used from multiple threads because the
methods now only all need `&self` rather than `&mut self`


carllerche:
Merged into `dev`, waiting for CI to pass. 578c2c5dd7de034edffebadff5ae4e8cafaaa67c

PR #426, opened 2016-07-09
Use `ReadinessQueue` on Windows

alexcrichton:
One of the major motivations for adding the `ReadinessQueue` was indeed using it
on Windows! The implementation in `poll` is more battle-tested than the
implementation in `src/sys/windows` already, so let's switch over to using it.

This should also end up being more performant as well as we're able to drop a
few locks here and there, yay!

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -1,8 +1,9 @@
 use {convert, sys, Evented, Token};
 use event::{self, EventSet, Event, PollOpt};
 use std::{fmt, io, mem, ptr, usize};
-use std::cell::UnsafeCell;
+use std::cell::{UnsafeCell, Cell};
 use std::isize;
+use std::marker;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicUsize, AtomicPtr, Ordering};
 use std::time::Duration;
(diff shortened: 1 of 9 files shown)


carllerche:
I tried to merge this into `dev`, but it seems like there are a ton of merge conflicts now :( I will take a look at them.

src/poll.rs
```diff
@@ -716,6 +747,7 @@ impl ReadinessQueue {
 }
 
 unsafe impl Send for ReadinessQueue { }
+unsafe impl Sync for ReadinessQueue {}
```
carllerche: Wait a minute... I must be missing some plot here. Why is this true?

carllerche: Specifically, my *assumptions* when building `ReadinessQueue` were MPSC.

carllerche:
Looking more at this PR, I think that there *should* be a way to implement this w/o exposing the readiness queue and only using the public API of `Poll`. Is there a reason why you think that this is not possible?

alexcrichton:
@carllerche I've rebased and managed to finagle it as well to retain the same API as before with `Poll` by just passing around `&Poll` more often.

carllerche:
Looks good to me! Unfortunately it looks like CI hung or something... let me check

carllerche:
This is on master now! Thanks.

PR #431, opened 2016-07-19
Finish completing accept/connect on Windows

### LINKED ISSUE #397: TcpStream::peer_addr and local_addr fail on Windows
I'm running into an issue where the same code behaves differently on Windows and Linux (big surprise :smile:). It's led one user to request changes to the API of WS-RS (see housleyjk/ws-rs#49), but I believe the underlying issue has to do with `peer_addr` and `local_addr` on Windows. This gist, derived from the simple TCP example in the docs and using only Mio, shows that the same code runs perfectly on Linux but fails on Windows: https://gist.github.com/housleyjk/fa0ab22ef2264818c43d25ab2ed76ffb. If possible, I would like to avoid needing to provide a Windows-specific API for obtaining the peer_addr in WS-RS, and I don't know how I can get the local_addr to work on Windows.

Thank you for your consideration!

alexcrichton:
Apparently there are extra steps that need to be taken after a socket finishes
being accepted and/or connected on Windows to prepare state like calling
getsockaddr and getpeeraddr. This updates the `miow` binding to where it has
these functions, then calls those functions in these situations.

Closes #397

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ libc   = "0.2.4"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.2.1"
-miow   = "0.1.1"
+miow   = "0.1.3"
 
 [dev-dependencies]
 env_logger = "0.3.0"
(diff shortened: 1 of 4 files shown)


alexcrichton:
For reference the associated runtime support is at https://github.com/alexcrichton/miow/commit/f987c98b3c0e30a390c3df13aeff528501b74ab3

carllerche:
Merged into dev: 3e18267ac71d26f47ad68baae87fb101d14fa987

PR #432, opened 2016-07-19
Only call `reuse_address` on Unix

### LINKED ISSUE #400: Binding twice on the same addr returns OK both times on Windows
Hello,

I wrote a test showing that on Windows, when using `std::net::TcpListener`, the second bind returns `Err` but using `mio::net::tcp::TcpListener`, both calls return `OK`. On linux, the tests are passing.

The [Appveyor build](https://ci.appveyor.com/project/blabaere/mio-win-inv).
The [Travis build](https://travis-ci.org/blabaere/mio-win-inv).
The [source code](https://github.com/blabaere/mio-win-inv/blob/master/src/lib.rs#L433).

Regards,
Benoît

alexcrichton:
The standard library only Does this on Unix, and otherwise it's a source of
divergent behavior on Windows.

Closes #400

### DIFF
--- a/src/net/tcp.rs
+++ b/src/net/tcp.rs
@@ -190,8 +190,10 @@ impl TcpListener {
             SocketAddr::V6(..) => TcpBuilder::new_v6(),
         });
 
-        // Set SO_REUSEADDR
-        try!(sock.reuse_address(true));
+        // Set SO_REUSEADDR, but only on Unix (mirrors what libstd does)
+        if cfg!(unix) {
+            try!(sock.reuse_address(true));
+        }
 
         // Bind the socket
         try!(sock.bind(addr));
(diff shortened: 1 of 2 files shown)


sateffen:
Just my optinion: But shouldn't rust behave the same on all oses, at least as much as possible?

I think mio shouldn't set reuse_address at all, but let the user set this, if he needs it. This way the behaviour on all platforms would be the same. Or is the resuse_address a default thing on unix systems?

carllerche:
Mio's defaults should result in portable code as much as possible. Aka, behavior that is as close as possible on all supported platforms. OS defaults diverge, so some different options will be required as defaults. If this is not the desired behavior, Mio allows the developer to configure the options as desired.

carllerche:
Thanks. Merged here 5d07d0b8ec14858129dfda8dc185ae58255eea11

PR #434, opened 2016-07-19
Use a monotonic time source in timer.

tmiasko:
Replace all uses of SystemTime with Instant to avoid too early and too
late timeouts when system time is changed (daylight saving time, etc.)

To wake up the internal thread after correct delay, the
thread::part_timeout function should also use a monotonic time source,
which unfortunately is not true across all supported platforms (but this
is a QOI issue of rust condition variables).


carllerche:
Thanks for catching this.

carllerche:
Merged into `dev`, will merge into master once CI passes: 5268c2184fe410f02a39ec345b172918d4b6a059

PR #435, opened 2016-07-19
Align UdpSocket with std::net::UdpSocket

alexcrichton:
* Remove the `bind` methods and change it into a constructor
* Remove `bound` as this is replaced by `bind`
* Add `from_socket` akin to `TcpStream::connect_stream`
* Re-synchronize all methods for `setsockopt` with libstd, removing some old,
  adding new ones, and adding accessors.


alexcrichton:
This doesn't yet currently add `UdpSocket::{connect, send, recv}` as that'll likely require some work on the Windows side which I didn't want to tackle just yet. Should be easy to add in the future though!

alexcrichton:
Ah and for reference all documentation was copied from libstd

carllerche:
On dev!

Issue #427, opened 2016-07-15
Mio memory usage - 2.5mb+ for running loop


tigranbs:
Hi I'm doing integrated library using Mio loop and after having some working demo, I've been surprised about memory usage just for running loop, it's around `2.5mb`. The good thing is that it's not growing fast during data transfer, but for my usecase it's really huge.

Main problem with my application is that I've created integratabtle async database, and I'm running event loops based on current CPU cores, so if server have 12 cores then only program initialization is getting `12*2.5 = 30mb` RAM memory. 

What can I do with this ?

P.S.  I've done some initial research for using `libuv` and found that it is very very cheap in terms of memory usage compared to Mio. Is there any downsides to using `libuv` (Performance or anything else) ?

sateffen:
What operating system are you on? On windows I've got some memory problems as well (implementation is pretty new), but on a linux machine my load doesn't execeed 7mb with ~5k Sockets transfering data (single core). Maybe this information could help?

tigranbs:
I'm on a mac,
It's not depend on connections, it's just program initialization.
Even basic TCP Echo server getting around 3mb of memory without any connections.

It's not a memory leak, because even on 100K connection load memory not getting bigger. 
I've just surprised why it's allocating that much memory at the beginning, and for my usecase it's really can be bigger on multi cpu servers.

alexcrichton:
@tigranbs thanks for the report! Could you share the code you're testing out as well? 

Right now mio doesn't allocate *that* much, but if you're using `EventLoop` then there are a few large-ish allocations:

* First, a [`Poll` is created](https://github.com/carllerche/mio/blob/c3069801da91ef6fe568a925a74f348b398c5726/src/event_loop.rs#L111) which contains a [buffer of events](https://github.com/carllerche/mio/blob/c3069801da91ef6fe568a925a74f348b398c5726/src/poll.rs#L58) which for epoll [allocate 1024 slots](https://github.com/carllerche/mio/blob/c3069801da91ef6fe568a925a74f348b398c5726/src/sys/unix/epoll.rs#L140-L149) and on kqueue [allocate even more](https://github.com/carllerche/mio/blob/c3069801da91ef6fe568a925a74f348b398c5726/src/sys/unix/kqueue.rs#L153-L166).
[...]

PR #436, opened 2016-07-19
Add an Events::with_capacity constructor

alexcrichton:
Should allow tuning how many events can get returned from one turn of an event
loop, in theory at least. Somewhat related to #427 where memory tuning may
matter in some situations.


carllerche:
Thanks, merged to dev

Issue #383, opened 2016-04-29
Timeouts trigger one tick late when there's a delay between creating and running the event loop


xalri:
This first manifested through mioco as https://github.com/dpc/mioco/issues/126, but it seems to be a bug in mio itself.

Here's what I _think's_ happening:
At the end of a call to `Timer::tick_to(t)` (https://github.com/carllerche/mio/blob/master/src/timer.rs#L202) where there are no timouts set to trigger on tick `t`, `self.tick` is equal to `t + 1`, one tick past the current time.
```rust
    while self.tick <= now {
        ...
        if curr == EMPTY {
            self.tick += 1;
```

Then, `next_tick_in_ms` returns the time in ms to the start of `self.tick + 1` (`== t + 2`), so any timeouts on tick `t + 1` don't trigger untill tick `t + 2`. I'm not sure why, but the effects of this are only visible when there's a delay between creating and starting the event loop. I made a small change to `tick_to` to test this, and it fixed the example below (https://gist.github.com/xalri/bc54030fa7ca48496fa253586d6f6a99).

Anyway, The effect is that timeouts trigger one tick late.
Here's an example which should timeout every 100ms (the default timer resolution), but, with a 10ms delay before starting the loop, the first takes 290ms, and all subsequent timeouts take 200ms.

``` rust
#[macro_use] extern crate log;
extern crate mio;

mod logger;
use std::time::Duration;

struct Timer;
impl mio::Handler for Timer{
    type Timeout = ();
    type Message = ();

    fn timeout(&mut self, evl: &mut mio::EventLoop<Timer>, _: ()){
		info!("timeout");
        evl.timeout((),Duration::from_millis(100)).unwrap();
    }
}

fn main() {
    logger::init().unwrap();
    let mut event_loop = mio::EventLoop::new().unwrap();
    std::thread::sleep(Duration::from_millis(10));
    event_loop.timeout((),Duration::from_millis(100)).unwrap();
    event_loop.run(&mut Timer{}).unwrap();
}
```
And the output, using mio master with a little more debug info:
```
mio::poll       32263544ms  registering with poller
mio::event_loop 32263554ms  event_loop.timeout Duration { secs: 0, nanos: 100000000 }
mio::timer      32263554ms  timeout_ms now: 32263553   delay: 100
mio::timer      32263554ms  timeout_at_ms at: 32263653
mio::timer      32263554ms  timeout_at_ms at-self.start: 110
[...]

PR #440, opened 2016-07-21
Use consistent rounding strategy when converting a duration to a tick.

tmiasko:
Ensure that the same strategy of rounding to the nearest tick is used
both when scheduling a timeout and in wakeup thread. Fixes issue #383.


carllerche:
Thanks. Merged: 18d16388b37f8d96618587c87cbd095e115e8e7d

PR #441, opened 2016-07-23
Remove the Unix Domain Socket bindings 

alexcrichton:
@carllerche expressed interested in removing these bindings in favor of an
external library, so this is the PR to do so! I've extracted the bindings to
mirror the standard library in a repo [mio-uds], and I intend to maintain the
functionality over time.

[mio-uds]: https://github.com/alexcrichton/mio-uds


carllerche:
I moved UDS to the `deprecated` directory, which is enough for now I think.

PR #444, opened 2016-07-30
windows: Finish pending writes when a socket is dropped

### LINKED ISSUE #423: Windows aborts TcpStreams on close
On Windows, dropping the stream results in a `WSACONNABORTED` error on the client side.

I've noticed this when running a hyper server on Windows, that all responses end in an abort: https://github.com/hyperium/hyper/issues/816

That issue also includes an attempt by a downstream user to fix it, I haven't been able to test those 'fixes' myself.

Test case: https://gist.github.com/seanmonstar/8829d325daa386ffe171b17e52d0038b

alexcrichton:
On Unix the kernel will do this for us, but on Windows we need to do it ourselves. Previously we would forcibly close a socket whenever a socket was dropped on Windows. This would leave around its associated Arc (for completion events), but it would cancel all I/O in flight.

Now we instead allow writes to finish (we don't cancel them), and we just cancel any reads that are in flight. This should ensure that we actually get around to closing the socket in a timely fashion (not blocked on a read that'll never come) but we'll still ensure that all the data we said we wrote actually gets written.

Closes #423 

cc @seanmonstar

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ libc   = "0.2.4"
 [target.'cfg(windows)'.dependencies]
 winapi = "0.2.1"
 miow   = "0.1.3"
+kernel32-sys = "0.2"
 
 [dev-dependencies]
 env_logger = "0.3.0"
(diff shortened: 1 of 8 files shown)


seanmonstar:
\o/

I can test later if you'd like

carllerche:
Thanks. Merged into `dev`

PR #468, opened 2016-09-12
prevent Io::into_raw_fd from dropping self

oconnor663:
Currently it closes its own fd immediately before returning. It looks like this affects `UnixSocket` too, since it contains an `Io`, though I'm not sure if either of these have any callers currently?

Related thought: Although I don't think it's exposed publicly, the `From<RawFd>` implementation for `Io` lets safe code create two `Io`'s from the same fd, which leads to a double close. I think that sort of thing is why `from_raw_fd` is unsafe. I'll put up a second PR to see what deleting that safe constructor looks like, and maybe you can let me know what you think?

### DIFF
--- a/src/sys/unix/io.rs
+++ b/src/sys/unix/io.rs
@@ -1,5 +1,6 @@
 use {io, poll, Evented, Ready, Poll, PollOpt, Token};
 use std::io::{Read, Write};
+use std::mem;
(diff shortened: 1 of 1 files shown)


carllerche:
Thanks for this. Merged here: f72e92d384b22f1f7dc2c4f6897b7d1952c4e72e

PR #469, opened 2016-09-12
get rid of the safe constructors for Io

oconnor663:
As mentioned in https://github.com/carllerche/mio/pull/468. It's kind of a drag to have `unsafe` everywhere, but on the other hand the function _is_ kind of unsafe. Curious for your thoughts.


carllerche:
Alright, that makes sense, thanks

carllerche:
Merged it into dev: 68aaf8b8d0d0ab093a203a5e84eef03aa38f841f

Issue #439, opened 2016-07-20
Windows implementation consumes lots of memory (linux doesn't)


sateffen:
Hey,

like mentioned in #415 windows consumes a huge lot of memory for a simple echo server with lots (5-10k) connections (sending and receiving data).

On [this](https://github.com/sateffen/rust-tcp-server) simple project creating around 10k connections consumes ~700mb of ram, on a linux machine it takes around 7mb. After all connections are closed the memory is freed again.

To reproduce:
- Clone the repo
- open a shell and execute _cargo run_
- open a shell and execute _node test.js_

Setup:
- Windows 10
- Node 6.2.2
- Rust 1.8.0
- mio 0.5.1

This will spin up 10k connections to localhost:8888 (the waiting rust echo server).

I'm experiencing this on a windows 10 laptop, simply observed with the taskmanager. As linux reference I've used a private rootserver and a cloud9 machine, none of them showed this behaviour.

As soon as I've got some time I'll update my compiler and try to create some simpler code to reproduce, but currently my time is somehow limited.

If any questions left, just ask :)

Cheers

alexcrichton:
Windows is implemented differently than Unix due to the IOCP model, so there's bound to be *some* more overhead than before. Right now Windows uses a blanket 64KB buffer for all reads on sockets, and if you have 10000 sockets that's 655 MB right there (close to the 700MB you're seeing)

The behavior on Windows should likely be smarter than just "always allocate 64KB buffers" along with the ability to tune the buffer size per connection ideally.

sateffen:
That sounds pretty much like the behaviour I'm experiencing. The 65mb + base memory + the bytes copied to my local buffers might go up to ~700mb (I don't remember the actual number, but lets assume it's around 650-700mb).

In my case I send around 14 byte payload, 64KB is pretty much too much *in this case*. Is there a chance to optimize this, or is the IOCP model that wrecked, that it has to be like this?

How is libuv doing the job not raising so much memory? Maybe we can check their tricks to learn from them?

alexcrichton:
Oh there's nothing inherently bad about IOCP here, we just have some heuristics about when to schedule I/O which probably need a large amount of tweaking or ability to configure.

tmiasko:
In libuv user is fully responsible for memory allocation and ensuring that
buffers are kept alive until IO operations complete, so that is not exactly
comparable with what is in MIO. 

But this makes me wonder, why is overlapped used in Windows implementation in
[...]

PR #471, opened 2016-09-13
windows: Don't allocate buffers for TCP reads

alexcrichton:
This commit employs a trick [1](https://github.com/carllerche/mio/issues/439#issuecomment-246602046) for avoiding allocation of an intermediate
buffer for in-flight calls to `TcpStream::read`. Previously this was done to
receive data, but apparently it works out to pass a 0-length buffer to the
kernel and get a notification back when there's internally buffered data.

This in turn allows us to avoid allocation any buffers while getting an
epoll-style readiness notification for TCP streams. Once a 0-byte read has
completed we just call the normal `TcpStream::read` function and wait for
`WouldBlock` to get returned. Note that this involves putting the socket into
nonblocking mode (which we managed to avoid before).

cc #439

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -454,7 +454,7 @@ impl RegistrationInner {
         // will set a `Release` barrier ensuring eventual consistency.
         self.node().events.store(event::as_usize(ready), Ordering::Relaxed);
 
-        trace!("readiness event {:?} {:?}", ready, self.node().token());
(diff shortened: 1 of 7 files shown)


alexcrichton:
Merged as https://github.com/carllerche/mio/commit/d8a79d6581a3cbe8b6ea19ccc6a86f0f29a960e7

PR #486, opened 2016-10-15
WIP: Vectored IO interface

Ralith:
This PR demonstrates a possible approach to exposing a portable interface to vectored IO. To guarantee binary compatibility with the syscalls without requiring copying and either dynamic allocation or a low limit on the number of buffers, the syscalls' native slice types (e.g. `iovec` or `WSABUF`) must be exposed publicly, so I defined a portable wrapper for Unix and Windows. I modified the *nix definition `UdpSocket::send_to` and its tests to demonstrate employment of this wrapper.

If this approach is deemed acceptable, the PR should be fleshed out with implementations of vectored IO for all network IO functions on all platforms. Due to the ugliness of the interface, these implementations should probably be provided alongside, rather than in place of, the existing slice-based functions.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -135,6 +135,7 @@ pub use poll::{
 pub use token::{
     Token,
 };
+pub use sys::IoVec;
 
 #[cfg(unix)]
 pub mod unix {
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -1,6 +1,6 @@
 //! Primitives for working with UDP
 
-use {io, sys, Evented, Ready, Poll, PollOpt, Token};
+use {io, sys, Evented, Ready, Poll, PollOpt, Token, IoVec};
 use super::SelectorId;
 use std::net::{self, Ipv4Addr, Ipv6Addr, SocketAddr};
 
@@ -56,10 +56,7 @@ impl UdpSocket {
 
     /// Sends data on the socket to the given address. On success, returns the
     /// number of bytes written.
-    ///
-    /// Address type can be any implementor of `ToSocketAddrs` trait. See its
-    /// documentation for concrete examples.
-    pub fn send_to(&self, buf: &[u8], target: &SocketAddr)
+    pub fn send_to(&self, buf: &[IoVec<&[u8]>], target: &SocketAddr)
                    -> io::Result<Option<usize>> {
         self.sys.send_to(buf, target)
     }
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -9,6 +9,7 @@ pub use self::unix::{
     TcpListener,
     UdpSocket,
     UnixSocket,
+    IoVec,
     pipe,
     set_nonblock,
 };
@@ -25,6 +26,7 @@ pub use self::windows::{
     TcpStream,
     TcpListener,
     UdpSocket,
+    IoVec,
 };
 
 #[cfg(windows)]
--- a/src/sys/unix/mod.rs
+++ b/src/sys/unix/mod.rs
@@ -95,3 +95,17 @@ mod nix {
         dup,
     };
 }
+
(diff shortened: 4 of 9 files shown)


alexcrichton:
Thanks for the PR! I wonder if we could perhaps try to keep the same level of ergonomics for the existing methods as well? (e.g. not require importing `IoVec`). Also, It may be worthwhile exploring `IoVec` vs `IoVec<T>` because `T` seems to not vary too much and it should basically almost always be bytes, right?

Ralith:
> I wonder if we could perhaps try to keep the same level of ergonomics for the existing methods as well? (e.g. not require importing IoVec)

I'm not sure what you mean. The existing methods should probably remain because they have a very nice, clean interface. IoVec is necessary for cross-platform zero-overhead vectored IO because the layouts of WSABUF, iovec, and rust slices are mutually inconsistent--only by exposing the existence of something that isn't a slice to the public API can we force the user to lay out their data in the required platform-specific form.

> Also, It may be worthwhile exploring IoVec vs IoVec<T> because T seems to not vary too much and it should basically almost always be bytes, right?

Mutability should be preserved somehow. Perhaps IoVec and IoVecMut?

alexcrichton:
Ah all I mean is that `send_to` and such methods are diverging from their counterparts in libstd, which I think we'll want to avoid. We could use a trait bound to get `send_to(&[1, 2, 3])` and `send_to(&[buf1, buf2])` to work, but we may also want to introduce a new `send_to_vector` or something like that for these methods.

> Mutability should be preserved somehow. Perhaps IoVec and IoVecMut?

I'd personally be more comfortable with that, yeah, or perhaps a generic `IoBuf` perhaps. I may play around with this.

alexcrichton:
Hm ok, we may be able to get away with `IoVec` being a DST, meaning we don't need a mut/not mut variant: https://gist.github.com/alexcrichton/c9b47fedd5f959d8d2d3bc73b7636d14.

Does that gist make sense? It's Unix right now but I think it'd be easily translatable to Windows.

Ralith:
Are you sure that's the gist you meant to link? I don't see anything relevant in it.

alexcrichton:
Oops, that was indeed the wrong gist! Let's try this one: https://gist.github.com/alexcrichton/a609558f6440703135e38170d743a539

Ralith:
That's a pretty nice looking interface, and it does seem to work. Probably because I'm still pretty new to rust, I have some trouble understanding how it can possibly be legal to transmute from `imp::iovec` to a reference, e.g. on line 135. Can you provide any insight there? I must be missing something. The actual `imp::iovec` storage has to exist somewhere...

It's also strange to see `&mut [&mut IoVec]` in the `readv` interface, since if the inner types were actually references then the outer reference wouldn't need to be mutable, would it? `readv` should only need to modify the bytes identified by the iovecs, not the iovec objects themselves.

alexcrichton:
Ah the transmute is actually valid because `&IoVec` is the same size as `iovec`. Technically the internal `[u8]` is "corrupt" because the pointer and length may be switched, but the size of the types works out so the transmute works.

Also yeah for `readv` I think we'll need two layers of mutability. Rust won't allow you to modify a `&[&mut [u8]]`, so this needs to preserve similar semantics.

Ralith:
Ah, `&IoVec` must be a fat pointer, I see. Clever hackery. It assumes that each platform's iovec analog is the same size as such a fat pointer, but considering that they're basically the same thing that seems reasonable, albeit sensitive to internal details of the rust implementation. I'll proceed with that approach when I get some more time, unless you beat me to it.

alexcrichton:
I've prototyped what I was thinking in https://github.com/carllerche/mio/pull/516

alexcrichton:
Ok, added in https://github.com/carllerche/mio/pull/516

PR #487, opened 2016-10-24
Update nix api for version 0.7.1 and later

asomers:
Git rev 45c82235a3b91939ea9a656c6afcf71be15411f3, to be released in
Nix 0.7.1 or later, changes the API of KEvent.

https://github.com/nix-rust/nix/pull/442 must be merged before this PR will build.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ slab     = "0.3.0"
 net2     = "0.2.19"
 
 [target.'cfg(unix)'.dependencies]
-nix    = "0.6.0"
+nix    = { git = "https://github.com/nix-rust/nix" }
 libc   = "0.2.14"
 
 [target.'cfg(windows)'.dependencies]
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! Right now master of this repo is the 0.6 release of mio, so we're hoping to not introduce git dependencies as that would block another point release of mio. Perhaps though this could be updated once the changes to nix are merged?

asomers:
In fact, the changes to nix just got merged yesterday.  Or do you mean we should update this PR once nix has a new release?  I'll certainly do that.
https://github.com/nix-rust/nix/commit/652e730f106fd3c103ba7549e64a7c992a72f11a

alexcrichton:
Yeah once there's a new release of nix we can pull that in.

posborne:
Since there are semver breaking changes in a few places, we are probably looking at a release 0.8.0 as the next release of nix.  We don't have a solid date on that in mind right now, but if there is urgency we can probably look to expedite.  I am doing a pass through nix improving (or just adding) documentation across the board which has led to some API changes where things were either broken or doing users no service.

alexcrichton:
@posborne sounds good to me! AFAIK there's no burning desire to get this updated, so mio can just update whenever 0.8.0 is released

carllerche:
The dependency on `nix` was removed as we weren't using much of it anyway and we ended up with custom bits.

Issue #483, opened 2016-10-13
timer thread is looping indefinitely


Geal:
Hi,

I have an issue with a timer, created like this:

``` rust
let timer  = Timer::default();
poll.register(&timer, Token(1), Ready::readable(), PollOpt::edge()).unwrap();
```

I can set a few timeouts on it, and they are triggered correctly, but the timer thread is busy looping (like, laptop heating up looping):

```
2016-10-13T18:10:32Z    299872025673192 7337    TRACE   sleeping; tick_ms=100; now_tick=74; sleep_until_tick=18446744073709551615; duration=18446744073709551615        |       mio::timer
2016-10-13T18:10:32Z    299872026053174 7337    TRACE   wakeup thread: sleep_until_tick=18446744073709551615; now_tick=74       |       mio::timer
2016-10-13T18:10:32Z    299872026071003 7337    TRACE   sleeping; tick_ms=100; now_tick=74; sleep_until_tick=18446744073709551615; duration=18446744073709551615        |       mio::timer
2016-10-13T18:10:32Z    299872026122879 7337    TRACE   wakeup thread: sleep_until_tick=18446744073709551615; now_tick=74       |       mio::timer
2016-10-13T18:10:32Z    299872026140557 7337    TRACE   sleeping; tick_ms=100; now_tick=74; sleep_until_tick=18446744073709551615; duration=18446744073709551615        |       mio::timer
2016-10-13T18:10:32Z    299872026168054 7337    TRACE   wakeup thread: sleep_until_tick=18446744073709551615; now_tick=74       |       mio::timer
2016-10-13T18:10:32Z    299872026176896 7337    TRACE   sleeping; tick_ms=100; now_tick=74; sleep_until_tick=18446744073709551615; duration=18446744073709551615        |       mio::timer
2016-10-13T18:10:32Z    299872026199532 7337    TRACE   wakeup thread: sleep_until_tick=18446744073709551615; now_tick=74       |       mio::timer
2016-10-13T18:10:32Z    299872026649172 7337    TRACE   sleeping; tick_ms=100; now_tick=74; sleep_until_tick=18446744073709551615; duration=18446744073709551615        |       mio::timer
2016-10-13T18:10:32Z    299872026693437 7337    TRACE   wakeup thread: sleep_until_tick=18446744073709551615; now_tick=74       |       mio::timer
[...]

PR #488, opened 2016-10-28
Fix infinite looping timer thread

nastevens:
Due to undefined behavior in pthread when passing in large timeout values, the park_timeout call of the timer thread was spinning very tightly and almost continuously causing poll wakeups. This commit changes the behavior to use a blocking park call when the timeout value is large.

See #483 for more information.


alexcrichton:
Merged as https://github.com/carllerche/mio/commit/42a41721afa441bf95846087737bb7f5fe9d4b9f, thanks for the PR!

PR #489, opened 2016-10-31
Fix compile on 32-bit OSX

alexcrichton:
Unfortunately there's a type error right now here.


alexcrichton:
Merged as https://github.com/carllerche/mio/commit/b7fc4504d3b1cc6a3453a68558dc4e61f686cefc

PR #493, opened 2016-11-08
Support SO_REUSEPORT option when feature 'reuse_port' enabled

choleraehyq:
On Linux, this socket option allows several sockets to be bound to one (address, port) pair, and the incoming requests will be distribute evenly across all sockets, which achieves a simple load balance and solves a kind of bottleneck of the single accepting thread.

But this option is only available on linux kernel 3.9+, so I implement it as an optional feature.

@carllerche @alexcrichton Please review this, thanks.

Signed-off-by: Cholerae Hu <choleraehyq@gmail.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,10 @@ env_logger = "0.3.0"
 tempdir    = "0.3.4"
 bytes      = "0.3.0"
 
+[features]
+default = []
+reuse_port = []
+
 [[test]]
 name = "test"
 path = "test/mod.rs"
--- a/src/net/tcp.rs
+++ b/src/net/tcp.rs
@@ -281,6 +281,13 @@ impl TcpListener {
             try!(sock.reuse_address(true));
         }
 
+        // Set SO_REUSEPORT, but only on Unix
+        // And on Linux, this socket option is only available on kernel 3.9+
+        if cfg!(unix) && cfg!(feature = "reuse_port") {
+            use net2::unix::UnixTcpBuilderExt;
+            try!(sock.reuse_port(true));
+        }
+
         // Bind the socket
         try!(sock.bind(addr));
 
--- a/test/test_tcp.rs
+++ b/test/test_tcp.rs
@@ -267,8 +267,12 @@ fn test_tcp_sockets_are_send() {
 }
 
 #[test]
-fn bind_twice_bad() {
+fn bind_twice() {
     let l1 = TcpListener::bind(&"127.0.0.1:0".parse().unwrap()).unwrap();
     let addr = l1.local_addr().unwrap();
-    assert!(TcpListener::bind(&addr).is_err());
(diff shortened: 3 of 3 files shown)


tailhook:
I think this shouldn't be enabled by default. Because it changes how kernel load-balances connections between sockets and may lead to unexpected results.

Also, it's often bad idea to detect kernel features at compile time rather than at runtime.

So I think it should be configured at runtime (i.e. a method) rather than at compile time. Also in many cases you can just skip the error if kernel doesn't support this feature instead of failing to bind socket.

choleraehyq:
Yes, it's not enabled by default.
I've thought about add a method, but `net2::TcpStream` doesn't provide method `reuse_port`, and right now the option `SO_REUSEADDR` is set in method `bind`, so I also set `SO_REUSEPORT` in `bind`.
Maybe I should add APIs `set_reuse_port` and `reuse_port` like `set_nodelay` and `delay` .
Do I need to open a new issue for this?

choleraehyq:
Besides, SO_REUSEPORT must be set before `sock.bind(addr)`, so if we want to control SO_REUSEPORT option via a new API, we need to use a boolean global flag to present whether the user want to use this option, and use the new API to set this global flag.

tailhook:
> and right now the option SO_REUSEADDR is set in method bind, so I also set SO_REUSEPORT in bind

I understand this. But they are very different. I don't know use cases for not having `SO_REUSEADDR` except superfluous caution.

But `SO_REUSEPORT` is very different beast. It should not be used unless you know what are you doing. I.e. it's very tempting to use `SO_REUSEPORT` instead of old "fork-and-pass-fd" procedure, but it's not the same, and reuse port is worse than latter in most cases.

> Besides, SO_REUSEPORT must be set before sock.bind(addr), so if we want to control SO_REUSEPORT option via a new API, we need to use a boolean global flag to present whether the user want to use this option, and use the new API to set this global flag.

There is `net2::TcpBuilder` interface for that case. For example [here is how you can disable SO_REUSEADDR](https://doc.rust-lang.org/net2-rs/net2/struct.TcpBuilder.html#method.reuse_address)

choleraehyq:
En...My meaning is that both `SO_REUSEADDR` and `SO_REUSEPORT` need to be set before `sock.bind(addr)`, but we don't have any init function, `TcpListener::bind` is the first and the only one function to get a bound listener, users know nothing about `TcpBuilder`. 
What about adding a new API called `TcpListener::bind_reuse_port` or something?

alexcrichton:
I agree with @tailhook that this shouldn't be a feature of the crate but rather something that's configured at runtime per socket. In a sense this would be a breaking change to activate this feature for crates which were not expecting it.

Note that this is also why mio allows construction of objects from `std::net` variants, which can notably be constucted with net2 with these various options.

choleraehyq:
I see. Close this PR.

PR #497, opened 2016-11-23
Fixed looping timer

rphmeier:
Pulling the patch described in https://github.com/ethcore/parity/pull/3592 upstream.

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -225,8 +225,12 @@ impl<T> Timer<T> {
         self.poll_to(target_tick)
     }
 
-    fn poll_to(&mut self, target_tick: Tick) -> Option<T> {
+    fn poll_to(&mut self, mut target_tick: Tick) -> Option<T> {
         trace!("tick_to; target_tick={}; current_tick={}", target_tick, self.tick);
+
+		if target_tick < self.tick {
+			target_tick = self.tick;
+		}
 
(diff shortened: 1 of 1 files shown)


rphmeier:
Hi, just wondering if the mio team has seen this at all? I really don't like to maintain alternative forks of repos for parity which is why I'm trying to upstream this fix. Any response would be appreciated.

src/timer.rs
```diff
@@ -225,9 +225,13 @@ impl<T> Timer<T> {
         self.poll_to(target_tick)
     }
 
-    fn poll_to(&mut self, target_tick: Tick) -> Option<T> {
+    fn poll_to(&mut self, mut target_tick: Tick) -> Option<T> {
         trace!("tick_to; target_tick={}; current_tick={}", target_tick, self.tick);
 
+		if target_tick < self.tick {
```
arthurprs: Nit: indentation is off.

arthurprs:
LGTM but I wonder if this is band aiding it instead of fixing the root cause.

alexcrichton:
Ah sorry this slipped under the radar! I merged this as https://github.com/carllerche/mio/commit/ef182bae193a9c7457cd2cf661fcaffb226e3eef and did a whitespace fixup last night, so closing.

PR #503, opened 2016-12-13
getting-started.md: Remove note about Windows

jonte:
README.md states that Windows is indeed supported, so remove this text from the tutorial


carllerche:

PR #507, opened 2016-12-28
Updating usage comments for Mio

reubencornel:
I found the original comments to be confusing because EventLoop and related structures have been deprecated. Updated the comments in an attempt to provide more clarity on its usage.

The other changes have been made by rustfmt.


alexcrichton:
Thanks!

PR #516, opened 2017-01-12
Implement readv/writev support for TcpStream

alexcrichton:
This commit implements `readv` and `writev` support for TCP sockets in mio. This
corresponds to the `readv` and `writev` syscalls on Unix and a bit of custom
logic on Windows. On Windows writes are just concatenated into one buffer and
then handed off to IOCP, and each buffer to read into incurs a `recv` syscall
(currently). This may be optimizable in the future for Windows.

The other major factor in this commit is that a new `IoVec` type is added to
this library to represent a buffer passed to these functions. This type is a
dynamically sized type to allow `&IoVec` to represent `&[u8]` and `&mut IoVec`
to represent `&mut [u8]`, allowing the same type to be used in the read/write
versions of these functions. The in-memory representation of `IoVec` is
equivalent to `libc::iovec` on Unix and `winapi::WSABUF` on Windows, asserted
with various transmutes. This all allows us to build up 0-cost abstractions for
readv/writev.

### DIFF
--- a/src/iovec.rs
+++ b/src/iovec.rs
@@ -1,0 +1,33 @@
+use std::mem;
+
+use sys;
+
+pub struct IoVec {
+    data: sys::IoVec,
+}
+
+impl<'a> From<&'a [u8]> for &'a IoVec {
+    fn from(bytes: &'a [u8]) -> &'a IoVec {
+        unsafe {
+            mem::transmute(<&sys::IoVec>::from(bytes))
+        }
+    }
+}
+
+impl<'a> From<&'a mut [u8]> for &'a mut IoVec {
+    fn from(bytes: &'a mut [u8]) -> &'a mut IoVec {
+        unsafe {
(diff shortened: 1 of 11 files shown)


carllerche:
👍

arthurprs:
That's some cool stuff. Could windows portion ever use WSASend and WSARecv instead?

alexcrichton:
@arthurprs I think `WSASend` won't ever be usable unfortunately due to how writes work on Windows. In that case we submit a custom buffer to IOCP that mio allocates and tracks internally. We're already copying bytes once, so for `write_bufs` we just copy all the buffers into the buffer to write. 

For `WSARecv` I left a TODO, I'm uncertain of the semantics of calling `WSARecv` on an overlapped but nonblocking socket. I suspect that it's not valid (e.g. requires an overlapped object), so we may not be able to do so.

PR #517, opened 2017-01-12
Remove dependency on nix

alexcrichton:
This commit removes the dependency on the `nix` crate in favor of local
bindings. Most of the local bindings are slated to be deleted with the eventual
removal of the `deprecated` module and implementing Unix sockets and pipes out
of tree anyway.

The `nix` crate has grown quite large over time and is a bit of a weighty
dependency, and future extensions to the mio crate, such as adding eventfd
support, would likely require bypassing `nix` regardless to deal with
conditional support for the feature. Additionally mio already has raw logic for
functions like `epoll_create1` and such.


carllerche:
👍

PR #519, opened 2017-01-19
Implement Display for SendError<T> and TrySendError<T>.

hicqu:
I want to convert SendError<Something> to String, but such error reported,
```
the method `to_string` exists but the following trait bounds were not satisfied: `mio::channel::SendError<Box<stub::Task +std::marker::Send>> : std::fmt::Display`
```
I believe this PR has resolved the issue. What do you think?

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -4,6 +4,7 @@
 
 use {io, Evented, Ready, Poll, PollOpt, Registration, SetReadiness, Token};
 use lazycell::{LazyCell, AtomicLazyCell};
+use std::fmt;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Looks good to me, thanks! Want to throw in an implementation of the `Error` trait while you're at it?

hicqu:
It build failed. Could you help me deal with that? Thank you very much.

alexcrichton:
@hicqu the error here is just for Rust 1.9.0, the oldest version of Rust that mio supports. You can fix this by adding `Any` bounds to the type parameters.

Issue #476, opened 2016-09-26
Use SetFileCompletionNotificationModes on Windows


alexcrichton:
I/O operations on Windows tend to have the ability to "complete immediately". For example something like [`read_overlapped`](http://alexcrichton.com/miow/miow/net/trait.TcpStreamExt.html#tymethod.read_overlapped) returns `io::Result<bool>` where `Ok(false)` means the operation is pending and `Ok(true)`  means the operation completed immediately. These correspond to the underlying primitives on Windows (with various error codes).

Unfortunately, though, "completes immediately" means that by default a completion message is enqueued on the completion port. That means that if we were to service the read immediately we'd have to arrange for that later completion message to get ignored, which isn't currently easy to do. This means that although the read finished immediately we return `WouldBlock` and let the trip through the event loop complete the read at a later date.

As you can imagine, this seems like it can definitely be a performance hit! Thankfully Windows has an API for this, [`SetFileCompletionNotificationModes`](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365538%28v=vs.85%29.aspx), which will cause a completion status to _not_ get enqueued if an operation finishes immediately. We should use this on Windows at least for TCP/UDP read/write to help throughput.

Some caveats:
- The API is not supported on Windows XP
- There's apparently [bugs in the `WSARecv` implementation](https://github.com/libuv/libuv/blob/v1.x/src/win/winsock.c#L271) which libuv works around at least. I think this is only for UDP, though?
- We should benchmark the actual gains here to make sure it's what we expect

carllerche:
Closing due to inactivity.

PR #520, opened 2017-01-19
windows: use SetFileCompletionNotificationModes for TCP sockets (#476)

steffengy:
## SetFileCompletionNotificationModes allows us to tell the kernel not to enqueue the completion message, if it returns `SUCCESS` immediately.

This removes the **CPU-overhead** of having to enter the event loop to handle this completion message.

### Less memory overhead
You do not need as much internal (or external) buffering,
which can be quite significant, especially in a context where the event loop is only run after buffering the content in an external buffer (up to 3x less, 512 MB insteadof 1.5GB for my case).
In that regard it also behaves **similar to the POSIX (epoll) implementation**:
You should be able to repeatedly write to the underlying IO, which will still allow it to progress, without requiring any event loop interaction (similar as epoll, which returns `eagain`, miow returns true).

### **general issues**, some of them already mentioned in the original issue:
- The API is not supported on Windows XP **(unsure it this needs additional attention?)**
- We might want to check for [non-IFS LSPs](https://support.microsoft.com/de-de/help/2568167/setfilecompletionnotificationmodes-api-causes-an-i-o-completion-port-not-to-work-correctly-if-a-non-ifs-lsp-is-installed) , since if one of them is installed, **SetFileCompletionNotificationModes will not work properly if such is installed**.
[Golang](https://github.com/golang/go/blob/d0cf0421717a93b705efcbce0770a24361582445/src/net/fd_windows.go#L47) does perform this check, while **libuv** doesn't seem to.
(This is probably because non-IFS LSPs are deprecated for ages)

### some **issues** or rather hurdles using it with **UDP-sockets**:
- There seems to be a bug in the underlying WSARecv implementation (which according to [this blogpost](http://www.lenholgate.com/blog/2010/01/file-skip-completion-port-on-success-and-datagram-socket-read-errors.html) which is a summary of the issue and links to the [original MSDN post](https://blogs.technet.microsoft.com/winserverperformance/2008/06/25/designing-applications-for-high-performance-part-iii/), where Microsoft confirmed the issue in the comment section.
[Golang](https://github.com/golang/go/blob/d0cf0421717a93b705efcbce0770a24361582445/src/net/fd_windows.go#L299) disables UDP because of this, while [LIBUV](https://github.com/libuv/libuv/blob/v1.x/src/win/winsock.c#L271) uses a (slightly limited) workaround.
[.NET Core](https://github.com/dotnet/corefx/pull/15141/files#diff-6c4b087518282ebcba0a74cb9bd073c3R22) disables it for <= Windows 7.
- Also the MIO api seems slightly different, which should not be any issue, but for the sake of completeness I still list it here.

This is the main reason this implements this for TCP sockets first.

### DIFF
--- a/src/sys/windows/tcp.rs
+++ b/src/sys/windows/tcp.rs
@@ -2,6 +2,7 @@ use std::fmt;
 use std::io::{self, Read, ErrorKind};
 use std::mem;
 use std::net::{self, SocketAddr};
+use std::os::windows::prelude::*;
 use std::sync::{Mutex, MutexGuard};
 
 use miow;
@@ -71,12 +72,17 @@ struct StreamInner {
     deferred_connect: Option<SocketAddr>,
     read: State<(), ()>,
     write: State<(Vec<u8>, usize), (Vec<u8>, usize)>,
+    /// whether we are instantly notified of success
+    /// (FILE_SKIP_COMPLETION_PORT_ON_SUCCESS,
+    ///  without a roundtrip through the event loop)
+    instant_notify: bool,
 }
 
 struct ListenerInner {
     iocp: ReadyBinding,
     accept: State<net::TcpStream, (net::TcpStream, SocketAddr)>,
     accept_buf: AcceptAddrsBuf,
+    instant_notify: bool,
 }
 
 enum State<T, U> {
@@ -101,6 +107,7 @@ impl TcpStream {
                         deferred_connect: deferred_connect,
                         read: State::Empty,
                         write: State::Empty,
+                        instant_notify: false,
                     }),
                 }),
             },
@@ -301,8 +308,7 @@ impl StreamImp {
     fn schedule_connect(&self, addr: &SocketAddr) -> io::Result<()> {
         unsafe {
             trace!("scheduling a connect");
-            let overlapped = miow::Overlapped::from_raw(self.inner.read.as_mut_ptr());
-            try!(self.inner.socket.connect_overlapped(addr, overlapped));
+            try!(self.inner.socket.connect_overlapped(addr, &[], self.inner.read.as_mut_ptr()));
         }
         // see docs above on StreamImp.inner for rationale on forget
         mem::forget(self.clone());
@@ -330,31 +336,31 @@ impl StreamImp {
 
         trace!("scheduling a read");
         let res = unsafe {
-            let overlapped = miow::Overlapped::from_raw(self.inner.read.as_mut_ptr());
-            self.inner.socket.read_overlapped(&mut [], overlapped)
+            self.inner.socket.read_overlapped(&mut [], self.inner.read.as_mut_ptr())
         };
         match res {
-            // TODO: investigate better handling `Ok(true)`
-            //
             // Note that `Ok(true)` means that this completed immediately and
             // our socket is readable. This typically means that the caller of
             // this function (likely `read` above) can try again as an
             // optimization and return bytes quickly.
             //
-            // Unfortunately, though, although the read completed immediately
+            // Normally, though, although the read completed immediately
             // there's still an IOCP completion packet enqueued that we're going
-            // to receive. As an ease of implementation for now we just let the
-            // completion packet drive the read completion.
+            // to receive.
             //
-            // Apparently you can configure this behavior with
+            // You can configure this behavior (miow) with
             // SetFileCompletionNotificationModes to indicate that `Ok(true)`
-            // does **not** enqueue a completion packet. We should test this out
-            // and see if it works for us.
+            // does **not** enqueue a completion packet. (This is the case
+            // for me.instant_notify)
             //
             // Note that apparently libuv has scary code to work around bugs in
             // `WSARecv` for UDP sockets apparently for handles which have had
             // the `SetFileCompletionNotificationModes` function called on them,
             // worth looking into!
+            Ok(Some(_)) if me.instant_notify => {
+                me.read = State::Ready(());
+                self.add_readiness(me, Ready::readable());
+            }
             Ok(_) => {
                 // see docs above on StreamImp.inner for rationale on forget
                 me.read = State::Pending(());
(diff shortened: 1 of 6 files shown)


nayato:
@steffengy FYI, recently merged in .NET Core: https://github.com/dotnet/corefx/pull/15141/files#diff-6c4b087518282ebcba0a74cb9bd073c3R22
The check is being done for Windows version if socket type is UDP.

src/sys/windows/tcp.rs
```diff
@@ -397,6 +405,10 @@ impl StreamImp {
             self.inner.socket.write_overlapped(&buf[pos..], overlapped)
         };
         match err {
+            Ok(true) if me.instant_notify => {
+                me.write = State::Empty;
+                self.add_readiness(me, Ready::writable());
```
alexcrichton: Here I think we should check/verify the number of bytes written, right?

steffengy: yeah, that'll need modification to MIOW though, since we'll need to pass `lpNumberOfBytesSend` to WSASend in `write_overlapped`.

```
A pointer to the number, in bytes, sent by this call if the I/O operation completes immediately.
Use NULL for this parameter if the lpOverlapped parameter is not NULL to avoid potentially erroneous results.
```

Golang and LIBUV seem to use this, which is surprising since the `errorneous results` doesn't sound good.

alexcrichton: I think you should be able to get that through `WSAGetOverlappedResult`, right? That is currently [exposed in miow](https://docs.rs/miow/0.1.5/i686-pc-windows-msvc/miow/net/trait.TcpStreamExt.html#tymethod.result) as of recently.

steffengy:
I'm also not quite sure how to add an explicit test for this, since by default
it should simply be used for everything in test_tcp?
(Maybe it would make sense to run with and without, but I don't quite see how you'd feasily implement that in those tests)

alexcrichton:
Oh yeah not so much for correctness (as you're right existing tcp tests take care of that) but more a test of right after a write you can write again.

src/sys/windows/tcp.rs
steffengy: Ok, seems to work.
Still feels slightly slower (might be worth to investigate it that's just a placebo or if doing
it like libuv/go is advantageous performance-wise.

alexcrichton: What's the benchmark you're using? I wouldn't mind trying to poke around it in my spare time to see what's up

steffengy:
@alexcrichton 
Test case is added and verified to only work with these changes.

src/sys/windows/tcp.rs
```diff
@@ -392,12 +400,26 @@ impl StreamImp {
...
         };
-        match err {
-            Ok(_) => {
+        match ret {
+            Ok(Some((transfered_bytes, _))) => {
+                me.write = if transfered_bytes == buf.len() - pos {
+                    self.add_readiness(me, Ready::writable());
+                    State::Empty
+                } else {
+                    State::Pending((buf, pos + transfered_bytes))
```
alexcrichton: In theory if this happens (which I'm still not 100% sure that it actually can) I think we need to loop here, right? Otherwise nothing's scheduled to move this from the pending to the writing state?

alexcrichton:
>  Test case is added and verified to only work with these changes.

Woo thanks!

src/sys/windows/tcp.rs
steffengy: unfortunately the code I'm currently using is to slow for cargo bench to output
an average. (~6 seconds/iteration which writes 1GB).
Since this is implemented now, the next step for me would be to profile it and get it hopefully much faster.

What you might do is bench the test I just added (maybe with writing more data tough) since that should be sufficient to see a difference?

src/sys/windows/tcp.rs
steffengy: I also have my doubts that this is actually a valid case but in theory yes.
I'll add the loop if we have some conclusive benchmark results for the `wsagetoverlappedresult` stuff (reducing the commit spam a bit by combining those changes, if the benchmarks show they make sense ).

src/sys/windows/tcp.rs
steffengy: @alexcrichton 
did some benchmarking to simply deterine the overhead of the `.result` call.
code: https://gist.github.com/steffengy/598350018d4827887b287ed0ea798850

mio/tcp.rs:
```patch
-self.inner.socket.result(overlapped).map(Some)
+Ok(Some((buf.len() - pos, 3u32)))
```


results (with .result call):

```
1,943,652 ns/iter (+/- 525,906)
2,009,193 ns/iter (+/- 1,153,769)
1,943,501 ns/iter (+/- 565,776)
```
without the .result call:
```
1,753,306 ns/iter (+/- 1,052,692)
1,724,071 ns/iter (+/- 589,609)
1,792,201 ns/iter (+/- 537,950)
```

alexcrichton:
Figured we could continue the perf discussion out here rather than on the now-hidden thread.

So it looks like you passed a pointer to [`WSASend`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms742203(v=vs.85).aspx), but according to that documentation:

> Use NULL for this parameter if the lpOverlapped parameter is not NULL to avoid potentially erroneous results. This parameter can be NULL only if the lpOverlapped parameter is not NULL.

So that makes me think that we've gotta call `WSAGetOverlappedResult` maybe? I'm not sure that that means "potentially erroneous results" though...

steffengy:
@alexcrichton 
That's what startled me (I posted it in some comment before).
[LibUv](https://github.com/libuv/libuv/blob/1a96fe33343f82721ba8bc93adb5a67ddcf70ec4/src/win/tcp.c#L882)
[Golang](https://github.com/golang/go/blob/d0cf0421717a93b705efcbce0770a24361582445/src/net/fd_windows.go#L503)
[.NET](https://referencesource.microsoft.com/#System/net/System/Net/Sockets/Socket.cs,4046)
[.NET Core](https://github.com/dotnet/corefx/blob/684af05e70b345eecd503e23d0d6d4efd67f9130/src/System.Net.Sockets/src/System/Net/Sockets/SocketAsyncEventArgs.Windows.cs#L671)

So if even Microsoft uses it in it's .NET implementations, I guess the documentation
overstates the importance.

steffengy:
MIOW is refactored to return the size directly, when valid.
So we just saved some more overhead.

Related PR: https://github.com/alexcrichton/miow/pull/8

PR #525, opened 2017-01-23
Add from_stream methods

GuillaumeGomez:
I'll need it for `tokio`. For the record, the conversation about this is [here](https://github.com/tokio-rs/tokio-core/pull/152).

### DIFF
--- a/src/net/tcp.rs
+++ b/src/net/tcp.rs
@@ -71,6 +71,15 @@ impl TcpStream {
         })
     }
 
+    /// Creates a new `TcpStream` from a standard `net::TcpStream`.
+    pub fn from_stream(stream: net::TcpStream) -> io::Result<TcpStream> {
+        try!(stream.set_nonblocking(true));
+        Ok(TcpStream {
+            sys: sys::TcpStream::from_stream(stream),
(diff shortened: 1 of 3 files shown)


tailhook:
There is a precedent `UdpSocket::from_socket`. Maybe this is a better name?

`from_stream` looks like it relates to `futures::Stream` which it doesn't.

GuillaumeGomez:
Sure.

alexcrichton:
These methods are not unsafe, and I'd personally advocate for a `from_stream` name because we already have `connect_stream` on `TcpStream` and there's already a `TcpListener::from_listener`. You should not add `TcpListener::from_socket`, it already exists.

GuillaumeGomez:
@alexcrichton: I don't see any `from_socket` equivalent for `TcpListener`. Or maybe I missed it?

GuillaumeGomez:
Updated.

alexcrichton:
Looks like there's a compilation error on Windows and this also needs to call `set_nonblocking`, likely on both sides.

GuillaumeGomez:
Seems all good now.

PR #528, opened 2017-01-28
Fix broken links and an outdated usage in mio's doc.

0gajun:
I found broken links and an outdated usage in mio's documentation.

Since v0.6, EventLoop and Handler are deprecated.
However, top level of mio's documentation still shows v0.5.x's usage instruction in v0.6.4 doc.

So, I fixed in this way.
Please review this!  🙂

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,9 +12,8 @@
 //!
 //! # Usage
 //!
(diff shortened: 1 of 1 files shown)


carllerche:
Thanks 👍 On dev: 2a2ed71be1edbd99aabbdfaaed5f43372ef2abf0

PR #529, opened 2017-01-28
Assorted kqueue fixes/improvements

nagisa:
See the commits for description

### DIFF
--- a/src/sys/unix/kqueue.rs
+++ b/src/sys/unix/kqueue.rs
@@ -260,14 +260,11 @@ fn does_not_register_rw() {
         type Message = ();
     }
 
-
-    // registering kqueue fd will fail if write is requested
+    // registering kqueue fd will fail if write is requested (On anything but some versions of OS
+    // X)
     let kq = unsafe { libc::kqueue() };
     let kqf = EventedFd(&kq);
     let mut evtloop = EventLoopBuilder::new().build::<Nop>().expect("evt loop builds");
-    evtloop.register(&kqf, Token(1234), Ready::readable() | Ready::writable(),
-                     PollOpt::edge() | PollOpt::oneshot()).unwrap_err();
-    evtloop.deregister(&kqf).unwrap();
     evtloop.register(&kqf, Token(1234), Ready::readable(),
                      PollOpt::edge() | PollOpt::oneshot()).unwrap();
 }
--- a/src/sys/unix/kqueue.rs
+++ b/src/sys/unix/kqueue.rs
@@ -1,5 +1,4 @@
 use std::{cmp, fmt, ptr};
-use std::cell::RefCell;
 use std::os::raw::c_int;
 use std::os::unix::io::RawFd;
 use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


src/sys/unix/kqueue.rs
```diff
@@ -35,7 +36,7 @@ impl Selector {
         Ok(Selector {
             id: id,
             kq: kq,
-            changes: RefCell::new(KeventList(Vec::new())),
+            changes: RefCell::new(KeventList(Vec::with_capacity(32))),
```
nagisa: I feel like a better option here would be to use a statically sized ring-like buffer, but that would increase size of Selector significantly, which I’m not sure is desired.

carllerche:
Thanks for the PR.

The commit that removes the RefCell looks fine to me.

I do have a question about removing the `EV_DISABLE`. The reason it was originally used was to handle `reregister`. Specifically, if you register a socket w/ `Readable + Writable`, then you reregister it with only `Readable`, that should prevent any writable events from being fired. Since kqueue tracks readable and writable separately, it seems that this change would mean that there is no way to "reregister" a FD to remove an interest. Is this correct?

nagisa:
> I do have a question about removing the EV_DISABLE. The reason it was originally used was to handle reregister. Specifically, if you register a socket w/ Readable + Writable, then you reregister it with only Readable, that should prevent any writable events from being fired. Since kqueue tracks readable and writable separately, it seems that this change would mean that there is no way to "reregister" a FD to remove an interest. Is this correct?

In the 2nd commit I take care to ensure mio `EV_DELETE`s the writable event in the scenario you describe, essentially preserving the semantics of the code previously.

Since in the 2nd commit code also checks registration errors on a per-`kevent` basis,  so it is able to precisely ignore the errors that arise when event that’s being `EV_DELETE`d does not exist. This is still preferable over the `EV_DISABLE` approach, due to problem described in the 1st commit (not all kinds of descriptors can be registered with `EV_WRITE|EV_DISABLE`).

`EV_DELETE` approach may also be more efficient during actual polling, but I haven’t confirmed that.

carllerche:
Ok, that makes sense. I think this is a good change.

carllerche:
Hmm, so the tests don't pass locally for me (OS X 10.12.2)

It looks like registering a kqueue handle w/ write interest works?

```

running 1 test
test sys::unix::kqueue::does_not_register_rw ... FAILED

failures:

---- sys::unix::kqueue::does_not_register_rw stdout ----
	thread 'sys::unix::kqueue::does_not_register_rw' panicked at 'called `Result::unwrap_err()` on an `Ok` value: ()', ../src/libcore/result.rs:837
note: Run with `RUST_BACKTRACE=1` for a backtrace.

nagisa:
It passes on Travis’ 10.11 OS X and all of the BSDs I’ve tested (which is FreeBSD). Must be a OS X bug or maybe undocumented (apple likes those) and intended change.

I’m fine with removing the failing check, as really only the other one matters.

carllerche:
Thanks, merged 11e6cb32df2152cd2b0fd9668f42547f55bcdbc5

PR #534, opened 2017-02-03
Fixes a race in Registration drop

### LINKED ISSUE #532: Panic when Registration is dropped from non-main thread
We have a project, which uses mio + tokio-core + futures + futures-cpupool to implement some network protocol (basically - sending and responding to messages). Everything works fine, but on rare occasion (too rare actually -it is a big problem to reproduce it), our main thread (which polls tokio-core event loop) panics with next stack (it is about v0.6.4, but on latest master issue remains):
```
  10:     0x5579b378c5af - mio::poll::Poll::poll::hc960007ac1ed502f
                        at /buildslave/rust-buildbot/slave/stable-dist-rustc-linux/build/obj/../src/libcore/macros.rs:21
                        at /home/svyatonik/.cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.6.4/src/poll.rs:816
                        at /home/svyatonik/.cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.6.4/src/poll.rs:744
                        at /home/svyatonik/.cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.6.4/src/poll.rs:654
                        at /home/svyatonik/.cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.6.4/src/poll.rs:251
  11:     0x5579b37863df - tokio_core::reactor::Core::poll::h43f4b499158d6eeb
                        at /home/svyatonik/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-core-0.1.4/src/reactor/mod.rs:272
```

So this unwrap leads to panic:
https://github.com/carllerche/mio/blob/master/src/poll.rs#L816
Giving that `head` in this `unlink` method is actually `ReadinessQueueInner::head_all_nodes`, I have examined `poll.rs` to find how it is changed. Here I have found one suspicious method.  `head_all_nodes` is set to `None` here:
https://github.com/carllerche/mio/blob/master/src/poll.rs#L702
And it only filled back with `Some` couple of lines above:
https://github.com/carllerche/mio/blob/master/src/poll.rs#L710

I have a test for this situation in this commit:
https://github.com/svyatonik/mio/commit/58e535e6081573db98b546247f3af6c1c1f0b974
If you'll run it with `reset && cargo test -- drop_registration_from_non_main_thread`, it will fail eventually.

But I'm still not sure - if that's a `RegistrationQueue` issue, or is it improper usage? Is it OK to drop `Registration`-s from cpupool threads, or not?
If it is not OK, then I will probably move this issue to `futures-rs`, because sometimes it is actually dropped from other threads - I have reproduced it while debugging. Drop is starting here:
https://github.com/alexcrichton/futures-rs/blob/futures-cpupool-0.1.2/futures-cpupool/src/lib.rs#L267

carllerche:
In certain cases, a race condition caused registration nodes to be
unlinked twice.  # Please enter the commit message for your changes.
Lines starting. Instead of communicating drop via the events atomic. use
the last atomic that will be touched by the SetRegistration half. This
could either be the readiness queue pointer OR the queued flag.

(Hopefully) fixes #532

The test cases provided in the issue does not fail anymore, but this may just means that it only "works for me". @svyatonik would you be able to see if you can reproduce the issue with this PR?

cc @alexcrichton

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -172,12 +172,6 @@ impl Ready {
         Ready(0x008)
     }
 
-    // Private
-    #[inline]
-    fn drop() -> Ready {
-        Ready(0x10)
-    }
-
     #[inline]
     pub fn all() -> Ready {
         Ready::readable() |
@@ -188,7 +182,7 @@ impl Ready {
 
     #[inline]
     pub fn is_none(&self) -> bool {
-        (*self & !Ready::drop()) == Ready::none()
+        *self == Ready::none()
     }
 
     #[inline]
@@ -284,8 +278,7 @@ impl fmt::Debug for Ready {
             (Ready::readable(), "Readable"),
             (Ready::writable(), "Writable"),
             (Ready::error(),    "Error"),
-            (Ready::hup(),      "Hup"),
-            (Ready::drop(),     "Drop")];
+            (Ready::hup(),      "Hup")];
 
         try!(write!(fmt, "Ready {{"));
 
@@ -354,14 +347,6 @@ pub fn from_usize(events: usize) -> Ready {
 /// set.
 pub fn is_empty(events: Ready) -> bool {
     events.0 == 0
-}
-
-pub fn is_drop(events: Ready) -> bool {
-    events.contains(Ready::drop())
-}
-
-pub fn drop() -> Ready {
-    Ready::drop()
 }
 
 // Used internally to mutate an `Event` in place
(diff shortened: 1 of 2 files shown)


svyatonik:
I'll check it out tonight. But it also takes a lot of time for me to reproduce this issue in our project.

carllerche:
I would hold off for now. It looks like there are issues. I will
investigate soon.

carllerche:
I'm going to close this PR in favor of #537, but that is a pretty aggressive change, so I would not recommend trying it out yet either...

PR #537, opened 2017-02-06
Rewrite the readiness queue

### LINKED ISSUE #532: Panic when Registration is dropped from non-main thread

carllerche:
The current readiness queue implementation has some concurrency bugs. So,
instead of fixing the bug, let's rewrite everything! :)

But more seriously, the current implementation has some problems (besides
having bugs):

* It's complicated
* It is not fair when using level-triggered notifications
* It is !Sync

This implementation will hopefully resolve the bugs, as well as simplifying
the implementation by reducing all the critical concurrent state for a
given node to a single atomic variable. It also changes the queueing
strategy from a stack to an actual queue based on the 1024cores MPSC
queue. However, this queue is able to reuse the nodes.

# TODO

* [x] Rewrite queue
* [x] Synchronize call to `poll`
* [x] Get tests to pass
* [x] Add custom evented tests
* [x] Check for performance regression.
* [x] Add multi-threaded poll tests
* [x] Code review

Fixes #532

### DIFF
--- a/benches/bench_poll.rs
+++ b/benches/bench_poll.rs
@@ -1,0 +1,52 @@
+#![feature(test)]
+
+extern crate mio;
+extern crate test;
+
+use mio::*;
+use test::Bencher;
+use std::sync::Arc;
+use std::thread;
+
+#[bench]
+fn bench_poll(bench: &mut Bencher) {
+    const NUM: usize = 10_000;
+    const THREADS: usize = 4;
+
+    let poll = Poll::new().unwrap();
+    let mut events = Events::with_capacity(1024);
+
+    let mut registrations = vec![];
+    let mut set_readiness = vec![];
+
+    for i in 0..NUM {
+        let (r, s) = Registration::new(
+            &poll,
+            Token(i),
+            Ready::readable(),
+            PollOpt::edge());
+
+        registrations.push(r);
+        set_readiness.push(s);
+    }
+
+    let set_readiness = Arc::new(set_readiness);
+
+    bench.iter(move || {
+        for mut i in 0..THREADS {
+            let set_readiness = set_readiness.clone();
+            thread::spawn(move || {
+                while i < NUM {
+                    set_readiness[i].set_readiness(Ready::readable()).unwrap();
+                    i += THREADS;
+                }
+            });
+        }
+
+        let mut n = 0;
+
+        while n < NUM {
+            n += poll.poll(&mut events, None).unwrap();
+        }
+    })
+}
--- a/src/event.rs
+++ b/src/event.rs
@@ -7,28 +7,28 @@ pub struct PollOpt(usize);
 
 impl PollOpt {
     #[inline]
-    pub fn edge() -> PollOpt {
-        PollOpt(0x020)
-    }
-
-    #[inline]
     pub fn empty() -> PollOpt {
         PollOpt(0)
     }
 
     #[inline]
+    pub fn edge() -> PollOpt {
+        PollOpt(0b0001)
+    }
+
+    #[inline]
     pub fn level() -> PollOpt {
-        PollOpt(0x040)
+        PollOpt(0b0010)
     }
 
     #[inline]
     pub fn oneshot() -> PollOpt {
-        PollOpt(0x080)
+        PollOpt(0b0100)
     }
 
     #[inline]
     pub fn urgent() -> PollOpt {
-        PollOpt(0x100)
+        PollOpt(0b1000)
     }
 
     #[inline]
@@ -154,28 +154,22 @@ impl Ready {
 
     #[inline]
     pub fn readable() -> Ready {
-        Ready(0x001)
+        Ready(0b0001)
     }
 
     #[inline]
     pub fn writable() -> Ready {
-        Ready(0x002)
+        Ready(0b0010)
     }
 
     #[inline]
     pub fn error() -> Ready {
-        Ready(0x004)
+        Ready(0b0100)
     }
 
     #[inline]
     pub fn hup() -> Ready {
-        Ready(0x008)
-    }
-
-    // Private
-    #[inline]
-    fn drop() -> Ready {
-        Ready(0x10)
+        Ready(0b1000)
     }
 
     #[inline]
@@ -188,7 +182,12 @@ impl Ready {
 
     #[inline]
     pub fn is_none(&self) -> bool {
-        (*self & !Ready::drop()) == Ready::none()
+        *self == Ready::none()
+    }
+
+    #[inline]
+    pub fn is_some(&self) -> bool {
+        !self.is_none()
     }
 
     #[inline]
@@ -284,8 +283,7 @@ impl fmt::Debug for Ready {
             (Ready::readable(), "Readable"),
             (Ready::writable(), "Writable"),
             (Ready::error(),    "Error"),
-            (Ready::hup(),      "Hup"),
-            (Ready::drop(),     "Drop")];
+            (Ready::hup(),      "Hup")];
 
         try!(write!(fmt, "Ready {{"));
 
@@ -342,26 +340,20 @@ impl Event {
  *
  */
 
-pub fn as_usize(events: Ready) -> usize {
+pub fn ready_as_usize(events: Ready) -> usize {
     events.0
 }
 
-pub fn from_usize(events: usize) -> Ready {
+pub fn opt_as_usize(opt: PollOpt) -> usize {
+    opt.0
+}
+
+pub fn ready_from_usize(events: usize) -> Ready {
     Ready(events)
 }
 
-/// Returns true if the `Ready` does not have any public OR private flags
-/// set.
-pub fn is_empty(events: Ready) -> bool {
-    events.0 == 0
-}
-
-pub fn is_drop(events: Ready) -> bool {
-    events.contains(Ready::drop())
-}
-
-pub fn drop() -> Ready {
-    Ready::drop()
+pub fn opt_from_usize(opt: usize) -> PollOpt {
+    PollOpt(opt)
 }
 
 // Used internally to mutate an `Event` in place
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -1,14 +1,62 @@
 use {sys, Evented, Token};
 use event::{self, Ready, Event, PollOpt};
-use std::{fmt, io, mem, ptr, usize};
-use std::cell::{UnsafeCell, Cell};
-use std::isize;
-use std::marker;
-use std::sync::Arc;
-use std::sync::atomic::{AtomicUsize, AtomicPtr, Ordering};
-use std::time::Duration;
+use std::{fmt, io, ptr, usize};
+use std::cell::UnsafeCell;
+use std::{ops, isize};
+use std::sync::{Arc, Mutex, Condvar};
+use std::sync::atomic::{self, AtomicUsize, AtomicPtr, AtomicBool};
+use std::sync::atomic::Ordering::{self, Acquire, Release, AcqRel, Relaxed, SeqCst};
+use std::time::{Duration, Instant};
 
-const MAX_REFCOUNT: usize = (isize::MAX) as usize;
+// Poll is backed by two readiness queues. The first is a system readiness queue
+// represented by `sys::Selector`. The system readiness queue handles
+// notifications provided by the system, such as TCP and UDP. The second
+// readiness queue is implemented in user space by `ReadinessQueue`. It provides
+// a way to implement purely userspace `Evented` types.
+//
+// `ReadinessQueue` is is backed by a MPSC queue that supports reuse of linked
+// list nodes. This significantly reduces the number of required allocations.
+// Each `Registration` / `SetReadiness` pair allocates a single readiness node
+// that is used for the lifetime of the registration.
+//
+// The readiness node also includes a single atomic variable, `state` that
+// tracks most of the state associated with the registration. This includes the
+// current readiness, interest, poll options, and internal state. When the node
+// state is mutated, it is queued in the MPSC channel. A call to
+// `ReadinessQueue::poll` will dequeue and process nodes. The node state can
+// still be mutated while it is queued in the channel for processing.
+// Intermediate state values do not matter as long as the final state is
+// included in the call to `poll`. This is the eventually consistent nature of
+// the readiness queue.
+//
+// The readiness node is ref counted using the `ref_count` field. On creation,
+// the ref_count is initialized to 3: one `Registration` handle, one
+// `SetReadiness` handle, and for the readiness queue. Since the readiness queue
+// doesn't *always* hold a handle to the node, we don't use the Arc type for
+// managing ref counts (this is to avoid constantly incrementing and
+// decrementing the ref count when pushing & popping from the queue). When the
+// `Registration` handle is dropped, the `dropped` flag is set on the node, then
+// the node is pushed into the registration queue. When Poll::poll pops the
+// node, it sees the drop flag is set, and decrements it's ref count.
+//
+// The MPSC queue is a modified version of the intrusive MPSC node based queue
+// described by 1024cores [1].
+//
+// The first modification is that two markers are used instead of a single
+// `stub`. The second marker is a `sleep_marker` which is used to signal to
+// producers that the consumer is going to sleep. This sleep_marker is only used
+// when the queue is empty, implying that the only node in the queue is
+// `end_marker`.
+//
+// The second modification is an `until` argument passed to the dequeue
+// function. When `poll` encounters a level-triggered node, the node will be
+// immediately pushed back into the queue. In order to avoid an infinite loop,
+// `poll` before pushing the node, the pointer is saved off and then passed
+// again as the `until` argument. If the next node to pop is `until`, then
+// `Dequeue::Empty` is returned.
+//
+// [1] http://www.1024cores.net/home/lock-free-algorithms/queues/intrusive-mpsc-node-based-queue
+
 
 /// Polls for readiness events on all registered values.
 ///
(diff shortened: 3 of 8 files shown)


carllerche:
Another win of this implementation, I added a simple benchmark, and this PR was 60% faster than dev:

```
**New impl**
test bench_poll ... bench:     581,817 ns/iter (+/- 14,068)

**Current impl**
test bench_poll ... bench:     968,455 ns/iter (+/- 94,963)
```

src/poll.rs
```diff
@@ -68,95 +113,155 @@ pub struct Registration {
...
-struct ReadyRef {
-    ptr: *mut ReadinessNode,
+    // Tail of the readiness queue.
+    //
+    // Only accessed by Poll::poll. Coordination will be handled by the poll fn
+    tail_readiness: UnsafeCell<*mut ReadinessNode>,
+
+    // Fake readiness node used to punctuate the end of the readiness queue.
+    // Before attempting to read from the queue, this node is inserted in order
+    // to partition the queue between nodes that are "owened" by the dequeue end
```
shepmaster: typo: owened.

benches/bench_poll.rs
```diff
@@ -0,0 +1,53 @@
...
+
+use mio::*;
+use test::Bencher;
+use std::sync::Arc;
+use std::thread;
+
+#[bench]
+fn bench_poll(bench: &mut Bencher) {
+    const NUM: usize = 10_000;
+    // const NUM: usize = 8;
```
shepmaster: Commented-out code

src/poll.rs
```diff
@@ -68,95 +113,155 @@ pub struct Registration {
...
+    // Each call to `update` will attempt to swap `update_lock` from `false` to
+    // `true`. If the CAS succeeds, the thread has obtained the update lock. If
+    // the CAS fails, then the `update` call returns immediately and the update
+    // is discarded.
+    update_lock: AtomicBool,
 
-    // Tracks if the node is queued for readiness using the MSB, the
-    // rest of the usize is the readiness delay.
-    queued: AtomicUsize,
+    // Number of outstandin Registration nodes
```
shepmaster: Typo: outstandin

src/poll.rs
```diff
@@ -393,115 +710,202 @@ impl fmt::Debug for Registration {
...
+        if self.update_lock.compare_and_swap(false, true, Acquire) {
+            // The lock is already held. Discard the update
             return Ok(());
         }
 
-        if self.queue_for_processing() {
-            try!(self.queue.wakeup());
-        }
+        // Relaxed ordering is acceptable here as the only memory that needs to
+        // be visible as part of the update are hte `token_*` variables, and
```
shepmaster: Typo: `hte`

src/poll.rs
```diff
@@ -393,115 +710,202 @@ impl fmt::Debug for Registration {
...
+            // to Poll::poll.
+            if curr_token != token {
+                next.set_token_write_pos(next_token_pos);
+            }
+
+            // Update readiness and poll opts
+            next.set_interest(interest);
+            next.set_poll_opt(opt);
+
+            // If there is effecitve readiness, the node will need to be queued
```
shepmaster: Typo: `effecitve`

src/poll.rs
```diff
@@ -564,107 +960,122 @@ impl Drop for RegistrationInner {
...
-                            break;
-                        }
+        // `until` is set with the first node that gets re-enqueued due to being
+        // set to have level-triggered notifications. This prevents an infinite
+        // loop where `Poll::poll` will keep dequeuing nodes it enqueues.
+        let mut until = ptr::null_mut();
+
+        'outer:
+        while dst.len() < dst.capacity() {
+            // Dequeue the a node. If the queue is in an inconsistent state,
```
shepmaster: Typo? "the a"

src/poll.rs
```diff
@@ -673,289 +1084,411 @@ impl ReadinessQueue {
...
+                *self.inner().tail_readiness.get() = next;
+                tail = next;
+                next = (*next).next_readiness.load(Acquire);
+            }
 
-            // Update the ref, use release ordering to ensure that mutations to
-            // previous atomics are visible if the mutation to the head pointer
-            // is.
-            let next_head = self.inner().head_readiness.compare_and_swap(curr_head, node.ptr, Ordering::Release);
+            // Only ned to check `until` at this point. `until` is either null,
```
shepmaster: typo: `ned`

src/poll.rs
```diff
@@ -673,289 +1084,411 @@ impl ReadinessQueue {
...
+                return Dequeue::Data(tail);
+            }
+
+            Dequeue::Inconsistent
         }
     }
 
-    fn unlink_node(&self, mut node: ReadyRef) -> Box<ReadinessNode> {
-        node.as_mut().unwrap().unlink(&mut self.inner_mut().head_all_nodes)
+    // Attempts to state to sleeping. This involves changing `head_readiness`
```
shepmaster: Wording? `Attempts to state to sleeping.`

src/event.rs
```diff
@@ -8,7 +8,7 @@ pub struct PollOpt(usize);
 impl PollOpt {
     #[inline]
     pub fn edge() -> PollOpt {
-        PollOpt(0x020)
+        PollOpt(0x01)
```
nagisa: May I suggest using bit literals (`0b0001`) here to better convey the intent?

src/poll.rs
```diff
@@ -52,14 +90,21 @@ const MAX_REFCOUNT: usize = (isize::MAX) as usize;
...
-
     // Platform specific IO selector
     selector: sys::Selector,
 
     // Custom readiness queue
     readiness_queue: ReadinessQueue,
+
+    // Use an atomic to first check if a full lock will be required. This is a
+    // fast-path check for single threaded cases avoiding the extra syscall
+    lock_state: AtomicUsize,
```
nagisa: I find this fast-path check dubious. Does it really have any significant performance benefits? 

I’m asking this, firstly because the `Mutex` is supposed to do this itself – user-space mutexes are supported on all major systems except macOS (which still does not syscall if no contention is present anyway) and `Mutex` impl should already be using them. And secondly, because implementing locks correctly is hard and if we absolutely must do this, its probably better to just use [parking_lot](https://crates.io/crates/parking_lot).

nagisa: I ran the benchmark with this lock_state “optimisation” naively removed and it was actually a tiny bit faster than when this fast-path is present on my system.

src/poll.rs
```diff
@@ -393,115 +717,202 @@ impl fmt::Debug for Registration {
...
+            // The node is only dropped once all `Registration` handles are
+            // dropped. Only `Registration` can call `update`.
+            debug_assert!(!state.is_dropped());
 
-        if !self.queue.identical(readiness_queue) {
-            return Err(io::Error::new(io::ErrorKind::Other, "registration registered with another instance of Poll"));
+            // Update the write token position, this will also release the token
+            // to Poll::poll.
+            if curr_token != token {
+                next.set_token_write_pos(next_token_pos);
```
alexcrichton: I think I'm missing something, but could this unconditionally happen? We only change `next_token_pos` in the case the tokens are different, so I'd naively expect this to be a noop if the tokens were equal (e.g. could happen unconditionally)

src/poll.rs
```diff
@@ -673,289 +1091,412 @@ impl ReadinessQueue {
...
-            registration_data: UnsafeCell::new(RegistrationData::new(token, interest, opts)),
-            next_readiness: ReadyRef::none(),
-            events: AtomicUsize::new(0),
-            queued: AtomicUsize::new(0),
-            ref_count: AtomicUsize::new(ref_count),
+            state: AtomicState::new(interest, opt),
+            // Only the first token is set, the others are initialized to 0
+            token_0: UnsafeCell::new(token),
+            token_1: UnsafeCell::new(Token(0)),
+            token_2: UnsafeCell::new(Token(1)),
```
alexcrichton: Was this intended to be 0?

src/poll.rs
```diff
@@ -68,95 +113,155 @@ pub struct Registration {
...
+    // `true`. If the CAS succeeds, the thread has obtained the update lock. If
+    // the CAS fails, then the `update` call returns immediately and the update
+    // is discarded.
+    update_lock: AtomicBool,
 
-    // Tracks if the node is queued for readiness using the MSB, the
-    // rest of the usize is the readiness delay.
-    queued: AtomicUsize,
+    // Number of outstanding Registration nodes
+    num_registration: AtomicUsize,
```
alexcrichton: Reading over all this again, I think this field may no longer be used? It's never incremented and is only decremented in `Drop for Registration`, so I wonder if this was added at a point where it was intended that `Registration` is `Clone`?

src/poll.rs
```diff
@@ -360,29 +634,79 @@ pub fn selector(poll: &Poll) -> &sys::Selector {
...
+        is_send::<SetReadiness>();
+        is_sync::<SetReadiness>();
+
+        // Clone handle to the readiness queue, this bumps the ref count
+        let queue = poll.readiness_queue.clone();
+
+        // Allocate the registration node. The new node will have `ref_count`
+        // set to 3, and `num_registration` set to
+        // 1. The 3 ref_counts represent ownership by one SetReadiness, one
+        // Registration, and the Poll handle.
```
alexcrichton: Hm ok so reading some more, I wonder if this is a source of a memory leak? If I create a `Registration` and `SetReadiness` pair, never call `update`, and then drop them, I think the refcount will still be at one and it'll leak?

Out of curiosity, was `Arc<ReadinessNode>` not possible here?

src/poll.rs
```diff
@@ -68,95 +113,155 @@ pub struct Registration {
...
+    // Unsafe pointer to the registration's node. The node is ref counted. This
+    // cannot "simply" be tracked by an Arc because `Poll::poll` has an implicit
+    // handle though it isn't stored anywhere. In other words, `Poll::poll`
+    // needs to decrement the ref count before the node is freed.
+    node: *mut ReadinessNode,
 }
 
 #[derive(Clone)]
 struct ReadinessQueue {
     inner: Arc<UnsafeCell<ReadinessQueueInner>>,
```
alexcrichton: Is this `UnsafeCell` required? It looks like `ReadinessQueueInner` already has all the proper mutability related to `UnsafeCell` and such, so this seems like it should be able to be `Arc<ReadinessQueueInner>`

src/poll.rs
```diff
@@ -564,107 +967,122 @@ impl Drop for RegistrationInner {
...
-                        if queued == next {
-                            break;
-                        }
+        // `until` is set with the first node that gets re-enqueued due to being
+        // set to have level-triggered notifications. This prevents an infinite
+        // loop where `Poll::poll` will keep dequeuing nodes it enqueues.
+        let mut until = ptr::null_mut();
+
+        'outer:
+        while dst.len() < dst.capacity() {
```
alexcrichton: This actually seems like an interesting starvation problem, I wonder if we should always reserve at least a few slots for custom events when passing `Events` down to the `sys::Selector` to ensure that a few custom events can get returned if there are any?

(this is totally fine to leave as a TODO)

src/poll.rs
```diff
@@ -564,107 +967,122 @@ impl Drop for RegistrationInner {
...
+            // Dequeue a node. If the queue is in an inconsistent state, then
+            // stop polling. `Poll::poll` will be called again shortly and enter
+            // a syscall, which should be enough to enable the other thread to
+            // finish the queuing process.
+            let ptr = match self.dequeue_node(until) {
+                Dequeue::Empty | Dequeue::Inconsistent => break,
```
alexcrichton: This handling of `Inconsistent` will make for some interesting documentation I think. We basically need to document zero guarantees about the return value of poll. This clause means that code like this may fail:

```rust
node.set_readiness(Read::readable());
poll.poll(&mut events, Some(Duration::new(0))); // may return 0
```

That is, this inconsistent state could be induced by another thread which blocks all other threads from having their notifications seen, including the current thread.

Now those semantics seem totally ok to me as any practical usage of mio is just spinning in `poll` anyway, but we may want to be careful (when documenting everything) to mention this.

src/poll.rs
```diff
@@ -673,289 +1091,412 @@ impl ReadinessQueue {
...
         }
 
-        ReadyList { head: ReadyRef::new(head) }
+        Ok(())
     }
 
-    fn new_readiness_node(&self, token: Token, interest: Ready, opts: PollOpt, ref_count: usize) -> ReadyRef {
-        let mut node = Box::new(ReadinessNode::new(token, interest, opts, ref_count));
-        let ret = ReadyRef::new(&mut *node as *mut ReadinessNode);
+    fn enqueue_node(&self, node: &ReadinessNode) -> bool {
```
alexcrichton: Could you add a comment here that this is the 1024 cores push?

src/poll.rs
```diff
@@ -673,289 +1091,412 @@ impl ReadinessQueue {
...
-        ret
+            prev == self.sleep_marker()
+        }
     }
 
-    /// Prepend the given node to the head of the readiness queue. This is done
-    /// with relaxed ordering. Returns true if `Poll` needs to be woken up.
-    fn prepend_readiness_node(&self, mut node: ReadyRef) -> bool {
-        let mut curr_head = self.inner().head_readiness.load(Ordering::Relaxed);
+    /// Must only be called in `poll`
```
alexcrichton: Could you mention here as well that this is the 1024 cores pop?

I'd also be fine if you wanted to make this fn unsafe and document that it's only safe to call from `poll`

src/poll.rs
alexcrichton: I think maybe a solution to this would be:

* The refcount in this function should be 2
* In `enqueue_node` the refcount is increases
* In `poll` when it's pulled off the refcount is decreased

That matches the semantics of if `Arc` were used though, so may be good to use that (if possible, still unsure if it is)

src/poll.rs
```diff
@@ -564,107 +967,122 @@ impl Drop for RegistrationInner {
...
+                if opt.is_edge() {
+                    // Mark the node as dequeued
+                    next.set_dequeued();
+
+                    if opt.is_oneshot() && readiness.is_some() {
+                        next.disarm();
                     }
+                } else if readiness.is_none() {
+                    // The event needs to be queued up again. This typically
+                    // happens when using level-triggered notifications.
```
alexcrichton: Is this comment still applicable?

alexcrichton:
Oh one other minor thing I noticed is that this probably leaks memory if you drop a `Poll` with queued events (the registration queue isn't cleared), but that doesn't matter too much I think in terms of implementation details and could just be a TODO for now.

src/poll.rs
carllerche: I would say that the main "problem" with using `Mutex` directly is mostly that the implementation lives in a dylib, so it can't be inlined, optimized, etc...

src/poll.rs
```diff
@@ -226,24 +336,188 @@ impl Poll {
...
+
+                lock = match timeout {
+                    Some(to) => {
+                        let now = Instant::now();
+
+                        // Wait to be notified
+                        let (l, result) = self.condvar.wait_timeout(lock, to).unwrap();
+
+                        // Wait timed out, stop trying to wait
+                        if result.timed_out() {
```
carllerche: You are correct that there is an issue... I will rework this bit.

src/poll.rs
carllerche: As mentioned in IRC, when the dropped flag is toggled, I should have enqueued the node so that the consumer half sees that the node is dropped.

src/poll.rs
carllerche: Yes, it could happen unconditionally. The guard is probably left over

src/poll.rs
```diff
@@ -393,115 +717,202 @@ impl fmt::Debug for Registration {
...
+
+    /// Set the node's dropped flag, informing `Poll::poll` that it should
+    /// decrement its ref count.
+    fn flag_as_dropped(&self) {
+        self.state.flag_as_dropped();
+    }
+}
+
+impl ops::Deref for RegistrationInner {
+    type Target = ReadinessNode;
```
carllerche: It's mostly to not have to always have an unsafe block to deref the node. Not sure what you mean by "doesn't have any more inherent methods".

src/poll.rs
carllerche: Yes, definitely a starvation potential. Another option would be to alternate which queue gets polled first. So, first call to `poll` would poll the selector, then the readiness queue, the second would poll the readiness queue then the selector...

src/poll.rs
carllerche: I added an issue to document this. I plan on doing a full doc pass shortly after this PR lands.

nagisa:
That is not a problem with parking_lot though? And it also exposes API for
timed mutex locks as well.

PR #543, opened 2017-02-16
Fix timer thread never exit

hicqu:
See: https://github.com/carllerche/mio/issues/541


hicqu:
Introduced new bug! Sorry for that. 👎

PR #545, opened 2017-02-20
Fix test_multicast in shared-IP FreeBSD jails

asomers:
Traditional Jails, aka shared-ip jails, do not have a fully virtualized
network stack.  One result is that attempts to bind 127.0.0.1 actually
bind the jail's IP instead.  The test should not assume that localhost's
address is 127.0.0.1.


carllerche:
Thanks. I merged this here: f0cce6f83400c9f228d6c9fe527deab3411c27d2

PR #546, opened 2017-02-20
Fix and enhance documentation of TcpListener::accept()

JanZerebecki:


carllerche:
Thanks for the PR. The docs are definitely out of date.

I would suggest instead of `Err(would_block())`, maybe `Err(e)` where e is an io::Error of kind WouldBlock?

Also, I would suggest removing this ` May block if there is nothing ready to be accepted.` as it isn't true (since the socket is in non-blocking mode).

JanZerebecki:
Right, I looked at the wrong place for the call making it non-blocking. I adopted your suggestions and clarified a bit more.

carllerche:
Thanks, merged: 9792b74752c68b8c1b1263110e62edeab27503ce

PR #548, opened 2017-02-21
Restructure more code in preparation for the next release.

carllerche:


src/event_imp.rs
```diff
@@ -200,14 +227,20 @@ impl Ready {
         self.contains(Ready::writable())
     }
 
+    #[deprecated(since = "0.6.5", note = "use unix::ReadyExt instead")]
```
alexcrichton: This may be painful for a bit b/c the inherent method will trump the trait by default, so you'd have to call this like `ReadyExt::is_error(&ready)`

(just FYI)

src/poll.rs
```diff
@@ -556,7 +563,7 @@ pub struct Events {
 }
 
 /// Iterate an Events structure
-pub struct EventsIter<'a> {
+pub struct Iter<'a> {
```
alexcrichton: This naming typically works best when the `poll` module itself is public because we can't lift multiple `Iter` types into the root namespace. For `std::collections`, for example, the collections are all lifted to that module but all iterators live inside `std::collections::$relevant_module::Iter` (for example)

Making `poll` public may be somewhat difficult though

carllerche: As mentioned in irc, `Iter` is actually re-exported in an `event` module, so it should be fine

src/poll.rs
```diff
@@ -297,7 +527,98 @@ impl Poll {
     }
 
     /// Register an `Evented` handle with the `Poll` instance.
-    pub fn register<E: ?Sized>(&self, io: &E, token: Token, interest: Ready, opts: PollOpt) -> io::Result<()>
+    ///
+    /// Once registerd, the `Poll` instance will monitor the `Evented` handle
```
alexcrichton: "registerd"

src/poll.rs
alexcrichton: 👍

src/poll.rs
```diff
@@ -1289,7 +1754,7 @@ fn release_node(ptr: *mut ReadinessNode) {
         // Because `fetch_sub` is already atomic, we do not need to synchronize
         // with other threads unless we are going to delete the object. This
         // same logic applies to the below `fetch_sub` to the `weak` count.
-        if (*ptr).ref_count.fetch_sub(1, Release) != 1 {
+        if (*ptr).ref_count.fetch_sub(1, AcqRel) != 1 {
```
alexcrichton: 👻

src/poll.rs
```diff
@@ -1310,7 +1775,7 @@ fn release_node(ptr: *mut ReadinessNode) {
         // > "acquire" operation before deleting the object.
         //
         // [1]: (www.boost.org/doc/libs/1_55_0/doc/html/atomic/usage_examples.html)
-        atomic::fence(Acquire);
+        // atomic::fence(Acquire);
```
alexcrichton: intentional?

alexcrichton:
Everything looks great to me!

src/poll.rs
carllerche: Yeah, the `Release` is correct (I sure as hell hope so, because it is what `Arc` does), but this change makes the new thread sanitizer lint happy.

src/poll.rs
carllerche: Yeah, the AcqRel above... I'm going to delete this line.

carllerche:
Merged here: 1c22d7b15a42d88040e02431ac3ed82313fe1387

PR #549, opened 2017-02-22
Cargo.toml: Change documentation to point to newest version

C4K3:
So the documentation link doesn't have to be manually updated for new
releases, but rather will always point to the newest version.

It's currently pointing to a (slightly) outdated version.


carllerche:
Thanks, merged: 0e3492ad3e1c347142e53648ee7f850c7dceb33b

PR #552, opened 2017-02-24
Tweak documentation of Ready

JanZerebecki:
Current master is much better than https://docs.rs/mio/0.6.4/mio/struct.Ready.html, which confused me into thinking of e.g. an event on a specific connection until I looked at the source. Feel free to just close this PR if you don't think this improves the wording.


carllerche:
Thanks, merged: 7ea8375fc54a3eab4d0742e0258338a60a5c37fb

PR #557, opened 2017-03-04
Fix Poll::new2 changing Registration Token

### LINKED ISSUE #556: "Registration::new2()" may have a bug
Look at this code, this is an example from the [document](http://carllerche.github.io/mio/mio/struct.Registration.html): 
```rust
extern crate mio;

use mio::{Events, Ready, Registration, Poll, PollOpt, Token};
use std::thread;

fn main() {
    
    let (registration, set_readiness) = Registration::new2();

    thread::spawn(move || {
        use std::time::Duration;
        thread::sleep(Duration::from_millis(500));

        set_readiness.set_readiness(Ready::readable());
    });

    let poll = Poll::new().unwrap();
    poll.register(&registration, Token(0), Ready::all(), PollOpt::edge()).unwrap();

    let mut events = Events::with_capacity(256);

    loop {
        poll.poll(&mut events, None);

        for event in &events {
            if event.token() == Token(0) && event.readiness().is_readable() {
                return;
            }
            println!("{:?}", event.token());
        }
    }
}
```

This is no problem, we will see "Token(0)" on the screen. If we register "Token" to a different value, instead of "Token (0)":

```rust
poll.register(&registration, Token(123), Ready::all(), PollOpt::edge()).unwrap();
```
We still see the screen output "Token (0)", instead of "Token (123)". Why ?

Look at this code, which comes from the code in the [test](https://github.com/carllerche/mio/blob/master/test/test_custom_evented.rs)

```rust
fn set_readiness_before_register() {
    use std::sync::{Arc, Barrier};
    use std::thread;

    let poll = Poll::new().unwrap();
    let mut events = Events::with_capacity(128);

    for i in 0..5_000 {
        let (r, set) = Registration::new2();

        let b1 = Arc::new(Barrier::new(2));
        let b2 = b1.clone();

        let th = thread::spawn(move || {
            b2.wait();
            set.set_readiness(Ready::readable()).unwrap();
        });

        b1.wait();

        poll.register(&r, Token(0), Ready::readable(), PollOpt::edge()).unwrap();

        loop {
            let n = poll.poll(&mut events, None).unwrap();

            if n == 0 {
                continue;
            }

            assert_eq!(n, 1);
            assert_eq!(events.get(0).unwrap().token(), Token(0));
            break;
        }

        th.join().unwrap();
    }
}
```

Similarly, if we change the Token (0) to the other, this test will not pass:

```rust
        poll.register(&r, Token(456), Ready::readable(), PollOpt::edge()).unwrap();

        loop {
            let n = poll.poll(&mut events, None).unwrap();

            if n == 0 {
                continue;
            }

            assert_eq!(n, 1);
            assert_eq!(events.get(0).unwrap().token(), Token(456));
            break;
        }
```

```
failures:

---- test_custom_evented::set_readiness_before_register stdout ----
    thread 'test_custom_evented::set_readiness_before_register' panicked at 'assertion failed: `(left == right)` (left: `Token(0)`, right: `Token(456)`)', test_custom_evented.rs:53
```

carllerche:
Fixes #556

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -2392,7 +2392,7 @@ impl ReadinessState {
     #[inline]
     fn update_token_read_pos(&mut self) {
         let val = self.token_write_pos();
-        self.set(val, MASK_2, TOKEN_WR_SHIFT);
+        self.set(val, MASK_2, TOKEN_RD_SHIFT);
     }
 }
 
--- a/test/test_custom_evented.rs
+++ b/test/test_custom_evented.rs
@@ -27,7 +27,7 @@ fn set_readiness_before_register() {
     let poll = Poll::new().unwrap();
     let mut events = Events::with_capacity(128);
 
-    for i in 0..5_000 {
+    for _ in 0..5_000 {
         let (r, set) = Registration::new2();
 
         let b1 = Arc::new(Barrier::new(2));
@@ -40,7 +40,7 @@ fn set_readiness_before_register() {
 
         b1.wait();
 
-        poll.register(&r, Token(0), Ready::readable(), PollOpt::edge()).unwrap();
+        poll.register(&r, Token(123), Ready::readable(), PollOpt::edge()).unwrap();
 
         loop {
             let n = poll.poll(&mut events, None).unwrap();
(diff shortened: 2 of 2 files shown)


carllerche:
Fixed here: 9f8bb5cf9ffbccf87aed7d52c3f3bbcaa42a0769

PR #558, opened 2017-03-05
POSIX AIO support

asomers:
On BSD-based operating systems, add low-level support for queuing AIO
operations in the event loop.  This commit adds 2 new public methods:
Ready::aio and Ready::is_aio.  Full support will be added in a separate
crate.

### DIFF
--- a/src/event_imp.rs
+++ b/src/event_imp.rs
@@ -538,13 +538,34 @@ impl fmt::Debug for PollOpt {
 #[derive(Copy, PartialEq, Eq, Clone, PartialOrd, Ord)]
 pub struct Ready(usize);
 
-const READABLE: usize = 0b0001;
-const WRITABLE: usize = 0b0010;
-const ERROR: usize    = 0b0100;
-const HUP: usize      = 0b1000;
-const READY_ALL: usize = READABLE | WRITABLE | ERROR | HUP;
+const READABLE: usize = 0b00001;
+const WRITABLE: usize = 0b00010;
+const ERROR: usize    = 0b00100;
+const HUP: usize      = 0b01000;
+const AIO: usize      = 0b10000;
+const READY_ALL: usize = READABLE | WRITABLE | ERROR | HUP | AIO;
 
 impl Ready {
+    /// Returns a `Ready` representing AIO completion readiness
+    ///
+    /// See [`Poll`] for more documentation on polling.
+    ///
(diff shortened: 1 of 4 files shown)


carllerche:
The idea is fine at a high level, however there are a few questions.

If AIO is used to `Registration`, it will panic (4 bit max for Ready). Another question is how to deal w/ platforms specific readiness in the first place. This has been a question that I've been trying to sort for a bit...

asomers:
@carllerche where do you think AIO will cause a Registration panic?  I've got everything working; the full AIO support is at https://github.com/asomers/mio-aio .  My thinking regarding platform-specific differences is that `mio` should be cross-platform, `mio-X` should be platform-specific (for example, `mio-aio` for BSD and `mio-libaio` for Linux), and `tokio-file` should be cross-platform, consuming all platform-specific `mio-X` crates.

asomers:
@carllerche I see what you're talking about now.  Using `Registration` with AIO might be useful on Linux, though it wouldn't allow mixing AIO and network calls in the same event loop.  I'll push a commit to expand the bitfield size for `Readiness`.  While I'm at it, I'll expand the bitfield for `PollOpt` too, so we don't run into this problem again soon.  `kevent`supports 17 different polling flags (not all on the same operating system), so it's likely that `mio` will eventually need more than 4.

asomers:
The Appveyor failures look like some sort of test cleanup failure.  `set_readiness_before_register`is getting an event with Token(123), but the only tests that create such a Token are in `test_poll_channel.rs` and `test/test_tcp_level.rs`.  Any ideas, @carllerche ?

carllerche:
A quick update, I'm splitting out portable `Ready` events from platform specific readiness events. The PR is here: https://github.com/carllerche/mio/pull/562

Also, since this is a BSD specific bit of code, I would really like to get BSD CI running. It looks like trust has some support for it: https://github.com/japaric/trust

Regarding the test failure, maybe try rebasing off of `dev`?

asomers:
In order to test on FreeBSD, Trust needs to spin up a QEMU VM, which is extremely inefficient.  I'm considering setting up a Buildbot cluster to test certain Rust crates on FreeBSD.

Regarding the test failure, I've managed to reproduce it on Linux.  I'll probably be able to fix it in the next few days.

asomers:
I rebased this PR as #586

PR #565, opened 2017-03-09
Fix a use-after-free in the readiness queue 

alexcrichton:
It looks like registrations created with `new2` can end up causing segfaults due
to a use-after-free, and after tracking it down it looks like it's because a
`RegistrationInner` expects to have an owning reference to the underlying
readiness queue, but when we lazily initialize the `readiness_queue` field the
`new2` branch forgot to create an owning reference (unlike the `new` constructor
which does this).

This commit adds an appropriate `mem::forget` to transfer ownership of the
reference count.


carllerche:
This has been merged into dev.

PR #571, opened 2017-03-15
Fix bug in custom readiness queue

### LINKED ISSUE #570: deadlock/livelock in 0.6.5
Steps to reproduce: 

  1. Check out [this commit](https://github.com/dwrensha/zillions/commit/f849f9d7c1e94070013ae7a966b1ff9988c4d5eb) of zillions
  2. Build the tokio-core impl: `cd impls/tokio-core && cargo build --release`
  3. Build the `stresstest` executable: `cargo build --release`
  4. Run with:
```
./target/release/stresstest ./impls/tokio-core/target/release/server --server 127.0.0.1:12345 -p 1 -s 2
```

When I do this on my MacOS machine, it sometimes succeeds as I expect it to, but it also sometimes deadlocks, and it sometimes hangs while pegging a cpu.

By adding some printlns to mio, I've observed that in the livelock case, `Poll::poll()` is getting called in a tight loop, with `events.len() == 0` and `timeout == None`. The readiness queue gets polled and then immediately returns [here](https://github.com/carllerche/mio/blob/2d58be9a54da8a9ffa737e6276a4a23085bb66b1/src/poll.rs#L1907) after seeing `Dequeue::Empty`. I don't have a strong understanding of how these internals work, but this feels to me like an indication of corrupted state.

Before the commit https://github.com/carllerche/mio/commit/43a69f218957ea7f43bb821734dbfbad09078c6b everything seems to always work as I expect.

It's possible that there is a bug in my code, but I can't think of any way that the livelock would happen unless there's a bug in mio.

carllerche:
A call to `Poll::poll` could corrupt the readiness queue's linked list
under certain conditions. This commit fixes the logic around inserting
the `sleep_token`, which allows producers to know when they need to
wakeup `Poll`.

Fixes #570

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -1057,8 +1057,6 @@ impl Poll {
 
     #[inline]
     fn poll2(&self, events: &mut Events, timeout: Option<Duration>) -> io::Result<usize> {
-        let mut sleep = false;
-
         // Compute the timeout value passed to the system selector. If the
         // readiness queue has pending nodes, we still want to poll the system
         // selector for new events, but we don't want to block the thread to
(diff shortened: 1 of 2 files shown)


carllerche:
Actually, I don't think that this is correct yet

PR #574, opened 2017-03-20
Move TCP & UDP types into `net` module.

carllerche:
Motivation for this is twofold.

1) Mirror the layout in `std`
2) Fix UDP API without breaking backwards compatibility.

### DIFF
--- a/src/net/mod.rs
+++ b/src/net/mod.rs
@@ -10,5 +10,5 @@
 mod tcp;
 mod udp;
 
-pub use self::tcp::{TcpListener, TcpStream, Shutdown};
+pub use self::tcp::{TcpListener, TcpStream};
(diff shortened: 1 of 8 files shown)


src/net/mod.rs
```diff
@@ -1,41 +1,14 @@
...
+mod udp;
 
-impl Clone for SelectorId {
-    fn clone(&self) -> SelectorId {
-        SelectorId {
-            id: AtomicUsize::new(self.id.load(Ordering::SeqCst)),
-        }
-    }
-}
+pub use self::tcp::{TcpListener, TcpStream, Shutdown};
```
alexcrichton: Perhaps `Shutdown` could be left out? (I think it's just a libstd reexport, right?)

PR #583, opened 2017-04-04
kqueue: Ignore EPIPE coming out of `kevent`

### LINKED ISSUE #582: suspicion: EPIPE coming out of `register` on OSX
We're seeing some odd behavior on OSX in rust-lang/rust right now. We're using [sccache](https://github.com/mozilla/sccache) to compile LLVM, which executes a whole lot of tokio + mio on each run. We've also got a lot of [spurious failures](https://github.com/rust-lang/rust/issues/40240), one of the most notorious right now is "broken pipe", which arises during [this `spawn`](https://github.com/mozilla/sccache/blob/bbb850245430cfb444b4f4d222ce3e6dff2c5f59/src/compiler/compiler.rs#L621-L627). 

From here on out, a lot of this is guesswork unfortunately. I've never been able to personally reproduce this, but here's my hunches. The main error happening here is that *spawning a process* somehow returns `EPIPE`. Now that's pretty odd because reviewing the syscalls that `Command::spawn` takes it turns out none of them can return `EPIPE`! I would also find it quite odd  if this is the first time we find this error in process spawning, surely it seems like it would have happened before now I'd guess.

So digging more the `spawn` here actually translates to [`CommandExt::spawn_async`](https://github.com/alexcrichton/tokio-process/blob/aae23522a07400275b7263ff3836d3c97a9e0c64/src/lib.rs#L182-L200). Note that the function there can fail for more than one reason, not just `Command::spawn`. The [`register_*`](https://github.com/alexcrichton/tokio-process/blob/aae23522a07400275b7263ff3836d3c97a9e0c64/src/unix.rs#L57-L70) functions are also [fallible](https://github.com/alexcrichton/tokio-process/blob/aae23522a07400275b7263ff3836d3c97a9e0c64/src/unix.rs#L180-L203). They can either [fail in `fcntl`](https://github.com/alexcrichton/tokio-process/blob/aae23522a07400275b7263ff3836d3c97a9e0c64/src/unix.rs#L194) (which I don't think can return EPIPE) or they can [fail](https://github.com/alexcrichton/tokio-process/blob/aae23522a07400275b7263ff3836d3c97a9e0c64/src/unix.rs#L201) in [`PollEvented::new`](https://github.com/tokio-rs/tokio-core/blob/29144e6c9bb4165df31a319f0b30d72ae690366a/src/reactor/poll_evented.rs#L62) which internally bottoms out to [registering an object](https://github.com/tokio-rs/tokio-core/blob/29144e6c9bb4165df31a319f0b30d72ae690366a/src/reactor/mod.rs#L476-L481).

So at this point my best suspicion is that `register` is returning EPIPE, although I have no idea why. One interesting observation is that we've only witnessed this on OSX, not on Linux or Windows. That itself is quite interesting as well! Trawling around for information on the internet related to this I stumbled across [a commit in libevent](https://github.com/libevent/libevent/commit/1fd34ab424b45f48dfeece108910978c79a8b7f6) which translates EPIPE from `kevent` to a writable notification. (e.g. doesn't punt it back to the user). That commit claims that if you register half of a broken pipe you'll get an EPIPE. Sure enough it looks like [xnu does return EPIPE](https://github.com/opensource-apple/xnu/blob/8cf668b09f7c419fadf6081d1f1bdd6c5033e708/bsd/kern/sys_pipe.c#L1570-L1578) on a broken pipe.

So that sounds great and all until it turns out I can't even [reproduce this with a broken pipe](https://gist.github.com/alexcrichton/aad9eeace869c1b04daafc734ff624b8). That program works successfully! (no errors)

So all in all there's something that's:

* OSX specific
* Likely related to tokio/mio, but not 100% certain
* Likely related to `Poll::register`, but not certain
* Likely related to broken pipes (as this only happens under high load)

Where this leaves me I'm not entirely sure. I wanted to jot down thoughts and ideas to make sure they're written, but I'm curious what you think as well @carllerche. I could add some code to hopefully handle `EPIPE` and somehow translate it to writable/readable (as appropriate), but I have no way of testing such a change. We could deploy it to rust-lang/rust and pray it fixes things, but we wouldn't really know for a few weeks whether it actually worked.

alexcrichton:
This commit updates mio to ignore `EPIPE` coming out of `kevent`. It
turns out that older versions of OSX (10.10 and 10.11 have been
confirmed) will return an `EPIPE` when registering one half of a pipe
where the other half has been closed. On other platforms this is an ok
operation which just returns that the pipe is readable/writable (to
return an error/eof), so this brings the OSX behavior in line by
ignoring the error and moving to the next event.

Closes #582

### DIFF
--- a/src/sys/unix/kqueue.rs
+++ b/src/sys/unix/kqueue.rs
@@ -95,13 +95,37 @@ impl Selector {
                                            ::std::ptr::null())));
             for change in changes.iter() {
                 debug_assert_eq!(change.flags & libc::EV_ERROR, libc::EV_ERROR);
-                if change.data != 0 {
-                    // there’s some error, but we want to ignore ENOENT error for EV_DELETE
-                    let orig_flags = if change.filter == libc::EVFILT_READ { r } else { w };
-                    if !(change.data as i32 == libc::ENOENT && orig_flags & libc::EV_DELETE != 0) {
-                        return Err(::std::io::Error::from_raw_os_error(change.data as i32));
-                    }
+
+                // Test to see if an error happened
+                if change.data == 0 {
+                    continue
                 }
+
+                // Older versions of OSX (10.11 and 10.10 have been witnessed)
+                // can return EPIPE when registering a pipe file descriptor
+                // where the other end has already disappeared. For example code
+                // that creates a pipe, closes a file descriptor, and then
+                // registers the other end will see an EPIPE returned from
+                // `register`.
+                //
+                // It also turns out that kevent will still report events on the
+                // file descriptor, telling us that it's readable/hup at least
+                // after we've done this registration. As a result we just
+                // ignore `EPIPE` here instead of propagating it.
+                //
+                // More info can be found at carllerche/mio#582
+                if change.data as i32 == libc::EPIPE &&
+                   change.filter == libc::EVFILT_WRITE {
+                    continue
+                }
+
+                // ignore ENOENT error for EV_DELETE
+                let orig_flags = if change.filter == libc::EVFILT_READ { r } else { w };
+                if change.data as i32 == libc::ENOENT && orig_flags & libc::EV_DELETE != 0 {
+                    continue
+                }
+
+                return Err(::std::io::Error::from_raw_os_error(change.data as i32));
             }
             Ok(())
         }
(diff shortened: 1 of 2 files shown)


carllerche:
Closed by #584

PR #586, opened 2017-04-15
POSIX AIO support

asomers:
On BSD-based operating systems, add low-level support for queuing AIO
operations in the event loop.  This commit adds 3 new public methods:
UnixReady::aio, UnixReady::is_aio, and Poll.as_raw_fd.  Full support
will be added in a separate crate.

This PR is a rebase of and replaces #558

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -543,23 +543,22 @@ struct AtomicState {
 
 const MASK_2: usize = 4 - 1;
 const MASK_4: usize = 16 - 1;
-const MASK_8: usize = 256 - 1;
 const QUEUED_MASK: usize = 1 << QUEUED_SHIFT;
 const DROPPED_MASK: usize = 1 << DROPPED_SHIFT;
 
 const READINESS_SHIFT: usize = 0;
-const INTEREST_SHIFT: usize = 8;
-const POLL_OPT_SHIFT: usize = 16;
-const TOKEN_RD_SHIFT: usize = 20;
-const TOKEN_WR_SHIFT: usize = 22;
-const QUEUED_SHIFT: usize = 24;
-const DROPPED_SHIFT: usize = 25;
+const INTEREST_SHIFT: usize = 4;
+const POLL_OPT_SHIFT: usize = 8;
+const TOKEN_RD_SHIFT: usize = 12;
+const TOKEN_WR_SHIFT: usize = 14;
+const QUEUED_SHIFT: usize = 16;
+const DROPPED_SHIFT: usize = 17;
 
 /// Tracks all state for a single `ReadinessNode`. The state is packed into a
 /// `usize` variable from low to high bit as follows:
 ///
-/// 8 bits: Registration current readiness
-/// 8 bits: Registration interest
+/// 4 bits: Registration current readiness
+/// 4 bits: Registration interest
 /// 4 bits: Poll options
 /// 2 bits: Token position currently being read from by `poll`
 /// 2 bits: Token position last written to by `update`
@@ -2290,7 +2289,7 @@ impl ReadinessState {
         let interest = event::ready_as_usize(interest);
         let opt = event::opt_as_usize(opt);
 
-        debug_assert!(interest <= MASK_8);
+        debug_assert!(interest <= MASK_4);
         debug_assert!(opt <= MASK_4);
 
         let mut val = interest << INTEREST_SHIFT;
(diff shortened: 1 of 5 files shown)


asomers:
The builds have all passed.  I don't know why appveyor hasn't reported status yet.

asomers:
@carllerche have you had a chance to look at this yet?

src/poll.rs
```diff
@@ -537,22 +543,23 @@ struct AtomicState {
...
-const TOKEN_RD_SHIFT: usize = 12;
-const TOKEN_WR_SHIFT: usize = 14;
-const QUEUED_SHIFT: usize = 16;
-const DROPPED_SHIFT: usize = 17;
+const INTEREST_SHIFT: usize = 8;
+const POLL_OPT_SHIFT: usize = 16;
+const TOKEN_RD_SHIFT: usize = 20;
+const TOKEN_WR_SHIFT: usize = 22;
+const QUEUED_SHIFT: usize = 24;
+const DROPPED_SHIFT: usize = 25;
```
carllerche: I'd be very wary of changing this here. Could you elaborate more about why you are changing these constants? I'm assuming it is to enable the AIO readiness to be used in the custom readiness queue, however, I don't think that is necessary.

I was reserving the extra bits in the atomic to potentially implement other features. I would prefer it if more readiness states didn't need to be added here.

I would guess that a better option would be to panic if invalid readiness bits are passed to a `Registration` or `SetReadiness`.

src/poll.rs
```diff
@@ -1098,13 +1105,24 @@ impl Poll {
...
+}
+
 fn validate_args(token: Token, interest: Ready) -> io::Result<()> {
     if token == AWAKEN {
         return Err(io::Error::new(io::ErrorKind::Other, "invalid token"));
     }
 
-    if !interest.is_readable() && !interest.is_writable() {
-        return Err(io::Error::new(io::ErrorKind::Other, "interest must include readable or writable"));
+    if !registerable(interest) {
```
carllerche: This wouldn't be necessary if `AIO` isn't permitted in this poll path.

src/poll.rs
asomers: I'm not using AIO in the custom readiness queue, but I think it might be useful on Linux.  Linux's epoll doesn't support AIO operations, so the easiest way to make them work with mio would be to associate a signalfd with the libaio requests and register that signalfd with mio for Readable.  Whenever something arrives in the signalfd, you would parse the struct and then set an Aio event in the custom readiness queue.  Do you still think I should remove this part?

src/poll.rs
asomers: Even when using the system readiness queue, you still must call Poll.register.

src/poll.rs
carllerche: I would say, remove it in this PR and we can evaluate if it should added back at a later time when someone asks for it. I would guess that there is most likely a better way to model it on linux, but I guess we will see then :)

carllerche: We can always add more functionality, but removing is more difficult.

src/poll.rs
carllerche: hmm... ok, so FDs that trigger AIO readiness do not also have readable / writable? I guess this would be fine then...

src/poll.rs
asomers: Ok.  BTW, I know of three possible ways to do AIO with mio on Linux, but none of them is very good:
1) Use POSIX AIO.  Use a separate thread to call `aio_suspend` (similar to `select` but strictly for POSIX AIO).  When it returns, shove an event into mio's custom readiness queue.  This is the slowest option.
2) Use LibAIO.  Use a separate thread to call `io_getevents`.  When it returns, shove an event into mio's custom readiness queue.
3) Use LibAIO with a signalfd.  Register the signalfd with mio.  When it's ready, read a `signalfd_siginfo` out of it.  Either shove an AIO event into mio's custom readiness queue, or perhaps translate it inline.  This is probably the best approach, though there are complications relating to signal delivery.

src/poll.rs
asomers: That's correct.  AIO events aren't actually associated with file descriptors; they're associated with `struct aiocb` callback structures.  There's nothing readable or writable about them.

asomers:
@carllerche are you satisfied with my latest commit?

carllerche:
Looks good to me.

/cc @alexcrichton

alexcrichton:
Looks good to me!

asomers:
@carllerche @alexcrichton So is it good to merge then?  Or do you want me to squash first or something?

alexcrichton:
Sorry for the delay and thanks @asomers!

PR #590, opened 2017-04-24
Fix the timer thread never exits and can not deconstruct the pipe in …

sapsan4eg:
…poll.
@zonyitoo Are you right. Your solution in https://github.com/carllerche/mio/issues/541 solves the problem. This PR with like your example. 
@carllerche may be for you is the minor bug. But for ws-rs crate this is really bug https://github.com/housleyjk/ws-rs/issues/134.

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -55,6 +55,15 @@ struct Inner {
     wakeup_thread: thread::JoinHandle<()>,
 }
 
+impl Drop for Inner {
+    fn drop(&mut self) {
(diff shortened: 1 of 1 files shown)


carllerche:
I don't know why appveyor is hanging...

carllerche:
Fixed by #591

stusmall:
When can we expect a release including this fix? I'm hitting the ws-rs bug and would love to get it cleared up.

carllerche:
Done

PR #595, opened 2017-04-28
Make mio work for fuchsia

aleksandrpak:
This PR includes 2 changes:
1) Use ```fcntl``` instead of ```ioctl``` for ```set_nonblock``` and ```set_cloexec```
2) Use ```epoll``` implementation for fuchsia

PR #596, opened 2017-04-28
Fix usage of O_CLOEXEC in SETFL

aleksandrpak:
Fixes potential bug introduced in previous PR [#595](https://github.com/carllerche/mio/pull/595).

From SETFL man page: "On Linux this command  can  change  only  the  O_APPEND,  O_ASYNC, O_DIRECT, O_NOATIME, and O_NONBLOCK flags"

Sorry for this :)

@raphlinus knows more about our mio effort

### DIFF
--- a/src/sys/unix/io.rs
+++ b/src/sys/unix/io.rs
@@ -18,7 +18,7 @@ pub fn set_nonblock(fd: libc::c_int) -> io::Result<()> {
 
 pub fn set_cloexec(fd: libc::c_int) -> io::Result<()> {
     unsafe {
-        let flags = libc::fcntl(fd, libc::F_GETFL);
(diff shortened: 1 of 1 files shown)


src/sys/unix/io.rs
```diff
@@ -19,7 +19,7 @@ pub fn set_nonblock(fd: libc::c_int) -> io::Result<()> {
 pub fn set_cloexec(fd: libc::c_int) -> io::Result<()> {
     unsafe {
         let flags = libc::fcntl(fd, libc::F_GETFL);
```
teisenbe: F_GETFL here should be F_SETFD

alexcrichton:
Thanks!

PR #600, opened 2017-05-06
fix doc link

FuGangqiang:


carllerche:
Thanks! I merged this as part of #608 (I rebased)

PR #603, opened 2017-05-17
Fix handling of failed TCP connections on windows

Ralith:
Previously, mio unconditionally called setsockopt after a TCP connection attempt completed. If the connection had failed, this resulted in a WSAENOTCONN error being yielded in place of the actual error.

~~This PR depends on https://github.com/alexcrichton/miow/pull/10.~~

### DIFF
--- a/src/sys/windows/tcp.rs
+++ b/src/sys/windows/tcp.rs
@@ -477,7 +477,12 @@ fn read_done(status: &OVERLAPPED_ENTRY) {
     // If a read didn't complete, then the connect must have just finished.
     trace!("finished a connect");
 
-    match me2.inner.socket.connect_complete() {
+    // By guarding with socket.result(), we ensure that a connection
+    // was successfully made before performing operations requiring a
+    // connected socket.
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! Isn't this supposed to be handled by `WSAGetOverlappedResult` though? (which doesn't use unstable internals?) I figured this'd call [`TcpStream::result`](https://docs.rs/miow/0.2/x86_64-pc-windows-msvc/miow/net/trait.TcpStreamExt.html#tymethod.result) to get the result?

Ralith:
Good catch. I've updated the PR to take that approach instead.

alexcrichton:
Awesome, thanks! Note that this I think is related to https://github.com/carllerche/mio/issues/521 where we should in theory be calling this method in more locations as well.

Also, would it be possible to add a test for this?

Ralith:
I've modified the existing test for connection refused to verify that the correct `ErrorKind` is produced.

alexcrichton:
Re-kicking CI due to some spurious failures

alexcrichton:
Thanks @Ralith!

PR #605, opened 2017-05-18
Fix a memory leak caused by an Arc cycle

### LINKED ISSUE #604: ReadinessNode and ReadinessQueueInner leak due to a reference cycle
There's a logical reference cycle between `ReadinessNode` and `ReadinessQueueInner`.

First, every ReadinessNode registered to some queue has a clone of the Arc<ReadinessQueueInner> (referenced via an AtomicPtr). It is kept even after the node is deregistered.

Second, the queue always holds a logical reference to every ReadinessNode registered with it (extra 1 in ref_count). This one is also kept even after the node is deregistered.

The Drop impl for ReadinessQueueInner attempts to break this reference cycle, but `ReadinessQueueInner::drop` cannot run until the nodes are already released. Furthermore, it releases only the nodes that are in the readiness queue, but some of them may be absent from it at the moment of dropping.

Tested using `tokio-core` with the following test, which leaks a `ReadinessQueueInner`, several `ReadinessNode`s and two file descriptors on every iteration, running into the default file descriptor limit of 1024 and breaking.
```rust
#[test]
fn tokio_fd_leak() {
    let handle = thread::spawn(move || {
        for _ in 0..1000 {
            let core = tokio_core::reactor::Core::new().unwrap();
        }
    });
    handle.join().unwrap();
}
```

carllerche:
Fixes #604

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -481,6 +481,11 @@ struct ReadinessQueueInner {
     // Similar to `end_marker`, but this node signals to producers that `Poll`
     // has gone to sleep and must be woken up.
     sleep_marker: Box<ReadinessNode>,
+
+    // Similar to `end_marker`, but the node signals that the queue is closed.
+    // This happens when `ReadyQueue` is dropped and signals to producers that
+    // the nodes should no longer be pushed into the queue.
+    closed_marker: Box<ReadinessNode>,
 }
 
 /// Node shared by a `Registration` / `SetReadiness` pair as well as the node
@@ -1891,6 +1896,7 @@ impl ReadinessQueue {
 
         let end_marker = Box::new(ReadinessNode::marker());
         let sleep_marker = Box::new(ReadinessNode::marker());
+        let closed_marker = Box::new(ReadinessNode::marker());
 
         let ptr = &*end_marker as *const _ as *mut _;
 
@@ -1901,6 +1907,7 @@ impl ReadinessQueue {
                 tail_readiness: UnsafeCell::new(ptr),
                 end_marker: end_marker,
                 sleep_marker: sleep_marker,
+                closed_marker: closed_marker,
             })
         })
     }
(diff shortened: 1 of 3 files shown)


alexcrichton:
Just to make sure I understand this, isn't actually removing the cyclic reference, right? The main readiness queue will still stay allocated until all readiness nodes are dropped?

alexcrichton:
I've implemented an [alternate fix](https://github.com/alexcrichton/mio/commit/6c3d92756e5411e6d10b72ccd74a9bbb285207a7) for this which relies on using `Weak<ReadinessQueueInner>` and switches readiness nodes themselves to `Arc<ReadinessNode>` which I believe handles the "destroy things as soon as possible" problem, but may not handle something else?

carllerche:
@alexcrichton I'm not sure what you meany by "actually removing the cycle", but when `ReadyQueue` is dropped, there is no more cycle since the readiness queue is empty and remains so.

Re: dropping the memory early, I looked and it didn't seem like there was a huge win. The only thing that would make sense is closing the awakener, but that could just as easily be done by explicitly calling `close` in `ReadyQueue::drop`

alexcrichton:
Oh I just mean that there's still a strong cyclic reference count between the readiness queue and the nodes it contains. I see now though (missed this before) that the destructor moving locations should provide a deterministic destruction point for the queue at least. 

I'd personally prefer to work with standard known-safe libstd abstractions though, if possible. Understanding this code is already super hard and relying on libstd-isms should help increase maintainability in theory at least. Do you have thoughts on the commit I had? I think it doesn't really come at a perf loss, should hopefully be a little less unsafe, and should also deallocate memory for lots of these nodes and such asap.

carllerche:
I'm not opposed to a larger cleanup, but I would prefer handling the leak first, then evaluate any other changes separately.

alexcrichton:
Ok that sounds reasonable, let's merge and I'll send a PR

alexcrichton:
I've sent a PR as https://github.com/carllerche/mio/pull/606

PR #606, opened 2017-05-19
Leverage `Arc` in the `ReadinessQueue` implementation

alexcrichton:
This commit moves memory mangement over to using the `Arc` type in libstd rather
than manual memory management through manual reference counting. This also
solves the memory leak from #604 by storing weak pointers inside of each
`ReadinessNode`.

The memory management around enqueueing nodes in the mpsc queue is still a bit
sketchy, but the hope is that it's isolated enough to be easily auditable.

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -2,12 +2,12 @@ use {sys, Token};
 use event_imp::{self as event, Ready, Event, Evented, PollOpt};
 use std::{fmt, io, ptr, usize};
 use std::cell::UnsafeCell;
-use std::{mem, ops, isize};
+use std::{mem, isize};
 #[cfg(unix)]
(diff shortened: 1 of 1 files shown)


carllerche:
Thanks. I'll take a look at the code soon.

Did you try comparing the performance using the (single :'() benchmark in the Mio repo? I did a quick check and this PR seemed to be 35-40% slower than compared against master.

carllerche:
Specifically (since I suck at the maths)

Master: 1,067,352 ns/iter
PR: 1,482,217 ns/iter

alexcrichton:
Oh heh didn't know that existed! I'll dig into that and see what's up.

alexcrichton:
I won't have the time to get back to this for awhile.

PR #601, opened 2017-05-09
add CI for aarch64-linux-android

chshawkn:
@MortimerGoro Moved CI integration to a separate PR as you asked.

PR #607, opened 2017-05-19
Fix ci aarch64 linux android

carllerche:
This is a continuation of #601. It has been rebased.


carllerche:
Merged as part of #601

PR #610, opened 2017-05-26
Expose TCP socket options

### LINKED ISSUE #566: Set-linger and other socket configs
Could we have [set_buffer_sizes](https://doc.rust-lang.org/net2-rs/net2/trait.TcpStreamExt.html#tymethod.set_send_buffer_size), [set_linger](https://doc.rust-lang.org/net2-rs/net2/trait.TcpStreamExt.html#tymethod.set_linger) etc for live sockets like we have for `std::net` counterparts through `net2` crate ? Right now if i need to change anything on a running socket, i need to use unsafe functions to obtain raw-fd, wrap it to `std::net::TcpStream/UdpSocket`, perform the changes, get the `mio` equivalent again from raw-fd.

So basically the extensions provided by `net2` crate.

carllerche:
This is done by exposing the options provided by net2 directly instead of implementing `TcpStreamExt`.

Fixes #566

### DIFF
--- a/src/net/tcp.rs
+++ b/src/net/tcp.rs
@@ -10,6 +10,7 @@
 
 use std::io::{Read, Write};
 use std::net::{self, SocketAddr, SocketAddrV4, SocketAddrV6, Ipv4Addr, Ipv6Addr};
+use std::time::Duration;
 
 use net2::TcpBuilder;
 use iovec::IoVec;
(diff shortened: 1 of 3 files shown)


carllerche:
Since Mio's TcpStream is expected to be non-blocking, I'm going to remove the read & write timeout options.

PR #612, opened 2017-05-27
Initial Emscripten websocket support

tones111:
This set of changes provides basic backend support for running mio within the Emscripten runtime.  It still needs some polish but I wanted to get some feedback on the general approach.

Some general notes...

- Emscripten only supports creating TCP client connections.  As such I disabled the unsupported capability so builds would fail instead of failing at runtime.
- Userspace events should be possible but I haven't tried them yet.
- Logic to simulate edge/level triggering is needed.
- Only one Selector is currently supported
- I haven't done any work to try and support the test suite since the build outputs javascript that needs to run in the browser.  I'm curious what other Emscripten projects due.

I have a repo demonstrating this functionality at https://github.com/tones111/mio-example

Thanks for taking a look.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,9 @@ iovec    = "0.1.0"
 [target.'cfg(unix)'.dependencies]
 libc   = "0.2.19"
 
+[target.'cfg(target_os="emscripten")'.dependencies]
+lazy_static = "0.2"
+
 [target.'cfg(windows)'.dependencies]
 winapi = "0.2.1"
 miow   = "0.2.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ default = ["with-deprecated"]
 lazycell = "0.4.0"
 log      = "0.3.1"
 slab     = "0.3.0"
-net2     = "0.2.29"
+net2 = { git = "https://github.com/rust-lang-nursery/net2-rs.git", rev="80f1bd17" }
 iovec    = "0.1.0"
 
 [target.'cfg(unix)'.dependencies]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,6 +88,10 @@ extern crate iovec;
 #[cfg(unix)]
 extern crate libc;
 
+#[cfg(target_os="emscripten")]
+#[macro_use]
+extern crate lazy_static;
+
 #[cfg(windows)]
 extern crate miow;
 
--- a/src/net/mod.rs
+++ b/src/net/mod.rs
@@ -12,7 +12,10 @@ mod tcp;
 #[cfg(not(target_os="emscripten"))]
 mod udp;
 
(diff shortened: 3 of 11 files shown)


carllerche:
I love it! I don't know much about emscripten, but I would be definitely interested in learning more.

Could you explain more about why you want Mio able to target emscripen? Also, what's the difference between emscripten and webassembly (wasm?)?

tones111:
Thanks.  This is actually a pretty unfamiliar road for me as well.  My interested started due to a lack of good rust gui options and I'd seen some people claiming that they've had success writing the frontend in html/css/js and have it communicate with the backend over a websocket (as I understand it, essentially a TCP abstraction over HTTP).

Not being a web guy I would like to maximize the amount of rust running in the browser, but while exploring that avenue I found it very hard to pass data between my rust in the browser and the websocket instantiated via javascript.  So, I went searching for a way to cut out all the overhead and eventually came across the emscripten docs for the socket callback registration stuff.

There seems to be some interest in the community both in GUI options and running rust in the browser, so having an easier mechanism to pass data seemed like a missing opportunity.  At one point I'd tried to use Tokio and quickly found that it depends on mio which was failing to compile.  So, here we are. ;)  

Emscripten's runtime model seems to be a good match for mio.  Instead of blocking in the select call you use Emscripten's event loop to determine the polling rate.  This can be seen in my example repo referenced above.  It's quite a bit different from the other backends but getting the functionality upstream in mio would prevent other users from having to separately implement all of the unsafe callback registration and event registration/handling.

wasm and asm-js are two different compilation targets that use emscripten to link to either a wasm binary or a javascript library that can be run from within the browser.  I intend to ultimately target wasm, but figured it would be easier to start by targeting asm-js which seems to have more information out there.  I'm not sure if wasm is at feature parity yet.

So, this backend is still a bit rough around the edges but I wanted to get a working proof-of-concept out there in a presentable state before sinking _even_more time into it.  Thanks

Aside:
It looks like the windows build is unhappy with an unrelated upstream change to net2.  I had to make a small change because the socket creation was failing due to a missing syscall.

tones111:
I updated my previous wasm hacking to incorporate these mio changes and have pushed that to the master branch at https://github.com/tones111/mio-example.  There is an asmjs branch that exercises that target.

The wasm example needs a bunch more infrastructure because the application needs to be served over http (as opposed to opening an html file directly from disk).  So it ends up needing to run a web server and run a websocket server to communicate with the code running in the browser.

In any event, hopefully this at least proves that it can work and provides a couple client-side examples in case anyone else out there finds this useful.  To run it execute **make run** (you'll need a working emscripten setup - http://www.hellorust.com/emscripten/) and then point your browser at localhost:8080

It's probably also worth pointing out that this all works on stable rust.

carllerche:
I'm going to close this for now since it doesn't seem like there is any new activity, but feel free to open a new PR.

PR #618, opened 2017-06-15
Remove unrelated documentation

rom1v:
Remove the SockAddr part from the send() documentation that was copied
from send_to().


carllerche:
Thanks

PR #621, opened 2017-06-17
Fix NetBSD build failure due to kevent type difference

iquiw:
The build error before fix can be seen [here](https://travis-ci.org/iquiw/mio/jobs/243910261).

I am not sure this is proper fix. At least, build and test passed on NetBSD.


carllerche:
I rebased off of master, waiting for CI.

PR #623, opened 2017-06-24
Listen backlog size is decided from SOMAXCONN on Linux.

kubo39:


bluetech:
I feel that the backlog should be configurable and not hardcoded or set to the max.

But if you do want to set it to the max, you can avoid the extra work by using `i32::MAX` since it is truncated to max. This holds for linux and probably holds for other OSes as well. The POSIX man page says this:

> The implementation may have an upper limit on the length of the listen queue—either global or per accepting socket. If backlog exceeds this limit, the length of the listen queue is set to the limit.

kubo39:
@bluetech
We can customize the backlog size by `sysctl -w net.core.somaxconn=XXX`.

[man listen(2)](http://man7.org/linux/man-pages/man2/listen.2.html) says:
> If the backlog argument is greater than the value in /proc/sys/net/core/somaxconn, then it is silently truncated to that value; the default value in this file is 128.  In kernels before 2.4.25, this limit was a hard coded value, SOMAXCONN, with the value 128.

mio's listen backlog size is `min(SOMAXCONN, 1024)`. However, I want to tune the backlog size greater than 1024 for performance reason.

bluetech:
Different applications on the same system may want different values. I agree it's niche.

I also agree that using the system max is better than hardcoding 1024, which may be too low or too high. Therefore I think your patch is an improvement. But you can achieve the same effect more simply (and more portably and without fallback code etc.) by just passing `i32::MAX` since as the passage you quoted says it will be truncated to `net.core.somaxconn`.

kubo39:
@bluetech 

> Different applications on the same system may want different values. I agree it's niche.

Looks nice, but there's no means to set this. How about create an issue?

> I also agree that using the system max is better than hardcoding 1024, which may be too low or too high. Therefore I think your patch is an improvement. But you can achieve the same effect more simply (and more portably and without fallback code etc.) by just passing i32::MAX since as the passage you quoted says it will be truncated to net.core.somaxconn.

Ah, but how about `u16::MAX` ? Linux stores the backlog in a ushort.

alexcrichton:
Thanks for the PR! This may not be best suited for mio, however, due to the surprising behavior it may have. This is why the methods exist, though, to construct streams from libstd streams/listeners. You can use crates like `net2` and/or `socket2` to configure sockets in a manual fashion, and then they can be converted into the respective mio types.

carllerche:
I agree w/ @alexcrichton that the correct way to configure the backlog would be going via `net2`.

However, I think that the argument here is that a better default than hardcoding 1024 would be to use `u16::MAX` or whatever as the OS caps it to an internal limit?

How does this work on other platforms?

* FreeBSD: ???
* OS X: ???
* Windows: ???
* [other]BSD: ???

raggi:
Posix specifies that the value given to listen is a hint only, and so large values should not cause errors.

carllerche:
So, on posix platforms, pass u16::MAX. Does windows do the same?

raggi:
I'd advise against always setting this unconditionally high. If you prevent push-back then you potentially expose your systems to high memory load in the network stack, leading to other, harder to diagnose problems. While most systems will apply other limits, the accept backlog really doesn't need to be super long most of the time. Having an option for making it long makes sense, but having it always be super long is a security issue by way of a DoS vector.

raggi:
@carllerche re windows, i forget the exact details, but if I remember correctly winsock2 clamps to around 200 and might errors with values that are super large. More modern versions work with much higher values.

carllerche:
@raggi so what's the conclusion?

carllerche:
I'm going to close this due to inactivity.

It is currently possible to create a listener with a custom backlog setting, so changing the default isn't critical.

Thomasdezeeuw:
@kubo39 Mio has been using 1024 for some time: https://github.com/tokio-rs/mio/blob/d4ce420212cda9471302adda5c12da820111ca08/src/net/tcp/listener.rs#L66

If you want to change this number can you open another issue? Or if you want to do it now you can use the [socket2 crate](https://docs.rs/socket2/0.4.2/socket2/struct.Socket.html#method.listen) for this.

PR #627, opened 2017-06-27
Implement assignment operators for Ready

rom1v:
Several operators (`|`, `^`, `&`, `-`) are implemented for `mio::Ready`.

In addition, implement the corresponding assignment versions (`|=`, `^=`, `&=`, `-=`).

(Note that I really don't like `-` nor `-=` on a bitset, they're inconsistent with the other ops)


carllerche:
Rebased against master

PR #628, opened 2017-07-12
Consider adding support for reading into uninitialized buffers via the `buffer` crate

tbu-:
This will allow users of this library to receive data from the network
into uninitialized buffers, like an empty `ArrayVec` or a `Vec` with
remaining capacity.

See also https://www.reddit.com/r/rust/comments/6md04j/how_to_create_variable_length_byte_buffer/dk2hjmz/?context=3.

This is just a suggestion, what do you think about it?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,11 +23,12 @@ with-deprecated = []
 default = ["with-deprecated"]
 
 [dependencies]
+buffer   = "0.1.8"
+iovec    = "0.1.0"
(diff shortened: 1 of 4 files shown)


carllerche:
Thanks for the suggestion. I'd say that this is outside the scope of Mio.

It looks like the buffer crate has a similar goal as the `bytes` crate, which mio initially depended on. Support was removed to provide a more minimal API close to std.

PR #629, opened 2017-07-13
Fuchsia port

cramertj:
There are still a few bits and bobs that could take some cleaning up (the commit history is nonsense, and there are a few broken `std` functions that should be fixed rather than worked-around), but I think this is ready for review.

Tests are passing locally in QEMU.

cc @raphlinus, @rtsuk, @trishume, @kulakowski


cramertj:
@carllerche I can pick out the changes to the existing tests into a separate PR, if you'd prefer.

src/sys/fuchsia/selector.rs
```diff
@@ -0,0 +1,355 @@
...
+        1
+    }
+    pub fn is_empty(&self) -> bool {
+        self.event_opt.is_none()
+    }
+    pub fn get(&self, idx: usize) -> Option<Event> {
+        if idx == 0 { self.event_opt } else { None }
+    }
+    pub fn push_event(&mut self, event: Event) {
+        assert!(::std::mem::replace(&mut self.event_opt, Some(event)).is_none(),
```
alexcrichton: I think the custom readiness queue in mio uses this `Events`, in which case this may be called multiple times during normal operations?

src/sys/fuchsia/selector.rs
```diff
@@ -0,0 +1,355 @@
...
+
+    pub ( in sys::fuchsia) fn register_handle<H>(&self,
+                                                 handle: &H,
+                                                 token: Token,
+                                                 interests: Ready,
+                                                 poll_opts: PollOpt) -> io::Result<()>
+        where H: magenta::HandleBase
+    {
+        if poll_opts.is_level() && !poll_opts.is_oneshot() {
+            panic!("Repeated level-triggered events are not supported on Fuchsia handles.");
```
alexcrichton: Perhaps an I/O error instead of a panic?

src/sys/fuchsia/selector.rs
```diff
@@ -0,0 +1,355 @@
...
+enum RegType {
+    Fd,
+    Handle,
+}
+
+fn key_from_token_and_type(token: Token, reg_type: RegType) -> u64 {
+    let key = token.0 as u64;
+    let msb = 1u64 << 63;
+    if (key & msb) != 0 {
+        panic!("Most-significant bit of token must remain unset.");
```
alexcrichton: Could this return an error instead of panicking? It's sort of unfortunate that a bit has to be reserved b/c `Token` in general is intended to be a bag of bits for everyone else, but it seems the implementation currently necessitates this restrictoin.

src/sys/fuchsia/eventedfd.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    fn reregister(&self,
+                  poll: &Poll,
+                  token: Token,
+                  interest: Ready,
+                  opts: PollOpt) -> io::Result<()>
+    {
+        self.deregister(poll)?;
+        self.register(poll, token, interest, opts)
```
alexcrichton: I think technically this sequence can be incorrect in the face of concurrent updates b/c these steps don't happen atomically, but this may not be a practical problem in the ecosystem using mio.

src/sys/fuchsia/eventedfd.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    fn deregister(&self, poll: &Poll) -> io::Result<()> {
+        let mut registration_lock = self.inner.registration.lock().unwrap();
+        let old_registration = registration_lock.take()
+            .expect("Tried to deregister on unregistered handle.");
```
alexcrichton: I/O error instead of panic?

src/sys/fuchsia/selector.rs
```diff
@@ -0,0 +1,355 @@
...
+    /// `Async::repeating` registrations.
+    ///
+    /// When a level-triggered `Async::repeating` event is seen, its token is added to this list so
+    /// that it will be reregistered before the next `port::wait` call, making `port::wait` return
+    /// immediately if the signal was high during the reregistration.
+    tokens_to_rereg: Mutex<Vec<Token>>,
+
+    /// Map from tokens to weak references to `EventedFdInner`-- a structure describing a
+    /// file handle, its associated `mxio` object, and its current registration.
+    token_to_fd: ConcHashMap<Token, Weak<EventedFdInner>>,
```
alexcrichton: Did you find this necessary for performance? (over just a `Mutex<HashMap<...>>`) With the liberal usage of `Mutex` elsewhere here I'd be tempted to avoid picking up this dependency and just stick to a simple construction for now.

src/sys/fuchsia/selector.rs
```diff
@@ -0,0 +1,355 @@
...
+
+    /// Register event interests for the given IO handle with the OS
+    pub ( in sys::fuchsia) fn register_fd(&self,
+                                          handle: &magenta::Handle,
+                                          fd: &EventedFd,
+                                          token: Token,
+                                          signals: magenta::Signals,
+                                          poll_opts: PollOpt) -> io::Result<()>
+    {
+        self.token_to_fd.insert(token, Arc::downgrade(&fd.inner));
```
alexcrichton: I think it's technically valid on other platforms to register two handles with the same token, which seems like it could cause problems here as one would overwrite the other. In practice I don't think this happens in the ecosystem much though.

src/net/tcp.rs
```diff
@@ -119,7 +121,16 @@ impl TcpStream {
...
     pub fn from_stream(stream: net::TcpStream) -> io::Result<TcpStream> {
-        try!(stream.set_nonblocking(true));
+
+        #[cfg(not(target_os = "fuchsia"))]
+        { try!(stream.set_nonblocking(true)); }
+
+        #[cfg(target_os = "fuchsia")]
+        {
+            try!(sys::set_nonblock(
+                ::std::os::unix::io::AsRawFd::as_raw_fd(&stream)));
```
cramertj: Yes-- this and recv_from are broken in libstd.

src/net/tcp.rs
```diff
@@ -618,24 +629,24 @@ impl Evented for TcpListener {
  *
  */
 
-#[cfg(unix)]
+#[cfg(all(unix, not(target_os = "fuchsia")))]
 use std::os::unix::io::{IntoRawFd, AsRawFd, FromRawFd, RawFd};
 
-#[cfg(unix)]
+#[cfg(all(unix, not(target_os = "fuchsia")))]
```
cramertj: It is, but we want to keep that hidden-- eventually we want to move away from the file-descriptor-based approach to something built on the native channels. That interface isn't ready to use, yet, though.

src/sys/fuchsia/eventedfd.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+/// An event-ed file descriptor. The file descriptor is owned by this structure.
+#[derive(Debug)]
+pub(in sys::fuchsia) struct EventedFdInner {
+    /// Properties of the current registration.
+    pub registration: Mutex<Option<EventedFdRegistration>>,
+
+    /// Owned file descriptor.
+    pub fd: RawFd,
```
cramertj: It _could_, but since the `Selector` keeps a map of references to `EventedFd`s for different types (`TcpSocket`, `TcpStream`, `UdpSocket`), we'd need to figure out how to handle that case. Perhaps an enum of `TcpSocket`, `TcpStream`, `UdpSocket` would work. What do you think?

src/sys/fuchsia/eventedfd.rs
```diff
@@ -0,0 +1,187 @@
...
+            (handle, raw_handle, signals)
+        };
+
+
+        let needs_rereg = opts.is_level() && !opts.is_oneshot();
+
+        {
+            let mut registration_lock = self.inner.registration.lock().unwrap();
+            if registration_lock.is_some() {
+                panic!("Called register on an already registered file descriptor.");
```
cramertj: Yes.

src/sys/fuchsia/eventedfd.rs
cramertj: This was just me being lazy-- you're absolutely right. Fixing it now.

src/sys/fuchsia/selector.rs
cramertj: Yeah, we can make it an error. When we eventually switch over to a fully handle-based approach with no file descriptors, this limitation will go away.

src/sys/fuchsia/selector.rs
cramertj: Sure, I can swap that out.

src/sys/fuchsia/selector.rs
cramertj: Hm. Interesting. I'll trigger an error then instead of overwriting.

src/sys/fuchsia/selector.rs
cramertj: Sounds good :)

src/sys/fuchsia/selector.rs
cramertj: Interesting-- is that not used in the tests at all?

src/sys/fuchsia/eventedfd.rs
```diff
@@ -0,0 +1,187 @@
...
+use magenta;
+use magenta::HandleBase;
+use sys::fuchsia::{DontDrop, poll_opts_to_wait_async, sys};
+use std::mem;
+use std::os::unix::io::RawFd;
+use std::sync::{Arc, Mutex};
+
+/// Properties of an `EventedFd`'s current registration
+#[derive(Debug)]
+pub(in sys::fuchsia) struct EventedFdRegistration {
```
cramertj: Yeah, that's true-- it's not exported from `fuchsia/mod.rs`, so I guess I can change it. During development, I wanted to make sure I wasn't exporting it publicly since you can trigger unsafety by swapping out the `RawFd` (which is closed on `Drop`).

cramertj: I'll add a comment explaining that.

src/sys/fuchsia/selector.rs
```diff
@@ -0,0 +1,355 @@
...
+                  _awakener: Token,
+                  timeout: Option<Duration>) -> io::Result<bool>
+    {
+        evts.event_opt = None;
+
+        self.reregister_handles()?;
+
+        let deadline = match timeout {
+            Some(duration) => {
+                let nanos = (duration.as_secs() * 1_000_000_000) +
```
cramertj: Done.

cramertj:
> Awesome work! I wonder, is it feasible to change the standard library to stop using file descriptors? It seems like it'd be a muche asier implementation to only work with handles and not file descriptors, and it seems like the file descriptor dispatch could maybe be in a mio-fuchsia library rather than baked into core (ideally).

Yes! This is absolutely the goal-- at some point. This part of Fuchsia is still very much WIP, so these changes will have to wait for the implementations to stabilize (or even be written at all :smile:).

This messy file-descriptor-based implementation of the net primitives will be replaced as soon as I can help it.

src/sys/fuchsia/eventedfd.rs
```diff
@@ -0,0 +1,246 @@
...
+#[derive(Debug)]
+pub struct EventedFdInner {
+    /// Properties of the current registration.
+    pub registration: Mutex<Option<EventedFdRegistration>>,
+
+    /// Owned file descriptor.
+    ///
+    /// Note that, while this is public, it should not be exported from the crate, as
+    /// `fd` is closed on `Drop`, so modifying `fd` is a memory-unsafe operation.
+    pub fd: RawFd,
```
trishume: Why is this not `pub (crate)` or `pub (in sys::fuchsia)` given the above comment?

src/sys/fuchsia/eventedfd.rs
```diff
@@ -0,0 +1,246 @@
...
+                    io::ErrorKind::NotFound,
+                    "Called rereregister on an unregistered file descriptor."))
+            };
+
+            poll::selector(poll)
+                .deregister_fd(old_registration.handle.inner_ref(), old_registration.token)?;
+        }
+
+        // Reregister
+        let epoll_events = ioevent_to_epoll(interest, opts);
```
trishume: Could this stuff that's duplicated with `register` be factored out into a helper that takes in the `MutexGuard` for the registration lock and the `self.inner.mxio`? Then both `register` and `reregister` could call it, just with different stuff around it.

Alternatively, the block that gets the handle and signals and the block that sets up the registration lock could be factored into two separate helpers that just take the stuff they need.

src/sys/fuchsia/eventedfd.rs
```diff
@@ -0,0 +1,246 @@
...
+        registered
+    }
+
+    fn reregister(&self,
+                  poll: &Poll,
+                  token: Token,
+                  interest: Ready,
+                  opts: PollOpt) -> io::Result<()>
+    {
+        // Take out the registration lock
```
trishume: Could this be a short explanation of why the lock needs to be taken out across both deregistration and reregistration? Right now it's kind of redundant with the code. 

I don't understand `mio` enough to know why this implementation of `reregister` takes out the lock once for both the deregistration and the reregistration, instead of the entire implementation just being `self.deregister(...);self.register(...)`. 

Is it because a concurrent call to `deregister` between those two would crash in `reregister` instead of the concurrent `register`? Or because a concurrent call to `deregister` in the middle would crash when it wouldn't afterwards (although this is a race condition because if it is scheduled before then `reregister` will crash)? Is it just a performance optimization?

src/sys/fuchsia/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+mod selector;
+
+use self::eventedfd::{EventedFd, EventedFdInner};
+
+pub use self::awakener::Awakener;
+pub use self::handles::EventedHandle;
+pub use self::net::{TcpListener, TcpStream, UdpSocket};
+pub use self::selector::{Events, Selector};
+
+// Set non-blocking (workaround since the std version doesn't work in fuchsia
```
trishume: missing close paren 😛

src/sys/fuchsia/selector.rs
```diff
@@ -0,0 +1,369 @@
...
+    Handle,
+}
+
+fn key_from_token_and_type(token: Token, reg_type: RegType) -> io::Result<u64> {
+    let key = token.0 as u64;
+    let msb = 1u64 << 63;
+    if (key & msb) != 0 {
+        return Err(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            "Most-significant bit of token must remain unset."));
```
trishume: Aren't these tokens provided by consumers of `mio`? Do important users like `tokio` not use `Token`s with the most significant bit set? I'm particularly worried about consumers that might generate random numbers for their tokens. Although if this is a problem I don't have any good ideas how to solve it.

src/sys/fuchsia/selector.rs
```diff
@@ -0,0 +1,369 @@
...
+    /// used to prevent having to lock `tokens_to_rereg` when it is empty.
+    has_tokens_to_rereg: AtomicBool,
+
+    /// List of `Token`s corresponding to registrations that need to be reregistered before the
+    /// next `port::wait`. This is necessary to provide level-triggered behavior for
+    /// `Async::repeating` registrations.
+    ///
+    /// When a level-triggered `Async::repeating` event is seen, its token is added to this list so
+    /// that it will be reregistered before the next `port::wait` call, making `port::wait` return
+    /// immediately if the signal was high during the reregistration.
```
trishume: Is that really how it behaves in practice? If so the Magenta docs should probably be clarified.

I got the opposite impression when reading the Magenta documentation. I interpreted it as saying that packets are only sent to a port when the signal *changes* state, which would mean if you register a handle using `object_wait_async` when that signal was already high then it wouldn't change and no packet would be sent to the port.

If I'm wrong that's actually great because I had no idea how to do things efficiently without race conditions with the behavior I interpreted it as having.

src/sys/fuchsia/eventedfd.rs
cramertj: @alexcrichton [indicated above](https://github.com/carllerche/mio/pull/629#discussion_r127262007) that `reregister` should probably be atomic, and I sort of agree. And yes, it has the additional performance benefit of only taking the lock out once.

src/sys/fuchsia/selector.rs
cramertj: Yeah, as I mentioned above this is unfortunate, but (1) this limitation shouldn't be a problem when using most mio-based libraries (i.e. tokio :smile:) and (2) this won't be a problem once we eventually switch away from the file-descriptor-based approach to the networking primitives.

src/sys/fuchsia/selector.rs
cramertj: Yeah, I had a long conversation with swetland about this on IRC-- I can pull up the logs if you want. Basically, mio is edge-triggered, _except_ that "registration when level is high" counts as a posedge. So if you reregister, it acts as a level trigger.

src/sys/fuchsia/eventedfd.rs
cramertj: I cleaned this up a bunch, yeah.

src/sys/fuchsia/mod.rs
cramertj: Done.

src/sys/fuchsia/eventedfd.rs
cramertj: See the conversation [here](https://github.com/carllerche/mio/pull/629#discussion_r127229539). I kind of tend towards agreeing with you, since then you can determine the visibility from just looking at the declaration site, but it is unnecessary and more verbose.

cramertj: Translation: *shrug*

src/sys/fuchsia/selector.rs
alexcrichton: Ah maybe not precisely like this? This'd also be a problem though if an I/O event comes in as well as a custom readiness event. The `push_event` [happens here](https://github.com/carllerche/mio/blob/a5415ba319e1bef09980f972caf4ffc9bf92d5b4/src/poll.rs#L2008) currently FWIW.

It should be ok to change this to a `Vec` right and just push at most one item during the call to `select`? It may be worth exploring writing a test for this as well, but that's ok to defer until later.

src/sys/fuchsia/selector.rs
alexcrichton: Yeah Tokio currently assigns (AFAIK at least) a unique token per event and tokens that start counting from 0. So at least for tokio's use case we won't hit this error I believe, but it's true that for arbitrary usage of `mio` this'd cause problems (but returning an error I think is fine)

src/sys/fuchsia/handles.rs
```diff
@@ -0,0 +1,66 @@
+use {io, poll, Evented, Ready, Poll, PollOpt, Token};
+use magenta::HandleBase;
+use std::sync::Mutex;
+
+/// Wrapper for registering a `HandleBase` type with mio.
+#[derive(Debug)]
+pub struct EventedHandle<T> where T: HandleBase {
```
alexcrichton: Since this is a publicly exported structure under `mio::fuchsia`, could we trim this back a bit to make it more conservative? Perhaps make the fields private and maybe remove the generic for now?

(or is the generic necessary?)

src/sys/fuchsia/handles.rs
```diff
@@ -0,0 +1,66 @@
...
+            *this_token = None;
+            poll::selector(poll).register_handle(&self.handle, token, interest, opts)?;
+            *this_token = Some(token);
+        }
+        Ok(())
+    }
+
+    fn deregister(&self, poll: &Poll) -> io::Result<()> {
+        let mut this_token = self.token.lock().unwrap();
+        let token = this_token.expect("Attempted to deregister an unregistered handle.");
```
alexcrichton: I'd return an error here instead of panicking

src/sys/fuchsia/eventedfd.rs
```diff
@@ -0,0 +1,246 @@
...
+use std::mem;
+use std::os::unix::io::RawFd;
+use std::sync::{Arc, Mutex};
+
+/// Properties of an `EventedFd`'s current registration
+#[derive(Debug)]
+pub struct EventedFdRegistration {
+    pub token: Token,
+    pub (in sys::fuchsia) handle: DontDrop<magenta::Handle>,
+    pub rereg_signals: Option<(magenta::Signals, magenta::WaitAsyncOpts)>,
```
alexcrichton: Could these fields all be private with public accessors?

src/sys/fuchsia/selector.rs
cramertj: Yeah, swapping that out would be fine.

src/sys/fuchsia/eventedfd.rs
```diff
@@ -0,0 +1,246 @@
...
+            },
+        }
+    }
+}
+
+/// An event-ed file descriptor. The file descriptor is owned by this structure.
+#[derive(Debug)]
+pub struct EventedFdInner {
+    /// Properties of the current registration.
+    pub registration: Mutex<Option<EventedFdRegistration>>,
```
alexcrichton: Similar to above, could the fields here be private with public accessors for what's needed?

src/sys/fuchsia/handles.rs
cramertj: The goal is to be able to use any [HandleBase](https://docs.rs/magenta/0.1.0/magenta/trait.HandleBase.html) here, so that we can write a tokio-based library using `PollEvented<EventedHandle<T>>`. We need visibility into the handle, although I suppose that could be done using a `get_handle`/`get_handle_mut` methods. The token can be private, too.

trishume: In Fuchsia's Rust stuff a lot of things are typed wrappers around handles, so for example if you have a handle to a channel to an IPC service speaking a protocol you can have a wrapper struct with Rust functions to send the relevant messages over the channel, and that wrapper will implement `HandleBase`.

Ideally we'll want to plumb the `EventedHandle` stuff up through Tokio, and expose a futures interface for the IPC system.

I haven't thought about what that would look like yet but I have a suspicion that it will be nicer with the generic on `EventedHandle`. On the other hand, given that most `HandleBase` structs are very thin wrappers, I wonder if there is a way to avoid the generic since it'll probably cause a lot of code bloat where it is instantiated almost identically for every wrapper.

I'll think about this and look at `tokio-core` and I'll try and come up with a definitive answer to how useful the generic is.

cramertj:
Test are still green on QEMU.

src/sys/fuchsia/handles.rs
cramertj: @trishume The value of the generic (rather than a `HandleRef` or similar) is that you can safely store the typed `EventedHandle<MyHandleType>` in tokio-core's `PollEvented`. We can then write a wrapper library that exposes e.g. `Channel` with operations that return `Future`s (where `Channel` is some wrapper around `PollEvented<EventedHandle<magenta::Channel>>`, similar to how TCP/UDP structures are implemented [in tokio-core](https://github.com/tokio-rs/tokio-core/blob/master/src/net/udp/mod.rs#L13)).

cramertj: Alternatively, we could just convert the type into its inner `Handle` to store in `EventedHandle` and make it the responsibility of our wrapper crate to do the conversion. However, I think this approach is strictly more flexible-- if we go with this approach, in our wrapper crate we could just only ever use `PollEvented<EventedHandle<Handle>>` if that turns out to be a better approach (e.g. it could reduce the amount of code generated by monomorphization).

alexcrichton: Oh yeah this seems great to expose, having the ability to do `PollEvented<CustomType>` is crucial for exposing platform goodies! (this is how `tokio-process` works at all). What I mostly meant was that this was exposed pretty differently than other platforms. On Unix we've got `EventedFd` which notably doesn't take ownership and also just works with a raw `&i32`. On Windows is... slightly different but basically works with `AsRawHandle` in the standard library.

That being said I don't mind having the design here basically be up to y'all. I had a brief chat with @cramertj on IRC about API stability here and there's not too many concerns. I think it'd be fine to just try to make the fields private here for now and worry about stability later.

alexcrichton:
I chatted briefly with @cramertj on IRC about the stability story here, and it's probably worth recappping. My thinking is that because the underlying *OS* is unstable still (e.g. fuchsia syscalls are changing every now and then) it's not really feasible to provide a stability guarantee of mio-on-fuchsia. I think the conclusion here is that mio just won't provide any backwards compatibility guarantees or semver guarantees on the Fuchsia platform for the near future. That'll buy us the freedom to change whatever we need here under the hood and notably break the public API as necessary, all the while without requiring `mio` to make a major version bump.

src/sys/fuchsia/eventedfd.rs
```diff
@@ -0,0 +1,257 @@
...
+
+impl EventedFdInner {
+    pub fn rereg_for_level(&self, port: &magenta::Port) {
+        let registration_opt = self.registration.lock().unwrap();
+        if let Some(ref registration) = *registration_opt {
+            if let Some((rereg_signals, rereg_opts)) = registration.rereg_signals {
+                let _res =
+                    registration
+                        .handle.inner_ref()
+                        .wait_async(port,
```
trishume: I'm somewhat worried about this behaving correctly but Magenta not handling it gracefully from a performance perspective, it's unclear from the docs. Might be good to check with kernel people on IRC, or just switch to your better idea of handling repeating level-triggered waits as one-shots that get re-queued.

If they don't check if the handle is already registered you might end up with 10000 repeating kernel waits stored in the kernel. They would all trigger whenever the signal is asserted but all except the first one see that there is already a packet in the queue and do nothing. This would lead to terrible performance and memory use but no observable difference in behavior.

src/sys/fuchsia/selector.rs
```diff
@@ -0,0 +1,370 @@
...
+    /// Magenta object on which the handles have been registered, and on which events occur
+    port: Arc<magenta::Port>,
+
+    /// Whether or not `tokens_to_rereg` contains any elements. This is a best-effort attempt
+    /// used to prevent having to lock `tokens_to_rereg` when it is empty.
+    has_tokens_to_rereg: AtomicBool,
+
+    /// List of `Token`s corresponding to registrations that need to be reregistered before the
+    /// next `port::wait`. This is necessary to provide level-triggered behavior for
+    /// `Async::repeating` registrations.
```
trishume: Do you know if `tokio` tends to use repeating level-triggered registrations? Because if so doing that performance optimization / potentially cleaner one-shot re-queueing optimization you mentioned becomes more important.

src/sys/fuchsia/eventedfd.rs
trishume: This concern is now moot since @cramertj switched to using one shots after I wrote this.

src/sys/fuchsia/selector.rs
cramertj: It does not, but i already added the optimization :smile:

src/sys/fuchsia/selector.rs
```diff
@@ -0,0 +1,370 @@
...
+        let observed_signals = match packet.contents() {
+            magenta::PacketContents::SignalOne(signal_packet) => {
+                signal_packet.observed()
+            }
+            magenta::PacketContents::SignalRep(signal_packet) => {
+                signal_packet.observed()
+            }
+            magenta::PacketContents::User(_user_packet) => {
+                // User packets are only ever sent by an Awakener
+                return Ok(true);
```
cramertj: mio just cares that it was awakened by the `Awakener`-- it doesn't care about the token.

src/sys/fuchsia/eventedfd.rs
cramertj: :smile:

src/sys/fuchsia/selector.rs
```diff
@@ -0,0 +1,370 @@
...
+        self.id
+    }
+
+    /// Returns a reference to the underlying port `Arc`.
+    pub fn port(&self) -> &Arc<magenta::Port> { &self.port }
+
+    /// Reregisters all registrations pointed to by the `tokens_to_rereg` list
+    /// if `has_tokens_to_rereg`.
+    fn reregister_handles(&self) -> io::Result<()> {
+        if self.has_tokens_to_rereg.load(Ordering::Relaxed) {
```
cramertj: Originally, I had hand-waved my way through this under the justification that this is a best-effort system: items coming off of port_wait race aren't guaranteed to be reregistered on a different thread, but they definitely will appear on the current thread. However, on second consideration I absolutely agree with you-- the first `load` should definitely be an `Acquire`, and the stores should be `Release`. I've made this change.

trishume: Cool, I spent a while puzzling over that. When you make the change it would be nice to include a comment about why that ordering was chosen, something like:

```rust
// The orderings guarantee that this won't trigger before the push is observed and
// thus not handle the pushed token once it is visible.
```

trishume:
Okay I'm ready to okay this if someone could confirm that my understanding of edge triggering behavior on Fuchsia is correct (described in previous top-level review comment). Also I'd be happy if a comment was added on the atomic ordering, like I described in a now-folded reply.

cramertj:
@trishume Yes, your understanding matches mine. I've added a comment about the ordering as well.

src/sys/fuchsia/selector.rs
```diff
@@ -0,0 +1,374 @@
...
+
+                        registration_lock
+                            .as_ref()
+                            .and_then(|r| r.rereg_signals())
+                            .is_some()
+                    };
+
+                    if needs_to_rereg {
+                        let mut tokens_to_rereg_lock = self.tokens_to_rereg.lock().unwrap();
+                        tokens_to_rereg_lock.push(token);
```
trishume: Wait, I just realized I have no idea why the handle isn't just re-registered right here and is re-registered later instead.

cramertj: If we wait eagerly, a new packet will _always_ be queued because we just received an event-- we know that the level is high.

trishume: Okay that makes sense thanks 👍

src/net/tcp.rs
alexcrichton: Oh sorry forgot to comment on this earlier, but could this logic be pushed into the fuchsia module? Ideally this module would have no `#[cfg]` in theory

src/net/tcp.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unreachable_code, unused_variables)]
+// TODO: remove
```
alexcrichton: Should this be removed?

src/poll.rs
```diff
@@ -1353,6 +1353,7 @@ impl fmt::Debug for Events {
 
 // ===== Accessors for internal usage =====
 
+#[allow(dead_code)]
```
alexcrichton: Is this still dead code?

src/sys/fuchsia/handles.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// in order to receive event notifications.
+    pub fn new(handle: T) -> Self {
+        EventedHandle {
+            handle: handle,
+            token: Mutex::new(None),
+        }
+    }
+
+    /// Get a reference to the underlying `HandleBase`.
+    pub fn get_handle(&self) -> &T {
```
alexcrichton: Idiomatically we tend to call these accessors `get_ref`, `get_mut`, and `into_inner`, want to add the third one here as well?

src/sys/fuchsia/selector.rs
```diff
@@ -0,0 +1,374 @@
...
+
+        if wait_res.is_err() {
+            self.token_to_fd.lock().unwrap().remove(&token);
+        }
+
+        wait_res
+    }
+
+    /// Deregister event interests for the given IO handle with the OS
+    pub ( in sys::fuchsia) fn deregister_fd(&self,
```
alexcrichton: the `in sys::fuchsia` can be removed here and in surrounding methods I think, right?

cramertj: Ah, I didn't realize `Selector` was inaccessible from mio's public API.

src/net/tcp.rs
cramertj: I'm not sure how to do this without a cfg in a way that doesn't require adding special `set_nonblock` functions to all the other sys modules. Would you be willing to accept this for the time being, with the understanding that it will be removed when the fix to libstd lands?

src/net/tcp.rs
cramertj: Yes :smile:

src/sys/fuchsia/handles.rs
cramertj: @alexcrichton I made the change, but just out of curiosity: why `get_ref` and `get_mut` instead of `as_ref` and `as_mut`?

alexcrichton: @cramertj oh by this point it's mostly convention, but the thinking is that this is sort of a container rather than a conversion, and we use `get` for contains and `as` for conversions.

src/net/tcp.rs
alexcrichton: Ok sure, that seems fine. Mind leaving a comment?

src/sys/fuchsia/handles.rs
cramertj: Gotcha. Good to know. Thanks!

cramertj:
@trishume The other test is failing because I used features that weren't available in Rust 1.9. @alexcrichton How would you prefer I handle that? is Rust 1.9-compatibility still a goal?

alexcrichton:
Ah yeah it'd be ideal to not worry too much about the compat for now, shouldn't be too hard to fix though?

cramertj:
@alexcrichton Okay, fixed.

alexcrichton:
Looks like the 1.9 compat [may need one more change](https://travis-ci.org/carllerche/mio/jobs/254540721)?

cramertj:
Oh whoops I didn't try building tests. You're absolutely right.

cramertj:
@alexcrichton Do you have any idea what's going on with the Android failure?

alexcrichton:
Hm that worked locally for me so let's see if it's spurious.

cramertj:
Hm... Looks like it passed this time around.

alexcrichton:
👍

PR #636, opened 2017-07-26
Correct test target file path in iOS CI

iquiw:
There exists target file like `x86_64-apple-ios/debug/test-b4dc53cf1da916aa`, instead of `target/x86_64-apple-ios/debug/test-b4dc53cf1da916aa`.

### DIFF
--- a/ci/run-ios.sh
+++ b/ci/run-ios.sh
@@ -20,7 +20,7 @@ case "$TARGET" in
 
 
     # Find the file to run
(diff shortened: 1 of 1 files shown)


iquiw:
hmm, different TARGET (=aarch64-linux-android) from what I changed failed.

carllerche:
It was a spurious fail, thanks for the fix.

PR #637, opened 2017-07-27
Bump to 0.6.10

alexcrichton:


carllerche:
Still trying to get all pending PRs landed... Slow CI is making this tricky!

carllerche:
I landed a bunch of PRs, so the changelog needs updating (I can try to do it in a bit). Besides that, it should be ready to land.

alexcrichton:
updated

PR #640, opened 2017-08-02
Allow registration with empty interest

rom1v:
In order to temporarily disable event listening for a specific handle, it may be convenient to reregister it with an empty interest.

This avoids to temporarily deregister the handle before registering it again (which, by the way, [is not supported on Windows][1]).

Therefore, remove the check that prohibited registration with empty interest.

Note that ERROR and HUP may still be triggered in that case.

[1]: https://github.com/carllerche/mio/issues/633

### DIFF
--- a/test/test_register_deregister.rs
+++ b/test/test_register_deregister.rs
@@ -93,15 +93,15 @@ pub fn test_register_deregister() {
 }
 
 #[test]
-pub fn test_register_with_no_readable_writable_is_error() {
+pub fn test_register_with_no_readable_writable_is_ok() {
     let poll = Poll::new().unwrap();
     let addr = localhost();
 
     let sock = TcpListener::bind(&addr).unwrap();
 
-    assert!(poll.register(&sock, Token(0), Ready::empty(), PollOpt::edge()).is_err());
+    poll.register(&sock, Token(0), Ready::empty(), PollOpt::edge()).unwrap();
 
-    poll.register(&sock, Token(0), Ready::readable(), PollOpt::edge()).unwrap();
+    poll.reregister(&sock, Token(0), Ready::readable(), PollOpt::edge()).unwrap();
 
-    assert!(poll.reregister(&sock, Token(0), Ready::empty(), PollOpt::edge()).is_err());
+    poll.reregister(&sock, Token(0), Ready::empty(), PollOpt::edge()).unwrap();
 }
--- a/test/test_register_deregister.rs
+++ b/test/test_register_deregister.rs
@@ -1,5 +1,6 @@
-use {localhost, TryWrite};
-use mio::*;
+use {expect_events, localhost, TryWrite};
+use mio::{Events, Poll, PollOpt, Ready, Token};
+use mio::event::Event;
 use mio::net::{TcpListener, TcpStream};
 use bytes::SliceBuf;
 use std::time::Duration;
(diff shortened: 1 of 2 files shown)


carllerche:
I think that I would rather get `deregister` to work (as I think it probably should). Internally, it probably is just a `register` w/ no args... Is that something you could try?

rom1v:
> I think that I would rather get `deregister` to work (as I think it probably should).

I agree that `deregister` (and `reregister`) should be [implemented](https://github.com/carllerche/mio/blob/23cc470ac5a4dc93ce8f636523caa9f178dee2e7/src/sys/windows/selector.rs#L203-L259) for Windows.

But even in the case `deregister` were working as expected, I think that empty interest should be allowed, since they are supported by the backends. For what it is worth, other APIs allow it (e.g. [Java NIO](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/SelectableChannel.html#register-java.nio.channels.Selector-int-java.lang.Object-)).

In my client code using _mio_, for convenience I implemented a `Poll` wrapper that exposes an API accepting an empty `Ready` (and call `register`, `deregister`, `reregister` as necessary), which is disappointing since the low-level backends accept empty interest.

rom1v:
@carllerche On IRC, you were wondering if using empty interest was portable across all platforms.

So we need to know if it works for the [backends](https://docs.rs/mio/0.6.10/mio/struct.Poll.html#implementation-notes) supported by _mio_: _epoll_, _kqueue_, _IOCP_, and also the backend used _fuchsia_ (I saw some commits recently about _fuchsia_).

[`man epoll`](http://man7.org/linux/man-pages/man2/epoll_ctl.2.html) suggests that _epoll_ supports it:
> The events member is a bit mask composed by ORing together **zero** or more of the following available event types

In order to test the "empty interest" behavior on each platform, I wrote a minimal sample that polls for _readable_ events, then with empty interest, then for _readable_ events again, so we can check that it behaves as expected: https://github.com/rom1v/mio-empty-interest

I executed it on Linux, Windows and Mac OS, I get the expected output:

```
interest: {readable}
0 -->
      --> 0
1 -->
      --> 1
interest: {}
2 -->
3 -->
4 -->
5 -->
6 -->
interest: {readable}
      --> 23456
7 -->
      --> 7
8 -->
      --> 8
```

I didn't test on _fuchsia_ (@cramertj – or someone else –, could I ask you to execute this sample on _fuchsia_, please?).

The sample is maybe too minimal. Can you think of any specific situations were empty interest may lead to incorrect behavior?

_I also use this patch in a "real-world" – not a sample – application where I often switch from/to non-empty/empty interests, it behaves as expected on Linux and Windows (I will test on Mac OS soon)._

carllerche:
So, what does it mean to 'register' w/ no interest? I could see `reregister` w/ no interest, but `register` is only supposed to be called when there is no interest... is a `register` w/ no interest basically a no-op?

rom1v:
Concretely, it means calling `epoll_ctl` with op `EPOLL_CTL_ADD` and a `struct epoll_event` having a `events` field bitmask with `EPOLLIN` and `EPOLLOUT` unset.

This is convenient when you initialize your interests conditionally, then call the same code whatever its value (without taking care of deregistering/reregistering):
 - may I read? → if yes, enable readable
 - may I write? → if yes, enable writable
 - update interests

carllerche:
You have a thought @alexcrichton ?

alexcrichton:
If epoll itself implements this it seems fine to me at least!

rom1v:
_rebased to solve conflicts with `master`_

test/test_register_deregister.rs
```diff
@@ -93,15 +93,15 @@ pub fn test_register_deregister() {
...
     let sock = TcpListener::bind(&addr).unwrap();
 
-    assert!(poll.register(&sock, Token(0), Ready::empty(), PollOpt::edge()).is_err());
+    poll.register(&sock, Token(0), Ready::empty(), PollOpt::edge()).unwrap();
 
-    poll.register(&sock, Token(0), Ready::readable(), PollOpt::edge()).unwrap();
+    poll.reregister(&sock, Token(0), Ready::readable(), PollOpt::edge()).unwrap();
 
-    assert!(poll.reregister(&sock, Token(0), Ready::empty(), PollOpt::edge()).is_err());
+    poll.reregister(&sock, Token(0), Ready::empty(), PollOpt::edge()).unwrap();
```
carllerche: Would you mind expanding the test to show that when a socket is received, no readable event is generated then once the listener is reregistered w/ readable interest, the readable event is generated (representing a pending accept).

rom1v: Yes, of course. Done :)

carllerche:
Alright, PR looks good, now it's just a question of getting CI to pass.

carllerche:
It seems like windows doesn't like the change :-/ I'm not sure exactly why yet. This may be a bigger issue w/ the custom event queue not being happy about registering w/ no interest.

rom1v:
The registration of the client socket was missing, which is a problem on Windows.

> The connect is not guaranteed to have started until it is registered
https://docs.rs/mio/0.6.10/mio/struct.Poll.html#registering-handles

I amended the commit to fix the issue.

PR #644, opened 2017-08-12
Fix x86_64-android socklen_t in bind()

mattico:
https://travis-ci.org/rust-lang-nursery/rustup.rs/jobs/263713429#L1612

x86_64-linux-android also uses u32 for socklen_t, for whatever reason.


mattico:
AppVeyor failure is #645

PR #650, opened 2017-08-17
Make does_not_register_rw test use Poll API

spastorino:
Remove the use of deprecated API from does_not_register_rw test only. This is part of a larger work and I'd do just one PR with all the commits but @carllerche said he preferred a PR per file for easier review, so here it is :).

There's a lot of tests that now that EventLoop is going to be removed are much more like an integration test for the whole crate.
I've said Carl the possibilities that come to my mind are:

1. Leave the tests that are using EventLoop for now and when EventLoop is gone we could have an EventLoop implementation to be used just in those tests for them to allow this kind of coverage and tests still happening.
2. Move this kind of tests to examples as a way of showing and documenting how to use mio, but we would probably need more coverage on Poll to be safer.
3. Move EventLoop and it's tests to mio-more.
4. Just remove the tests when EventLoop is gone if you consider safe the kind of coverage you have.

### DIFF
--- a/src/sys/unix/kqueue.rs
+++ b/src/sys/unix/kqueue.rs
@@ -326,23 +326,17 @@ impl fmt::Debug for Events {
 
 #[test]
 fn does_not_register_rw() {
-    #![allow(deprecated)]
+    use {Poll, Ready, PollOpt, Token};
+    use unix::EventedFd;
 
-    use ::deprecated::{EventLoopBuilder, Handler};
-    use ::unix::EventedFd;
-    struct Nop;
-    impl Handler for Nop {
-        type Timeout = ();
-        type Message = ();
-    }
+    let kq = unsafe { libc::kqueue() };
+    let kqf = EventedFd(&kq);
+    let poll = Poll::new().unwrap();
 
     // registering kqueue fd will fail if write is requested (On anything but some versions of OS
     // X)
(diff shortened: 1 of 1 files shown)


alexcrichton:
I personally like the idea of migrating away from `EventLoop` in all tests and showcasing `Poll` and its APIs. Mainly in the sense that they should in theory serve for "pretty good" documentation for the crate for the curious. Curious to hear what @carllerche thinks though!

carllerche:
Yeah, I'd probably want to just migrate away as well, even if it ends up being a bit more verbose. People look through the tests to see how things should be used.

spastorino:
Cool, I'm curious of what you guys prefer to do from the options I gave. I mean, I can start migrating out of EventLoop in tests but what would you do in cases where tests heavily depend on it?.
Just move EventLoop implementation to tests and reuse in all those cases or what?.

alexcrichton:
I'd personally advocate for migrating the tests away from `EventLoop` to directly calling `Poll`, although it can sometime perhaps require some relatively significant changes locally. Anything testing specific features of `EventLoop` (and not `Poll`) can probably just get deleted.

If a test is particularly difficult to migrate away from `EventLoop` it's probably ok to vendor a copy to the tests for now at least.

alexcrichton:
In any case this PR itself looks good, so merged!

PR #688, opened 2017-08-20
Method order in mio::net.

### LINKED ISSUE #664: Method order in mio::net
[`TcpListener`](https://docs.rs/mio/0.6.10/mio/net/struct.TcpListener.html) and [`TcpStream`](https://docs.rs/mio/0.6.10/mio/net/struct.TcpStream.html) have setters before getters, [`UdpSocket`](https://docs.rs/mio/0.6.10/mio/net/struct.UdpSocket.html) has getters before setters. I don't have a preference, but reading the docs would be a smoother experience if all three were consistent.

```rust
// TcpListener
fn set_ttl(&self, ttl: u32) -> Result<()>;
fn ttl(&self) -> Result<u32>;
```

```rust
// UdpSocket
fn ttl(&self) -> Result<u32>;
fn set_ttl(&self, ttl: u32) -> Result<()>;
```

(Same for all the getters and setters, not just ttl.)

ahmedcharles:
Fixes #664.

### DIFF
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -101,6 +101,14 @@ impl UdpSocket {
         self.sys.connect(addr)
     }
 
+    /// Sets the value of the `SO_BROADCAST` option for this socket.
+    ///
+    /// When enabled, this socket is allowed to send packets to a broadcast
+    /// address.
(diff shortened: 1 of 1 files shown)


alexcrichton:
:+1:

Issue #655, opened 2017-08-18
Drop the with-deprecated cfg and everything behind it


dtolnay:
As I understand it, the only thing blocking this is reimplementing the test cases that currently rely on deprecated functionality.

The best way to approach this would be running `cargo test --no-default-features` to see what breaks:

```
...

error: aborting due to 180 previous errors
```

then reworking small self-contained pieces to use the new APIs. I imagine this will be split across lots of PRs. It also sounds like a great way to learn the ins and outs of mio in general!

Once `cargo test --no-default-features` passes, we can rip out all the deprecated functionality.

ahmedcharles:
I'm working on this, (perhaps slowly).

PR #692, opened 2017-08-20
Remove uses of the deprecated tcp module.

ahmedcharles:
#655

### DIFF
--- a/src/event_imp.rs
+++ b/src/event_imp.rs
@@ -32,7 +32,7 @@ use std::{fmt, io, ops};
 /// ```
 /// use mio::{Ready, Poll, PollOpt, Token};
 /// use mio::event::Evented;
-/// use mio::tcp::TcpStream;
+/// use mio::net::TcpStream;
 ///
 /// use std::io;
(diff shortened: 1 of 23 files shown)


alexcrichton:
Thanks! For backwards compatibility though could this defer the actual deletion until later? For now I think we mainly just want to remove it's usage, not the deprecated pieces themselves yet

ahmedcharles:
Sure.

ahmedcharles:
I decided to split it into a few PR's to make reviews easier.

alexcrichton:
Thanks so much for doing this! Looks like the Windows tests are failing though?

ahmedcharles:
I don't see any correlation between the failures and any changes I made. I'll probably push PR's for each commit so I can see which one causes the failure and perhaps that will bring more insight.

ahmedcharles:
I'll just have a PR for each commit, at least for ones without merge conflicts.

ahmedcharles:
Rerunning the tests, since there was a docker failure.

PR #698, opened 2017-08-21
Associated Constants for Ready and PollOpt constructors

### LINKED ISSUE #661: Consider using associated constants for the bitflag-like types
The various Ready constructors currently work like this:

```rust
impl Ready {
    pub fn readable() -> Ready { /* ... */ }
    pub fn writable() -> Ready { /* ... */ }
}
```

With the associated constants feature of Rust 1.20, it could work like this:

```rust
impl Ready {
    pub const READABLE: Ready = /* ... */;
    pub const WRITABLE: Ready = /* ... */;
}
```

The constructors never do any work so they shouldn't need to be functions.

One advantage of constants is that they can be used in the pattern of a match statement.

AndyGauge:
Fixes #661


ahmedcharles:
This seems to be nightly only.

AndyGauge:
@ahmedcharles  That's right.  Associated Constraints are planned to drop in 1.20.

carllerche:
Looks like associated constants landed on stable. We'll have to wait 12 weeks before merging it into Mio though...

AndyGauge:
I think the issue was 'consider using associated constants'.  I still haven't seen a discussion about it.

AndyGauge:
So at this point the only failure is for rust 1.13.0 which will obviously fail, as will 1.19.0

carllerche:
Yeah, we are following the rust-lang lib policy of supporting the 2 latest stable versions of Rust, so we can merge this in ~3 mo :)

carllerche:
It would probably just be worth waiting for 0.7.

carllerche:
If you are up for it, could you rebase this against the `v0.7.x` branch? I am using that as staging for v0.7.x

AndyGauge:
Do we need to do anything to get Travis to build 1.14?  Should I update the .travis.yml#27 with 1.20  

> error: associated constants are experimental (see issue #29646)

carllerche:
Yes, I think it would be fine to update the 0.7 branch to depend on 1.20. Would you mind submitting a separate PR for that? I like to keep a clear marker in the git history of the decision to update the min version.

AndyGauge:
OK, that looks ready for 0.7.x

carllerche:
@AndyGauge Thanks! Are the commits based on the 0.7.x branch? I tried switching the PR base but it looked like it pulled in additional commits.

AndyGauge:
It took a while to get master and 0.7x merged together.  I think the extra commits are from the master branch.  If you want I can just create a new PR with 0.7x as the base and ignore the changes to master.

AndyGauge:
I'm closing this PR.  It is easier to use IDE tools to find and replace than keeping this branch current.

PR #713, opened 2017-08-22
Index access depreciated

### LINKED ISSUE #697: Remove Events::get
We want to deprecate indexed access in favor of an iterator only API.

This came up in #681.

AndyGauge:
Fixes #697

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -935,7 +935,10 @@ impl Poll {
     ///
     /// `poll` returns the number of readiness events that have been pushed into
     /// `events` or `Err` when an error has been encountered with the system
-    /// selector.
+    /// selector.  The value returned is deprecated and will be removed in 0.7.0.
+    /// Accessing the events by index is also deprecated.  Events can be
+    /// inserted by other events triggering, thus making sequential access
+    /// problematic.  Use the iterator API instead.  See [`iter`].
     ///
     /// See the [struct] level documentation for a higher level discussion of
     /// polling.
@@ -943,6 +946,7 @@ impl Poll {
     /// [`readable`]: struct.Ready.html#method.readable
     /// [`writable`]: struct.Ready.html#method.writable
     /// [struct]: #
+    /// [`iter`]: struct.Events.html#method.iter
     ///
     /// # Examples
     ///
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -1166,7 +1166,7 @@ impl Poll {
         self.readiness_queue.poll(&mut events.inner);
 
         // Return number of polled events
-        Ok(events.iter().count())
+        Ok(events.inner.len())
     }
 }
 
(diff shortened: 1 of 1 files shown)


ahmedcharles:
It should be possible to provide an impl of `ExactSizeIterator` which would retain access to the `len` function.

carllerche:
Ah, specifically I do not want the exact size to be exposed. This allows the iterator to skip or insert events during iteration.

src/poll.rs
```diff
@@ -1161,7 +1161,7 @@ impl Poll {
         self.readiness_queue.poll(&mut events.inner);
 
         // Return number of polled events
-        Ok(events.len())
+        Ok(events.iter().count())
```
AndyGauge: Should we not return the number of polled events since we do not know the exact number?

carllerche: Yeah, we should definitely not return the number... We have to leave it for now, but that return value should be deprecated.

AndyGauge: What about a 0 or 1?  Return 0 for no events (like a timeout) and 1 if there is at least one?

carllerche: Yeah, I think some return value like that could be good, but it will have to be punted until 0.7. We can open an issue to track that change. The ret value today will have to stay the same :(

carllerche:
Thanks! Looks good. Would you mind also updating the documentation for `Poll::poll` to indicate that the `usize` return variable is deprecated and will be removed in 0.7?

AndyGauge:
Sure, once the tests finish I'll get that done.  I meant to ask how to deprecate the return value!

carllerche:
@AndyGauge I think it has to just be part of the fn docs as well as do a pass of the rest of the docs / tests to ensure it isn't being used.

Maybe @alexcrichton has a better idea?

alexcrichton:
Nah yeah there's no way I know of to deprecate just the return value, but you could temporarily switch it to a `bool` and that'd flag every location that needs to be fixed up internally at least

AndyGauge:
Found an example that needs to be reworked.  Poll::deregister example (poll.rs#892
```
    /// // Set a timeout because this poll should never receive any events.
    /// let n = poll.poll(&mut events, Some(Duration::from_secs(1)))?;
    /// assert_eq!(0, n);
```
Looking for high level guidance how this should be addressed.

carllerche:
I think checking for zero right now is OK. It's the only way to test if `poll` returned due to event OR due to timeout. When the return value is changed, it will have some equivalent API to check if `n == 0`.

AndyGauge:
That's why I wanted to bring it to your attention.  We likely need to create a convenience for this case like `.empty` or whatever makes sense for your API

carllerche:
Ah, so basically add a test fn on the `Events` instead of rely on the return value. I think that might be an interesting idea.

carllerche:
Ok, I think the answer will be to add `Events::is_empty()`.

This PR is good as is, so I'll create a tracking issue for this.

PR #714, opened 2017-08-23
Remove ops::Not impl's.

### LINKED ISSUE #656: Remove the implementations of std::ops::Not
From what I can tell, the only reason these exists is to be used in the implementation of Sub and SubAssign.

```rust
impl<T: Into<Ready>> ops::Sub<T> for Ready {
    type Output = Ready;

    fn sub(self, other: T) -> Ready {
        Ready(self.0 & !other.into().0)
    }
}

impl<T: Into<Ready>> ops::SubAssign<T> for Ready {
    fn sub_assign(&mut self, other: T) {
        self.0 &= !other.into().0;
    }
}
```

Other than in Sub, it never makes sense to apply Not to a Ready. You would never call Poll::register and pass the Not of a Ready, for example. Let's reimplement Sub in a different way and remove the semantically meaningless Not operation.

Applies to Ready, UnixReady, FuchsiaReady, and PollOpt.

ahmedcharles:
Fixes #656.

### DIFF
--- a/src/event_imp.rs
+++ b/src/event_imp.rs
@@ -470,6 +470,9 @@ impl ops::Sub for PollOpt {
     }
 }
 
+#[deprecated(since = "0.6.10", note = "removed")]
+#[cfg(feature = "with-deprecated")]
+#[doc(hidden)]
 impl ops::Not for PollOpt {
     type Output = PollOpt;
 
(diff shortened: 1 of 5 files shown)


carllerche:
Thanks for the PR.

This can't actually be deleted yet as it would be a breaking change. It can be flagged as deprecated and scheduled for removal later. 

That said, before this is actually deprecated, I'd like to check the repo history to see if there is any info about it when it was added.

carllerche:
Alright. Seems fine to mark as deprecated.

ahmedcharles:
I somehow pushed the wrong code last time, this should only deprecate the impl's, not remove them.

carllerche:
I rebased against master, which includes the CI fix.

PR #722, opened 2017-09-03
Prefer ErrorKind::WouldBlock

### LINKED ISSUE #616: Take advantage of ErrorKind::WouldBlock.into()
Stable since 1.14 and it's probably faster (https://github.com/rust-lang/rust/pull/42426)

arthurprs:
Fix #616 

Not a big difference but an improvement nevertheless.

Also includes a version bump in a separated commit.

```
pub fn before2() -> io::Error {
    std::io::Error::from_raw_os_error(111111)
}

pub fn after() -> io::Error {
    io::ErrorKind::WouldBlock.into()
}


example::before2:
        push    rbp
        mov     rbp, rsp
        push    rbx
        push    rax
        mov     rbx, rdi
        mov     esi, 111111
        call    std::io::error::Error::from_raw_os_error@PLT
        mov     rax, rbx
        add     rsp, 8
        pop     rbx
        pop     rbp
        ret

example::after:
        push    rbp
        mov     rbp, rsp
        mov     word ptr [rdi], 2561
        mov     rax, rdi
        pop     rbp
        ret

str.0:
        .ascii  "would block"
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,12 +19,12 @@ matrix:
 
     # Test compatibility
     #
-    # 1.13.0 is the oldest supported version of Rust. This value should NOT be
+    # 1.14.0 is the oldest supported version of Rust. This value should NOT be
     # changed without prior discussion.
     #
     # This build also deploys docs
     - os: linux
-      rust: 1.13.0
+      rust: 1.14.0
       before_script:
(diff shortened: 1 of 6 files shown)


carllerche:
Thanks!

PR #723, opened 2017-09-04
Document fuchsia module instability

### LINKED ISSUE #659: Document that the mio::fuchsia module is unstable
As I understand it, it is fair game for mio to make breaking changes to the mio::fuchsia module in a minor version of mio. This is important because mio::fuchsia depends on types from the magenta-sys crate which is far from stable.

This needs to be clearly documented in the module-level documentation of mio::fuchsia.

nivkner:
fixes #659


alexcrichton:
Thanks!

PR #736, opened 2017-09-28
lio_listio(2) support

asomers:
On FreeBSD and DragonflyBSD, kevent(2) has a distinct filter type for
use with lio_listio(2).  This commit adds 2 new public methods:
UnixReady::lio and UnixReady::is_lio.

### DIFF
--- a/src/event_imp.rs
+++ b/src/event_imp.rs
@@ -550,10 +550,10 @@ fn test_debug_pollopt() {
 #[derive(Copy, PartialEq, Eq, Clone, PartialOrd, Ord)]
 pub struct Ready(usize);
 
-const READABLE: usize = 0b00001;
-const WRITABLE: usize = 0b00010;
-const ERROR: usize    = 0b00100;
-const HUP: usize      = 0b01000;
+const READABLE: usize = 0b000001;
+const WRITABLE: usize = 0b000010;
+const ERROR: usize    = 0b000100;
+const HUP: usize      = 0b001000;
 
 impl Ready {
     /// Returns the empty `Ready` set.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -189,6 +189,12 @@ pub mod unix {
         EventedFd,
     };
     pub use sys::unix::UnixReady;
+    #[cfg(any(target_os = "bitrig", target_os = "dragonfly",
+              target_os = "freebsd", target_os = "ios", target_os = "macos",
+              target_os = "netbsd", target_os = "openbsd"))]
+    pub use sys::unix::BSDReady;
+    #[cfg(any(target_os = "dragonfly", target_os = "freebsd"))]
+    pub use sys::unix::FreeBSDReady;
 }
 
 #[cfg(target_os = "fuchsia")]
--- a/src/sys/unix/bsd/freebsd/mod.rs
+++ b/src/sys/unix/bsd/freebsd/mod.rs
@@ -1,0 +1,1 @@
+pub mod ready;
--- a/src/sys/unix/bsd/freebsd/ready.rs
+++ b/src/sys/unix/bsd/freebsd/ready.rs
@@ -1,0 +1,206 @@
+use event_imp::{Ready, ready_as_usize, ready_from_usize};
+use unix::UnixReady;
+use super::super::ready::BSDReady;
+
+use std::ops;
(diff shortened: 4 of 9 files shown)


carllerche:
Hmm... if these flags proliferate, I wonder if we should start adding other `Ready` variants. The danger lies in if the flags have the same value on two platforms but mean something different.

asomers:
Currently, I'm not aware of any conflicts between Linux and the BSDs (including OSX).  The various BSDs all agree on the meaning of the different kqueue filters.  And Linux, for the most part, uses totally different APIs to report the stuff that BSDs use kqueue for.  I think the likeliest route to conflict would be if mio decides to support different types of reactor on the same platform.  For example, if mio supports both epoll(2) based and io_get_events(2) based reactors on Linux.

asomers:
Since my pull request only affects the Unix code, I don't think it's responsible for the Appveyor failure.  Here's the error message from the failing test:

```
---- test_timer::test_cancel_timeout stdout ----
	thread 'test_timer::test_cancel_timeout' panicked at 'iterated too many times', test\test_timer.rs:267:12
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

asomers:
Which way would you like me to do it, @carllerche ?  Use `UnixReady` or add another `Readiness` type?

carllerche:
I think that I would rather add a new `Readiness` type. This is the safest strategy with regards to forwards compatibility.

carllerche:
cc @alexcrichton

asomers:
Ok, I created two new types: `BSDReady` and `FreeBSDReady`.  I'm not very happy about it though, because they create hundreds of lines of boiler plate in mio, and dozens in other consumers like tokio-core.

asomers:
The code duplication will only get worse as mio gains support for more event types.  For example, kevent(2) on FreeBSD supports `EVFILT_EMPTY`, `EVFILT_VNODE`, `EVFILT_PROC`, `EVFILT_PROCDESC`, `EVFILT_SIGNAL`, `EVFILT_TIMER`, and `EVFILT_USER` in addition to the types already supported by mio or by this pull request.  OSX has `EVFILT_PROC`, `EVFILT_SIGNAL`, `EVFILT_TIMER`, `EVFILT_MACHPORT`, `EVFILT_FS`, `EVFILT_USER`, and `EVFILT_VM`.  Handling all those event types as done in this PR would end up requiring either a separate `Ready` class per OS, or per event type.  Either way kind of defeats the purpose of having a single `Ready` class.

Instead of using separate classes for every platform-specific `Ready` bit, I have two suggestions for how to reduce code duplication while still maintaining a programmatic distinction between universally portable `Ready` bits and nonportable ones:
1) Throw everything into the base `Ready` class, but feature gate some of them on a `platform-specific-extensions` feature flag.  Applications intended to be fully portable can leave that flag cleared, providing instant compile failures if they inadvertently use a nonportable bit.
2) Throw everything into a `SysReady` class, with individual bits `#[cfg()]` gated as appropriate.  Make `Ready` a NewType for `SysReady` that will only expose methods for handling the portable bits.  Applications will be able to access `SysReady` directly if they must, but `Ready` will be sufficient for basic networking code.

Does either of those options sound appealing?

asomers:
BTW, the Travis failure is due to what looks like a QEMU error on one of the Android targets.

carllerche:
Sorry for the delay (you should feel free to harass me on IRC / Gitter).

I think that you are probably right with regards to the combinatorial explosion of permutations. I think that we can revisit just having a single `UnixReady` struct with conditionally defined values per platform.

Looking through existing code, the `Ready` values are actually not the ones provided by the OS. They are mapped in the OS specific selector implementation. Anyway, I think that my [original comment](https://github.com/carllerche/mio/pull/736#issuecomment-332878862) was incorrect.

asomers:
So are you in favor of something similar to my first commit, but with the `lio`-related functions disabled on platforms that don't provide a way to support them?

carllerche:
@asomers yes, the original commit + conditionally defining everything to only be available on platforms that provide those flags.

Thanks for actually implementing the version w/ `BSDReady` though, it was useful to see how bad my original thought was 👍

carllerche:
This PR is replaced by #780.

Issue #738, opened 2017-09-28
CI for FreeBSD


asomers:
Would you be interested in running CI tests on FreeBSD?  I volunteer to host it.  I'm already hosting CI for the [nix](https://github.com/nix-rust/nix) and [mio-aio](https://github.com/asomers/mio-aio) crates, and it would be no trouble to add this one.  If you agree, all that you'll need to do will be to add the GitHub commit hooks.

carllerche:
Yes, I think we'd be happy to add freebsd CI to the run.

asomers:
Ok, I'll set it up in polling mode.  I don't think I need any special access to the mio repository in order to post status reports.  Once it's working, I'll ask you to enable commit hooks.

asomers:
Here's an example pull request: https://github.com/nix-rust/nix/pull/663 .  And if you look at https://github.com/nix-rust/nix/commits/master you can see the status posted for each commit.  Nix uses bors, but otherwise it's similar to mio.  BuildBot will post its own status notification alongside Travis and Appveyor.

PR #739, opened 2017-09-29
Remove the FreeBSD build from Travis

asomers:
It uses BuildBot now
Issue #738

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,7 +89,7 @@ matrix:
       install:
         - rustup target add $TARGET
 
-    # NetBSD 
+    # NetBSD
     - os: linux
       env: TARGET=x86_64-unknown-netbsd DISABLE_TESTS=1
       install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,7 +89,7 @@ matrix:
       install:
         - rustup target add $TARGET
 
-    # NetBSD
+    # NetBSD 
     - os: linux
       env: TARGET=x86_64-unknown-netbsd DISABLE_TESTS=1
       install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,15 +89,7 @@ matrix:
       install:
         - rustup target add $TARGET
 
-    # FreeBSD
-    - os: linux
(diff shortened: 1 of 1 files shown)


asomers:
Buildbot swears that it updated the status for this issue, but it's not showing up.  Perhaps I don't have permission to post status updates?

carllerche:
@asomers ah, weird I missed this PR.

I can look in a bit.

asomers:
There are two problems ATM:
1) BuildBot isn't able to post status updates, probably because @asomers isn't a collaborator, though I can't find any GitHub documentation that says I need to be.
2) The FreeBSD tests are failing due to https://github.com/rust-lang/rust/issues/45038 .  I'm leaning toward disabling doc tests on FreeBSD to workaround that issue.

asomers:
The problem with the doc tests seems to be fixed in Rust 1.22.0.  I ran 2214 times in a row with no problems.  It was also fixed in FreeBSD's patched rust a long time ago, but the patches seem to have finally percolated upstream.

carllerche:
I added you as a collaborator, so you should be able to get things posting back.

carllerche:
Let me know if there is anything else that I have to do.

kpp:
@asomers Hello. Is there any progress on the issue?

asomers:
Everything works 99% of the time, but occasionally the mio tests cause my VM to lock up.  It's rare enough that I haven't been able to debug it, but common enough that it's a problem.  I'll try enabling it again with a newer rustc and see if the problem persists.

asomers:
rebased

carllerche:
@asomers looks like it is hanging on CI in this build?

I am trying to rebuild...

asomers:
It's not hanging; it's failing.  And it's failing because it needs https://github.com/carllerche/mio/pull/829 to be merged first.  And I _think_ I've fixed the hangs in my VM just by adding more RAM.  But we'll probably have to run for a full month to be certain.

carllerche:
Ah I missed that PR. I will get that one merged.

Issue #657, opened 2017-08-18
Document that UdpSocket::recv and recv_from do not read from the buffer


dtolnay:
```rust
fn recv(&self, buf: &mut [u8]) -> Result<usize>;
fn recv_from(&self, buf: &mut [u8]) -> Result<(usize, SocketAddr)>;
```

It is important for these to document that they only write to the buffer, not read from it, because that means it is safe to pass an uninitialized buffer. It would be wasteful to construct a buffer, initialize it to contain all 0 bytes, and pass it to UdpSocket.

frehberg:
And add to the docs, if recv returns with Ok(N), the following code must not use iterate over complete array, but iterate over a slice of the array buf[0,..,N-1].

frehberg:
can be closed?

dtolnay:
I don't think so. The docs in mio still need some work.

https://github.com/carllerche/mio/blob/5b9d20c4e01a63115ebdb6b87970fc9dd361a29b/src/net/udp.rs#L79-L81

frehberg:
Finally the patch has been rebased, and could be merged

carllerche:
The original PR got messed up somehow. I moved the commit to #775.

carllerche:
Thanks. This has been merged (f6f293f0785d737fc7b18aac5c5f2b1368490d80).

zheland:
The commit https://github.com/tokio-rs/mio/commit/f6f293f0785d737fc7b18aac5c5f2b1368490d80 updated the documentation for the wrong deprecated function that was doc-hidden and removed later. So the fix never reached the docs.

The commit updates docs for `recv_from` in `src/udp.rs` which was in:
```rust
#[deprecated(since = "0.6.6", note = "use net module instead")]
#[cfg(feature = "with-deprecated")]
#[doc(hidden)]
pub mod udp;
```
and removed in https://github.com/tokio-rs/mio/commit/e739c8a8a76a62e1939e24a0de6204d1720ee9a7 (remove deprecated code)

However, it is worth noting that the function by itself calls `std::net::UdpSocket` that doens't provide any "only write to the buffer" guarantees and, on the contrary, contain a requirement for "valid byte array" in docs:
```rust
    /// The function must be called with valid byte array `buf` of sufficient size to
    /// hold the message bytes. If a message is too long to fit in the supplied buffer,
    /// excess bytes may be discarded.
    /// ...
[...]

PR #740, opened 2017-10-03
Document that UdpSocket::recv and recv_from do not read from the buff…

frehberg:
…er (#657)

### DIFF
--- a/src/udp.rs
+++ b/src/udp.rs
@@ -80,8 +80,17 @@ impl UdpSocket {
         self.sys.send_to(buf, target).map_non_block()
     }
 
-    /// Receives data from the socket. On success, returns the number of bytes
-    /// read and the address from whence the data came.
+    /// Receives data from the socket and stores data in the supplied buffer `buf`. On success,
+    /// returns the number of bytes read and the address from whence the data came.
+    ///
+    /// The function must be called with valid byte array `buf` of sufficient size to
+    /// hold the message bytes. If a message is too long to fit in the supplied buffer,
+    /// excess bytes may be discarded.
+    ///
+    /// The function does not read from `buf`, but is overwriting previous content of `buf`.
+    ///
+    /// Assuming the function has read `n` bytes, slicing `&buf[..n]` provides
+    /// efficient access with iterators and boundary checks.
     pub fn recv_from(&self, buf: &mut [u8])
                      -> io::Result<Option<(usize, SocketAddr)>> {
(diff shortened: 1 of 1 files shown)


src/udp.rs
```diff
@@ -82,6 +82,13 @@ impl UdpSocket {
 
     /// Receives data from the socket. On success, returns the number of bytes
     /// read and the address from whence the data came.
+    ///
+    /// The function must be called with valid byte array `buf` of sufficient size to
+    /// hold the message bytes. If a message is too long to fit in the supplied buffer,
+    /// excess bytes may be discarded.
+    ///
+    /// The function is overwriting previous content of `buf`. Slicing
+    /// `&buf[..NumRead]` allows efficient access and boundary checking at compile time.
```
KodrAus: Assuming `NumRead` is the result of `recv_from` I think a more idiomatic name for it is `num_read`, so we know it's some local variable.

src/udp.rs
```diff
@@ -82,6 +82,13 @@ impl UdpSocket {
 
     /// Receives data from the socket. On success, returns the number of bytes
     /// read and the address from whence the data came.
+    ///
+    /// The function must be called with valid byte array `buf` of sufficient size to
```
KodrAus: I think we still need to mention that this function only writes into the buffer and never reads from it.

src/udp.rs
```diff
@@ -95,7 +102,14 @@ impl UdpSocket {
     }
 
     /// Receives data from the socket previously bound with connect(). On success, returns
-    /// the number of bytes read and the address from whence the data came.
+    /// the number of bytes read.
+    ///
+    /// The function must be called with valid byte array `buf` of sufficient size to
```
KodrAus: I think we still need to mention that this function only writes into the buffer and never reads from it.

frehberg:
API-Doc has been extended, adding these aspects. I had to re-arrange the text a bit, does it please?

carllerche:
Thanks... this looks good to me. Unfortunately, it looks like Android CI broke... which going to block me from merging until it is fixed 😢

carllerche:
A CI fix has been merged to master. Would you mind rebasing when you have a moment? Thanks!

KodrAus:
@frehberg How's it going? Would you like a hand rebasing this?

carllerche:
Something wonky happened with this PR. The commit was rebased, but didn't make it to the PR. I moved it here: #775

PR #742, opened 2017-10-05
Poll::poll(): Retry select() on EINTR

### LINKED ISSUE #494: `poll.poll()` and `EINTR`
`poll.poll()` can currently return `io::ErrorKind::Interrupted` when an inner system call is being interrupted. 

As all examples tend to ignore this, I was wondering whether this behavior was intentional, or if the function should silently retry when it happens. I'd be glad to submit a pull request for the later.

asayers:
Usually the user should retry poll() on EINTR; however, this is an
easily-overlooked case. Let's improve the reliability of mio-based
programs by handing EINTR by default.

Closes #494.

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -1133,32 +1133,48 @@ impl Poll {
     }
 
     #[inline]
-    fn poll2(&self, events: &mut Events, timeout: Option<Duration>) -> io::Result<usize> {
+    fn poll2(&self, events: &mut Events, mut timeout: Option<Duration>) -> io::Result<usize> {
         // Compute the timeout value passed to the system selector. If the
         // readiness queue has pending nodes, we still want to poll the system
         // selector for new events, but we don't want to block the thread to
         // wait for new events.
-        let timeout = if timeout == Some(Duration::from_millis(0)) {
+        if timeout == Some(Duration::from_millis(0)) {
             // If blocking is not requested, then there is no need to prepare
             // the queue for sleep
-            timeout
         } else if self.readiness_queue.prepare_for_sleep() {
             // The readiness queue is empty. The call to `prepare_for_sleep`
             // inserts `sleep_marker` into the queue. This signals to any
             // threads setting readiness that the `Poll::poll` is going to
             // sleep, so the awakener should be used.
-            timeout
         } else {
             // The readiness queue is not empty, so do not block the thread.
-            Some(Duration::from_millis(0))
-        };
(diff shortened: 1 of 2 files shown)


carllerche:
Thanks. I believe that the timeout would need to be updated before `poll` is attempted again?

asayers:
Sorry for the slow response. Yes, you're right.

carllerche:
A CI fix has been merged to master. Would you mind rebasing when you have a moment? Thanks!

asayers:
Rebased. The iOS CI run timed out after starting the test program. I can't tell if this failure is genuine - is it possible to rerun the test?

carllerche:
It probably is spurious. iOS CI is really janky.

asayers:
Added a line to the changelog.

src/poll.rs
```diff
@@ -1132,32 +1132,48 @@ impl Poll {
...
+                    self.readiness_queue.inner.awakener.cleanup();
+                    break;
+                }
+                Ok(false) => break,
+                Err(ref e) if e.kind() == io::ErrorKind::Interrupted => {
+                    // Interrupted by a signal; update timeout if necessary and retry
+                    if let Some(to) = timeout {
+                        let elapsed = now.elapsed();
+                        timeout = if elapsed >= to {
+                            Some(Duration::from_millis(0))
```
carllerche: At this point, would you not break? While probably an implausible situation, if each call to `select` returns w/ an interrupt, this would cause an infinite loop?

asayers: I assumed that calling `select` with a timeout of 0 is non-blocking and therefore can't return EINTR, but to my surprise a bit of googling revealed that "non-blocking syscalls don't return EINTR" is not actually documented behaviour. I'll change to a break.

asayers:
* Rebased on master
* Skip retry when timeout has expired

carllerche:
🎉 CI passes.

Thanks for hanging in there.

asayers:
Thanks!

bbshelper:
My application is required to gracefully shutdown itself upon `SIGTERM`. It does so by changing a global variable in signal handler, and check the variable after `poll` returns `ErrorKind::Interrupted`. With this change, `poll` won't pass that information, breaking my application. Will it be possible to see the old behavior, maybe with a new API?

carllerche:
Yeah, it would be fine to add a `poll_interruptible`. Sorry about that.

PR #747, opened 2017-10-14
Avoid runtime panics in IoVec on Windows

### LINKED ISSUE #746: Iovec API change causing panics
The iovec crate recently made an api change in 0.1.1 that is causing panics (see https://github.com/carllerche/iovec/commit/007031f4b4f7f518ea6b1888b0bc189604430dbd). Using the deprecated methods panics when a zero length buffer is passed.

alexcrichton:
Closes #746

### DIFF
--- a/src/sys/windows/tcp.rs
+++ b/src/sys/windows/tcp.rs
@@ -246,7 +246,10 @@ impl TcpStream {
     }
 
     pub fn read(&self, buf: &mut [u8]) -> io::Result<usize> {
-        self.readv(&mut [buf.into()])
+        match IoVec::from_bytes_mut(buf) {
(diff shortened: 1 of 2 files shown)


src/sys/windows/tcp.rs
```diff
@@ -246,7 +246,11 @@ impl TcpStream {
     }
 
     pub fn read(&self, buf: &mut [u8]) -> io::Result<usize> {
-        self.readv(&mut [buf.into()])
+        if buf.len() > 0 {
+            self.readv(&mut [buf.into()])
```
jolhoeft: Would it be better to IoVec::from_bytes_mut here?

alexcrichton: Indeed!

PR #754, opened 2017-11-01
Ensure html_root_url is kept in sync

### LINKED ISSUE #613: Outdated html_root_url
The html_root_url of version 0.6.8 [here](https://github.com/carllerche/mio/blob/v0.6.8/src/lib.rs#L78) points to the documentation for version 0.6.1.

I would recommend adding a reminder in Cargo.toml like https://github.com/serde-rs/serde/commit/dc7ab2696a14fe5ca219ac09f8496bdc613ff602.

mgeisler:
This adds two new unit tests: one that checks that the html_root_url
attribute always uses the current crate version number, and one that
checks that the TOML code blocks in the README use the correct version
number.

This is implemented by my version-sync crate -- I hope it's useful!

Fixes #613.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,9 +41,10 @@ miow   = "0.2.1"
 kernel32-sys = "0.2"
 
 [dev-dependencies]
-env_logger = { version = "0.4.0", default-features = false }
-tempdir    = "0.3.4"
-bytes      = "0.3.0"
+env_logger   = { version = "0.4.0", default-features = false }
+version-sync = "0.4"
+tempdir      = "0.3.4"
+bytes        = "0.3.0"
 
 [[test]]
 name = "test"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -75,7 +75,7 @@
 //!
 //! ```
 
-#![doc(html_root_url = "https://docs.rs/mio/0.6.1")]
+#![doc(html_root_url = "https://docs.rs/mio/0.6.11")]
 #![crate_name = "mio"]
 
 #![deny(warnings, missing_docs, missing_debug_implementations)]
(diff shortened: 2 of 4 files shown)


mgeisler:
This might be more complicated than I hoped since version-sync doesn't support Rust 1.14 (version 1.15 is the oldest it can be compiled on).

mgeisler:
Please let me know if there's any interest in the general approach — then I can see about removing the version-sync dependency on versions and targets where it doesn't work (remove with `sed` in the Travis build).

carllerche:
Thanks for the PR.

I'd rather not add another dependency and instead deal with this by having a `RELEASE` doc and wait for this to get fixed up stream. But, again, thanks for the attempt :).

twmb:
As an alternative, I've used 

```yaml
before_install: |
  version=$(grep "^version =" Cargo.toml | sed -r  's/^[^0-9]*([0-9]+.[0-9]+.[0-9]+).*/\1/') &&
  docs_version=$(grep -R 'docs\.rs' src | head -n 1 | sed -r 's/^[^0-9]*([0-9]+.[0-9]+.[0-9]+).*/\1/') &&
  test "$version" == "$docs_version"
```

in my .travis.yml in the past.

mgeisler:
@carllerche Sure, no problem. Note that it's only a new dev-dependency, so only people who run the MIO tests are affected by this.

I'll see about making another PR that at least fixes the outdated version number :-)

carllerche:
@twmb I think a smoke test like that would be good.

@mgeisler Sure re: PR that fixes the version :)

mgeisler:
I've submitted #760 to fix the version number.

Issue #665, opened 2017-08-18
Rustdoc examples for mio::net methods


dtolnay:
Per the [`C-EXAMPLE`](https://rust-lang-nursery.github.io/api-guidelines/documentation.html#c-example) guideline.

Keep in mind that the purpose of an example is not always to show *how to use* the item. Users can be expected to understand the syntax of how to invoke a method or match on an enum. Rather, an example is often intended to show *why someone would want to use* the item.

---

- [ ] [TcpListener](https://docs.rs/mio/*/mio/net/struct.TcpListener.html)::bind
- [ ] TcpListener::from_listener
- [ ] TcpListener::accept
- [ ] TcpListener::local_addr
- [ ] TcpListener::try_clone
- [ ] TcpListener::set_ttl
- [ ] TcpListener::ttl
- [ ] TcpListener::set_only_v6
- [ ] TcpListener::only_v6
- [ ] TcpListener::take_error

---

- [ ] [TcpStream](https://docs.rs/mio/*/mio/net/struct.TcpStream.html)::connect
- [ ] TcpStream::connect_stream
- [ ] TcpStream::from_stream
- [ ] TcpStream::peer_addr
- [ ] TcpStream::local_addr
- [ ] TcpStream::try_clone
- [ ] TcpStream::shutdown
- [ ] TcpStream::set_nodelay
- [ ] TcpStream::nodelay
- [ ] TcpStream::set_recv_buffer_size
- [ ] TcpStream::recv_buffer_size
- [ ] TcpStream::set_send_buffer_size
- [ ] TcpStream::send_buffer_size
- [ ] TcpStream::set_keepalive
- [ ] TcpStream::keepalive
- [ ] TcpStream::set_ttl
- [ ] TcpStream::ttl
- [ ] TcpStream::set_only_v6
- [ ] TcpStream::only_v6
- [ ] TcpStream::set_linger
- [ ] TcpStream::linger
- [ ] TcpStream::take_error
- [ ] TcpStream::read_bufs
- [ ] TcpStream::write_bufs

---

- [ ] [UdpSocket](https://docs.rs/mio/*/mio/net/struct.UdpSocket.html)::bind
- [ ] UdpSocket::from_socket
- [ ] UdpSocket::local_addr
- [ ] UdpSocket::try_clone
- [ ] UdpSocket::send_to
- [ ] UdpSocket::recv_from
- [ ] UdpSocket::send
- [ ] UdpSocket::recv
- [ ] UdpSocket::connect
[...]

PR #762, opened 2017-11-16
Added some examples fo mio::net::UdpSocket.

lanedraex:
Some rustdoc examples: (#665)

Includes examples for: UdpSocket, bind, local_addr, send_to, recv_from, set_broadcast, broadcast.

### DIFF
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -52,9 +52,9 @@ use std::net::{self, Ipv4Addr, Ipv6Addr, SocketAddr};
 /// poll.register(&sender_socket, SENDER, Ready::writable(), PollOpt::edge())?;
 /// poll.register(&echoer_socket, ECHOER, Ready::readable(), PollOpt::edge())?;
 ///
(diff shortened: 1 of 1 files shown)


carllerche:
Thanks for the PR!

I think including a full example for the struct level docs is probably good, however, for each individual function, would it be possible to reduce down the example to focus on the function as question? This should make it easier to see the relevant bits. Ideally, none of the fn level examples would include a full event loop.

lanedraex:
Hey, should the examples hide the event loop with `#` or should I simply "fake" calls to the functions like:

```rust
# Ok(())
socket.send_to(&buffer, &addr);
```

Or something else entirely?

carllerche:
@lanedraex sticking with fake calls is fine. You can use the `no_run` annotation I think it is.

carllerche:
@lanedraex Thanks, these examples look great.

lanedraex:
I'll add examples to the other methods.

Also, just noticed that I made a typo in the pull request title, sorry.

PR #764, opened 2017-11-18
Split `Poll` into `Poll` and `Register`

### LINKED ISSUE #757: Split `Poll` into `Poll` and `Register`.
As part of [rewriting the readiness queue](https://github.com/carllerche/mio/commit/43a69f218957ea7f43bb821734dbfbad09078c6b), `Poll::poll` acquired a lock. This is because `Poll` needed to be `Sync` so that I/O resources could be merged from other threads.

However, `Poll` really provides two separate capabilities:

* Registering I/O resources.
* Polling for I/O events.

Only I/O registration needs to be `Sync`. If the `Poll` type is split into two types, then `Poll::poll` can take `&mut self` and the lock can be removed.

** But, what about polling from multiple threads? **

Well, that isn't supported today anyway. I also suspect that the answer isn't to make `Poll` `Sync` but to add some clone type functionality so that each thread can get its own `Poll` handle. This would also better support the max concurrency settings that IOCP has.

carllerche:
This also makes `Poll::poll` require `&mut self` since `Poll` is no
longer expected to be stored in a `Sync` context.

Resolves #757

### DIFF
--- a/src/event_imp.rs
+++ b/src/event_imp.rs
@@ -1,11 +1,11 @@
-use {Poll, Token};
+use {Register, Token};
 use std::{fmt, io, ops};
 
-/// A value that may be registered with `Poll`
+/// A value that may be registered with `Register`
 ///
-/// Values that implement `Evented` can be registered with `Poll`. Users of Mio
-/// should not use the `Evented` trait functions directly. Instead, the
-/// equivalent functions on `Poll` should be used.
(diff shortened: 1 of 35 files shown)


carllerche:
cc @alexcrichton

carllerche:
I'm going to merge this soonish unless there are any objections.

PR #768, opened 2017-12-01
bump slab to 0.4

ignatenkobrain:


carllerche:
Thanks for the PR. It doesn't look like CI is happy with it. However, since slab is only used in deprecated code (I think), maybe it isn't worth trying to update the crate?

ignatenkobrain:
Okay, seems that many crates depend on slab 0.3, so I will create additionalversion of that in Fedora.

ignatenkobrain:
But it would be still nice to update it, because tokio-core requires slab0.4 and mio which requires slab0.3..

ignatenkobrain:
@carllerche so the new version of mio didn't remove dependency and it is blocking me from packaging mio in Fedora... So I spent some time and ported it.. Please review this PR ;)

src/timer.rs
```diff
@@ -94,24 +91,9 @@ const TICK_MAX: Tick = u64::MAX;
...
-type Slab<T> = slab::Slab<T, ::Token>;
-
-pub type Result<T> = ::std::result::Result<T, TimerError>;
+pub type Result<T> = ::std::result::Result<T, io::Error>;
 // TODO: remove
 pub type TimerResult<T> = Result<T>;
-
-
-#[derive(Debug)]
-pub struct TimerError {
```
carllerche: Why is this removed?

briansmith:
@ignatenkobrain Any chance that you could update this? If not, do you mind if I take it over?

carllerche:
Thanks for the initial work. These changes have landed as part of #824.

PR #770, opened 2017-12-01
bump bytes to 0.4

ignatenkobrain:


carllerche:
Thanks for the PR. It looks like CI doesn't like this?

carllerche:
Sorry, accidentally closed.

ignatenkobrain:
Let me take look at it..

ignatenkobrain:
@carllerche I'm totally lost with changes in `bytes`. I've fixed some errors, but I don't know how to replace RingBuf and SliceBuf...

ignatenkobrain:
@carllerche if you could port it yourself, would be very helpful for me since I need this in order to package for Fedora.

carllerche:
@ignatenkobrain I'm unsure if there will be an effort to update bytes in mio 0.6 as most of (if not all of) the usage of `bytes` will be removed in 0.7 I think?

PR #772, opened 2017-12-12
RFC: Balancing system evens and custom events

frehberg:
Hi,  I would like to ask you for comments, please do not merge yet.

After analysation of the Poll functionality I did a bit refactoring to balance system events and custom events. 

Now the sleep_marker is rotating token, the end_marker became obsolete. When the sleep_marker is back at tail, the routine will check for system events. Afterwards setting sleep_marker back to head of ReadinessQueue, defining the set of pending custom events for that period.

The test "with_small_events_collection" is still failing, seems a condition is not met yet. Will need some more time to finalize it.
Comments are welcome, maybe you see the flaw causing this test fail?

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -2029,6 +2029,7 @@ impl ReadinessQueue {
         // sleep. If it is not, then the queue is not empty and there should be
         // no sleeping.
         if tail == sleep_marker {
+            // False assumption: having 3 meta nodes the queue may be empty even if head!=tail
             return self.inner.head_readiness.load(Acquire) == sleep_marker;
         }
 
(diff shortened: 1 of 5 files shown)


carllerche:
Hi, thanks for the PR. I tried looking through it. I'm not following everything, but it seems like this would invert the problem and put too much emphasis on the custom readiness queue instead of the OS queue.

Maybe you could write some pseudo code explaining the logic.

frehberg:
Hi Carl, you are right, if the capacity of the events-container is low, this implementation is emphasing custom events; but non will starve either one. it is organised using time-slices or call it 'cycles'. The sys-selector starts a cycle and the sleep_marker is used to define a set of  custom-events in the queue that belong to this cycle, that must be consumed before starting the next cycle. Each cycle is processing system-events _and_ custom events.

The pseudo-code might look like:

```
poll([out]events):
  if tail!=sleep_marker {
     // continue a cycle
     while events.has_capcity && tail!=sleep_marker {
        push-custom-event-to-user-container-events;
     }
     // POSt: if tail==sleep_marker, cycle is ending
  } else {
     // begin a new cycle
     blocking-wait-in-sys-selector-and-fill-sys-events-container;
     consume-sleep-marker-and-re-attach-to-head; 
     // POST: all events between tail and sleep-marker do belong
     //       to this sys-selector-cycle
     if awakener-token-present {
        consume-single-token-from awakener; // due to concurrency just one
        while events.has_capcity && tail!=sleep_marker {
           push-custom-event-to-user-container-events;
        }
        // POST: if tail==sleep_marker, cycle is ending
     }
  }
}

```

frehberg:
Wondering why CI is failing  for "test_register_deregister" on Windows; the branch works for me  on Win7-x64!

carllerche:
CI tends to catch more concurrency bugs due to the unpredictable execution characteristics of running on a low power / over subscribed VM.

PR #773, opened 2017-12-13
add TcpStream::peek

seanmonstar:


src/sys/windows/tcp.rs
```diff
@@ -252,6 +252,44 @@ impl TcpStream {
         }
     }
 
+    pub fn peek(&self, buf: &mut [u8]) -> io::Result<usize> {
+        let mut me = self.inner();
+
+        match me.read {
```
carllerche: It looks like this block is copied verbatim from `readv`? Would you mind extracting this to a `before_read` fn or something like that?

seanmonstar: Done.

seanmonstar:
Ah, so this requires a newer version of Rust than what is being tested on Travis. Breaking change needing 0.7, or just Deal With It?

seanmonstar:
CI passed!

Issue #658, opened 2017-08-18
Winapi in the public API of mio


dtolnay:
On windows, the `mio::windows::Overlapped::new` and `as_mut_ptr` methods refer to winapi in their signature.

```rust
fn new(cb: fn(_: &winapi::minwinbase::OVERLAPPED_ENTRY)) -> Overlapped;
fn as_mut_ptr(&self) -> *mut winapi::minwinbase::OVERLAPPED;
```

Since winapi is pre-1.0, this makes it risky to stabilize mio. If a future version of winapi were to change these structs in certain ways, mio may be forced into a breaking change to pick them up.

Let's either figure out a way to keep these winapi types out of mio's public API, or lay out a plan for how we would deal with a hypothetical future release of winapi. If the latter, the plan should be clearly documented in comments in the windows module.

dtolnay:
From IRC:

> **\<WindowsBunny>** winapi 0.3 is coming very soon and you better want to upgrade to it

carllerche:
Basically, we should figure out how to remove winapi from all public apis (unless it plans to hit 1.0 soon)

retep998:
winapi has no plans for 1.0 at the moment due to several critical Rust features still being either unstable or unimplemented.

steffengy:
@carllerche 
This can be closed since the winapi 0.3 update.

Since this still hasn't landed, and we can't really backport the 0.3 migration,
are there news for mio on the 0.7 front?
Any plans to push some open points that won't see progress to another milestone,
insteadof delaying this indefinitely?

carllerche:
The main issue right now is time on my part. My current focus is on getting the Tokio stack stable.

I can try to triage the remaining 0.7 issues to see if others can jump on them. The biggest one is going to be to figure out multi thread polling.

Is there urgency on getting a winapi update released? The impact should be minimal and mostly be limited to having multiple versions of the crate compiled.

steffengy:
That's understandable and totally reasonable. In regards to impact there's
- crate-duplication (which affects most async things, mostly showing in **build times**) 
- a few mostly internal changes

So nothing that warrants making this a critical priority (so far).

**I personally just don't like that it's not remotely foreseeable when this will land**
It could be just a few months, but it also could be towards the end of 2018 already.

Currently it doesn't seem realistic to me, 
that mio 0.7 will be ready before the next (major stable) tokio release. 
[...]

PR #777, opened 2017-12-21
migrate to winapi 0.3

steffengy:
This is not ready as-is.
- https://github.com/alexcrichton/miow/pull/12 needs to land first.
- https://github.com/carllerche/iovec/pull/11 should land first 
(the representation change from https://github.com/carllerche/iovec/commit/db866454eb31b3dafe0308fff3e8551744f6c654 is required)


carllerche:
Unfortunately, this is a breaking change as  the winapi is exposed.

Ideally, winapi would be hidden from the user as per #658.

steffengy:
Yeah sure, it's also tracked as breaking change for miow.

I don't see a way to abstract winapi away here since that only shifts the winapi-update-problem away:
- Using own wrapper types would break third-party crates on updates
- Not exposing it at all would require unsafe casts and lead to sudden failures that are annoying to track down

Personally I currently see no other way than a breaking release for such cases and then 
updating all dependant crates one by one
(which in any case would be necessary since they need to use a matching winapi version if they depend on this "feature")

Of course if somebody has a practical idea to tackle this, we can still address that after this has landed, so this is the last breaking release of this sort, even though I thinks it's unlikely we'll come up with a solution there any time soon.

carllerche:
I am currently favoring having Mio expose its own windows API types. Since the struct layout is defined by the actual windows API and C ABI, unsafe cast between should be safe. And, either way, since we are dealing w/ FFI types, unsafety is required.

carllerche:
@steffengy This PR can't land until 0.7, which means that iovec 0.2 is needed as well.

It will probably take a little bit or this change to happen.

steffengy:
@carllerche 
Yeah sure, no rush on it.

> I am currently favoring having Mio expose its own windows API types

Definitely also an option, even though it's a bit of additional code overhead.
What benefit has this from simply removing winapi from the public API-surface and just
returning/taking a `*mut c_void` (for public API methods)?
(The user will then have to deal with the casting as if we'd use our own types and we internally
do not have to mess with the declarations, linking, etc.)

carllerche:
@steffengy yeah, actually returning / taking `*mut c_void` is probably the best option.

steffengy:
@carllerche 
Implemented a first version using `*mut c_void`.
Not quite sure yet if it's safe as-is (transmute).

test/test_echo_server.rs
```diff
@@ -22,28 +21,23 @@ impl EchoConn {
...
 
-                self.buf = Some(buf);
                 self.interest.insert(Ready::writable());
             }
             Ok(Some(r)) => {
                 debug!("CONN : we wrote {} bytes!", r);
 
-                self.mut_buf = Some(buf.flip());
-
+                self.buf.clear();
```
carllerche: This should be `self.buf.drain(..r)`. It is possible for `r != self.buf.len()`.

test/test_echo_server.rs
```diff
@@ -168,20 +157,19 @@ impl EchoClient {
...
-                while buf.has_remaining() {
-                    let actual = buf.read_byte().unwrap();
-                    let expect = self.rx.read_byte().unwrap();
-
-                    assert!(actual == expect, "actual={}; expect={}", actual, expect);
+                    assert!(*actual == expect, "actual={}; expect={}", actual, expect);
                 }
 
-                self.mut_buf = Some(buf.flip());
+                buf.clear();
```
carllerche: Same as above, I think this needs to be `buf.drain(..r)`.

test/test_unix_echo_server.rs
```diff
@@ -25,27 +24,23 @@ impl EchoConn {
...
                 debug!("client flushing buf; WOULDBLOCK");
 
-                self.buf = Some(buf);
                 self.interest.insert(Ready::writable());
             }
             Ok(Some(r)) => {
                 debug!("CONN : we wrote {} bytes!", r);
 
-                self.mut_buf = Some(buf.flip());
+                self.buf.clear();
```
carllerche: `self.buf.drain(..r)`.

test/test_unix_echo_server.rs
```diff
@@ -163,21 +154,21 @@ impl EchoClient {
...
+                for actual in buf.iter() {
+                    let expect = self.rx[0];
+                    self.rx = &self.rx[1..];
 
-                    assert!(actual == expect, "actual={}; expect={}", actual, expect);
+                    assert!(*actual == expect, "actual={}; expect={}", actual, expect);
                 }
 
-                self.mut_buf = Some(buf.flip());
+                buf.clear();
```
carllerche: If I'm reading this correctly, this `clear` is fine because the `for` loop above checks every element.

If I am incorrect, please let me know :)

steffengy: Yeah that's the highest layer, where we received a "full message" and check if it's what we expected.

test/test_echo_server.rs
steffengy: This should be okay, as below, highest layer with same for-loop above.

carllerche:
I was about to merge, and I noticed this was based on master and not v0.7.x :(

I will take a stab to fix the merge conflicts.

carllerche:
Alright, I merged... hopefully I didn't botch it!

steffengy:
Oh, sorry about that.
I don't spot anything obviously wrong after the merge.

Interestingly, those annoying-to-port unix tests don't even exist anymore in 0.7, 
so even that would've been easier.

Eijebong:
This didn't end up on master with the other breaking changes, any reason why ?

PR #780, opened 2017-12-22
lio_listio(2) support

asomers:
This PR adds support for lio_listio(2) on supported platforms.  It also `#[cfg()]` gates the existing aio-related methods for supported platforms only.

This replaces PR #736 .  I'm leaving that old PR online to maintain a record of the `BSDReady` approach.


carllerche:
Looks good. I pushed a fix to hopefully fix CI.

The AIO functions were not previously guarded by platform, so we have to leave them for backwards compatibility.

PR #782, opened 2017-12-22
Add function Events::clear(&mut self)

frehberg:
Add function Events::clear(&mut self) and expose to other layers of the module or user-app

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -1419,6 +1419,39 @@ impl Events {
             pos: 0
         }
     }
+
+    /// Clearing all `Event` values from container explicitly.
+    ///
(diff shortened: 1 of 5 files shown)


frehberg:
In kqueue.rs the Selector::select() method does not perform any clearing. So I am not sure the implementation Events::clear() is correct there!?

frehberg:
looks good to me. Next time I will not copy style of the legacy code ;)

carllerche:
@frehberg no worries, Mio is an old lib that has evolved across changing Rust code styles.

Once rustfmt is stable, I plan to do a bulk reformat of everything.

PR #798, opened 2018-01-07
adapt bench_poll for split of poll and register

frehberg:
Hi Carl, the fix looks a bit ugly. Any reason there is no Events::len() method?

### DIFF
--- a/benches/bench_poll.rs
+++ b/benches/bench_poll.rs
@@ -46,10 +46,7 @@ fn bench_poll(bench: &mut Bencher) {
 
         while n < NUM {
             if poll.poll(&mut events, None).is_ok() {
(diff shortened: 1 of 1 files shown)


carllerche:
You can always use [`Iterator::count`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.count).

`len` is deprecated as there is no way for Mio to get the number of polled events in constant time. This is because Mio reserves the right to internally make registrations which will be stored in the events struct.

frehberg:
using `Iterator::count` looks much better, thanks for the hint

frehberg:
Ready for merge! (?)

frehberg:
Please let me know if anything is missing.

carllerche:
Thanks!

PR #800, opened 2018-01-08
Bench bandwidth

frehberg:
First draft of bandwidth benchmarks, comments are welcome. Some questions:

- Should these bench-tests be placed in separate bench-file?
- Do you think the measurements are comparable to each other?
- Could the measurements be stored in DB?
- Could the measured benchmarks be converted to diagram and embedded into MD markup-file?
- ff

Code is going to be cleaned up in near future and going to be commented.


carllerche:
Thanks.

I'd probably move these to `bench_bandwidth.rs` file. The other questions can be punted to later work I would guess. I'm not sure what "ff" means.

PR #805, opened 2018-01-11
Remove lio_listio(2) support for DragonFly - Fix compile

mneumann:
There is no EVFILT_LIO on DragonFly. Commit 00d85730 failed to
compile on DragonFly. There seems to be no (official) LIO support,
at least it is not mentioned anywhere in manpages etc.

With this commit, mio can be build again on DragonFly and all tests pass.

### DIFF
--- a/src/sys/unix/kqueue.rs
+++ b/src/sys/unix/kqueue.rs
@@ -299,7 +299,7 @@ impl Events {
                     event::kind_mut(&mut self.events[idx]).insert(UnixReady::aio());
                 }
             }
-#[cfg(any(target_os = "dragonfly", target_os = "freebsd"))]
+#[cfg(any(target_os = "freebsd"))]
             {
                 if e.filter == libc::EVFILT_LIO {
                     event::kind_mut(&mut self.events[idx]).insert(UnixReady::lio());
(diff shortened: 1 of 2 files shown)


carllerche:
Is there a way to get DragonFly added to the CI run?

asomers:
Not easiliy.  Neither rustup nor japaric/cross support Dragonfly.  I suppose that means that for Dragonfly, the toolchain is build from ports?  Adding Dragonfly to CI would require creating a buildbot/jenkins instance, and it may not be possible to always use the same version of Rust for Dragonfly as for Travis.

mneumann:
Yes, rust on DragonFly is build from ports (we call them "dports"). There is a docker script in the rust repo which cross-compiles Rust for DragonFly on Linux, but due to missing compute resources it never got enabled. So there are no "official" builds for DragonFly (yet).

mneumann:
@carllerche: can we merge this, despite having no CI for DragonFly?

carllerche:
Could you remove Dragonfly from the list of supported platforms?

Unfortunately, I can't really claim it is supported unless there is CI. I can merge the PR otherwise.

mneumann:
@carllerche removed DragonFly from the list and "merged" the two commits into one.

PR #807, opened 2018-01-12
Illumos support 

cneira:
I have a PR to add EPOLL(5) to libc so mio and iovec could be built in Illumos.   As mio and iovec are using libc tag 0.2.19.  It's needed to wait until the PR is upstream in Master and that tag.
In the meantime, if some needs to build mio or iovec in Illumos he or she needs to use this dep.

libc =  { git = "https://github.com/cneira/libc.git", branch = "mio" }

### DIFF
--- a/src/sys/unix/mod.rs
+++ b/src/sys/unix/mod.rs
@@ -3,10 +3,10 @@ use libc::{self, c_int};
 #[macro_use]
 pub mod dlsym;
 
-#[cfg(any(target_os = "linux", target_os = "android"))]
+#[cfg(any(target_os = "linux", target_os = "android", target_os = "solaris"))]
 mod epoll;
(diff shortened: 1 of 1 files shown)


carllerche:
Ok, can you let me know when the PR lands?

Also, is there any way to get Ci support? Otherwise, it can't be an officially supported platform.

cneira:
PR has landed. you'll need to update to the latest libc.
As far as CI integration, Travis does not provide an illumos image, so at the moment this is not possible.

carllerche:
It looks like `cross` has *some* support for sun: https://github.com/japaric/cross

Maybe we could at least get CI to try compiling?

bgermann:
The libc changes are released.

PR #803, opened 2018-01-09
Add bench bandwidth

frehberg:
Same PR as before, just the commits have been squashed

PR #810, opened 2018-01-26
Bench bandwidth

carllerche:
This is a continuation of #798 and #803

/cc @frehberg


frehberg:
Seems, the error is not related to the patch! Performing an "adb push" failed, pushing files onto the android test-target.

carllerche:
Gone stale.

PR #813, opened 2018-01-28
Cleanup some more deprecated things.

ahmedcharles:


src/sys/unix/ready.rs
```diff
@@ -126,9 +126,7 @@ impl UnixReady {
 
     #[cfg(not(any(target_os = "dragonfly",
         target_os = "freebsd", target_os = "ios", target_os = "macos")))]
-    #[deprecated(since = "0.6.12", note = "this function is now platform specific")]
-    #[doc(hidden)]
-    pub fn aio() -> UnixReady {
+    fn aio() -> UnixReady {
```
carllerche: Why is this fn still needed?

src/sys/fuchsia/net.rs
```diff
@@ -245,12 +245,10 @@ impl TcpListener {
         })
     }
 
-    #[allow(deprecated)]
     pub fn set_only_v6(&self, only_v6: bool) -> io::Result<()> {
```
carllerche: This fn and `only_v6` probably can be deleted now?

src/sys/unix/ready.rs
```diff
@@ -384,8 +378,8 @@ impl fmt::Debug for UnixReady {
             (UnixReady(Ready::writable()), "Writable"),
             (UnixReady::error(), "Error"),
             (UnixReady::hup(), "Hup"),
-            #[allow(deprecated)]
-            (UnixReady::aio(), "Aio")];
+            (UnixReady::aio(), "Aio"),
```
carllerche: aio and lio probably shouldn't be included if the platform doesn't support it.

src/sys/fuchsia/net.rs
ahmedcharles: Removed. What about the other `only_v6` and `set_only_v6` functions in the crate? Should they all be removed?

src/sys/unix/ready.rs
ahmedcharles: It's used below in the Debug impl. Note, it's not public.

src/sys/unix/ready.rs
ahmedcharles: Assuming the Debug impl should print Aio and Lio on supported platforms, then the only question is how to implement it. This should work, since on unsupported platforms, `aio()` and `lio()` return `empty()`, which prints nothing. Do you prefer an alternative implementation?

Thomasdezeeuw:
Fuschia is no longer supported and `UnixReady` is gone as well, so I'm closing this. Thanks @ahmedcharles.

PR #823, opened 2018-02-28
fixed doc formatting for mio::net::UdpSocket::local_addr function, added more examples for mio::net::UdpSocket.

lanedraex:
References #665 

Fixes a doc formatting displacement caused by #815 (and because I forgot to use closing \`\`\`) and adds examples for `try_clone`, `set_ttl` and `ttl`.

### DIFF
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -150,7 +150,6 @@ impl UdpSocket {
     // so simply disable the test on FreeBSD.
     #[cfg_attr(not(target_os = "freebsd"), doc = " ```")]
     #[cfg_attr(target_os = "freebsd", doc = " ```no_run")]
-    /// ```
     /// # use std::error::Error;
     /// #
     /// # fn try_main() -> Result<(), Box<Error>> {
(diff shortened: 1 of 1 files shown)


src/net/udp.rs
```diff
@@ -150,7 +150,6 @@ impl UdpSocket {
     // so simply disable the test on FreeBSD.
     #[cfg_attr(not(target_os = "freebsd"), doc = " ```")]
     #[cfg_attr(target_os = "freebsd", doc = " ```no_run")]
-    /// ```
```
carllerche: I'm a bit confused (I'm not used to doc examples guarded by cfg_attr).

Are the triple ticks not needed here? Is it because of the cfg_attr? Huh... who knew.

lanedraex: I was surprised as well to see this, but if you take a look at the [local_addr](https://carllerche.github.io/mio/mio/net/struct.UdpSocket.html#method.local_addr) function docs, removing the triple ticks fixes the issue.

lanedraex:
Would you prefer to have every function tagged as `no_run` so we can avoid having `cfg_attr`s in the example docs, but then I can only guarantee that the examples would run on my machine (Windows 10).

PR #824, opened 2018-03-02
Update slab dependency from 0.3 to 0.4

briansmith:
Many other projects closely related to mio are already using slab 0.4. use slab 0.4 like they do, so that they don't have to download and build slab 0.3.


briansmith:
I rebased @ignatenkobrain's excellent PR #768 on top of the current master. Then I reverted the breaking API changes.

PR #830, opened 2018-03-29
Fix inconsistency of README vs. source code documentation

ChristophWurst:
While reading the [readme](https://github.com/carllerche/mio) and then the [source code documentation](https://carllerche.github.io/mio/mio/index.html) I discovered this inconsistency. According to the source code timers are deprecated, so I removed it from the goals.

While at it, it seemed that the list of supported platform was also not in sync, so I updated it as well.


dimbleby:
`mio` doesn't provide any of "a notification channel, buffer abstractions, and a slab" either, does it?

So I think you could safely remove that entire line.

PR #831, opened 2018-04-04
Fix spurious failures

carllerche:
Testing multi threading is hard...

### DIFF
--- a/test/test_tcp.rs
+++ b/test/test_tcp.rs
@@ -578,8 +578,8 @@ fn connect_error() {
     let poll = Poll::new().unwrap();
     let mut events = Events::with_capacity(16);
 
-    // Pick a "random" port that shouldn't be in use.
-    let l = TcpStream::connect(&"127.0.0.1:38381".parse().unwrap()).unwrap();
+    let l = TcpStream::connect(&"10.245.111.123:38381".parse().unwrap()).unwrap();
+
     poll.register(&l, Token(0), Ready::writable(), PollOpt::edge()).unwrap();
 
     'outer:
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -185,7 +185,7 @@ impl UdpSocket {
     /// use mio::net::UdpSocket;
     ///
     /// // We must bind it to an open address.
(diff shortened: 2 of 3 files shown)


asomers:
What about the other tests that bind to port 7777 on lines 41, 106, 158, 223, 305, 442, and 454?

carllerche:
Just playing wack-a-mole at this point :P

test/test_tcp.rs
```diff
@@ -578,8 +578,8 @@ fn connect_error() {
     let poll = Poll::new().unwrap();
     let mut events = Events::with_capacity(16);
 
-    // Pick a "random" port that shouldn't be in use.
-    let l = TcpStream::connect(&"127.0.0.1:38381".parse().unwrap()).unwrap();
+    let l = TcpStream::connect(&"10.245.111.123:38381".parse().unwrap()).unwrap();
```
asomers: This could be a valid address in the environment where the tests are run.  If you want an address that will never be in use, you should use an RFC5737 address, reserved for documentation, like 192.0.2.1.
https://tools.ietf.org/html/rfc5737

asomers:
There are also two tests in udp.rs that both bind to port 11100.

test/test_tcp.rs
carllerche: ah, good thought.

test/test_tcp.rs
```diff
@@ -578,7 +578,16 @@ fn connect_error() {
     let poll = Poll::new().unwrap();
     let mut events = Events::with_capacity(16);
 
-    let l = TcpStream::connect(&"10.245.111.123:38381".parse().unwrap()).unwrap();
+    // Pick a "random" port that shouldn't be in use.
+    let l = match TcpStream::connect(&"127.0.0.1:38381".parse().unwrap()) {
```
asomers: There's still no guarantee that this port won't be in use.  How about an RFC5737 address?  Those will never be routable.

carllerche: I'm going to merge this as is, but if you have thoughts on how to improve this test, we can follow up in another PR.

PR #833, opened 2018-04-07
Bitflagsify {PollOpt,Ready}

### LINKED ISSUE #661: Consider using associated constants for the bitflag-like types

tamird:
Uses [bitflags] to provide associated consts for bitflag-like types.

Fixes #661.
Closes #698, which looks trapped in history hell.

I found it useful to separate the manual steps from the automated steps; see commits.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ exclude       = [
 ]
 
 [dependencies]
+bitflags = "1.0"
 lazycell = "0.6.0"
 log      = "0.3.1"
 slab     = "0.3.0"
--- a/benches/bench_poll.rs
+++ b/benches/bench_poll.rs
@@ -23,7 +23,7 @@ fn bench_poll(bench: &mut Bencher) {
         let (r, s) = Registration::new(
             &poll,
             Token(i),
-            Ready::readable(),
+            Ready::READABLE,
             PollOpt::EDGE);
 
         registrations.push(r);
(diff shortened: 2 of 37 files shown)


tamird:
@carllerche this is green except for a (known?) failure on freebsd. Looks like that test was fixed on master but not v0.7.x.

carllerche:
Thanks for doing this work.

I hesitate to use an external crate to generate any API that is exported as a public type. The original issue (#661) only mentions the bitflags crate as an example of associated constants.

tamird:
What's the hesitation? I know the issue didn't suggest this, I thought it'd be a nice reduction in code, though.

carllerche:
The hesitation is not having any control over exactly what fns are generated by the third party crate. Any fn that is generated by the crate becomes part of the mio public API and must be supported. This means that if one ever decides to move away from `bitflags` because it doesn't support a case, then the entire set of fns generated by `bit_flags` must also be ported.

carllerche:
The plan for 0.7 will be to avoid a bitflag style API as possible. As such, this PR is not a direction Mio will take. That said, thanks so much for the work. The contribution is appreciated either way.

Refs: #901

PR #834, opened 2018-04-09
Observe kernel limit of 1024 items for readv/writev calls

sifkarov:
This commit takes into account the kernel maximum of 1024 iov
items passed to a readv/writev call.

`man 2 readv`:

> To deal with the fact that IOV_MAX was so low on early versions of
>        Linux, the glibc wrapper functions for readv() and writev() did some
>        extra work if they detected that the underlying kernel system call
>        failed because this limit was exceeded.

> Starting with glibc version 2.9,
>        the wrapper functions provide this behavior only if the library
>        detects that the system is running a Linux kernel older than version
>        2.6.18 (an arbitrarily selected kernel version).  And since glibc
>        2.20 (which requires a minimum Linux kernel version of 2.6.32), the
>        glibc wrapper functions always just directly invoke the system calls.

### DIFF
--- a/src/sys/fuchsia/net.rs
+++ b/src/sys/fuchsia/net.rs
@@ -135,7 +135,7 @@ impl TcpStream {
     pub fn readv(&self, bufs: &mut [&mut IoVec]) -> io::Result<usize> {
         unsafe {
             let slice = iovec::as_os_slice_mut(bufs);
-            let len = cmp::min(<libc::c_int>::max_value() as usize, slice.len());
+            let len = cmp::min(1024, slice.len());
             let rc = libc::readv(self.io.as_raw_fd(),
                                  slice.as_ptr(),
                                  len as libc::c_int);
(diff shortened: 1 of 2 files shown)


asomers:
This should be `libc::IOV_MAX` instead of a magic number.

carllerche:
@sifkarov  what are your thoughts on @asomers' comment?

sifkarov:
@carllerche I can't remember my original rationale but I agree with @asomers on second thought.

After a quick glance it doesn't look like `libc` exposes the constant, but I'm not familiar with that crate's structure off hand. I'll have to dive a bit deeper and possibly land something there first?

carllerche:
@asomers Do you have thoughts on how to get `MAX_VALUE`?

asomers:
[`POSIX`](http://pubs.opengroup.org/onlinepubs/007904875/basedefs/limits.h.html) says that `IOV_MAX` may be variable, and should be checked at runtime.  The correct way to do that is to use sysconf.  With Nix, you would do it like this:
https://play.rust-lang.org/?gist=e3d4fb15c73b371fced35c29548279d3&version=stable&mode=debug

However, for certain operating systems, `IOV_MAX` is hard-coded.  This is the case for FreeBSD and Linux; I haven't checked any of the others.  For those operating systems, the value is defined in `limits.h`, and should be added to libc.

carllerche:
This PR has gone stale. I'm going to close it. If there is interest in pursuing a change, I would recommend opening an issue to discuss the best way to do this. That said, thanks so much for the work! It is appreciated either way.

PR #840, opened 2018-04-24
Add Evented impls for container types

passcod:
This makes `Arc<Evented>` and `Box<Evented>` both implementations of `Evented`.

I need this in Notify to be able to generically use different `Evented` drivers from a common core. Basically, I'm getting both unix-kernelspace `EventedFd` and userspace `Registration` (and possibly other kinds) over a trait method, so I firstly needed them to be compatible with a `Box<Evented>` trait objects. Also due to implementation details I needed an `Arc<Evented>` so I included it.

Could possibly also add `Rc<Evented>`? I don't have a use for it but while we're at it...

### DIFF
--- a/src/event_imp.rs
+++ b/src/event_imp.rs
@@ -154,6 +154,48 @@ pub trait Evented {
     fn deregister(&self, poll: &Poll) -> io::Result<()>;
 }
 
+impl Evented for Box<Evented> {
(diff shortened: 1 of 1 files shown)


carllerche:
CI failed due to flaky test. Fixed in #831. I restarted CI.

passcod:
Thanks!

PR #844, opened 2018-05-23
Redox OS support

jD91mZM2:
This was a part of my RSoC project of [porting tokio](https://www.redox-os.org/news/rsoc-porting-tokio-to-redox/) and I think it's complete now :)

Tested:
 - Two small custom examples
 - 11/13 tokio examples work (udp servers might not be working but I'm getting `0.0.0.0:0` straight from rust which indicates this is a problem with either my setup or rust itself)
 - 6/7 hyper examples work (`web_api` is trying to connect to itself, which I think is generally broken - you can't even communicate between two netcat instances running on redox)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ libc = "0.2.19"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.3.4"
-miow   = { git = "https://github.com/redox-os/miow", branch = "latest" }
+miow   = { git = "https://github.com/redox-os/miow", rev = "5b10500" }
 
 [dev-dependencies]
 env_logger = { version = "0.4.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,8 +30,6 @@ default = ["with-deprecated"]
 log      = "0.4"
 slab     = "0.4.0"
 net2     = { git = "https://github.com/redox-os/net2-rs.git" }
-
-[target.'cfg(not(target_os = "redox"))'.dependencies]
 iovec    = { git = "https://github.com/redox-os/iovec.git" }
 
 [target.'cfg(target_os = "fuchsia")'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ default = ["with-deprecated"]
 [dependencies]
 iovec    = "0.1.2"
 log      = "0.4"
-net2     = { git = "https://github.com/redox-os/net2-rs.git" }
+net2     = { git = "https://github.com/redox-os/net2-rs.git", rev = "8f537b3" }
 slab     = "0.4.0"
 
 [target.'cfg(target_os = "fuchsia")'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ default = ["with-deprecated"]
 [dependencies]
 iovec    = "0.1.2"
 log      = "0.4"
-net2     = { git = "https://gitlab.redox-os.org/redox-os/net2-rs" }
+net2     = { git = "https://github.com/redox-os/net2-rs" }
 slab     = "0.4.0"
 
 [target.'cfg(target_os = "fuchsia")'.dependencies]
@@ -44,7 +44,7 @@ libc = "0.2.19"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.3.4"
-miow   = { git = "https://gitlab.redox-os.org/redox-os/miow" }
+miow   = "0.3.1"
 
 [dev-dependencies]
 env_logger = { version = "0.4.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ default = ["with-deprecated"]
 [dependencies]
 iovec    = "0.1.2"
 log      = "0.4"
-net2     = { git = "https://github.com/redox-os/net2-rs", rev = "8f537b3" }
+net2     = { git = "https://github.com/redox-os/net2-rs", rev = "0eae919" }
 slab     = "0.4.0"
 
 [target.'cfg(target_os = "fuchsia")'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,8 +43,9 @@ redox_syscall = "0.1"
 libc = "0.2.19"
 
 [target.'cfg(windows)'.dependencies]
-winapi = "0.3.4"
-miow   = "0.3.1"
+winapi = "0.2.6"
+miow   = "0.2.1"
+kernel32-sys = "0.2"
 
 [dev-dependencies]
 env_logger = { version = "0.4.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ default = ["with-deprecated"]
 [dependencies]
 iovec    = "0.1.2"
 log      = "0.4"
-net2     = { git = "https://github.com/redox-os/net2-rs" }
+net2     = "0.2.33"
 slab     = "0.4.0"
 
 [target.'cfg(target_os = "fuchsia")'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ fuchsia-zircon-sys = "0.3.2"
 redox_syscall = "0.1"
 
 [target.'cfg(unix)'.dependencies]
-libc = "0.2.19"
+libc = "0.2.42"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.2.6"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,9 +29,10 @@ default = ["with-deprecated"]
 [dependencies]
 log      = "0.4"
 slab     = "0.4.0"
+net2     = { git = "https://github.com/redox-os/net2-rs.git" }
 
 [target.'cfg(not(target_os = "redox"))'.dependencies]
-iovec    = "0.1.0"
+iovec    = { git = "https://github.com/redox-os/iovec.git" }
 
 [target.'cfg(target_os = "fuchsia")'.dependencies]
 fuchsia-zircon = "0.3.2"
@@ -41,7 +42,7 @@ fuchsia-zircon-sys = "0.3.2"
 redox_syscall = "0.1"
 
 [target.'cfg(unix)'.dependencies]
-libc   = "0.2.42"
+libc   = { git = "https://github.com/redox-os/liblibc.git" }
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.2.6"
@@ -56,7 +57,3 @@ bytes      = "0.3.0"
 [[test]]
 name = "test"
 path = "test/mod.rs"
-
-[replace]
-"libc:0.2.30" = { git = "https://github.com/redox-os/liblibc.git" }
-"net2:0.2.31" = { git = "https://github.com/redox-os/net2-rs.git" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ default = ["with-deprecated"]
 [dependencies]
 iovec    = "0.1.2"
 log      = "0.4"
-net2     = { git = "https://github.com/redox-os/net2-rs" }
+net2     = { git = "https://gitlab.redox-os.org/redox-os/net2-rs" }
 slab     = "0.4.0"
 
 [target.'cfg(target_os = "fuchsia")'.dependencies]
@@ -44,7 +44,7 @@ libc = "0.2.19"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.3.4"
-miow   = { git = "https://github.com/redox-os/miow", rev = "5b10500" }
+miow   = { git = "https://gitlab.redox-os.org/redox-os/miow", rev = "5b10500" }
 
 [dev-dependencies]
 env_logger = { version = "0.4.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ libc = "0.2.19"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.3.4"
-miow   = { git = "https://gitlab.redox-os.org/redox-os/miow", rev = "5b10500" }
+miow   = { git = "https://gitlab.redox-os.org/redox-os/miow" }
 
 [dev-dependencies]
 env_logger = { version = "0.4.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,8 +29,10 @@ default = ["with-deprecated"]
 [dependencies]
 log      = "0.4"
 slab     = "0.4.0"
+
+[target.'cfg(not(target_os = "redox"))'.dependencies]
+iovec    = "0.1.0"
 net2     = "0.2.29"
-iovec    = "0.1.1"
 
 [target.'cfg(target_os = "fuchsia")'.dependencies]
 fuchsia-zircon = "0.3.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ default = ["with-deprecated"]
 [dependencies]
 iovec    = "0.1.2"
 log      = "0.4"
-net2     = { git = "https://github.com/redox-os/net2-rs", rev = "0eae919" }
+net2     = { git = "https://github.com/redox-os/net2-rs" }
 slab     = "0.4.0"
 
 [target.'cfg(target_os = "fuchsia")'.dependencies]
@@ -43,9 +43,8 @@ redox_syscall = "0.1"
 libc = "0.2.19"
 
 [target.'cfg(windows)'.dependencies]
-winapi = "0.2.6"
-miow   = { git = "https://github.com/redox-os/miow" }
-kernel32-sys = "0.2"
+winapi = "0.3.4"
+miow   = { git = "https://github.com/redox-os/miow", branch = "latest" }
 
 [dev-dependencies]
 env_logger = { version = "0.4.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,6 @@ slab     = "0.4.0"
 
 [target.'cfg(not(target_os = "redox"))'.dependencies]
 iovec    = "0.1.0"
-net2     = "0.2.29"
 
 [target.'cfg(target_os = "fuchsia")'.dependencies]
 fuchsia-zircon = "0.3.2"
@@ -57,3 +56,7 @@ bytes      = "0.3.0"
 [[test]]
 name = "test"
 path = "test/mod.rs"
+
+[replace]
+"libc:0.2.30" = { git = "https://github.com/redox-os/liblibc.git" }
+"net2:0.2.31" = { git = "https://github.com/redox-os/net2-rs.git" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,10 +27,10 @@ with-deprecated = []
 default = ["with-deprecated"]
 
 [dependencies]
+iovec    = "0.1.2"
 log      = "0.4"
-slab     = "0.4.0"
 net2     = { git = "https://github.com/redox-os/net2-rs.git" }
-iovec    = { git = "https://github.com/redox-os/iovec.git" }
+slab     = "0.4.0"
 
 [target.'cfg(target_os = "fuchsia")'.dependencies]
 fuchsia-zircon = "0.3.2"
@@ -40,7 +40,7 @@ fuchsia-zircon-sys = "0.3.2"
 redox_syscall = "0.1"
 
 [target.'cfg(unix)'.dependencies]
-libc   = { git = "https://github.com/redox-os/liblibc.git" }
+libc = "0.2.19"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.2.6"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,9 @@ iovec    = "0.1.1"
 fuchsia-zircon = "0.3.2"
 fuchsia-zircon-sys = "0.3.2"
 
+[target.'cfg(target_os = "redox")'.dependencies]
+redox_syscall = "0.1"
+
 [target.'cfg(unix)'.dependencies]
 libc   = "0.2.42"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ default = ["with-deprecated"]
 [dependencies]
 iovec    = "0.1.2"
 log      = "0.4"
-net2     = { git = "https://github.com/redox-os/net2-rs.git", rev = "8f537b3" }
+net2     = { git = "https://github.com/redox-os/net2-rs", rev = "8f537b3" }
 slab     = "0.4.0"
 
 [target.'cfg(target_os = "fuchsia")'.dependencies]
@@ -44,7 +44,7 @@ libc = "0.2.19"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.2.6"
-miow   = "0.2.1"
+miow   = { git = "https://github.com/redox-os/miow" }
 kernel32-sys = "0.2"
 
 [dev-dependencies]
--- a/src/deprecated/redox.rs
+++ b/src/deprecated/redox.rs
@@ -22,14 +22,14 @@ pub struct PipeReader {
 
 impl PipeReader {
     pub fn from_stdout(stdout: process::ChildStdout) -> io::Result<Self> {
-        match sys::set_nonblock(&stdout) {
+        match sys::set_nonblock(stdout.as_raw_fd()) {
             Err(e) => return Err(e),
             _ => {},
         }
         return Ok(PipeReader::from(unsafe { Io::from_raw_fd(stdout.into_raw_fd()) }));
     }
     pub fn from_stderr(stderr: process::ChildStderr) -> io::Result<Self> {
-        match sys::set_nonblock(&stderr) {
+        match sys::set_nonblock(stderr.as_raw_fd()) {
             Err(e) => return Err(e),
             _ => {},
         }
@@ -76,7 +76,7 @@ pub struct PipeWriter {
 
 impl PipeWriter {
     pub fn from_stdin(stdin: process::ChildStdin) -> io::Result<Self> {
-        match sys::set_nonblock(&stdin) {
+        match sys::set_nonblock(stdin.as_raw_fd()) {
             Err(e) => return Err(e),
             _ => {},
         }
--- a/src/deprecated/redox.rs
+++ b/src/deprecated/redox.rs
@@ -1,0 +1,169 @@
+use {io, sys, Evented, Ready, Poll, PollOpt, Token};
+use std::io::{Read, Write};
+use std::process;
+
+pub use sys::Io;
+
+/*
+ *
+ * ===== Pipe =====
+ *
+ */
+
+pub fn pipe() -> io::Result<(PipeReader, PipeWriter)> {
+    let (rd, wr) = try!(sys::pipe());
+    Ok((From::from(rd), From::from(wr)))
+}
+
+#[derive(Debug)]
+pub struct PipeReader {
+    io: Io,
+}
+
+impl PipeReader {
+    pub fn from_stdout(stdout: process::ChildStdout) -> io::Result<Self> {
+        match sys::set_nonblock(&stdout) {
+            Err(e) => return Err(e),
+            _ => {},
+        }
+        return Ok(PipeReader::from(unsafe { Io::from_raw_fd(stdout.into_raw_fd()) }));
+    }
+    pub fn from_stderr(stderr: process::ChildStderr) -> io::Result<Self> {
+        match sys::set_nonblock(&stderr) {
+            Err(e) => return Err(e),
+            _ => {},
+        }
+        return Ok(PipeReader::from(unsafe { Io::from_raw_fd(stderr.into_raw_fd()) }));
+    }
+}
+
+impl Read for PipeReader {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        self.io.read(buf)
+    }
+}
+
+impl<'a> Read for &'a PipeReader {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        (&self.io).read(buf)
+    }
+}
+
+impl Evented for PipeReader {
+    fn register(&self, poll: &Poll, token: Token, interest: Ready, opts: PollOpt) -> io::Result<()> {
+        self.io.register(poll, token, interest, opts)
+    }
+
+    fn reregister(&self, poll: &Poll, token: Token, interest: Ready, opts: PollOpt) -> io::Result<()> {
+        self.io.reregister(poll, token, interest, opts)
+    }
+
+    fn deregister(&self, poll: &Poll) -> io::Result<()> {
+        self.io.deregister(poll)
+    }
+}
+
+impl From<Io> for PipeReader {
+    fn from(io: Io) -> PipeReader {
+        PipeReader { io: io }
+    }
+}
+
+#[derive(Debug)]
+pub struct PipeWriter {
+    io: Io,
+}
+
+impl PipeWriter {
+    pub fn from_stdin(stdin: process::ChildStdin) -> io::Result<Self> {
+        match sys::set_nonblock(&stdin) {
+            Err(e) => return Err(e),
+            _ => {},
+        }
+        return Ok(PipeWriter::from(unsafe { Io::from_raw_fd(stdin.into_raw_fd()) }));
+    }
+}
+
+impl Write for PipeWriter {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        self.io.write(buf)
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        self.io.flush()
+    }
+}
+
+impl<'a> Write for &'a PipeWriter {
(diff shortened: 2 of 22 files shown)


jD91mZM2:
Oof, seems like our net2 patch breaks windows support. Thank you AppVeyor :)

jD91mZM2:
Sometimes I really wish cargo would let you apply a patch to all dependencies

jD91mZM2:
Sorry for the delay - it's finally upstreamed :smile:

jD91mZM2:
Initially I thought the version bumps were necessary to get miow to use the latest version of dependencies and therefore support redox. But Cargo is nice enough that `0.2.5` will also match `0.2.33` so a `cargo update` would suffice. I will separate it from this PR shortly.

Making the platform unstable for a while does not work very well with tokio since people would then indirectly use mio and can't control the features.

I guess I'll need to step up as maintainer for this.

carllerche:
We can continue the details of the process (stable vs. unstable) here: #853

jD91mZM2:
Bump?

src/sys/redox/awakener.rs
```diff
@@ -0,0 +1,74 @@
...
+            Ok(Awakener {
+                reader: rd,
+                writer: wr,
+            })
+        }
+
+        pub fn wakeup(&self) -> io::Result<()> {
+            match (&self.writer).write(&[1]) {
+                Ok(_) => Ok(()),
+                Err(e) => {
```
Thomasdezeeuw: Just an opinion, might not need to change, but I prefer the following code:

```rust
match (&self.writer).write(&[1]) {
    Ok(_) => Ok(()),
    Err(ref err) if err.kind() == io::ErrorKind::WouldBlock => Ok(()),
    Err(err) => Err(err),
}
```

src/sys/redox/awakener.rs
```diff
@@ -0,0 +1,74 @@
...
+            loop {
+                // Consume data until all bytes are purged
+                match (&self.reader).read(&mut buf) {
+                    Ok(i) if i > 0 => {},
+                    _ => return,
+                }
+            }
+        }
+
+        fn reader(&self) -> &redox::Io {
```
Thomasdezeeuw: This can be removed I think (see comment at `Evented`)?

src/sys/redox/awakener.rs
```diff
@@ -0,0 +1,74 @@
...
+            &self.reader
+        }
+    }
+
+    impl Evented for Awakener {
+        fn register(&self, poll: &Poll, token: Token, interest: Ready, opts: PollOpt) -> io::Result<()> {
+            self.reader().register(poll, token, interest, opts)
```
Thomasdezeeuw: Why not use the `reader` field directly, same for other methods.

src/sys/redox/awakener.rs
```diff
@@ -0,0 +1,74 @@
...
+                    if e.kind() == io::ErrorKind::WouldBlock {
+                        Ok(())
+                    } else {
+                        Err(e)
+                    }
+                }
+            }
+        }
+
+        pub fn cleanup(&self) {
```
Thomasdezeeuw: I don't see this being called anywhere. Can this be removed?

src/sys/redox/eventedfd.rs
```diff
@@ -0,0 +1,105 @@
+use {io, poll, Evented, Ready, Poll, PollOpt, Token};
+use std::os::unix::io::RawFd;
+
+/*
+ *
+ * ===== EventedFd =====
+ *
+ */
+
+#[derive(Debug)]
```
Thomasdezeeuw: Move this after the description, it's sort of floating around currently.

src/sys/redox/eventedfd.rs
```diff
@@ -0,0 +1,105 @@
...
+///     fn deregister(&self, poll: &Poll) -> io::Result<()> {
+///         EventedFd(&self.fd).deregister(poll)
+///     }
+/// }
+/// ```
+///
+/// [`Evented`]: ../event/trait.Evented.html
+/// [`Poll`]: ../struct.Poll.html
+/// [`Poll::register`]: ../struct.Poll.html#method.register
+pub struct EventedFd<'a>(pub &'a RawFd);
```
Thomasdezeeuw: Could the type definition of this be shared with the unix module, having only differente `Evented` implementations? This way we don't have to keep the documentation in sync.

src/sys/redox/io.rs
```diff
@@ -0,0 +1,104 @@
...
+
+/// Manages a FD
+#[derive(Debug)]
+pub struct Io {
+    fd: RawFd,
+}
+
+impl FromRawFd for Io {
+    unsafe fn from_raw_fd(fd: RawFd) -> Io {
+        Io { fd: fd }
```
Thomasdezeeuw: Can use `Io { fd }`.

src/sys/redox/io.rs
```diff
@@ -0,0 +1,104 @@
...
+use std::os::unix::io::{IntoRawFd, AsRawFd, FromRawFd, RawFd};
+use syscall::{fcntl, F_SETFL, F_GETFL, O_NONBLOCK};
+
+pub fn set_nonblock(fd: RawFd) -> io::Result<()> {
+    let flags = fcntl(fd, F_GETFL, 0).map_err(super::from_syscall_error)?;
+    fcntl(fd, F_SETFL, flags | O_NONBLOCK).map_err(super::from_syscall_error)
+                                                     .map(|_| ())
+}
+
+
```
Thomasdezeeuw: Double empty line.

src/sys/redox/io.rs
```diff
@@ -0,0 +1,104 @@
...
+impl Read for Io {
+    fn read(&mut self, dst: &mut [u8]) -> io::Result<usize> {
+        <&Io as Read>::read(&mut &*self, dst)
+    }
+}
+
+impl<'a> Read for &'a Io {
+    fn read(&mut self, dst: &mut [u8]) -> io::Result<usize> {
+        use syscall::read;
```
Thomasdezeeuw: Move this to the imports/uses at the top of the file.

src/sys/redox/io.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        Ok(())
+    }
+}
+
+impl<'a> Write for &'a Io {
+    fn write(&mut self, src: &[u8]) -> io::Result<usize> {
+        use syscall::write;
```
Thomasdezeeuw: Move this to the imports/uses at the top of the file.

src/sys/redox/io.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    fn deregister(&self, poll: &Poll) -> io::Result<()> {
+        poll::selector(poll).deregister(self.fd)
+    }
+}
+
+impl Read for Io {
+    fn read(&mut self, dst: &mut [u8]) -> io::Result<usize> {
+        <&Io as Read>::read(&mut &*self, dst)
```
Thomasdezeeuw: You can use `(&self).read(&mut dst)`, I think.

src/sys/redox/io.rs
```diff
@@ -0,0 +1,104 @@
...
+        use syscall::read;
+
+        read(self.as_raw_fd(), dst)
+            .map_err(super::from_syscall_error)
+    }
+}
+
+impl Write for Io {
+    fn write(&mut self, src: &[u8]) -> io::Result<usize> {
+        <&Io as Write>::write(&mut &*self, src)
```
Thomasdezeeuw: You can use `(&self).write(&src)`, I think.

src/sys/redox/io.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        Ok(())
+    }
+}
+
+impl Drop for Io {
+    fn drop(&mut self) {
+        use syscall::close;
```
Thomasdezeeuw: Move this to the imports/uses at the top of the file.

src/sys/redox/io.rs
```diff
@@ -0,0 +1,104 @@
...
+        let _ = close(self.as_raw_fd());
```
Thomasdezeeuw: Maybe add a note about why the error is ignored. Likely can be the same as in the unix module, or otherwise the once from the standard library.

src/sys/redox/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+pub use self::awakener::Awakener;
+pub use self::eventedfd::EventedFd;
+pub use self::io::{Io, set_nonblock};
+pub use self::selector::{Events, Selector};
+pub use self::tcp::{TcpStream, TcpListener};
+pub use self::udp::UdpSocket;
+
+use std::os::unix::io::FromRawFd;
+
+pub fn pipe() -> ::io::Result<(Io, Io)> {
```
Thomasdezeeuw: It's unclear to my which `Result` your using here, I assume it's `std::io::Result`, but starting with `::` I'm not sure.

src/sys/redox/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+    use syscall::{O_NONBLOCK, O_CLOEXEC};
```
Thomasdezeeuw: Move the imports to the top of the file with the rest.

src/sys/redox/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+    use syscall::pipe2;
+
+    let mut fds = [0; 2];
+    try!(pipe2(&mut fds, O_NONBLOCK | O_CLOEXEC).map_err(from_syscall_error));
```
Thomasdezeeuw: `?` is prefered.

src/sys/redox/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+pub use self::tcp::{TcpStream, TcpListener};
+pub use self::udp::UdpSocket;
+
+use std::os::unix::io::FromRawFd;
+
+pub fn pipe() -> ::io::Result<(Io, Io)> {
+    use syscall::{O_NONBLOCK, O_CLOEXEC};
+    use syscall::pipe2;
+
+    let mut fds = [0; 2];
```
Thomasdezeeuw: Maybe use `-1` so the code can't even mess with fd 0. Syscalls on fd `-1` should returns errors (right?).

src/sys/redox/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+    try!(pipe2(&mut fds, O_NONBLOCK | O_CLOEXEC).map_err(from_syscall_error));
+
+    unsafe {
+        Ok((Io::from_raw_fd(fds[0]), Io::from_raw_fd(fds[1])))
+    }
+}
+
+pub fn from_syscall_error(err: ::syscall::Error) -> ::io::Error {
```
Thomasdezeeuw: Again starting with `::` is somewhat unclear to which type this is refering, same for the body of the code.

src/sys/redox/selector.rs
```diff
@@ -0,0 +1,206 @@
...
+/// Each Selector has a globally unique(ish) ID associated with it. This ID
+/// gets tracked by `TcpStream`, `TcpListener`, etc... when they are first
+/// registered with the `Selector`. If a type that is previously associatd with
+/// a `Selector` attempts to register itself with a different `Selector`, the
+/// operation will return with an error. This matches windows behavior.
+static NEXT_ID: AtomicUsize = ATOMIC_USIZE_INIT;
+
+#[derive(Debug)]
+pub struct Selector {
+    id: usize,
```
Thomasdezeeuw: This doesn't seem to be used, can this be removed?

src/sys/redox/selector.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    pub fn id(&self) -> usize {
+        self.id
+    }
+
+    /// Wait for events from the OS
+    pub fn select(&self, evts: &mut Events, awakener: Token, timeout: Option<Duration>) -> io::Result<bool> {
+        let mut timeout_fd = None;
+        if let Some(timeout) = timeout {
+            let mut file = File::open(format!("time:{}", CLOCK_MONOTONIC))?;
```
Thomasdezeeuw: Do you want to allocate a String everytime select is called?

This could be resolved in a future pr.

src/sys/redox/selector.rs
```diff
@@ -0,0 +1,206 @@
...
+
+            // TODO: use try_from below when stable
+            if timeout.as_secs() > ::std::i64::MAX as u64 {
+                panic!("too high duration");
```
Thomasdezeeuw: Maybe mention it's above the timeout to `mio::Poll`, this message is too vague.

src/sys/redox/selector.rs
```diff
@@ -0,0 +1,206 @@
...
+            }
+
+            let mut time = syscall::TimeSpec::default();
+            file.read(&mut time)?;
+
+            //tv_sec += i64::try_from(timeout.as_secs()).expect("too high duration"),
```
Thomasdezeeuw: Remove?

src/sys/redox/selector.rs
```diff
@@ -0,0 +1,206 @@
...
+    sync::{
+        Mutex,
+        atomic::{AtomicUsize, Ordering, ATOMIC_USIZE_INIT}
+    },
+    time::Duration
+};
+use syscall::{self, CLOCK_MONOTONIC, EVENT_READ, EVENT_WRITE};
+use {io, Ready, PollOpt, Token};
+
+const TIMEOUT_TOKEN: Token = Token(::std::usize::MAX - 1);
```
Thomasdezeeuw: Documentated somewhere that this is not a valid token for use by the user of this crate.

src/sys/redox/selector.rs
```diff
@@ -0,0 +1,206 @@
...
+            time.tv_nsec += timeout.subsec_nanos() as i32;
+
+            file.write(&time)?;
+
+            self.inner_register(file.as_raw_fd(), TIMEOUT_TOKEN, EVENT_READ)?;
+            timeout_fd = Some(file);
+        }
+
+        let cnt;
+        unsafe {
```
Thomasdezeeuw: You can let this unsafe block return `cnt`:

```rust
let cnt = unsafe {
    // ...
    cnt
};
```

src/sys/redox/selector.rs
```diff
@@ -0,0 +1,206 @@
...
+            let bytes = (&self.efd).read(slice::from_raw_parts_mut(
+                evts.events.as_mut_ptr() as *mut u8,
+                evts.events.capacity() * mem::size_of::<syscall::Event>()
```
Thomasdezeeuw: I think this is incorrect. You're going beyond the bounds of the slice.

src/sys/redox/selector.rs
```diff
@@ -0,0 +1,206 @@
...
+/// registered with the `Selector`. If a type that is previously associatd with
+/// a `Selector` attempts to register itself with a different `Selector`, the
+/// operation will return with an error. This matches windows behavior.
+static NEXT_ID: AtomicUsize = ATOMIC_USIZE_INIT;
+
+#[derive(Debug)]
+pub struct Selector {
+    id: usize,
+    efd: File,
+    tokens: Mutex<BTreeMap<RawFd, BTreeSet<Token>>>
```
Thomasdezeeuw: Why not use `HashMap` and `HashSet`? Just curious.

src/sys/redox/tcp.rs
```diff
@@ -0,0 +1,285 @@
...
+    pub fn peer_addr(&self) -> io::Result<SocketAddr> {
+        self.inner.peer_addr()
+    }
+
+    pub fn local_addr(&self) -> io::Result<SocketAddr> {
+        self.inner.local_addr()
+    }
+
+    pub fn try_clone(&self) -> io::Result<TcpStream> {
+        self.inner.try_clone().map(|s| {
```
Thomasdezeeuw: `self.inner.try_clone().map(TcpStream::from_stream)`?

src/sys/redox/tcp.rs
```diff
@@ -0,0 +1,285 @@
...
+            total += n;
+            if n < buf.len() {
+                return Ok(total);
+            }
+        }
+        Ok(total)
+    }
+}
+
+impl<'a> Read for &'a TcpStream {
```
Thomasdezeeuw: As with other `Read`/`Write` implementations on references; this is ok to do in Redox?

src/sys/redox/udp.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    pub fn leave_multicast_v6(&self,
+                              multiaddr: &Ipv6Addr,
+                              interface: u32) -> io::Result<()> {
+        self.io.leave_multicast_v6(multiaddr, interface)
+    }
+
+    pub fn set_only_v6(&self, _only_v6: bool) -> io::Result<()> {
+        //self.io.set_only_v6(only_v6)
+        Err(io::Error::new(io::ErrorKind::Other, "Not implemented"))
```
Thomasdezeeuw: Maybe `unimplemented!()`? Since I would like to know this while developing, but I'm not sure about this.

src/deprecated/redox.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+
+impl Read for PipeReader {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        self.io.read(buf)
+    }
+}
+
+impl<'a> Read for &'a PipeReader {
```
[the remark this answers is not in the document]
jD91mZM2: I *think* it does. It has a SchemeMut trait and I'm sure that it locks that in a safe way

src/deprecated/redox.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+#[derive(Debug)]
+pub struct PipeReader {
+    io: Io,
+}
+
+impl PipeReader {
+    pub fn from_stdout(stdout: process::ChildStdout) -> io::Result<Self> {
+        match sys::set_nonblock(stdout.as_raw_fd()) {
```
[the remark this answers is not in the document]
jD91mZM2: Looks like most of this file was a copy paste of the unix code. Perhaps that should be cleaned up as well :stuck_out_tongue:

src/deprecated/redox.rs
```diff
@@ -0,0 +1,169 @@
...
+            Err(e) => return Err(e),
+            _ => {},
+        }
+        return Ok(PipeReader::from(unsafe { Io::from_raw_fd(stdout.into_raw_fd()) }));
```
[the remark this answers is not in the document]
jD91mZM2: Unix does this

src/net/tcp.rs
```diff
@@ -7,12 +7,16 @@
 //!
 /// [portability guidelines]: ../struct.Poll.html#portability
 
+#[cfg(windows)]
```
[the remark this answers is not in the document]
jD91mZM2: No it doesn't? It compiles, and only windows uses SocketAddrV4

src/net/tcp.rs
```diff
@@ -515,6 +528,7 @@ impl TcpListener {
     /// socket is desired then the `net2::TcpBuilder` methods can be used in
     /// combination with the `TcpListener::from_listener` method to transfer
     /// ownership into mio.
+    #[cfg(all(any(unix, windows), not(target_os = "redox")))]
```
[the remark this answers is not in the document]
jD91mZM2: I have no idea what went on in my head when I wrote that. :stuck_out_tongue:

src/sys/redox/awakener.rs
```diff
@@ -0,0 +1,74 @@
+pub use self::pipe::Awakener;
+
+/// Default awakener backed by a pipe
+mod pipe {
```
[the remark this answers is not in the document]
jD91mZM2: This seems to be mostly copy pasted from the unix module

src/sys/redox/awakener.rs
jD91mZM2: `self.readiness_queue.inner.awakener.cleanup();`
src/poll.rs

src/sys/redox/eventedfd.rs
jD91mZM2: You're welcome to suggest where I'd place it

src/sys/redox/selector.rs
jD91mZM2: `selector_id != poll.selector.id()`
src/poll.rs

src/sys/redox/selector.rs
jD91mZM2: Good point, this is static and could be cached. Optimally would be if Rust would just allow `concat!`ing with constants :(

src/sys/redox/selector.rs
jD91mZM2: When it becomes stable I want to use this code instead.

src/sys/redox/selector.rs
jD91mZM2: Is it ever documented that you can't use the awakener, `usize::MAX`?

src/sys/redox/selector.rs
jD91mZM2: ???

I'm taking the full vector, including the uninitialized contents, as a byte slice.
Then depending on how much the operating system read I `set_len` to update the vector's length.

src/sys/redox/selector.rs
jD91mZM2: No idea, I just went with what Jeremy did. I heard someone mention you for some reason couldn't use HashMaps inside the redox kernel, so perhaps he just got used to BTrees...

src/sys/redox/udp.rs
jD91mZM2: The idea is that an application can fall back to other error handling

src/net/tcp.rs
```diff
@@ -538,12 +551,15 @@ impl TcpListener {
         })
     }
 
-    #[deprecated(since = "0.6.13", note = "use from_std instead")]
-    #[cfg(feature = "with-deprecated")]
-    #[doc(hidden)]
-    pub fn from_listener(listener: net::TcpListener, _: &SocketAddr)
```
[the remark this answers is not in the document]
jD91mZM2: I have not got the slightest idea. I'm just gonna assume it was a problem with the rebase ¯\\\_(ツ)_/¯

src/sys/redox/selector.rs
jD91mZM2: Although, fun fact, `File::open` allocates on each occurance on linux. This is because it needs a terminating nul byte. So this isn't that expensive

src/deprecated/redox.rs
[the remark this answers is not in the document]
Thomasdezeeuw: Ok, then this looks good to me.

src/deprecated/redox.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    fn deregister(&self, poll: &Poll) -> io::Result<()> {
+        self.io.deregister(poll)
+    }
+}
+
+impl From<Io> for PipeReader {
+    fn from(io: Io) -> PipeReader {
+        PipeReader { io: io }
```
Thomasdezeeuw: `PipeReader { io }` works here.

src/deprecated/redox.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    fn deregister(&self, poll: &Poll) -> io::Result<()> {
+        self.io.deregister(poll)
+    }
+}
+
+impl From<Io> for PipeWriter {
+    fn from(io: Io) -> PipeWriter {
+        PipeWriter { io: io }
```
Thomasdezeeuw: `PipeWriter { io }`

jD91mZM2:
Could you restart the appveyor test? I really doubt I affected that :stuck_out_tongue:

src/net/tcp.rs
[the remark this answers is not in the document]
Thomasdezeeuw: Your right I was look at `TcpStream::connect`, but it uses `SocketAddr::V4`.

src/sys/redox/selector.rs
Thomasdezeeuw: Ok, nevermind then.

src/sys/redox/selector.rs
Thomasdezeeuw: Didn't about `File::open` allocating. Then lets leave this for a future pr.

src/sys/redox/eventedfd.rs
Thomasdezeeuw: Maybe crate a new file `src/sys/eventedfd.rs` with just the struct definition, then only use it with unix and redox. And implement `FromRawFd` etc. separately in the redox and unix modules.

src/sys/redox/mod.rs
Thomasdezeeuw: Ah, didn't know that.

src/sys/redox/mod.rs
Thomasdezeeuw: Well I would prefer it to be a bit clearer, but that can be left for a future pr then.

src/sys/redox/selector.rs
Thomasdezeeuw: Could you add a note about that? Now it seems like the code is just left in by accident.

src/sys/redox/udp.rs
Thomasdezeeuw: Ok. So I would be correct to assum it's not implemented in Redox?

src/sys/redox/selector.rs
Thomasdezeeuw: Ok, well then it's fine to leave it.

jD91mZM2:
I'd love to add CI support, but the policy for adding new platforms has not yet been completely resolved, and @asomers' suggested changes would imply that redox starts off without a CI until it gets a lower tier.

carllerche:
At the very least, it would be nice if there was some sort of cross compilation test to ensure that it compiled at the very least.

carllerche:
@hawkw could you review this? I would not focus on any of the implementation details. The only thing of import is that no existing code is impacted short of mod declarations & `use` statements that are scoped with a cfg for redox.

src/sys/unix/tcp.rs
```diff
@@ -54,11 +54,7 @@ impl TcpStream {
     }
 
     pub fn try_clone(&self) -> io::Result<TcpStream> {
-        self.inner.try_clone().map(|s| {
-            TcpStream {
-                inner: s,
-            }
-        })
+        self.inner.try_clone().map(Self::from_stream)
```
hawkw: This change seems unrelated from the rest of this branch?

src/net/tcp.rs
```diff
@@ -89,18 +93,26 @@ impl TcpStream {
...
         let sock = match *addr {
             SocketAddr::V4(..) => TcpBuilder::new_v4(),
             SocketAddr::V6(..) => TcpBuilder::new_v6(),
         }?;
         // Required on Windows for a future `connect_overlapped` operation to be
         // executed successfully.
-        if cfg!(windows) {
-            sock.bind(&inaddr_any(addr))?;
-        }
+        #[cfg(windows)]
```
hawkw: The change from `if cfg!(windows) { ... }` to `#[cfg(windows)]` is probably better, but seems unrelated to adding Redox support.

src/net/tcp.rs
```diff
@@ -389,6 +401,7 @@ impl TcpStream {
     }
 }
 
+#[cfg(windows)]
```
hawkw: Again, this seems unrelated to Redox support.

src/net/tcp.rs
```diff
@@ -523,9 +537,8 @@ impl TcpListener {
         }?;
 
         // Set SO_REUSEADDR, but only on Unix (mirrors what libstd does)
-        if cfg!(unix) {
-            sock.reuse_address(true)?;
-        }
+        #[cfg(unix)]
```
hawkw: Change from `if cfg!(...)` to `#[cfg(...)]` attribute seems unrelated.

src/net/tcp.rs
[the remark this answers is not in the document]
hawkw: Looks like an unrelated change?

Cargo.toml
```diff
@@ -26,18 +26,21 @@ with-deprecated = []
...
 [target.'cfg(target_os = "fuchsia")'.dependencies]
 fuchsia-zircon = "0.3.2"
 fuchsia-zircon-sys = "0.3.2"
 
+[target.'cfg(target_os = "redox")'.dependencies]
+redox_syscall = "0.1"
+
 [target.'cfg(unix)'.dependencies]
-libc   = "0.2.42"
+libc = "0.2.42"
```
hawkw: Unrelated formatting change (though, this did need to be fixed, I think).

Cargo.toml
```diff
@@ -26,18 +26,21 @@ with-deprecated = []
 default = ["with-deprecated"]
 
 [dependencies]
+iovec    = "0.1.2"
 lazycell = "1"
 log      = "0.4"
+net2     = "0.2.33"
```
hawkw: Were the dependency version bumps necessary for adding Redox support?

jD91mZM2: Yes, net2 only recently has redox support

jD91mZM2:
@carllerche Didn't want to set up a CI until I got thumbs up, especially with @asomers' proposed platform policy (which would only allow a CI at tier 2). With your permission I shall do so now.

jD91mZM2:
Guessing the current CI issue is unrelated?

src/lib.rs
```diff
@@ -198,12 +201,13 @@ pub use poll::Iter as EventsIter;
 #[doc(hidden)]
 pub use io::deprecated::would_block;
 
-#[cfg(all(unix, not(target_os = "fuchsia")))]
+#[cfg(any(all(unix, not(target_os = "fuchsia")), target_os = "redox"))]
```
carllerche: Is it possible to move this to a separate block? So, split out the `cfg(redox)` and have a separate `pub mod unix`? It would make it easier to manage IMO.

src/net/tcp.rs
```diff
@@ -7,12 +7,16 @@
 //!
 /// [portability guidelines]: ../struct.Poll.html#portability
 
+#[cfg(not(target_os = "redox"))]
```
carllerche: Could you explain on why this file needs changes?

I would prefer to avoid adding platform cfg guards in `src/net/tcp`.

src/net/tcp.rs
```diff
@@ -683,24 +709,24 @@ impl fmt::Debug for TcpListener {
  *
  */
 
-#[cfg(all(unix, not(target_os = "fuchsia")))]
+#[cfg(any(all(unix, not(target_os = "fuchsia")), target_os = "redox"))]
```
carllerche: I'm wondering if these impls should be moved out of this file in order to maintain the goal of "no cfgs" in this file.

src/net/tcp.rs
jD91mZM2: Back when I wrote this, I would argue that "redox doesn't support creating non-blocking sockets before an address is bound". I am extremely certain that I tried it, but... I've seen the source code of redox' libc now, and it seems to disagree.

It's still always better to use native rust where possible (redox' backend doesn't use libc), but I cannot safely say the old code won't wory. I'll try it again, later.

jD91mZM2: Just double checked, also checked the code. There *are* null sockets implemented, but they don't support fcntl. I'll ask Jeremy what he'd like to do here.

jD91mZM2:
The latest changes separate stuff into their respective platforms. This obviously also affects unix/windows, but not more than strictly necessary. I might be able to revert and keep redox consistent this if the redox netstack team gives me thumbs up to implement fcntl for unbound sockets, but then again that depends on libc so this is better anyway.

jD91mZM2:
Whatever I said about fcntl depending on libc was stupid because it does the direct system call and in no way does using that depend on another system call. And as I need unbound sockets for openssl anyway, I'm implementing that right now.

jD91mZM2:
There we go, unbound fcntls, no redox-specific code for connecting. This *does* mean it uses libc for making the connection (net2), so tell me if you want me to resolve this in some other way.

jD91mZM2:
BUMP?

carllerche:
Ok, here is the plan. Lets get this in the 0.7 release of Mio.

I commented with a proposed implementation strategy for adding new platforms [here](https://github.com/tokio-rs/mio/issues/853#issuecomment-490630832). This proposal is based on how this PR needs to touch so many files. I want to avoid this.

So, if my proposed strategy is accepted, the contents of `sys/redox` would be moved into a dedicated crate that mio would conditionally depend on (`repoll`?). Also, we would split out the cfg flags... I'll probably work on some of this restructuring on master.

wdyt?

jD91mZM2:
I appreciate the idea, it sounds pretty great! I'm afraid, however, I might just drop this PR in its entirety as https://github.com/rust-lang/rust/issues/60139 is merging. Redox is moving its focus away from updating the world to rely directly on Redox APIs, and instead making the C API cover everything. So I am sorry if I made you figure out this policy for smaller gain than expected, but I hope you can still use it for the other platforms as it sounds like a great idea in general.

carllerche:
Sorry about all that. The policy is still going to be valuable I think.

However, even w/ redox targettng `unix`, I would guess that some level of redox specific integration is going to be needed. There are linux, bsd, etc... specific code for implementing the selector.

jD91mZM2:
That's true, someone will probably open a small PR to add redox to the linux `cfg`s. We seem to be going the `epoll` route in our C standard library relibc. :smile:

Thomasdezeeuw:
@jD91mZM2 this is likely not the place for it, but I would recommend `kqueue` over `epoll`. The API in my opinion is much easier to use and supports I/O, signals, timers and use space events all with 2 system calls.

jD91mZM2:
@Thomasdezeeuw I will look up `kqueue` and hear with the BDFL if he thinks implementing it instead (or alongside?) of his `epoll` implementation is worth it. Thanks a lot!

PR #846, opened 2018-06-04
libc 0.2.42 changed definition of bind to use socklen_t on aarch64-linux-android

mehcode:
https://github.com/rust-lang-nursery/net2-rs/pull/74 is also needed to get mio compiling for aarch64-linux-android again


asomers:
Don't forget to update the libc dependency in Cargo.toml too.

carllerche:
Thanks for the PR. #850 has been merged and addresses the same issue. I will close this now.

PR #849, opened 2018-06-27
bump libc dependency to 0.2.42

bcko:


carllerche:
Thanks for the PR.

Could you explain the reason to bump the minimum version? The version in Cargo.toml should specify the oldest version that is needed to run Mio.

carllerche:
This change was included as part of #850, so I will close this now.

PR #850, opened 2018-06-29
Fix android-aarch64 build when using libc 0.2.42

fredszaq:
socketlen_t is defined as an u32 in libc 0.2.42, this conditionnal
compilation is not needed anymore and creates a compilation error

(I ran the test on a real aarch64 android using [dinghy](https://github.com/snipsco/dinghy/) and they pass)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,18 +79,15 @@ matrix:
         - rustup target add $TARGET
 
     # Android aarch64-linux-android
-    #
-    # TODO: Re-enable once net2 is fixed.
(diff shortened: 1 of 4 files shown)


fredszaq:
Oops I've just seen #846 that does exactly the same thing... this one reactivates the android build on travis too

fredszaq:
@carllerche would it be possible to merge this and make a release ? This currently forces us to stay on libc `0.2.41` and the number of libs in the ecosystem starting to use `0.2.42` is increasing  wich can make dependency management cumbersome...

#851 and #852 are about the same problem

PR #857, opened 2018-07-04
Bump mio version

jD91mZM2:
Warning: Only compilation is tested, runtime isn't

I first thought these changes were required for Redox support (#844) because I thought we'd need to modify miow. But we don't, since cargo is nice enough to automatically select newer, semver compatible, versions. Since I already did this, I'm sending it here. But really, I don't care if you don't merge it


carllerche:
`winapi` is part of the public API and cannot be bumped to 0.3.x as this would be a breaking change. I'm less sure off the top of my head about `miow`.

jD91mZM2:
miow depends on winapi, so no I guess

PR #865, opened 2018-08-16
Make iOS CI build compile-only

### LINKED ISSUE #863: Fix CI
iOS CI is no longer working. I am not sure why.

If it can be fixed, then it should be fixed. Otherwise, we should consider switching to a "compile only" CI mode.

hawkw:
Closes #863. Closes #864.

Running the tests in the iOS simulator seems to hang forever on Travis.
I'm unsure what's causing this, as everything works fine in the same
version of the iOS simulator on my Mac. However, increasing the timeout
hasn't seemed to help (see my attempts in #864), so it seems likely that
*something* about the CI environment is causing the tests to hang. 

Given that this is holding up a lot of other work, I've made the iOS
build compile-only for the time being, and removed the scripts used for
running the simulator. If we're eventually able to figure out what's
going wrong, they'll still be in the Git history if we need to bring
back the iOS build.


.travis.yml
```diff
@@ -63,7 +63,6 @@ matrix:
       rust: stable
       script:
         - cargo build
```
carllerche: I'm not 100% suree, but I do not think this does the right thing. This will build under linux. The `TARGET` env var was for the ci script. I think you have to explicitly pass an arg to toolchain.

hawkw: Ah, yeah, I think you're right --- fixed in 610e411a579d17425f3f8ec810a28d38343ad2e4.

hawkw:
Somehow this broke the aarch64-linux-android build:
```
adb: error: failed to get feature set: device still connecting
emulator: WARNING: Not saving state: emulator hasn't finished booting.
```

Seems unrelated, I'm guessing this is a transient failure?

Issue #860, opened 2018-07-27
timeout of None doesnt block


aep:
from the docs of Poll::poll

```
A `timeout` of `None` means that     /// `poll` will block until a readiness event has been received.
```

however, None actually makes it ready immediately while a timeout of Some(Duration::new(0,0)) will do the correct thing.

tobz:
What platform are you seeing this on, btw?

aep:
Nothing special. Linux x86_64, glibc.

carllerche:
That seems ... odd.

hawkw:
I wrote [some quick tests](https://github.com/carllerche/mio/blob/15bce04f5b7137fa5e41b008ff48ded859f2be09/test/test_poll.rs#L23-L62) to try and reproduce this, and I'm [not seeing this behaviour](https://travis-ci.org/carllerche/mio/builds/417029695) on any of the CI platforms (modulo a couple cases where the wait time before assertions had to be bumped up to compensate for slow CI boxes). 

@aep, if you have a minimal code sample that reproduces this, that would be very helpful!

carllerche:
We are unable to reproduce. Are you able to provide a repro case illustrating the bug?

aep:
let me try. unfortunately i dont have that code anymore (lots of refactor and force push, FML)

i was using  PollOpt::level()

hawkw:
@aep Okay, thanks for the update --- do let me know if you see any similar issues in the future.

PR #866, opened 2018-08-16
Add tests for poll blocking/unblocking

hawkw:
This branch adds tests in `tests/test_poll.rs` which assert that calls
to `poll` with various timeouts block and unblock the calling thread as
expected.

This is primarily an attempt to reproduce #860, and as I can't reproduce
the behavior therein described on my Mac, I'm opening this PR primarily
to get it to run on more platforms on CI; it doesn't need to be reviewed or
merged. However, these simple tests might be worth having on master 
as well, so I'd be happy to merge this branch if desired.

### DIFF
--- a/test/test_poll.rs
+++ b/test/test_poll.rs
@@ -1,9 +1,7 @@
 use mio::*;
 use std::time::Duration;
-use std::sync::{
-    atomic::{AtomicBool, Ordering},
-    Arc,
-};
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
 use std::thread;
 
 #[test]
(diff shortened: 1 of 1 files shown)


carllerche:
I'd like to hold off on this for now until we can confirm a bug. The main issue being that these kinds of tests are inherently flaky.

Once we have identified a bug, we can figure out how to add a regression test in the most robust way possible.

hawkw:
@carllerche 
> I'd like to hold off on this for now until we can confirm a bug. The main issue being that these kinds of tests are inherently flaky.

Yeah, I agree, I'm going to close this PR for now, since #860 has been closed. I opened the PR primarily just to trigger a CI build for these tests so that they could be run across all targets.

PR #867, opened 2018-08-17
Add EPOLLPRI readiness to UnixReady on supported platforms

### LINKED ISSUE #835: Missing way to specify EPOLLPRI with UnixReady
There doesn't currently appear to be any way to specify EPOLLPRI with UnixReady when registering for an event. This flag is necessary in order to be able to poll for GPIO events for the Linux GPIO files found under /sys/class/gpio

Documentation: https://www.kernel.org/doc/Documentation/gpio/sysfs.txt

hawkw:
Closes #835.

This branch adds `UnixReady::{priority, is_priority}` functions which
set and test for the `EPOLLPRI` flag, respectively, on platforms that
support `epoll`. Additionally, it modifies the `epoll` selector in
`mio::sys::unix::epoll` to support `UnixReady::priority()` as a distinct
interest type, and set the `priority` bit as well as the `readable` bit
on events where `EPOLLPRI` is set.

### DIFF
--- a/src/sys/unix/epoll.rs
+++ b/src/sys/unix/epoll.rs
@@ -145,6 +145,11 @@ fn ioevent_to_epoll(interest: Ready, opts: PollOpt) -> u32 {
         kind |= EPOLLRDHUP;
     }
 
+
+    if UnixReady::from(interest).is_priority() {
+        kind |= EPOLLPRI;
+    }
+
     if opts.is_edge() {
         kind |= EPOLLET;
     }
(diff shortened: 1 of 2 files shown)


hawkw:
@hetmankp, is this sufficient for the use case you described in #835?

hetmankp:
@hawkw , yes, this should be sufficient for my use case of polling Linux sysfs/kernfs attributes as per #835 . This solution still means it's not possible to distinguish EPOLLIN from EPOLLPRI but I'm not personally aware of any use cases where this matters so it's definitely a worthwhile change. Thanks!

hawkw:
@hetmankp 
> This solution still means it's not possible to distinguish EPOLLIN from EPOLLPRI  but I'm not personally aware of any use cases where this matters

Yeah, that's correct --- I thought it was better for this change to be additive only, rather than changing the current behaviour of setting the `readable` bit on EPOLLPRI. 

'm not sure how much existing code actually relies on that, but I wanted to avoid a breaking change if possible, especially because I also couldn't think of any use cases where you wouldn't want to treat EPOLLPRI as `readable`...

PR #869, opened 2018-08-20
Add UdpSocket::{peek, peek_from}

### LINKED ISSUE #749: Expose peek and peek_from
I was comparing mio's API to libstd, and while most of the omissions make sense, I couldn't figure out why `peek` and `peek_from` are absent in mio. Is there are reason I'm missing?

EDIT: One reason: you'd have to drop support for 1.14 <= rust < 1.18

hawkw:
Closes #749.

This branch adds `peek` and `peek_from` methods to `UdpSocket`, similar
to the methods of the same name on `std::net::UdpSocket`, and adds new
tests for these functions.

Because the Windows implementation of `UdpSocket` buffers data in the
`Ready` state, I couldn't get an implementation which called the
`peek_from` method on the socket's `std::net::UdpSocket` to work
correctly. Instead, the Windows `UdpSocket` implements `peek_from` by
providing access to the currently buffered data without draining it.
This means that it never actually calls `recvfrom` with the `MSG_PEEK`
flag. Hopefully, this implementation is acceptable; otherwise, I'd
appreciate some pointers on a better way to do it, as I'm not too
familiar with the Windows platform code.

### DIFF
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -542,6 +542,33 @@ impl UdpSocket {
     pub fn take_error(&self) -> io::Result<Option<io::Error>> {
         self.sys.take_error()
     }
+
+    /// Receives a single datagram on the socket from the remote address to which it is
+    /// connected, without removing the message from input queue. On success, returns
+    /// the number of bytes peeked.
+    ///
+    /// The function must be called with valid byte array `buf` of sufficient size to
+    /// hold the message bytes. If a message is too long to fit in the supplied buffer,
+    /// excess bytes may be discarded.
+    ///
+    /// Successive calls return the same data. This is accomplished by passing
+    /// `MSG_PEEK` as a flag to the underlying `recv` system call.
+    pub fn peek(&self, buf: &mut [u8]) -> io::Result<usize> {
+        self.sys.peek(buf)
+    }
+
+    /// Receives a single datagram message on the socket, without removing it from the
+    /// queue. On success, returns the number of bytes read and the origin.
+    ///
+    /// The function must be called with valid byte array `buf` of sufficient size to
+    /// hold the message bytes. If a message is too long to fit in the supplied buffer,
+    /// excess bytes may be discarded.
+    ///
+    /// Successive calls return the same data. This is accomplished by passing
+    /// `MSG_PEEK` as a flag to the underlying `recvfrom` system call.
+    pub fn peek_from(&self, buf: &mut [u8]) -> io::Result<(usize, SocketAddr)> {
+        self.sys.peek_from(buf)
+    }
 }
 
 impl Evented for UdpSocket {
--- a/src/sys/windows/udp.rs
+++ b/src/sys/windows/udp.rs
@@ -267,6 +267,44 @@ impl UdpSocket {
         self.imp.inner.socket.take_error()
     }
 
+    pub fn peek_from(&self, mut buf: &mut [u8]) -> io::Result<(usize, SocketAddr)> {
+        let mut me = self.inner();
+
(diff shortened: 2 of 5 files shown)


src/sys/windows/udp.rs
```diff
@@ -267,6 +267,44 @@ impl UdpSocket {
...
+                r
+            },
+        }
+    }
+
+    pub fn peek(&self, buf: &mut [u8]) -> io::Result<usize> {
+        self.peek_from(buf).map(|(size,_)| size)
+    }
+
+
```
Thomasdezeeuw: Double empty line.

src/sys/windows/udp.rs
```diff
@@ -267,6 +267,44 @@ impl UdpSocket {
...
+                                        "failed to parse socket address"))
+                };
+                me.read = State::Ready(data);
+                r
+            },
+        }
+    }
+
+    pub fn peek(&self, buf: &mut [u8]) -> io::Result<usize> {
+        self.peek_from(buf).map(|(size,_)| size)
```
Thomasdezeeuw: What happens on Windows if we call `peek` on an unconnected socket? I think on unix this should return `ENOTCONN`, but looking at this it seems to accept packets from anywhere. I don't run Windows so I can't try this myself, please let me know if I'm wrong here.

test/test_udp_socket.rs
```diff
@@ -173,3 +243,31 @@ pub fn test_udp_socket_discard() {
...
+
+#[test]
+pub fn test_udp_socket_peek() {
+    let addr = localhost();
+    let any = localhost();
+
+    let tx = UdpSocket::bind(&any).unwrap();
+    let rx = UdpSocket::bind(&addr).unwrap();
+
+        let tx_addr = tx.local_addr().unwrap();
```
Thomasdezeeuw: Doubl indentation.

carllerche:
@Thomasdezeeuw Do you think we should try to push forward with this now or wait until 0.7? I'm hoping the windows story will be clearer then.

src/net/udp.rs
```diff
@@ -542,6 +542,33 @@ impl UdpSocket {
...
         self.sys.take_error()
     }
+
+    /// Receives a single datagram on the socket from the remote address to which it is
+    /// connected, without removing the message from input queue. On success, returns
+    /// the number of bytes peeked.
+    ///
+    /// The function must be called with valid byte array `buf` of sufficient size to
+    /// hold the message bytes. If a message is too long to fit in the supplied buffer,
+    /// excess bytes may be discarded.
```
Thomasdezeeuw: Maybe discarded isn't the right word as the bytes will still remain in the input queue.

src/net/udp.rs
```diff
@@ -542,6 +542,33 @@ impl UdpSocket {
...
+    pub fn peek(&self, buf: &mut [u8]) -> io::Result<usize> {
+        self.sys.peek(buf)
+    }
+
+    /// Receives a single datagram message on the socket, without removing it from the
+    /// queue. On success, returns the number of bytes read and the origin.
+    ///
+    /// The function must be called with valid byte array `buf` of sufficient size to
+    /// hold the message bytes. If a message is too long to fit in the supplied buffer,
+    /// excess bytes may be discarded.
```
Thomasdezeeuw: Same as above regarding discarded.

Thomasdezeeuw:
@hawkw Is this a change you depended on? If so waiting on the rewrite may take a little too long? Otherwise I would say wait for the rewrite.

carllerche:
Closing due to inactivity.

PR #875, opened 2018-09-05
Reduce spurious awaken calls

### LINKED ISSUE #785: Pipe write occurs when epoll_wait() is not happening. (Linux)
As observed on Linux, when a `set_readiness()` happens, mio will write a 0x01 byte to a pipe, later observe the corresponding event on the pipe's read file descriptor in `epoll_wait()`, read the 0x01 from the pipe, and return the event.

[This Reddit comment from Carl Lerche](https://www.reddit.com/r/rust/comments/7klghl/tokio_internals_understanding_rusts_asynchronous/drhbjvp/) indicates that this is not expected to happen when an `epoll_wait()` is not in progress.  For example, in a single-thread scenario this should not be seen, since a `set_readiness()` could not happen concurrently with an in-progress `epoll_wait()`.  However, this pipe write does appear to happen outside of an `epoll_wait()`, at least in some circumstances.

I've posted [a small example program in this gist](https://gist.github.com/simmons/0e84e9d2d60ab3221db62004468d5fd8) that shows the issue with mio 0.6.11.  It registers a UDP socket and a `mio::Registration` with mio, sends a UDP packet to the socket, receives the packet, then calls `set_readiness()`.  An strace shows the pipe write happening outside of an `epoll_wait()`:

```
$ strace -f -s 1024 target/debug/mio-pipe
...
pipe2([5, 6], O_NONBLOCK|O_CLOEXEC) = 0
epoll_ctl(4, EPOLL_CTL_ADD, 5, {EPOLLIN|EPOLLET, {u32=4294967295, u64=18446744073709551615}}) = 0
epoll_ctl(4, EPOLL_CTL_ADD, 3, {EPOLLIN, {u32=0, u64=0}}) = 0
socket(PF_INET, SOCK_DGRAM|SOCK_CLOEXEC, IPPROTO_IP) = 7
bind(7, {sa_family=AF_INET, sin_port=htons(0), sin_addr=inet_addr("127.0.0.1")}, 16) = 0
sendto(7, "hello", 5, MSG_NOSIGNAL, {sa_family=AF_INET, sin_port=htons(2000), sin_addr=inet_addr("127.0.0.1")}, 16) = 5
epoll_wait(4, [{EPOLLIN, {u32=0, u64=0}}], 16, -1) = 1
recvfrom(3, "hello", 1500, 0, {sa_family=AF_INET, sin_port=htons(41815), sin_addr=inet_addr("127.0.0.1")}, [16]) = 5
write(1, "recv 5 bytes from 127.0.0.1:41815.\n", 35) = 35
write(6, "\1", 1)       = 1
epoll_wait(4, [{EPOLLIN, {u32=4294967295, u64=18446744073709551615}}], 16, 0) = 1
read(5, "\1", 128)      = 1
read(5, 0x7fff70ace168, 128) = -1 EAGAIN (Resource temporarily unavailable)
write(1, "mio::Registration readiness received.\n", 38) = 38
...
```

If this is indeed a bug, it's probably a low priority bug since functionality is not impacted.

carllerche:
When a registration is signalled from a side thread and the main thread
is blocke in `poll`, a pipe is used to interrupt the blocked `poll`
call. However, this pipe write is expensive and should only be made when
`poll` is blocked.

To avoid spurrious calls, the readiness queue tracks a `sleep_token`.
When a node is scheduled, it checks if the `sleep_token` is present. If
it is, it writes to the pipe to wake up the main thread.

Currently, the `sleep_token` isn't cleared when the blocking call to
`poll` wakes up naturally. This results in the next registration signal
to write to the pipe even though the main thread is not blocked.

This patch clears the `sleep_token` when the blocking call returns.

Fixes #785

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -1157,6 +1157,8 @@ impl Poll {
         if timeout == Some(Duration::from_millis(0)) {
             // If blocking is not requested, then there is no need to prepare
             // the queue for sleep
+            //
+            // The sleep_marker should be removed by readiness_queue.poll().
         } else if self.readiness_queue.prepare_for_sleep() {
             // The readiness queue is empty. The call to `prepare_for_sleep`
             // inserts `sleep_marker` into the queue. This signals to any
@@ -2116,6 +2118,13 @@ impl ReadinessQueue {
         // loop where `Poll::poll` will keep dequeuing nodes it enqueues.
         let mut until = ptr::null_mut();
 
+        if dst.len() == dst.capacity() {
+            // If `dst` is already full, the readiness queue won't be drained.
+            // This might result in `sleep_marker` staying in the queue and
+            // unecessary pipe writes occuring.
+            self.inner.clear_sleep_marker();
+        }
+
         'outer:
         while dst.len() < dst.capacity() {
             // Dequeue a node. If the queue is in an inconsistent state, then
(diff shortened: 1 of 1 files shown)


jonhoo:
See also #785

jonhoo:
This does seem to have eliminated all the spurious reads and writes from pipes that we were seeing in [`distributary`](https://github.com/mit-pdos/distributary)!

PR #883, opened 2018-10-30
Improve deprecation message and fix example for Registration

mikong:
Fixes carllerche/mio#817

Some of the suggestions in the issue discussion are:

1. Remove deprecation
1. Improve deprecation message
1. Update example in the documentation of `Registration` to use `deregister` of `Poll`

This PR does 2 and 3. It changes the deprecation message from "use `Evented` impl" to "use `Poll::deregister` instead". It also updates the example in the doc to use:

```
poll.deregister(&self.registration)
```

Please let me know if removing the deprecation is more preferable than improving the deprecation message.

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -424,7 +424,7 @@ pub struct Poll {
 ///     }
 ///
 ///     fn deregister(&self, poll: &Poll) -> io::Result<()> {
-///         self.registration.deregister(poll)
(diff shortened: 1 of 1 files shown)


carllerche:
Thanks so much for this and sorry for the delay. CI has not been a good state. We got CI working on master! would you be able to pull in the latest master?

PR #886, opened 2018-11-08
Fix TcpListener example

povilasb:
In real world we will register TcpListener for readable events. Let's make example reflect that as well.
See https://github.com/carllerche/mio/issues/884

### DIFF
--- a/src/net/tcp.rs
+++ b/src/net/tcp.rs
@@ -482,7 +482,7 @@ impl fmt::Debug for TcpStream {
 /// let mut events = Events::with_capacity(128);
 ///
 /// // Register the socket with `Poll`
-/// poll.register(&listener, Token(0), Ready::writable(),
+/// poll.register(&listener, Token(0), Ready::readable(),
 ///               PollOpt::edge())?;
 ///
 /// poll.poll(&mut events, Some(Duration::from_millis(100)))?;


povilasb:
Build fails only with Rust 1.18.0:
image

Probably `lazycell` crate doesn't support this Rust version anymore:
```
   Compiling lazycell v1.2.0
error: associated constants are experimental (see issue #29646)
   --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/lazycell-1.2.0/src/lib.rs:232:5
    |
232 | /     pub const NONE: Self = Self {
233 | |         inner: UnsafeCell::new(None),
234 | |         state: AtomicUsize::new(NONE),
235 | |     };
    | |______^
```

tobz:
@povilasb Could you bump the minimum Rust version in `.travis.yml` to 1.20? Curious to see if the associated consts bit is the only messed up part.

carllerche:
Unfortunately, the lazy cell dep broke the build...

carllerche:
This is the problem: https://github.com/carllerche/mio/commit/82066e545c3bf4ad9a0a6e270c4688d156b5a582

We need to go down to "1.1" or something like that.

carllerche:
We probably should remove the lazycell dependency...

povilasb:
Ok, I'll give it a try ;)

carllerche:
@povilasb could you submit a new PR that removes the lazycell dependency? Then we can rebase the other PRs.

povilasb:
I just replace `LazyCell` in timer with `inner: RefCell<Option<Inner>>` - all tests are still passing. Will do the rest of the work tomorrow morning - I'm gmt+2 ;)

carllerche:
@povilasb thanks!

A quick search shows we also use the atomic variant. It is also OK to just copy all the code from the lazycell crate into a `lazycell.rs` file in mio, then update it to work on Rust 1.18.

carllerche:
Would you be able to merge in master? That should make CI pass.

povilasb:
@carllerche CI is passing ;)

PR #893, opened 2018-12-09
Prefer pass-by-value over pass-by-reference for public methods

asomers:
pass-by-value is faster than pass-by-reference for types that are small
and Copy.  This is technically a breaking change since these methods are
public.  However, most crates should compile fine because the only
method argument that's changing is `&self`, which is provided
implicitly.  The only way to get into trouble would be something like
this:

```rust
let ready = Ready::new;
Ready::is_hup(&ready)
```

### DIFF
--- a/src/event/event.rs
+++ b/src/event/event.rs
@@ -112,9 +112,9 @@ impl Event {
     /// Create a reference to an `Event` from a platform specific event.
     pub(crate) fn from_sys_event_ref(sys_event: &sys::Event) -> &Event {
         unsafe {
-            // This is safe because only because the memory layout of `Event` is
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw:
Also something I would like for 0.7.

Thomasdezeeuw:
@asomers Can you rebase this on master then I think we get this merged.

Thomasdezeeuw:
@asomers I've created a pr https://github.com/asomers/mio/pull/1 that rebased this pr on the master branch so we can get this merged.

Thomasdezeeuw:
@asomers thanks.

PR #894, opened 2018-12-09
Pass Ipv4Addr arguments by value

asomers:
This is faster than passing them by reference, since they're Copy and
only 4 bytes big.  But it's a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased (Unreleased)
 
+* The `join_multicast_v4` and `leave_multicast_v4` methods now take their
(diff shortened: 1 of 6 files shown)


Thomasdezeeuw:
I think we should do this everywhere for 0.7.

carllerche:
+1

Thomasdezeeuw:
@carllerche you OK with merging this into master? Or do you want to move this into the 0.7 branch?

carllerche:
Lets fork off master to v0.6.x and merge this into master.

The v0.7.x branch stagnated, so the work already done probably needs to be cherry picked.

asomers:
Rebased.

PR #899, opened 2018-12-20
Implement AsRawSocket and FromRawSocket on Windows for TcpListener, T…

sdroege:
…cpStream and UdpSocket

IntoRawSocket can't be easily implemented as we do reference counting
for the underlying socket and as such can't move the socket out of the
struct.

### DIFF
--- a/src/net/tcp.rs
+++ b/src/net/tcp.rs
@@ -735,3 +735,46 @@ impl FromRawFd for TcpListener {
         }
     }
 }
+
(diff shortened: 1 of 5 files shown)


tobz:
Hmm, this looks like a clone of #859, which still requires specific work to deal with Windows.

sdroege:
Indeed, and that one is also nicer. Let's close mine then!

PR #903, opened 2019-01-02
Switch FreeBSD's CI to cirrus-ci.com

### LINKED ISSUE #895: Switch FreeBSD CI to Cirrus
cirrus-ci.com recently added a FreeBSD build option.  It integrates with Github, it's free for FOSS, and seems to work ok.  Switching from buildbot to cirrus-ci.com would increase mio's bus factor and save @asomers $$$.  Do you agree @carllerche ?

asomers:
This change does the following:
1) Adds a CI build on cirrus-ci.com
2) Switches FreeBSD's CI from a jail to a full VM

This change does _not_ disable buildbot.  That must happen outside of
git.

Fixes #895

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,21 @@
+freebsd_instance:
+  image: freebsd-11-2-release-amd64
+
+task:
+  name: FreeBSD 11.2 amd64
+  setup_script:
+    - pkg install -y curl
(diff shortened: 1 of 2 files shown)


asomers:
The Appveyor failure is due to a hang.  But it can't be caused by this PR; I didn't touch any files used on Windows.

Issue #907, opened 2019-01-12
Remove `Registration` and `SetReadiness` APIs


carllerche:
The `Registration` and `SetReadiness` APIs require maintaining a custom scheduler internal to Mio. The ecosystem is moving away from using the Mio scheduler. For example, Tokio provides a number of its own schedulers.

By removing these APIs, Mio on *nix platforms will (again) truely become zero cost on top of the OS APIs.

carllerche:
The windows selector uses `Registration` and `SetReadiness` internally. So, mio would have to maintain the code, but it can keep it private and only when Mio is built for windows.

Thomasdezeeuw:
@carllerche is the plan to keep the existing `Registration` queue implementation for Windows, or do want to use an external crate? For example [crossbeam](https://crates.io/crates/crossbeam) has a number of channels and queues. Not having to maintain a complicated multi-threaded queue would make maintenance easier.

eduardosm:
As I know of, `Registration` / `SetReadiness` is the only way to wake `poll` from another thread. Will there be an alternative?

Thomasdezeeuw:
@eduardosm The current plan (which is a work in progress, for which we're looking for feedback) is to reduce the API service of Mio. Anything that can be implemented outside of Mio, should be implemented outside mio. This includes user space notifications, which is currently implemented as `Registration`/`SetReadiness`.

If your goal is just to wake poll from another thread then your implementation of the `Awakener` in #909 would do the trick, which can be implemented outside the Mio crate.

carllerche:
@eduardosm @Thomasdezeeuw I think there will need to be some public awakener API. That is the only thing needed to impl `Registration` / `SetReadiness` outside of the Mio crate.

carllerche:
Thinking more about it, technically an awakener could be implemented in userland, but I am not sure if that is a good idea or not.

#788 proposes reserving token space for selector implementation. This would allow OS specific selectors to provide additional features as needed. Reserving a token for the awakener was also part of the intent for this.

I'm leaning towards keeping an awakener concept part of core mio because a) it is pretty much always needed to have a way to wake up an event loop b) the implementation it is pretty platform specific.

eduardosm:
> I'm leaning towards keeping an awakener concept part of core mio because a) it is pretty much always needed to have a way to wake up an event loop b) the implementation it is pretty platform specific.

[...]

PR #909, opened 2019-01-13
Used eventfd-backed awakener on linux.

eduardosm:
On Linux, instead of a pipe, we can use an eventfd, which uses less resources (a single fd and does not require buffers).

### DIFF
--- a/src/sys/unix/awakener.rs
+++ b/src/sys/unix/awakener.rs
@@ -1,6 +1,11 @@
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
 pub use self::pipe::Awakener;
 
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub use self::eventfd::Awakener;
+
 /// Default awakener backed by a pipe
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
 mod pipe {
     use sys::unix;
     use {io, Ready, Poll, PollOpt, Token};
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
@eduardosm Thank you for the pr, but we're likely going to remove the `ReadinessQueue` (see #907), which is the only use for the `Awakener` on unix, thus I'm not sure if we should accept this now.

carllerche:
Whether or not `Awakener` is needed is a good question... lets discuss in #907.

src/sys/unix/awakener.rs
```diff
@@ -72,3 +77,66 @@ mod pipe {
...
+        pub fn new() -> io::Result<Awakener> {
+            Ok(Awakener { fd: unix::eventfd()? })
+        }
+
+        pub fn wakeup(&self) -> io::Result<()> {
+            let data: [u8; 8] = unsafe { transmute(1u64) };
+            match (&self.fd).write(&data) {
+                Ok(_) => Ok(()),
+                Err(e) => {
+                    if e.kind() == io::ErrorKind::WouldBlock {
```
Thomasdezeeuw: I don't think hiding this error is a good idea. If this return a `WouldBlock` error it means that wake is called 2^64 times, without it being clean up. This is an user error in my opinion. Furthermore the pol instance won't actually be woken, another reason to return an error to let the user know.

src/sys/unix/awakener.rs
```diff
@@ -72,3 +77,66 @@ mod pipe {
...
+                    } else {
+                        Err(e)
+                    }
+                }
+            }
+        }
+
+        pub fn cleanup(&self) {
+            let mut buf = [0; 8];
+            let _ = (&self.fd).read(&mut buf);
```
Thomasdezeeuw: This can return an `WouldBlock` error if not calls to `wakeup` were made yet, we might want to hide that error.

src/sys/unix/awakener.rs
asomers: `WouldBlock` cannot be called 2^64 times.  Even if it could be called once per clock cycle, it would still take more than a century.

Thomasdezeeuw: @asomers right, so why bother with checking for the error in the first place?

eduardosm: For "completeness". Even though it is unthinkable that the counter can overflow, if it ever did, this is how it should be handled (the same way it is handled a full buffer in the pipe awakener). I just like be aware of corner cases, but feel free to remove the check.

Thomasdezeeuw:
Also see #969.

Thomasdezeeuw:
This is done in #969, thank you @eduardosm.

PR #927, opened 2019-04-24
Set up CI with Azure Pipelines

carllerche:


asomers:
What advantage does Azure pipelines have over Cirrus or Travis?  Not criticizing, just asking.

carllerche:
@asomers Cirrus will stay for `FreeBSD` testing. Compared to travis / appveyor, it is **much** faster. Switching to azure from travis reduced Tokio's CI time from 90 minutes -> 10.

With Azure, OSS projects get 10 parallel builds where each build gets a bigger VM.

carllerche:
Also, Azure has the benefit of not having just had all their senior staff fired (see travis being acquired by a private capital firm).

asomers:
Wow, that's pretty impressive.  Will tokio keep using travis for OSX ?

carllerche:
@asomers Azure supports Linux, OS X, and Windows. So, we will no longer use Travis once this merges.

asomers:
Did you intentionally leave `cargo doc --no-deps` out of the azure build?

carllerche:
It’s still a work in progress :-)

fkorotkov:
Cirrus also supports Linux, Windows and macOS. Have you considered experimenting with it? I think [Dockerfile as a CI environment](https://cirrus-ci.org/guide/docker-builder/#dockerfile-as-a-ci-environment) feature might be interesting for you so you don't need to build Docker containers each time in `run-docker.sh`.

carllerche:
@fkorotkov Azure pipelines provides 10 concurrent builds each on large VMs. Speeding up the CI build is the goal. In my experience, azure pipeline is often able to complete the Tokio build before Cirrus even though azure does ~15x more work (Cirrus only runs FreeBSD).

carllerche:
This should be ready to merge (assuming CI passes).

cc/ @asomers @Thomasdezeeuw

README.md
```diff
@@ -3,16 +3,21 @@
 Mio is a lightweight I/O library for Rust with a focus on adding as little
 overhead as possible over the OS abstractions.
 
-[![crates.io](http://meritbadge.herokuapp.com/mio)](https://crates.io/crates/mio)
-[![Build Status](https://travis-ci.org/carllerche/mio.svg?branch=master)](https://travis-ci.org/carllerche/mio)
-[![Build status](https://ci.appveyor.com/api/projects/status/ok90r1tcgkyndnvw/branch/master?svg=true)](https://ci.appveyor.com/project/carllerche/mio/branch/master)
-[![Build Status](https://api.cirrus-ci.com/github/carllerche/mio.svg)](https://cirrus-ci.com/github/carllerche/mio)
```
asomers: Are you intentionally removing the Cirrus build badge?

carllerche:
This should be ready to merge.

cc/ @asomers @Thomasdezeeuw

ci/azure-cross-compile.yml
```diff
@@ -0,0 +1,50 @@
...
+      parameters:
+        rust_version: stable
+
+    - script: rustup target add $(target)
+      displayName: "Add target"
+
+    - script: cargo check --target $(target)
+      displayName: Check source
+
+    - script: cargo check --tests --target $(target)
```
asomers: This series of commands won't check any examples.  It wouldn't hurt to add a `cargo check --examples`.

fkorotkov:
@carllerche there is a penalty of 1-2 minutes for starting a new FreeBSD VM and container-based tasks are scheduled within seconds. Anyways I just wanted to point you to the [Dockerfile as a CI environment](https://cirrus-ci.org/guide/docker-builder/#dockerfile-as-a-ci-environment) feature in case you didn't see it. It can help you to get faster Android builds by caching Docker images with everything pre-installed and not having `run-docker.sh` to build the images every single run.

carllerche:
@fkorotkov Ah thanks 👍 I'll take a look at that.

README.md
carllerche: No

carllerche:
@asomers Ok, I added a check for benches 👍

carllerche:
To note: This PR removes tests actually running under android. The current process is really slow and we probably do want to investigate an alternate strategy as mentioned above. We still cross compile though.

carllerche:
@asomers you have a strategy to set a reminder? I am really just pinning nightly to get bench support. We can work on a reminder later though :+1:

asomers:
Well, we could add a step to the tests that will fail if the nightly compiler's build date is more than six weeks old?

carllerche:
@asomers probably, but I would punt that for now (got 6 months to do it!)

PR #932, opened 2019-04-28
Split Poll into Poll and Register

### LINKED ISSUE #757: Split `Poll` into `Poll` and `Register`.

carllerche:
This also makes `Poll::poll` require `&mut self` since `Poll` is no
longer expected to be stored in a `Sync` context.

Resolves #757

This is a very large change and I would not look for perfection, just that nothing in the change is terribly broken. This will be but one of many PRs to transition Mio to 0.7 and I am attempting to make it as incremental as possible.

cc @Thomasdezeeuw @stjepang

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -328,12 +328,12 @@ use {sys, Token};
 /// [`SetReadiness`]: struct.SetReadiness.html
 /// [`Poll::poll`]: struct.Poll.html#method.poll
 pub struct Poll {
-    register: Register,
+    registry: Registry,
 }
 
 /// Registers I/O resources.
 #[derive(Clone)]
-pub struct Register {
+pub struct Registry {
     inner: Arc<Inner>,
 }
 
@@ -659,22 +659,22 @@ impl Poll {
             readiness_queue: ReadinessQueue::new()?,
         });
 
-        let register = Register { inner };
+        let registry = Registry { inner };
 
         // Register the notification wakeup FD with the IO poller
-        register.inner.readiness_queue.inner.awakener.register(
-            &register,
+        registry.inner.readiness_queue.inner.awakener.register(
+            &registry,
             AWAKEN,
             Ready::readable(),
             PollOpt::edge(),
         )?;
 
-        Ok(Poll { register })
+        Ok(Poll { registry })
     }
 
-    /// Return a reference to the associated `Register`.
-    pub fn register(&self) -> &Register {
-        &self.register
+    /// Return a reference to the associated `Registry`.
+    pub fn registry(&self) -> &Registry {
+        &self.registry
     }
 
     /// Wait for readiness events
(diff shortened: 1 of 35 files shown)


src/poll.rs
```diff
@@ -652,25 +654,233 @@ impl Poll {
...
+
+impl fmt::Debug for Poll {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.debug_struct("Poll").finish()
+    }
+}
+
+impl fmt::Debug for Register {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.debug_struct("Poll").finish()
```
Thomasdezeeuw: This uses `Poll`, but it's for `Register`.

test/test_custom_evented.rs
```diff
@@ -176,120 +197,6 @@ mod stress {
         }
     }
 
-    #[test]
-    fn multi_threaded_poll() {
```
Thomasdezeeuw: Is this test coming back?

carllerche: No since we are getting rid of multi threaded polling (removal of the locks).

carllerche:
@Thomasdezeeuw I agree with you about the tests and `poll.register().register(...)`. However, I hope to punt that until later.

- Many tests will be deleted (as we delete old functionality).
- Tests need a general overhaul. I think this would be best done after the Mio API changes are done.
- I am not sure exactly sure what the best strategy will be for us. We may just want to bring back `Poll::register`, `reregister`, `deregister` and rename `register()` -> `handle()` or something. I think we can better evaluate that after the changes we know we want are done.

Thoughts?

carllerche:
@seanmonstar @Thomasdezeeuw I don't have a huge opinion re: what `Register` should be called. If we want to change to `Registry`, we can do that.

hawkw:
FWIW, I'm +1 on renaming `Register` to `Registry`

stjepang:
Same here. Even though `Register` can be a noun, it sounds more like an action to me, while `Registry` would be unambiguous.

Thomasdezeeuw:
`Registry` is fine by me.

Issue #788, opened 2017-12-23
Consider reserving token space for internal use by the selectors


carllerche:
This would allow selectors to provide additional features.

Since pointers are (usually) to aligned memory locations, it should be easy enough to at least reserve half the token space for the internal selector.

The API could be that the MSB cannot be set by the end user.

Thomasdezeeuw:
Is half of the token space needed? On 32 bits (if that is supported) that might be too much.

carllerche:
Half is somewhat arbitrary, but should be more than enough even on 32bit platforms. The idea is to be able to use bit masking to quickly differentiate between internal vs. user tokens.

We can probably reduce to 1 quarter by checking two bits. This may even better to allow numerically increasing tokens (0, 1, 2, 3, 4, ...) as well as pointers as tokens. To do this, we would use the LSB as well as some upper bit to differentiate between internal and external.

uazu:
I was considering using a pointer as a mio Token. Searching around for info, I found this issue. Is pointer-as-Token supported long-term? In particular mio would have to guarantee that any likely pointer value generated by Rust would be a valid Token usize value, and also that once an Evented type is dropped, that a future poll would never generate the associated Token again. It would simplify my code if I could do this.

carllerche:
As long as the pointer is aligned, it could be cast into a token.

Regarding re-use, you should read the guarantees documented for `Poll`. The answer there is not straightforward.

uazu:
Thanks. I plan to drop the memory pointed to by the Token at the same time that the Evented instance is dropped. The thing that was concerning me was that if on receiving an Event, I cast the Token back to a pointer, I could get use-after-free if I'm not careful. I can see from the poll docs that I could get duplicate Tokens in the list resulting from a poll, so I need to deal with that. Apart from that, could I get the same Token in the results of a poll call made _after_ dropping the Evented instance (i.e. after the deregister)? I'm concerned that on some platform perhaps some excess events may still be queued internally, and so would come through with the old Token value and give me a use-after-free. That would break the whole plan.

carllerche:
You can 100% get the token back after **dropping** the Evented value. This is an OS thing, not Mio. 

The only way to guarantee you do not see the token again is to explicitly deregister the Evented and establish a happens-before relation to the next call to poll.

As you mentioned, an Evented may have multiple entries in the results from a single call to poll.

uazu:
[...]

PR #935, opened 2019-05-08
Document guarantee that after deregister, token is gone

uazu:
Following discussion on issue #788, document that after an explicit
deregister, token will not be returned again.  (Important if using a
pointer as a Token.)


uazu:
@carllerche I rephrased it to make it fit the context.

PR #937, opened 2019-05-08
Switch to use Rust Edition 2018

carllerche:


src/sys/unix/mod.rs
```diff
@@ -50,7 +50,7 @@ pub use iovec::IoVec;
 
 use std::os::unix::io::FromRawFd;
 
-pub fn pipe() -> ::io::Result<(Io, Io)> {
+pub fn pipe() -> std::io::Result<(Io, Io)> {
```
stjepang: Shouldn't this be `crate::io::Result<(Io, Io)>`? It's the same re-exported type, but I mean just for consistency...

carllerche: I think the `io` mod w/ the re-exports should go away. I am just leaving that for a different PR.

src/sys/unix/tcp.rs
```diff
@@ -1,19 +1,17 @@
+use crate::event::Evented;
+use crate::sys::unix::eventedfd::EventedFd;
+use crate::sys::unix::io::set_nonblock;
+use crate::sys::unix::uio::VecIo;
+use crate::{io, Interests, PollOpt, Registry, Token};
 use std::fmt;
-use std::io::{Read, Write};
-use std::net::{self, SocketAddr};
-use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
-use std::time::Duration;
```
stjepang: unintentional newline?

test/test_tcp.rs
```diff
@@ -8,10 +8,10 @@ use std::time::Duration;
 
 use net2::{self, TcpStreamExt};
 
+use crate::{TryRead, TryWrite};
```
stjepang: The imports here are separated by newline, while everywhere else they're squished together.

PR #939, opened 2019-05-13
Don't report `RDHUP` as `HUP`

carllerche:
Continuation of #933, adding a test and a fix for kqueue.

This should also fix https://github.com/tokio-rs/tokio/issues/449.

### DIFF
--- a/src/sys/unix/epoll.rs
+++ b/src/sys/unix/epoll.rs
@@ -6,7 +6,7 @@ use std::time::Duration;
 use std::{cmp, i32};
 
 use libc::{self, c_int};
-use libc::{EPOLLERR, EPOLLHUP, EPOLLRDHUP, EPOLLONESHOT};
+use libc::{EPOLLERR, EPOLLHUP, EPOLLONESHOT};
 use libc::{EPOLLET, EPOLLOUT, EPOLLIN, EPOLLPRI};
 
 use {io, Ready, PollOpt, Token};
@@ -141,11 +141,6 @@ fn ioevent_to_epoll(interest: Ready, opts: PollOpt) -> u32 {
         kind |= EPOLLOUT;
     }
 
-    if UnixReady::from(interest).is_hup() {
-        kind |= EPOLLRDHUP;
-    }
-
-
     if UnixReady::from(interest).is_priority() {
         kind |= EPOLLPRI;
     }
@@ -228,7 +223,7 @@ impl Events {
                 kind = kind | UnixReady::error();
             }
 
-            if (epoll & EPOLLRDHUP) != 0 || (epoll & EPOLLHUP) != 0 {
+            if (epoll & EPOLLHUP) != 0 {
                 kind = kind | UnixReady::hup();
             }
 
--- a/src/sys/unix/kqueue.rs
+++ b/src/sys/unix/kqueue.rs
@@ -291,6 +291,19 @@ impl Events {
                 event::kind_mut(&mut self.events[idx]).insert(Ready::readable());
             } else if e.filter == libc::EVFILT_WRITE as Filter {
                 event::kind_mut(&mut self.events[idx]).insert(Ready::writable());
+
+
+                // `EV_EOF` set with `EVFILT_WRITE` indicates the connection has been fully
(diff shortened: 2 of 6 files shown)


carllerche:
cc @behrat

test/test_tcp_shutdown.rs
```diff
@@ -0,0 +1,73 @@
...
+    #[cfg(unix)]
+    {
+        ready |= mio::unix::UnixReady::hup();
+    }
+
+    let client = TcpStream::connect(&addr).unwrap();
+    poll.register(&client,
+                  Token(0),
+                  ready,
+                  PollOpt::edge()).unwrap();
```
hawkw: Nit: weird formatting?

src/sys/unix/kqueue.rs
```diff
@@ -306,7 +306,7 @@ impl Events {
                 }
             }
 
-            if e.flags & libc::EV_EOF != 0 {
+            if e.filter == libc::EVFILT_WRITE as Filter && e.flags & libc::EV_EOF != 0 {
```
hawkw: IMHO this conditional is now complex enough that it would be nice to have a comment explaining it?

test/test_tcp_shutdown.rs
carllerche: 0.6 isn't formatted. Formatting will be fixed when merged into master.

hawkw: 👍

src/sys/unix/kqueue.rs
carllerche: Restructured + added a comment.

src/sys/unix/kqueue.rs
```diff
@@ -291,6 +291,19 @@ impl Events {
                 event::kind_mut(&mut self.events[idx]).insert(Ready::readable());
             } else if e.filter == libc::EVFILT_WRITE as Filter {
                 event::kind_mut(&mut self.events[idx]).insert(Ready::writable());
+
```
Thomasdezeeuw: Double empty line.

src/sys/unix/epoll.rs
```diff
@@ -141,11 +141,6 @@ fn ioevent_to_epoll(interest: Ready, opts: PollOpt) -> u32 {
         kind |= EPOLLOUT;
     }
 
-    if UnixReady::from(interest).is_hup() {
```
Thomasdezeeuw: So this means passing HUP readiness is a no-op on epoll platforms right?

carllerche: No, `HUP` is an implicit interest w/ epoll (you always get it). The error here was requesting `RDHUP` which means something different.

carllerche:
@Thomasdezeeuw There already is a warning:

>     /// **Note that only readable and writable readiness is guaranteed to be
>     /// supported on all platforms**. This means that `hup` readiness
>     /// should be treated as a hint. For more details, see [readiness] in the
>     /// poll documentation.

I can make it stronger though.

carllerche:
@Thomasdezeeuw feedback should be addressed.

src/sys/unix/epoll.rs
Thomasdezeeuw: Right `EPOLLHUP` is implicit but it won't be possible to get `EPOLLRDHUP` anymore right?

SerhoLiu:
> on kqueue, only set HUP when EV_EOF is received with the EVFILT_WRITE filter. 

This did not fix https://github.com/tokio-rs/tokio/issues/449. tokio-reactor (https://github.com/tokio-rs/tokio/blob/tokio-reactor-0.1.9/tokio-reactor/src/lib.rs#L428) will notify read task, but read still return`WouldBlock`. In other words, EV_EOF with EVFILT_WRITE does not indicate EVFILT_READ.

carllerche:
@SerhoLiu I believe it does, `EV_EOF with EVFILT_WRITE` will only trigger (for TCP) when the socket is fully disconnected, in which case the read half will be ready as well.

carllerche:
@SerhoLiu  If I am incorrect, could you submit a test case that demonstrates the bug after this PR is applied?

Issue #934, opened 2019-05-02
Make `Option<Interests>` the same size as `Interests`


FXTi:
In mio-st I've made this `NonZeroU8` to make `Option<Interests>` the same size as `Interests`. But that can come later (in a future pr).

_Originally posted by @Thomasdezeeuw in https://github.com/tokio-rs/mio/pull/925_

* Remove `Interests::both`
* Add `Interests::read_write()`

Thomasdezeeuw:
Also also decide on `Interests::both()` vs. `Interests::all()` and whether or not such a method should exist at all.

carllerche:
I'm wondering if we should **not** have `all()` or `both()`. It is unclear to me if `Interests::all()` would pull in operating specific interests or only pull in portable interests.

I'm OK w/ requiring the user to specify both read & write interests explicitly. Alternatively, we could have` `Interests::read_write()`

stjepang:
I agree with @carllerche. Perhaps for completeness we should also have `Interests::none()`.

Thomasdezeeuw:
> I agree with @carllerche. Perhaps for completeness we should also have `Interests::none()`.

The point of `Interests` was to not have empty interests as it doesn't make a lot of sense. I see no point in registering with empty interests, thus I suggested `Interests` as a non-empty `Ready`.

@carllerche I'm also against a `Interests::all/both()` in mio_st it's a private convenience method.

FXTi:
@carllerche @Thomasdezeeuw I agree.

So let's remove `Interests::both` and add `Interests::read_write` for convenience?

carllerche:
@FXTi for now, we can just remove `both` for now. Lets not add a new API yet.

FXTi:
Shall we add `ops::SubAssign` for `Option<Interests>`?
```rust
                if let Some(x) = self.interests.as_mut() {
                    *x -= Interests::readable();
                }
```

Here, if `self.interests` is just `READABLE`, then the result should be `None`, but not `Option<Interests(0)>` .

To do this, may be in this way?

```rust
self.interests -= Some(Interests::readable());
```

carllerche:
@FXTi Ah, you are correct, `sub` is a bad idea here.

FXTi:
@carllerche I thought two branches for just removing `Interests::readable()` are too much to write.

```rust
if self.interests == Some(Interests::readable()) {
    self.interests = None;
} else if let Some(x) = self.interests.as_mut() {
[...]

PR #940, opened 2019-05-14
Make `Option<Interests>` the same size as `Interests`

FXTi:
Issue #934

The problem in test case is still waiting. https://github.com/tokio-rs/mio/issues/934#issuecomment-492096665

### DIFF
--- a/src/event_imp.rs
+++ b/src/event_imp.rs
@@ -1,5 +1,5 @@
 use crate::{Registry, Token};
-use std::num::NonZeroUsize;
+use std::num::NonZeroU8;
 use std::{fmt, io, ops};
 
 /// A value that may be registered with `Registry`
@@ -963,17 +963,6 @@ fn test_debug_ready() {
     assert_eq!("Writable", format!("{:?}", Ready::writable()));
 }
 
-#[cfg(any(
-    target_os = "dragonfly",
-    target_os = "freebsd",
-    target_os = "ios",
-    target_os = "macos"
-))]
-const AIO: usize = 0b01_0000;
-
-#[cfg(any(target_os = "freebsd"))]
-const LIO: usize = 0b10_0000;
-
 /// Interests used in registering.
 ///
 /// Interests are used in registering [`Evented`] handles with [`Poll`],
@@ -987,7 +976,7 @@ const LIO: usize = 0b10_0000;
 /// use std::mem::size_of;
 /// use mio::Interests;
 ///
-/// assert_eq!(size_of::<Option<Interests>>(), size_of::<usize>());
+/// assert_eq!(size_of::<Option<Interests>>(), size_of::<Interests>());
 /// ```
 ///
 /// [`Poll`]: struct.Poll.html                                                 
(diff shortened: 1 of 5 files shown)


FXTi:
@Thomasdezeeuw Could you please review this?

Though, I thought the code newly needed for sub `Interests` would add to the coder's thinking burden.

test/test_echo_server.rs
```diff
@@ -87,7 +87,9 @@ impl EchoConn {
             }
             Err(e) => {
                 debug!("not implemented; client err={:?}", e);
-                if let Some(x) = self.interests.as_mut() {
+                if self.interests == Some(Interests::readable()) {
```
FXTi: Would this be too much for just remove `Interests::readable()`?

carllerche: Are you asking if you can remove the `if` condition here?

carllerche: I don't understand why this bit needs to change.

src/event_imp.rs
```diff
@@ -980,11 +981,21 @@ const LIO: usize = 0b10_0000;
...
+/// assert_eq!(size_of::<Option<Interests>>(), size_of::<usize>());
+/// ```
+///
 /// [`Poll`]: struct.Poll.html                                                 
 /// [`readable`]: #method.readable
 /// [`writable`]: #method.writable
 #[derive(Copy, PartialEq, Eq, Clone, PartialOrd, Ord)]
-pub struct Interests(usize);
+#[repr(transparent)]
+pub struct Interests(NonZeroUsize);
```
Thomasdezeeuw: Might as well use `NonZeroU8`.

src/event_imp.rs
```diff
@@ -1004,7 +1015,7 @@ impl Interests {
     /// [`Poll`]: struct.Poll.html
     #[inline]
     pub fn readable() -> Interests {
-        Interests(READABLE)
+        Interests(NonZeroUsize::new(READABLE).unwrap())
```
Thomasdezeeuw: It think it's ok to use `new_unchecked` here, this would allow it to be a `const fn` in the future

src/event_imp.rs
```diff
@@ -1174,14 +1185,14 @@ impl ops::BitOr for Interests {
 
     #[inline]
     fn bitor(self, other: Self) -> Self {
-        Interests(self.0 | other.0)
+        Interests(NonZeroUsize::new(self.0.get() | other.0.get()).unwrap())
```
Thomasdezeeuw: Again `new_unchecked` can be used.

test/test_echo_server.rs
carllerche: Ah, the answer is in [this](https://github.com/tokio-rs/mio/issues/934#issuecomment-492096665) comment.

src/event_imp.rs
carllerche: Is it worth to add the `unsafe` blocks in this context?

Thomasdezeeuw: I think so as the only possible invalid behaviour is passing 0 to `new_unchecked`, which is easy to check for. I haven't checked this but I think removing the `unwrap` should be good for performance and as I said it would allow the function to be constant.

test/test_echo_server.rs
FXTi: LGTM, just ignore it.

carllerche:
I'm good w/ this whenever @Thomasdezeeuw

Thomasdezeeuw:
@FXTi thanks!

PR #887, opened 2018-11-20
Incorporate lazycell

povilasb:
Incorporate lazycell 1.2.0 into the project
    
    ...rather than getting it from the crates.io. The problem is that latest
    lazycell doesn't work with rust 1.18.0 anymore that mio supports.  

See discussions at https://github.com/carllerche/mio/pull/886

PR #943, opened 2019-05-16
Use published lazycell

lnicola:


Thomasdezeeuw:
Pr #887 removed lazycell because 1.2 required Rust 1.20 (based on the comment in #886), but master is now on 1.34+ (Rust 2018) right? So I think this is fine.

lnicola:
Yes, that was my reasoning.

Thomasdezeeuw:
I'll leave this up to @carllerche to decide whether or not to merge since I haven't reviewed lazycell.

carllerche:
IMO there is a deeper problem w/ the lack of definition around `rustc` backwards compat and crate versions.

If anything, mio should strive to have fewer dependencies rather than more.

carllerche:
i.e. while `lazycell` may work today, there is no guarantee the dependency will maintain the backwards compat that `mio` would like.

lnicola:
We can choose to stick to an older version of `lazycell` if it ends up dragging another version of the compiler. Let's cross that bridge when we get there. It's not a large dependency, but I don't think it makes sense to vendor it like this.

> here is no guarantee the dependency will maintain the backwards compat that mio would like

Does `mio` have a policy wrt. the minimum compiler version?

lnicola:
See https://github.com/rayon-rs/rayon/issues/586 for a similar issue about supported compilers and upgrading dependencies.

lnicola:
Closing due to lack of interest.

Issue #946, opened 2019-05-17
mio 0.6.17 compilation error on apple ios targets


glebpom:
Mio version 0.6.17 fails to compile on `apple-ios-i386` and `armv7s-apple-ios` with the following error:

```
error[E0277]: the trait bound `i32: std::convert::From<u32>` is not satisfied
  --> /Users/jenkins/.cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.6.17/src/sys/unix/kqueue.rs:77:26
   |
77 |                 tv_nsec: libc::c_long::from(to.subsec_nanos()),
   |                          ^^^^^^^^^^^^^^^^^^ the trait `std::convert::From<u32>` is not implemented for `i32`
   |
   = help: the following implementations were found:
             <i32 as std::convert::From<bool>>
             <i32 as std::convert::From<i16>>
             <i32 as std::convert::From<i8>>
             <i32 as std::convert::From<std::num::NonZeroI32>>
           and 2 others
   = note: required by `std::convert::From::from`
```

Thomasdezeeuw:
@glebpom I've created #948 (and #947 for master) to fix this, can you try that change as I don't have any 32bit platforms.

@carllerche Do we have any CI around 32 bit kqueue platforms?

glebpom:
@Thomasdezeeuw your branch works well, thanks

Thomasdezeeuw:
Pr #948 (and #947 for the master branch) is now merged. @carllerche I think this needs a release.

PR #947, opened 2019-05-17
Fix kqueue for platform where C's long is i32

Thomasdezeeuw:
This for the master branch.

Updates #946.

### DIFF
--- a/src/sys/unix/kqueue.rs
+++ b/src/sys/unix/kqueue.rs
@@ -74,7 +74,11 @@ impl Selector {
     ) -> io::Result<bool> {
         let timeout = timeout.map(|to| libc::timespec {
             tv_sec: cmp::min(to.as_secs(), time_t::max_value() as u64) as time_t,
-            tv_nsec: libc::c_long::from(to.subsec_nanos()),
+            // `Duration::subsec_nanos` is guaranteed to be less than one
+            // billion (the number of nanoseconds in a second), making the
+            // cast to i32 safe. The cast itself is needed for platforms
+            // where C's long is only 32 bits.
+            tv_nsec: libc::c_long::from(to.subsec_nanos() as i32),
         });
         let timeout = timeout
             .as_ref()


src/sys/unix/kqueue.rs
```diff
@@ -74,7 +74,9 @@ impl Selector {
     ) -> io::Result<bool> {
         let timeout = timeout.map(|to| libc::timespec {
             tv_sec: cmp::min(to.as_secs(), time_t::max_value() as u64) as time_t,
-            tv_nsec: libc::c_long::from(to.subsec_nanos()),
+            // On some platforms C's long is i32, for those we need to ensure
+            // `subsec_nanos` doesn't overflow.
+            tv_nsec: cmp::min(to.subsec_nanos(), libc::c_long::max_value() as u32) as libc::c_long,
```
asomers: `libc::c_long::max_value() as u32` will overflow on many platforms.  How about `i32::max_value() as u32` instead?

Thomasdezeeuw: I was wondering this myself as well (https://github.com/tokio-rs/mio/pull/948#discussion_r285062700), I think `i32::max_values` would do the trick.

Thomasdezeeuw:
@asomers Fixed the overflow.

Thomasdezeeuw:
@asomers I forgot it can only be less than one billion, I've changed the code to your suggestion.

src/sys/unix/kqueue.rs
```diff
@@ -74,7 +74,11 @@ impl Selector {
     ) -> io::Result<bool> {
         let timeout = timeout.map(|to| libc::timespec {
             tv_sec: cmp::min(to.as_secs(), time_t::max_value() as u64) as time_t,
-            tv_nsec: libc::c_long::from(to.subsec_nanos()),
+            // `Duration::subsec_nanos` is guaranteed to be less than one
+            // billion (the number of nanoseconds in a second), making the
+            // cast to i32 safe. The cast itself is needed for platforms
+            // where C's long is only 32 bits.
```
asomers: The comment isn't quite right.  The cast is only needed for platforms where C's long is `i32`.  No cast is needed where long is `u32`.

Thomasdezeeuw: Isn't C `long` always signed? According to https://en.wikipedia.org/wiki/C_data_types: "Long signed integer type. Capable of containing at least the [−2,147,483,647, +2,147,483,647] range;[3][4] thus, it is at least 32 bits in size."

asomers: Yes it is always signed, except when I'm being stupid.

PR #950, opened 2019-05-20
Update links

atouchet:


Thomasdezeeuw:
Do we also want this in the v0.6 branch, if we're releasing another version for #946?

PR #952, opened 2019-05-23
Add more CI targets for cross compilation

carllerche:
This should hopefully catch most errors related to secondary platforms.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -13,9 +13,11 @@ task:
     - . $HOME/.cargo/env
     - cargo build
     - cargo build --no-default-features
-  test_script:
+  amd64_test_script:
     - . $HOME/.cargo/env
     - cargo test
-    - cargo test --no-default-features
+  i386_test_script:
+    - . $HOME/.cargo/env
+    - cargo test --target i686-unknown-freebsd
   before_cache_script:
     - rm -rf $HOME/.cargo/registry/index
--- a/ci/azure-cross-compile.yml
+++ b/ci/azure-cross-compile.yml
@@ -6,9 +6,17 @@ jobs:
   displayName: Cross
   strategy:
     matrix:
-      iOS:
+      iOS_64:
         vmImage: macOS-10.13
         target: x86_64-apple-ios
+
+      iOS_32:
+        vmImage: macOS-10.13
+        target: i386-apple-ios
+
+      iOS_ARM:
+        vmImage: macOS-10.13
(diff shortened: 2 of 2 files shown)


ci/azure-cross-compile.yml
```diff
@@ -26,6 +34,18 @@ jobs:
         vmImage: ubuntu-16.04
         target: x86_64-sun-solaris
 
+      FreeBSD_32:
+        vmImage: ubuntu-16.04
```
asomers: Why cross-compile FreeBSD i386 from Linux?  Better to do it from FreeBSD amd64.  That way we can run the tests, too.  See https://github.com/nix-rust/nix/blob/master/.cirrus.yml .

carllerche: @asomers either way, this cross compilation doesn't seem to catch the error?

asomers: What error are you expecting to catch?

carllerche: This fix: https://github.com/tokio-rs/mio/commit/aeaebb0a908f65910ef5cca2863e71734d77d0aa

carllerche:
Ok, I guess running the 32bit freebsd tests is good enough for now... I will rebase against master.

ci/azure-cross-compile.yml
```diff
@@ -26,6 +34,14 @@ jobs:
         vmImage: ubuntu-16.04
         target: x86_64-sun-solaris
 
+      Linux_ARM:
```
Thomasdezeeuw: Is the name accurate when using android? I don't what, if any, differences there are still.

.cirrus.yml
```diff
@@ -13,9 +13,11 @@ task:
...
     - cargo build
     - cargo build --no-default-features
-  test_script:
+  amd64_test_script:
     - . $HOME/.cargo/env
     - cargo test
-    - cargo test --no-default-features
+  i386_test_script:
+    - . $HOME/.cargo/env
+    - cargo test --target i686-unknown-freebsd
```
Thomasdezeeuw: CI fails, you might need to add `rustup target add i686-unknown-freebsd`.

ci/azure-cross-compile.yml
carllerche: i did not put much thought into the name...

Thomasdezeeuw:
Done in #1057.

PR #956, opened 2019-05-25
Upgrade winapi to 0.3.7 and miow to 0.3.3

aloucks:
winapi `0.2` is still being pulled in by `iovec`, but that would be fixed with https://github.com/carllerche/iovec/pull/24

### DIFF
--- a/test/test_udp_socket.rs
+++ b/test/test_udp_socket.rs
@@ -1,6 +1,5 @@
 use crate::localhost;
 use bytes::{Buf, MutBuf, RingBuf, SliceBuf};
-use iovec::IoVec;
 use mio::net::UdpSocket;
 use mio::{Events, Interests, Poll, PollOpt, Token};
 use std::io::ErrorKind;
--- a/test/test_udp_socket.rs
+++ b/test/test_udp_socket.rs
@@ -187,6 +187,8 @@ pub fn test_udp_socket_discard() {
 #[cfg(unix)]
 #[test]
 pub fn test_udp_socket_send_recv_bufs() {
+    use iovec::IoVec;
+
     let (tx, rx) = connected_sockets();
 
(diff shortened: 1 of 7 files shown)


test/test_udp_socket.rs
```diff
@@ -1,6 +1,5 @@
 use crate::localhost;
 use bytes::{Buf, MutBuf, RingBuf, SliceBuf};
-use iovec::IoVec;
```
Thomasdezeeuw: This is needed, it fails various build see the Azure pipeline.

Thomasdezeeuw:
Not sure if we want this in 0.6.x as well, or only in master (for 0.7). Also see #956.

Also who can review the Windows side of things?

Keruspe:
From what I got, mio currently exposes some winapi in its public interface, si this will have to wait for 0.7.
Given 0.7 has no clear schedule though, maybe the right call here would be to just s/0.7/0.8/ and do a 0.7 release changing just the winapi stuff?

Thomasdezeeuw:
Then I'm going to let @carllerche decide, I don't have a stake in Windows related API.

test/test_udp_socket.rs
aloucks: Ah, that test had `#[cfg(unix)]` which explains why it was showing up as an unused import. I added the `use` in the test function directly to prevent the unused import warning.

carllerche:
Thanks for the PR. Looks like a good first step.

This is indeed a breaking change, and as such will be bundled in with the 0.7 release.

It does not make sense to just release this as 0.7 and have another imminent 0.8 release as mio's types are heavily used as part of the public APIs throughout the ecosystem. Churning breaking releases will cause fractures.

One option would be for someone to release these changes using an alt name (`mio-preview` or whatever), but I will not personally do it as I don't think it is a critical priority.

PR #958, opened 2019-05-26
Remove hup kqueue

carllerche:


src/sys/unix/kqueue.rs
```diff
@@ -295,18 +295,6 @@ impl Events {
...
-
-                // `EV_EOF` set with `EVFILT_WRITE` indicates the connection has been fully
-                // disconnected (read and write), but only sockets...
-                if e.flags & libc::EV_EOF != 0 {
-                    event::kind_mut(&mut self.events[idx]).insert(UnixReady::hup());
-
-                    // When the read end of the socket is closed, EV_EOF is set on
-                    // flags, and fflags contains the error if there is one.
-                    if e.fflags != 0 {
-                        event::kind_mut(&mut self.events[idx]).insert(UnixReady::error());
```
Thomasdezeeuw: Won't we be missing out on error events here?

carllerche: Yes... but providing those have never been guaranteed. I think we need to take a step back and look at each event that is not "readable" and "writable" and define what they mean exactly. For example, if you get a `Ready::error`, does that mean a read will fail or maybe there is still some data that can be read...

carllerche: i.e. there is no value in additional events besides `readable` and `writable` unless the guarantees are clearly defined. Currently, there are no guarantees (as stated in the docs), so it seems best to remove it.

PR #963, opened 2019-06-02
Split Mio

Thomasdezeeuw:
This is a work in progress that will split Mio into multiple crates. There will be a number of platform specific crates, e.g. mio-epoll, mio-kqueue, that will wrap the platform specific selector which the main mio crate will use to provide a small, but cross-platform API to.

TODO:
- [x] Create a common crate with types such as `Interests`, `Event`, etc.
- [ ] Design the "backend" trait (the methods that the platform specific selector need to implement).
- [ ] Determine the required types beyond the `Selector`, e.g. `Awakener`.
- [ ] Moving of the code to separate crates:
  - [ ] Linux, Android (mio-epoll),
  - [ ] macOS, iOS, FreeBSD, OpenBSD, NetBSD (mio-kqueue),
  - [ ] Windows (mio-iocp).
  - [ ] Redox?
  - [ ] Fuschia?

/cc @carllerche @stjepang @asomers for some initial feedback.


carllerche:
I believe we opted to have each platform provide its own "core" types and have Mio's main API wrap them. If so, this PR is no longer applicable?

Thomasdezeeuw:
Mostly this doesn't apply, I do think we still have to more `Evented` and `Interests`. But I'm closing this for now.

Issue #968, opened 2019-06-07
Awakener design


Thomasdezeeuw:
This the issue to discuss the design of the `Awakener`. The `Awakener` allows for cross-thread waking of `Poller`.

carllerche:
My current thinking is to structure it so it can implement `Park` / `Unpark` from Tokio (which will be extracted into a separate crate as it is generally useful).

https://github.com/tokio-rs/tokio/blob/8d0f102de83f8aa59661b9d1839ecc57e50873a7/tokio-executor/src/park.rs#L58-L128

So, `Poll` would implement park, and it would be able to return an "unparker" that is `Sync + Send + 'static`. We can tweak the specifics. Maybe we want to include Clone on there...

Thomasdezeeuw:
For Gaea (previously mio-st) I use the following design ([docs.rs](https://docs.rs/gaea/0.3.0/gaea/os/struct.Awakener.html)):

```rust
struct Awakener { /* ... */ }

impl Awakener {
    /// Create a new `Awakener`.
    /// 
    /// `Poller` will get events with the provided `id` when `wake` is called.
    pub fn new(os_queue: &mut OsQueue, id: event::Id) -> io::Result<Awakener> {
        // ...
    }

    /// Attempts to clone the `Awakener`.
    pub fn try_clone(&self) -> io::Result<Awakener> {
        // ...
    }

    /// Wake up the [`Poller`] associated with this `Awakener`.
    pub fn wake(&self) -> io::Result<()> {
        // ...
}
```

Thomasdezeeuw:
After discussing with @carllerche, we've decided to use a design like mentioned in https://github.com/tokio-rs/mio/issues/968#issuecomment-499987787, but without the `try_clone` method.

Thomasdezeeuw:
Done in #969.

PR #969, opened 2019-06-07
Add Awakener

Thomasdezeeuw:
Awakener allows cross-thread waking of Poll.

This is aport of Gaea's Awakener.

Updates #968.

### DIFF
--- a/src/awakener.rs
+++ b/src/awakener.rs
@@ -4,8 +4,10 @@ use crate::{poll, sys, Registry, Token};
 
 /// Awakener allows cross-thread waking of [`Poll`].
 ///
-/// When created it will cause events with [`Ready::readable()`] and the provided
+/// When created it will cause events with [`Ready::READABLE`] and the provided
 /// `token` if [`wake`] is called, possibly from another thread.
+///
+/// [`Poll`]: crate::Poll
 ///
 /// # Notes
 ///
@@ -17,7 +19,7 @@ use crate::{poll, sys, Registry, Token};
 /// What happens if multiple `Awakener`s are registered with the same `Poll` is
 /// undefined.
 ///
-/// [`Ready::readable()`]: crate::Ready::readable
+/// [`Ready::READABLE`]: crate::Ready::READABLE
 /// [`wake`]: Awakener::wake
 ///
 /// # Implementation notes
@@ -86,6 +88,8 @@ impl Awakener {
     }
 
     /// Wake up the [`Poll`] associated with this `Awakener`.
+    ///
+    /// [`Poll`]: crate::Poll
     pub fn wake(&self) -> io::Result<()> {
         self.inner.wake()
     }
--- a/src/awakener.rs
+++ b/src/awakener.rs
@@ -69,6 +69,7 @@ use crate::{poll, sys, Registry, Token};
 ///
 /// // After about 500 milliseconds we should we awoken by the other thread we
 /// // started, getting a single event.
+/// assert!(!events.is_empty());
 /// for event in &events {
 ///     assert_eq!(event, Event::new(Ready::READABLE, WAKE_TOKEN));
 /// }
--- a/src/awakener.rs
+++ b/src/awakener.rs
@@ -1,0 +1,94 @@
+use std::io;
+
+use crate::{poll, sys, Registry, Token};
+
+/// Awakener allows cross-thread waking of [`Poll`].
+///
+/// When created it will cause events with [`Ready::readable()`] and the provided
+/// `token` if [`wake`] is called, possibly from another thread.
+///
+/// # Notes
+///
+/// The `Awakener` needs to be kept alive as long as wake up notifications are
+/// required. This is due to an implementation detail where if all copies of the
+/// `Awakener` are dropped it will also drop all wake up notifications from the
+/// system queue, including wake up notifications that have been added before
+/// the `Awakener` that was dropped, resulting the [`Poll`] not being woken up.
+///
+/// Only a single `Awakener` should active per [`Poll`], if multiple threads
+/// need access to the `Awakener` it can be shared via for example an `Arc`.
+/// What happens if multiple `Awakener`s are registered with the same `Poll` is
+/// undefined.
+///
+/// [`Ready::readable()`]: crate::Ready::readable
+/// [`wake`]: Awakener::wake
+///
+/// # Implementation notes
+///
+/// On platforms that support kqueue this will use the `EVFILT_USER` event
+/// filter, see [implementation notes of `Poll`] to see what platform supports
+/// kqueue. On Linux it uses [eventfd].
+///
+/// [implementation notes of `Poll`]: ../index.html#implementation-notes
+/// [eventfd]: http://man7.org/linux/man-pages/man2/eventfd.2.html
+///
+/// # Examples
+///
+/// Wake an [`Poll`] from another thread.
+///
+/// ```
+/// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+/// use std::io;
+/// use std::thread;
+/// use std::time::Duration;
+/// use std::sync::Arc;
+///
+/// use mio::event::Event;
+/// use mio::{Events, Ready, Token, Poll, Awakener};
+///
+/// const WAKE_TOKEN: Token = Token(10);
+///
+/// let mut poll = Poll::new()?;
+/// let mut events = Events::with_capacity(2);
+///
+/// let awakener = Arc::new(Awakener::new(poll.registry(), WAKE_TOKEN)?);
+///
+/// // We need to keep the Awakener alive, so we'll create a clone for the
+/// // thread we create below.
+/// let awakener1 = awakener.clone();
+/// let handle = thread::spawn(move || {
+///     // Working hard, or hardly working?
+///     thread::sleep(Duration::from_millis(500));
+///
+///     // Now we'll wake the queue on the other thread.
+///     awakener1.wake().expect("unable to wake");
+/// });
+///
+/// // On our current thread we'll poll for events, without a timeout.
+/// poll.poll(&mut events, None)?;
+///
+/// // After about 500 milliseconds we should we awoken by the other thread we
+/// // started, getting a single event.
+/// for event in &events {
+///     assert_eq!(event, Event::new(Ready::READABLE, WAKE_TOKEN));
+/// }
+/// # handle.join().unwrap();
+/// #     Ok(())
+/// # }
+/// ```
+#[derive(Debug)]
+pub struct Awakener {
+    inner: sys::Awakener,
+}
+
+impl Awakener {
+    /// Create a new `Awakener`.
+    pub fn new(registry: &Registry, token: Token) -> io::Result<Awakener> {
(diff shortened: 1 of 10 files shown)


Thomasdezeeuw:
@asomers Can you check kqueue's side of this?

Thomasdezeeuw:
I forgot to run Rustfmt, will do after some initial feedback.

Thomasdezeeuw:
Builds are failing because of fmt and various warnings not related to this change. I'll fix that in another pr I'm currently working on.

Thomasdezeeuw:
Solaris is Linux like right? Does that mean it has eventfd?

Thomasdezeeuw:
This also need the suppression of warnings in #970 for the tests to pass.

Thomasdezeeuw:
I've rebased this on #970, so hopefully the CI will be green.

Thomasdezeeuw:
I've used a unix pipe for Solaris. Now all that is left is Windows. It seems to get the wrong token: https://dev.azure.com/tokio-rs/Tokio/_build/results?buildId=867&view=logs&jobId=429a383f-e486-57ac-3982-9ac65641a32a.

src/awakener.rs
```diff
@@ -0,0 +1,94 @@
...
+///     // Now we'll wake the queue on the other thread.
+///     awakener1.wake().expect("unable to wake");
+/// });
+///
+/// // On our current thread we'll poll for events, without a timeout.
+/// poll.poll(&mut events, None)?;
+///
+/// // After about 500 milliseconds we should we awoken by the other thread we
+/// // started, getting a single event.
+/// for event in &events {
```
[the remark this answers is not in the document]
Thomasdezeeuw: You're right, but `Events` doesn't provide a length method. I'll add `assert!(!events.is_empty())`.

src/sys/unix/dlsym.rs
```diff
@@ -16,13 +16,15 @@ macro_rules! dlsym {
     )
 }
 
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
Thomasdezeeuw: On macOS this gave errors, but it's really unrelated to adding the Awakener.

test/test_awakener.rs
```diff
@@ -0,0 +1,101 @@
...
+    // Waking on the same thread.
+    awakener.wake().expect("unable to wake");
+    expect_events(
+        &mut poll,
+        &mut events,
+        1,
+        vec![Event::new(Ready::readable(), token)],
+    );
+
+    // Multiple wakes between polls.
```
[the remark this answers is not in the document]
Thomasdezeeuw: Will do.

Thomasdezeeuw:
> It strikes me that in the spirit of slimming mio down, Awakener could be moved to a separate crate. I'm not going to insist upon it, though.

@asomers You're right but it's fairly common to require this behaviour. And it does need add to the Selector internal on kqueue. So even if this would go into a separate crate it would be rather tightly coupled.

I'll address the comments and rebase on master.

src/sys/unix/dlsym.rs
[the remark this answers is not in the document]
asomers: It would be better to remove the symbol on platforms where its unused.  That's preferable to `#[allow(dead_code)]` IMHO.

asomers:
> > It strikes me that in the spirit of slimming mio down, Awakener could be moved to a separate crate. I'm not going to insist upon it, though.
> 
> @asomers You're right but it's fairly common to require this behaviour. And it does need add to the Selector internal on kqueue. So even if this would go into a separate crate it would be rather tightly coupled.
> 
> I'll address the comments and rebase on master.

Makes sense.

Thomasdezeeuw:
I've force pushed the changes, the last commit addresses @asomers comments.

src/sys/unix/dlsym.rs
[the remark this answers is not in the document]
Thomasdezeeuw: This change is also in https://github.com/tokio-rs/mio/commit/434405a66c66c1e2b15a07388dd0ed4edcc7e3c8, which is already committed to master. But I do agree it's not a great solution.

carllerche:
@asomers What to include in `mio` proper is a bit of a balance. Should TCP / UDP be included? It is common enough it seems like it. What about UDS? It is part of stdlib, i think we should consider it.

carllerche:
"Slimming down" is equally about maintenance as well as code in the hot path.

Thomasdezeeuw:
@asomers Sorry about the force push.

Windows doesn't work (see https://dev.azure.com/tokio-rs/Tokio/_build/results?buildId=894&view=logs&jobId=429a383f-e486-57ac-3982-9ac65641a32a), it returns the incorrect token. It seems to return the max value of usize.

Thomasdezeeuw:
Who has a Windows machine that can help out with getting the Windows side working?

src/awakener.rs
```diff
@@ -0,0 +1,95 @@
...
+use crate::{poll, sys, Registry, Token};
+
+/// Awakener allows cross-thread waking of [`Poll`].
+///
+/// When created it will cause events with [`Ready::readable()`] and the provided
+/// `token` if [`wake`] is called, possibly from another thread.
+///
+/// # Notes
+///
+/// The `Awakener` needs to be kept alive as long as wake up notifications are
```
carllerche: I would reword this as stating that wake up events are only guaranteed to be delivered while the `Awakener` value is alive.

carllerche: The implementation detail isn't relevant for the public API docs imo

Thomasdezeeuw:
@carllerche suggested the name `Waker` instead of `Awakener`, any thoughts @asomers?

src/awakener.rs
```diff
@@ -0,0 +1,95 @@
...
+/// required. This is due to an implementation detail where if all copies of the
+/// `Awakener` are dropped it will also drop all wake up notifications from the
+/// system queue, including wake up notifications that have been added before
+/// the `Awakener` that was dropped, resulting the [`Poll`] not being woken up.
+///
+/// Only a single `Awakener` should active per [`Poll`], if multiple threads
+/// need access to the `Awakener` it can be shared via for example an `Arc`.
+/// What happens if multiple `Awakener`s are registered with the same `Poll` is
```
carllerche: Why is it undefined? If they are registered with different tokens, it should be OK?

Thomasdezeeuw: I put that in to allow some room for future implementations that weren't ok with multiple `Awakener`s, but currently it's fine on all platforms I think

asomers:
> @carllerche suggested the name `Waker` instead of `Awakener`, any thoughts @asomers?

Either name is fine by me.

src/awakener.rs
Thomasdezeeuw: Keeping the mentioning of it being an implementation detail (second line)?

Thomasdezeeuw:
Ok then I'll rename it to `Waker`, but that means I'll also have to rename some files to I'll leave if we're all on agreement on the other stuff.

src/awakener.rs
Thomasdezeeuw: I can remove it if you want.

carllerche: I'm just wondering about it. I was thinking, since tokens are exposed, you could have multiple wakers that have different significance.

src/sys/unix/kqueue.rs
```diff
@@ -238,6 +238,78 @@ impl Selector {
...
+        if (kevent.flags & libc::EV_ERROR) != 0 && kevent.data != 0 {
+            Err(io::Error::from_raw_os_error(kevent.data as i32))
+        } else {
+            Ok(())
+        }
+    }
+
+    // Used by `Awakener`.
+    #[cfg(any(target_os = "freebsd", target_os = "ios", target_os = "macos"))]
+    pub fn try_clone(&self) -> io::Result<Selector> {
```
carllerche: Since this is kqueue specific, what if we don't return a `Selector`? That would prevent needing incrementing `NEXT_ID`. We could just return the RawFd, or some other private type.

src/sys/windows/awakener.rs
```diff
@@ -1,73 +1,39 @@
...
-        events: Interests,
-        opts: PollOpt,
-    ) -> io::Result<()> {
-        self.register(registry, token, events, opts)
-    }
-
-    fn deregister(&self, _registry: &Registry) -> io::Result<()> {
-        *self.inner.lock().unwrap() = None;
+        let inner = self.inner.lock().unwrap();
+        let status = CompletionStatus::new(0, usize::from(inner.token), 0 as *mut _);
```
carllerche: The windows implementation does not work as one would expect. Receiving completion status events in `select` does not result in a mio event. Instead, in `select`, the callback is called which then results in an event getting posted via the custom readiness queue.

The easiest way to hack in awakeners would probably be to use a hard coded awakener token (there already is one [here](https://github.com/tokio-rs/mio/blob/master/src/sys/windows/selector.rs#L87)). Then, use some other field to pass in the token... the best bet is probably to use the `lpOverlapped` pointer (set to 0 here).

Then, in `select`, when an awakener event is received, push an appropriate event into the events list.

Hopefully we can replace the windows impl soon...

src/sys/unix/kqueue.rs
Thomasdezeeuw: We can also just use `0`/max as id since it won't be used in the `Awakener` anyway.

src/sys/windows/awakener.rs
Thomasdezeeuw: Can't we use the `Vec` in `Events`? We detect `overlapped() == 0` in `select` already.

Thomasdezeeuw: I've just pushed https://github.com/tokio-rs/mio/pull/969/commits/37a6ba282aad1fe3ace8d26325f6387eed45f3d6, see if it works.

Thomasdezeeuw:
Ok that worked! It just broke other tests on Windows :)

src/sys/windows/awakener.rs
Thomasdezeeuw: This works, finally.

src/awakener.rs
Thomasdezeeuw: Want me to remove it?

Thomasdezeeuw:
CI is finally green.

src/sys/unix/kqueue.rs
Thomasdezeeuw: Reusing the id seems to work fine, so I'll do that.

src/awakener.rs
Thomasdezeeuw: I've tried adding a test for this and it doesn't work on kqueue platforms currently.

Thomasdezeeuw: I could try and fix that.

Thomasdezeeuw:
Don't want me to rename it to `Waker`?

carllerche:
@Thomasdezeeuw follow up PR? I mostly wanted to get it in to fix the merge conflicts

Thomasdezeeuw:
Sure

Thomasdezeeuw:
Pr for renaming: https://github.com/tokio-rs/mio/pull/976.

PR #974, opened 2019-06-11
[skip ci] Spellcheck a comment

asomers:


asomers:
I'm going to merge this by hand since Cirrus CI obeys `[skip ci]`, which means the status checks will never pass.

carllerche:
@asomers I can't either as I set it to enforce rules on admins too :) I will temp disable it.

PR #978, opened 2019-06-12
Move custom readiness queue into sys::windows

carllerche:
This frees us up to further cleanup the mio source without having to wait on the windows rewrite.


carllerche:
Yes, it was just a move. There was some minor structural changes to make it compile but no logic changes.

Issue #668, opened 2017-08-18
Accept ToSocketAddrs


dtolnay:
The methods that currently accept SocketAddr or &SocketAddr as an argument could take A: ToSocketAddrs, which is implemented for str and various other helpful types.

- This would match the signatures of std::net's TcpListener, TcpStream, and UdpSocket.
- It would save a parse().unwrap() in all of our example code.

alexcrichton:
Oh this is actually explicitly avoided in `mio` because of the semantics of `ToSocketAddrs`. Right now `ToSocketAddrs` will perform a blocking DNS call (e.g. `getaddrinfo`) which is something that `mio`, a crate for nonblocking I/O, in general wants to discourage. In that sense all we really want is `T: TryInto<SocketAddr>` in the sense that string parsing is fine but we don't want DNS lookups to happen.

That being said, this has come up a few times, so I think it'd at least be worth explicitly documenting!

dtolnay:
Okay, makes sense. I agree, let's document this!

carllerche:
Closing due to inactivity.

PR #979, opened 2019-06-13
tcp: make TcpStream accept any ToSocketAddrs implementor

Keruspe:
This mirrors what `std::net::TcpStream` does.

You can still pass a `&SocketAddr` as it implements `ToSocketAddrs`
so it is backward compatible in this regards.

Because of type inference though, this prevents from doing:
```
TcpStream::connect(&"127.0.0.1:80".parse())
```
But it allows this instead:
```
TcpStream::connect("127.0.0.1:80")
```
It also has the benefit of adding DNS support so this is ok too:
```
TcpStream::connect("localhost:80")
```

### DIFF
--- a/src/net/tcp.rs
+++ b/src/net/tcp.rs
@@ -13,7 +13,7 @@ use iovec::IoVec;
 use net2::TcpBuilder;
 use std::fmt;
 use std::io::{Read, Write};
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw:
This is done on purpose, `ToSocketAddrs` can block when it needs to do a DNS lookup, also see #668.

PR #980, opened 2019-06-13
 Add RUST_BACKTRACE env variable to CI build 

Thomasdezeeuw:
For obvious reasons the second commit shouldn't be commited and it's only to test the backtrace is shown properly.


Thomasdezeeuw:
Ready for review now.

Results can be seen here https://cirrus-ci.com/task/6260625395679232?command=test#L97 and https://dev.azure.com/tokio-rs/Tokio/_build/results?buildId=928&view=logs.

PR #981, opened 2019-06-14
Remove unused `io` module

carllerche:
Use `std::io` instead of `crate::io` re-exports.


Thomasdezeeuw:
LGTM, assuming CI is green of course.

PR #983, opened 2019-06-15
Make Event a wrapper around a platform event

Thomasdezeeuw:
Now `Event` is just a wrapper around a platform specific event to provide
a nice API, with methods such as is_readable.

This removes the `Ready` type and moves all its methods to the `Event` type.

TODO:
* [x] Make this work on all platforms:
  * [x] kqueue,
  * [x] epoll,
  * [x] Windows.
* [x] Fix tests:
  * [x] Fix `test_write_shutdown` test, the following fails `assert!(!event.is_hup())`.
  * [x] Update all tests in `test_waker` module (uses now removed `Event::new`).
* [x] Fix examples:
  * [x] in Waker.
* [x] Fix docs everywhere.

### DIFF
--- a/src/event_imp.rs
+++ b/src/event_imp.rs
@@ -287,11 +287,6 @@ impl Interests {
     pub fn is_lio(&self) -> bool {
         (self.0.get() & LIO) != 0
     }
-
-    #[cfg(windows)]
-    pub(crate) fn to_ready(&self) -> Ready {
-        Ready(self.0.get() as u8)
-    }
 }
 
 impl ops::BitOr for Interests {
@@ -398,15 +393,14 @@ fn test_debug_interests() {
 
 /// An readiness event returned by [`Poll::poll`].
 ///
-/// `Event` is a [readiness state] paired with a [`Token`]. It is returned by
+/// `Event` is a readiness state paired with a [`Token`]. It is returned by
 /// [`Poll::poll`].
 ///
 /// For more documentation on polling and events, see [`Poll`].
 ///
-/// [`Poll::poll`]: ../struct.Poll.html#method.poll
-/// [`Poll`]: ../struct.Poll.html
-/// [readiness state]: ../struct.Ready.html
-/// [`Token`]: ../struct.Token.html
+/// [`Poll::poll`]: crate::Poll::poll
+/// [`Poll`]: crate::Poll
+/// [`Token`]: crate::Token
 #[repr(transparent)]
 pub struct Event {
     inner: sys::Event,
@@ -418,19 +412,19 @@ impl Event {
         self.inner.token()
     }
 
-    /// Returns true if the `Ready` set contains readable readiness.
+    /// Returns true if the event contains readable readiness.
     #[inline]
     pub fn is_readable(&self) -> bool {
         self.inner.is_readable()
     }
 
-    /// Returns true if the `Ready` set contains writable readiness.
+    /// Returns true if the event contains writable readiness.
     #[inline]
     pub fn is_writable(&self) -> bool {
         self.inner.is_writable()
     }
 
-    /// Returns true if the `Ready` set contains error readiness.
+    /// Returns true if the event contains error readiness.
     ///
     /// Error events occur when the socket enters an error state. In this case,
     /// the socket will also receive a readable or writable event. Reading or
@@ -445,7 +439,7 @@ impl Event {
         self.inner.is_error()
     }
 
-    /// Returns true if the `Ready` set contains HUP readiness.
+    /// Returns true if the event contains HUP readiness.
     ///
     /// HUP events occur when the remote end of a socket hangs up. In the TCP
     /// case, this occurs when the remote end of a TCP socket shuts down writes.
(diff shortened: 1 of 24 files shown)


Thomasdezeeuw:
Windows is left, the other platforms work.

Thomasdezeeuw:
I've made an attempt at fixing Windows, basically moving `Event` and `Ready` into the windows module and pretend nothing changed. We already wanted a major rewrite of the Window's side of things, I just added to the mess here and hopefully that's acceptable for now.

Thomasdezeeuw:
waker_multiple_wakeups_same_thread fails on Windows because it explicitly check for a single event, but Windows creates three. @carllerche do we just want to make the test less strict?

Thomasdezeeuw:
With 31e271ae3f7f45fc59e5cb9e662e8a82e96357aa Windows also works.

src/sys/unix/epoll.rs
```diff
@@ -171,6 +168,69 @@ impl Drop for Selector {
...
+        Event { inner: epoll_event }
+    }
+}
+
+impl fmt::Debug for Event {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // TODO: add readiness.
+        f.debug_struct("Event")
+            .field("token", &self.token())
+            //.field("readiness", &self.readiness())
```
Thomasdezeeuw: Can we do with readiness? I could create a function `epoll_event`/`kevent` -> `Ready`, but is worth the effort?

src/sys/mod.rs
```diff
@@ -1,14 +1,16 @@
 #[cfg(unix)]
 pub use self::unix::{
-    pipe, set_nonblock, EventedFd, Events, Io, Selector, TcpListener, TcpStream, UdpSocket, Waker,
+    pipe, set_nonblock, Event, EventedFd, Events, Io, RawEvent, Selector, TcpListener, TcpStream,
```
Thomasdezeeuw: Not sure if I like the name `RawEvent`, would `SysEvent` be better?

test/test_tcp_shutdown.rs
```diff
@@ -61,11 +61,11 @@ fn test_write_shutdown() {
...
     let next = events.iter().next();
-    assert_eq!(next, None);
+    assert!(next.is_none());
 
     println!("SHUTTING DOWN");
     // Now, shutdown the write half of the socket.
     socket.shutdown(Shutdown::Write).unwrap();
 
-    wait!(poll, is_readable);
+    wait!(poll, is_readable, true);
```
Thomasdezeeuw: This is a function change.

test/test_waker.rs
```diff
@@ -91,26 +76,25 @@ fn waker_multiple_wakeups_different_thread() {
...
-    );
+    expect_waker_event(&mut poll, &mut events, token);
 
     expect_no_events(&mut poll, &mut events);
 
     handle1.join().unwrap();
     handle2.join().unwrap();
 }
+
+fn expect_waker_event(poll: &mut Poll, events: &mut Events, token: Token) {
```
Thomasdezeeuw: This because less strict, allow multiple waker events rather then only one.

test/test_tcp_shutdown.rs
Thomasdezeeuw: Before this wouldn't expect `hup` here.

src/sys/mod.rs
carllerche: SysEvent is probably best as we are going for "sys" terminology.

src/event_imp.rs
```diff
@@ -899,106 +398,118 @@ fn test_debug_interests() {
...
+        self.inner.is_lio()
+    }
 
-#[cfg(windows)]
-pub fn ready_as_usize(events: Ready) -> usize {
-    events.as_usize()
-}
+    /// Get access to the platform specific event, the returned value differs
+    /// per platform.
+    pub fn raw_event(&self) -> &sys::RawEvent {
```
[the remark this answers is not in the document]
Thomasdezeeuw: Yes, this allows users to access the underlying system event for platform specific extensions. You previously mentioned accessing `kevent` for optimising Tokio.

src/sys/unix/epoll.rs
Thomasdezeeuw: I've moved the `Debug` impl to `event_imp.rs`.

src/sys/mod.rs
Thomasdezeeuw: `sys::Event` and `SysEvent` are two different types. `sys::Event` is a wrapper around `SysEvent` to implement `token`, `is_readable`, etc. and `SysEvent` is just an alias for `epoll_event`/`kevent`.

src/event_imp.rs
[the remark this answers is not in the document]
carllerche: Ah, my intent was not to expose `RawEvent` but have kqueue only fns directly on `Event` (conditionally defined).  WDYT? Either way, we can punt exposing the raw event until we get to exposing kqueue specific data.

carllerche:
Very nice job :+1:

Thomasdezeeuw:
I've removed the system event from the public API and opened #990.

PR #984, opened 2019-06-17
Implement Evented for any type Deref<Target = Evented>

Thomasdezeeuw:


carllerche:
I think this strategy is fine. There is a slight forwards compatibility risk as adding a blanket impl could block some impl we might want in the future, but `T: Deref` is pretty conservative.

Thomasdezeeuw:
@carllerche So do you want to commit to this? Then I'll rebase.

carllerche:
Your decision either way.

Thomasdezeeuw:
I think it's worth it.

PR #985, opened 2019-06-17
Remove Sub and SubAssign from Interests

Thomasdezeeuw:
/cc @FXTi

### DIFF
--- a/src/event_imp.rs
+++ b/src/event_imp.rs
@@ -811,22 +811,6 @@ impl ops::BitOrAssign for Interests {
     }
 }
 
-impl ops::Sub for Interests {
-    type Output = Self;
-
-    #[inline]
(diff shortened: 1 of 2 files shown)


FXTi:
It looks good, should be right.

src/event_imp.rs
```diff
@@ -804,29 +804,6 @@ impl ops::BitOr for Interests {
     }
 }
 
-impl ops::BitOrAssign for Interests {
```
carllerche: If we keep `BitOr` (which I think we are), we probably can keep `BitOrAssign` as well. WDYT?

Thomasdezeeuw: Sure I'll drop that commit.

Thomasdezeeuw: I've dropped the commit.

PR #986, opened 2019-06-17
Add interests associated constants to net types

Thomasdezeeuw:
Useful to quickly register a type with the correct (or most used)
interests.


carllerche:
I'm a little hesitant to add this as it is mostly for convenience vs. required for some functionality. I can see how common registrations could be nice from an ergonomics POV, but if we are going w/ ergonomics, I don't know if a constant would be best.

For example, there could be `registry.register_default(&my_socket, my_token)`. These kinds of APIs we could experiment w/ first in an external crate before committing to them here?

Thomasdezeeuw:
Fair enough, then I'll close this.

PR #997, opened 2019-06-20
Change Iter to return &Event

Thomasdezeeuw:
Before Event was 16 bytes, but since it became a wrapper for platform specific event it grew on some platforms. For example kevent is 32 bytes. Furthermore all methods on Event only take a reference to, so we can do away with the cloning.

Removes IntoIter.


Thomasdezeeuw:
Always rust fmt... but I'm very close to enabling in my editor.

PR #1002, opened 2019-06-24
Remove SelectorId

### LINKED ISSUE #999: Remove SelectorId
Currently it only seems to be used check if a `Evented` type isn't (re)registered with different selectors (https://github.com/tokio-rs/mio/blob/2d8ecd63018010e5db4e4803ac684a83f6c77ed2/src/poll.rs#L802-L813).

Thomasdezeeuw:
Now TcpStream, TcpListener and UdpSocket are just a file descriptor on
unix.

Closes #999.


carllerche:
So, `SelectorId` was added to enforce _some_ level of platform consistency. I do think that it is not great to take up the struct space when not needed.

What if we did something like keep `SelectorId` in "debug" mode, but when compiled in release mode, it would be gone. We could use `#[cfg(debug_assertions)]` to guard, which seems appropriate as the check is basically a debug assertion.

Thomasdezeeuw:
@carllerche done that in #1004, closing this.

Issue #957, opened 2019-05-25
Support vectored I/O 0.7


Thomasdezeeuw:
Tracking issue: https://github.com/rust-lang/rust/issues/58452, stable in Rust 1.36.

New method: https://doc.rust-lang.org/nightly/std/io/trait.Write.html#method.write_vectored, using https://doc.rust-lang.org/nightly/std/io/struct.IoSlice.html.

We can replace the iovec crate with `IoSlice`.

(mio_st issue: https://github.com/Thomasdezeeuw/mio-st/issues/99)

carllerche:
👍

Thomasdezeeuw:
Closed by #1016.

Issue #977, opened 2019-06-12
Support for `peer_addr` for `UdpSocket`


MOZGIII:
I'm trying to test-use the unstable rust feature from here: https://github.com/rust-lang/rust/issues/59127
It would be great if this crate provided the `peer_addr` the same way the std socket does.

It's ok if we hide it with the feature flag.

Thomasdezeeuw:
If it moves to stable I think we can do this for 0.7.

Thomasdezeeuw:
@MOZGIII You want to create a pr for this?

MOZGIII:
Sure. I'd rather enable support for this even though the std part is in nightly.
It should be possible to hide the method under a feature flag - this was it'll be opt in.

I'll also look into if it's possible to enable it automatically if the nightly `#![feature(udp_peer_addr)]` is enabled somewhere else in the code, but I guess it's not possible currently.

Any advice?

Thomasdezeeuw:
You could add an empty `nightly` feature to `Cargo.toml` (for example see https://github.com/Thomasdezeeuw/gaea/blob/76b8f797b9416efdac396f9d9b99501ed143cdaf/Cargo.toml#L36-L37). 

Next enable the feature when nightly is available: `#![cfg_attr(feature = "nightly", feature(udp_peer_addr))]` in `src/lib.rs`.

And finally add `#[cfg(feature = "nightly")]` before the method (see https://github.com/Thomasdezeeuw/gaea/blob/76b8f797b9416efdac396f9d9b99501ed143cdaf/src/sys/unix/pipe.rs#L137-L140).

I think that should work.

carllerche:
The master branch is targeting nightly atm. 0.7 will be released when At least the iovec types are stable.

Thomasdezeeuw:
`IoSlice`(previously `IoVec`) will be stablised in 1.36.0, I've just asked if `peer_addr` can be stablised for 1.37.0 (https://github.com/rust-lang/rust/issues/59127#issuecomment-501747563).

carllerche:
My personal preference is to avoid adding feature flags. If you need access to that function, you can always impl it yourself w/ an ext trait or a custom type.

MOZGIII:
[...]

PR #1016, opened 2019-07-04
Implement vectored I/O for TcpStream

Thomasdezeeuw:
@carllerche I know you've said your not a fan of this approach but at least we can start using more nightly features on the crate without breaking the stable builds this way.

TODO:
- [x] Remove `TcpStream::{read_bufs, write_bufs}`.
- [x] Windows implementation, already uses `IoVec` so it shouldn't be too hard.
- [x] Remove `iovec` dependency.

Updates #957.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,6 @@ publish = false
 iovec = "0.1.2"
 log   = "0.4.6"
 net2  = "0.2.33"
-slab  = "0.4.2"
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2.58"
@@ -41,4 +40,5 @@ winapi = { version = "0.3.7", features = ["minwindef", "minwinbase", "ioapiset",
 [dev-dependencies]
 bytes      = "0.4.12"
 env_logger = { version = "0.6.1", default-features = false }
+slab       = "0.4.2"
 tempdir    = "0.3.7"
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,7 +20,7 @@ jobs:
     name: nightly
     displayName: Nightly
     # Pin nightly to avoid being impacted by breakage
-    rust_version: nightly-2019-04-22
+    rust_version: nightly-2019-07-16
     benches: true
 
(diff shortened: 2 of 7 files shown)


Thomasdezeeuw:
I want to take the same approach for #977.

Thomasdezeeuw:
@carllerche What do you think about it?

Thomasdezeeuw:
Well [1.36 was just released](https://blog.rust-lang.org/2019/07/04/Rust-1.36.0.html), so this doesn't have to work with nightly features any more. But this can still apply to `peer_addr`.

carllerche:
@Thomasdezeeuw my take is, let’s target a future rust release and use nightly until then. I just prefer avoiding the feature flag. I expect we won’t cut 0.7 until at least async / await is on stable.

carllerche:
Basically I’m fine with this change, just without the feature flag :-)

Thomasdezeeuw:
@carllerche Sounds good to me. Could you change the CI to use the nightly compiler only? I haven't figured out the Azure pipeline stuff.

Thomasdezeeuw:
I've forced push changes to just implemented {read, write}_vectored. But this will fail the CI on old stable compilers.

carllerche:
We can probably just switch stable to nightly in Ci for now. That is what we do for Tokio.

Thomasdezeeuw:
Looking at another build from earlier today (https://dev.azure.com/tokio-rs/Tokio/_build/results?buildId=1324&view=logs), it seems 1.36 is only used for macOS builds, other OSes are still on 1.35. I'll wait off on a rebase when the other OSes also use 1.36.

Thomasdezeeuw:
Pr #1028 fixes the problems with the CI, so I'll wait for that to be merged before rebasing this.

Thomasdezeeuw:
I've rebased on master, @carllerche can you give this a final look over to see if I didn't mess anything up while merging?

Thomasdezeeuw:
Great, both min Rust version and Windows fails because they both use 1.35.

Issue #995, opened 2019-06-20
Refactor tests


carllerche:
The mio tests are still structured as they were initially many years ago. They are in need of some cleanup.

- Remove `test_` from file name
- Remove `tests/mod.rs`.
- Remove boilerplate (see [this](https://github.com/tokio-rs/mio/blob/master/test/test_tcp_shutdown.rs) test).
- Other cleanup.

Thomasdezeeuw:
I also plan to port of all tests I have in Gaea (https://github.com/Thomasdezeeuw/gaea/tree/5103cf8ae4c9e54b118f50620a43b3df4605afa4/tests).

Thomasdezeeuw:
@carllerche I think this can be closed, or do you have more plans? #1071 is also still open.

carllerche:
This can be closed :+1:

PR #1019, opened 2019-07-06
Test refactor

Thomasdezeeuw:
Step one refactoring the tests.

This
* removes the `test_` prefix from the test files,
* renames the `test` directory to `tests`, and
* merges some test files.

The easiest way to review is to do it per commit as I tried to make small(ish) logical changes per commit. But combined they become rather big.

No change were functionally changed or removed.

Updates #995.


asomers:
The old setup compiled everything in the test directory into a single executable.  But your PR results in several different executables.  I find it harder to understand the output of "cargo test" when there are several executables in the tests directory.

carllerche:
@asomers this is the idiomatic way to do it w/ rust & cargo now (mio predates all this). could you elaborate more on what you find harder to understand?

Thomasdezeeuw:
@asomers I agree that the multiple lists of tests vs 1 single list of tests is somewhat harder to read, but as @carllerche mentioned it's the way to do it in Rust now. It's also the reason why I merged some of the files to reduce the number of lists of tests.

asomers:
For one thing, there's a lot more output this way.  For another, it's hard to see how many test cases there are in total.  And finally, it's harder to see the difference between tests, lib tests, and doc tests in the output.

Thomasdezeeuw:
@asomers I agree. We can reduce the number of files and thus binaries and sections of tests to make the output a bit more manageable. `cargo test -- --format terse` can also be used, which replaces each test name with a single dot, doesn't make the output much clearer but it does become shorter.

carllerche:
What's the conclusion?

Thomasdezeeuw:
I've looking making the output more usable and found some issues: https://github.com/rust-lang/cargo/issues/4324, https://github.com/rust-lang/cargo/issues/2832. Basically other people have the same problem but no one improved the situation yet.

But I believe this is still the way to go.

carllerche:
@asomers You ok to proceed?

asomers:
@Thomasdezeeuw how do you feel about moving files into a subdirectory?  That would still keep the files separate but reduce the number of binaries.  The file structure would look like this:
```
tests/
    functional/
        mod.rs
        interests.rs
        poll.rs
        ....
```
This is the direction I'm moving for my crates, to minimize the number of binaries.  It still allows separate test binaries when that's really called for.  For example, nix has a few dedicated binaries for tests that don't use the standard test harness, or that must run in their own process for some other reason.

carllerche:
@asomers I'm not disagreeing that there are issues w/ the idiomatic structure. My personal take is that I don't want to go against the grain here. I'd like mio to just follow idiomatic cargo structure and any discussion about alt layouts / improvements taken up w/ cargo.

Thomasdezeeuw:
@asomers files two layers deep, e.g. `tests/functional/poll.rs`, won't be turned into tests automatically. We further merge the files for example to `tcp.rs`, `udp.rs` (or even merge them into a single `net.rs` file), `poll.rs`, `event.rs` and `waker.rs`. But them we might end up with very large test files.

carllerche:
fwiw, i've always wished we could add dirs to tests... probably another thing to bring up w/ cargo.

asomers:
Perhaps `tests/functional/poll.rs` won't be added automatically (I haven't checked), but `tests/functional.rs` will, and that can include files from other subdirectories.  In any case, it's not worth fighting over.  While my preference is for fewer binaries, your change is certainly not wrong, and I won't object if you commit it as is.

Thomasdezeeuw:
@asomers I've experimented with your idea in #1030, only look at the [last commit](https://github.com/tokio-rs/mio/pull/1030/commits/dfab85104a99aca52d3fc6d896da07525d1f7b88).

Thomasdezeeuw:
Following some experimentation in #1030, I'm merging this with the indent to move to the structure @asomers laid out in https://github.com/tokio-rs/mio/pull/1019#issuecomment-510221353, once the rustfmt bugs have been resolved (see #1030).

PR #1021, opened 2019-07-08
System module cleanup

### LINKED ISSUE #911: deregister() sometimes fails with "No such file or directory" on MacOS
Several users reported this error on MacOS.

Today, I got more logs (see discussion from https://github.com/Genymobile/gnirehtet/issues/136#issuecomment-456893076) in which `poll.register()` returns a `Token(3)`, but passing this token to `poll.deregister()` fails with `No such file or directory`.

This happens only on MacOS (apparently), so I guess it is related to the kqueue backend.

### LINKED ISSUE #1015: Discussion around `cvt`
The `cvt` function (defined here: https://github.com/tokio-rs/mio/blob/aa84849c612a8887fab92581e8101369e0b2c4eb/src/sys/unix/mod.rs#L43-L66) is helper function to convert a system call return value into an `io::Result`.

I'm personally not a fan of it. It is just a single if statement and I think it's better to just inline this in the code.

At the very least it could do with a better name.

Thomasdezeeuw:
This introduces the `syscall!` macro, replacing `cvt` (see #1015).

On kqueue side of things it:

 * Fixes some type definitions on NetBSD, macOS and iOS.
 * Optimises Selector::register to only pass a single event to kevent if not both readable and writable interests are specified.
 * Documents when and why some errors returned by kevent are ignored.
 * Moves calling kevent and checking of errors into there own functions to allow for greater reuse.
 * This should also fix #911, but we need a test to make sure.

Closes #1015.
Updates #911.

### DIFF
--- a/src/sys/unix/epoll.rs
+++ b/src/sys/unix/epoll.rs
@@ -25,7 +25,7 @@ impl Selector {
         // According to libuv `EPOLL_CLOEXEC` is not defined on Android API <
         // 21. But `EPOLL_CLOEXEC` is an alias for `O_CLOEXEC` on all platforms,
         // so we use that instead.
-        syscall!(libc::epoll_create1(libc::O_CLOEXEC)).map(|ep| Selector {
+        syscall!(epoll_create1(libc::O_CLOEXEC)).map(|ep| Selector {
             #[cfg(debug_assertions)]
             id: NEXT_ID.fetch_add(1, Ordering::Relaxed),
             ep,
@@ -43,7 +43,7 @@ impl Selector {
             .unwrap_or(-1);
 
         evts.clear();
-        syscall!(libc::epoll_wait(
+        syscall!(epoll_wait(
             self.ep,
             evts.events.as_mut_ptr(),
             evts.events.capacity() as i32,
@@ -62,13 +62,7 @@ impl Selector {
             u64: usize::from(token) as u64,
         };
 
-        syscall!(libc::epoll_ctl(
-            self.ep,
-            libc::EPOLL_CTL_ADD,
-            fd,
-            &mut event
-        ))
-        .map(|_| ())
+        syscall!(epoll_ctl(self.ep, libc::EPOLL_CTL_ADD, fd, &mut event)).map(|_| ())
     }
 
     pub fn reregister(&self, fd: RawFd, token: Token, interests: Interests) -> io::Result<()> {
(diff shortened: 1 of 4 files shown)


Thomasdezeeuw:
For me it is saying that the cross NetBSD build is still queue, but Azure says it's done: https://dev.azure.com/tokio-rs/Tokio/_build/results?buildId=1305&view=logs. So the CI is green, ready for review.

src/sys/unix/epoll.rs
```diff
@@ -1,44 +1,34 @@
...
 
-/// Each Selector has a globally unique(ish) ID associated with it. This ID
-/// gets tracked by `TcpStream`, `TcpListener`, etc... when they are first
-/// registered with the `Selector`. If a type that is previously associated with
-/// a `Selector` attempts to register itself with a different `Selector`, the
-/// operation will return with an error. This matches windows behavior.
+/// Unique id for use as `SelectorId`.
 #[cfg(debug_assertions)]
-static NEXT_ID: AtomicUsize = AtomicUsize::new(0);
+static NEXT_ID: AtomicUsize = AtomicUsize::new(1);
```
carllerche: What is the reason to bump to 1?

Thomasdezeeuw: So we can remove the + 1 in `Selector::new`.

src/sys/unix/mod.rs
```diff
@@ -1,3 +1,17 @@
+/// Helper macro to execute a system call that returns an `io::Result`.
+//
+// Macro must be defined before any modules that uses them.
+macro_rules! syscall {
+    ($e:expr) => {{
+        let res = unsafe { $e };
```
carllerche: Ok, I see what you mean... I want to backtrack on my earlier thoughts :)

If the `unsafe` block is contained in the macro, I **do** think the macro should inject `libc::` in order to prevent passing arbitrary code blocks into the `unsafe` block.

Or, I would vote to remove `unsafe` from the macro.

Thomasdezeeuw: I added the unsafe block in the macro to attempt to make some of the longer calls (i.e. `kevent`) fit on a single line. I'm fine either either option.

@asomers any thoughts?

carllerche: My vote is to move `libc::` into the macro and keep `unsafe` in the macro (your original suggestion).

src/sys/unix/epoll.rs
carllerche: 👍

src/sys/unix/mod.rs
Thomasdezeeuw: Changed it.

PR #1023, opened 2019-07-08
Various small changes

Thomasdezeeuw:


carllerche:
I’d like to hold off on the length field for now unless there is a compelling use case for it and we decide that it can be supported by all platforms.

The main issue with a length field is that it would prevent lazily filtering out internal events if platforms need it. 

The raw pointer cast is fine.

Thomasdezeeuw:
Seems reasonable. As #893 already includes changing the transmute to a pointer cast, I'm closing this.

PR #1027, opened 2019-07-11
Change sys::Events to Vec<sys::Event> on Unix

Thomasdezeeuw:
I'm not sure we want to do this, but I just noticed we don't really need to wrapper on epoll and kqueue.


Thomasdezeeuw:
Force push to rebase on master.

PR #1029, opened 2019-07-11
Remove net2 dependency

### LINKED ISSUE #841: Switch from net2 to socket2
Any thoughts on migrating from net2 to socket2 instead?

https://docs.rs/socket2/0.3.5/socket2/

socket2 seems to be getting more attention now than net2, and may be the path forward.

### LINKED ISSUE #1045: Reduce TcpStream, TcpListener and UdpSocket API to the same as in libstd
Currently Mio provides the following methods on `TcpStream` which the standard library doesn't (includes both setters and getters):

 - linger.
 - keepalive.
 - send_buffer_size.
 - recv_buffer_size.

In addition I suggest we remove `TcpStream::connect_stream` and `TcpStream::from_stream` to reduce the API further. `FromRawFd` and `FromRawSocket` can be used instead. Same goes for `TcpListener::from_std`, `TcpListener::accept_std` and `UdpSocket::from_socket`.

This also a potential blocker for #1029, as we currently use net2's implementation to provide most of the methods.

Thomasdezeeuw:
This initial comment only remove net2 from `TcpStream::connect`, on Unix platforms (expect for iOS, macOS and Solaris) this reduces the number of system calls from three to two.

@carllerche this does increase the complexity a bit, do we still want to continue down this route?

Closes #841.
Closes #1045.

### DIFF
--- a/src/sys/unix/tcp.rs
+++ b/src/sys/unix/tcp.rs
@@ -1,14 +1,16 @@
-use crate::sys::unix::io::set_nonblock;
-use crate::sys::unix::SourceFd;
-use crate::{event, Interests, Registry, Token};
-
-use libc;
-use net2::TcpStreamExt;
 use std::fmt;
 use std::io::{self, IoSlice, IoSliceMut, Read, Write};
+use std::mem::size_of_val;
 use std::net::{self, SocketAddr};
 use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
 use std::time::Duration;
+
+use libc;
+use net2::TcpStreamExt;
+
+use crate::sys::unix::io::set_nonblock;
+use crate::sys::unix::SourceFd;
+use crate::{event, Interests, Registry, Token};
 
 pub struct TcpStream {
     inner: net::TcpStream,
@@ -19,7 +21,59 @@ pub struct TcpListener {
 }
 
 impl TcpStream {
-    pub fn connect(stream: net::TcpStream, addr: SocketAddr) -> io::Result<TcpStream> {
+    pub fn connect(address: SocketAddr) -> io::Result<TcpStream> {
+        let domain = match address {
+            SocketAddr::V4(..) => libc::AF_INET,
+            SocketAddr::V6(..) => libc::AF_INET6,
+        };
+        #[cfg(any(
+            target_os = "ios", // Darwin doesn't have SOCK_NONBLOCK or SOCK_CLOEXEC.
+            target_os = "macos",
+            target_os = "solaris" // Not sure about Solaris, couldn't find anything online.
+        ))]
+        let socket_type = libc::SOCK_STREAM;
+        #[cfg(any(
+            target_os = "android",
+            target_os = "bitrig",
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "linux",
+            target_os = "netbsd",
+            target_os = "openbsd"
+        ))]
+        let socket_type = libc::SOCK_STREAM | libc::SOCK_NONBLOCK | libc::SOCK_CLOEXEC;
+
+        let socket = syscall!(socket(domain, socket_type, 0));
+
+        #[cfg(any(target_os = "ios", target_os = "macos", target_os = "solaris"))]
+        let socket = socket.and_then(|socket| {
+            // For platforms that don't support flags in socket, we need to
+            // set the flags ourselves.
+            syscall!(fcntl(
+                socket,
+                libc::F_SETFL,
+                libc::O_NONBLOCK | libc::O_CLOEXEC
+            ))
+            .map(|_| socket)
+        });
+
+        socket
+            .and_then(|socket| {
+                let (raw_address, raw_address_length) = socket_address(&address);
+                syscall!(connect(socket, raw_address, raw_address_length))
+                    .or_else(|err| match err {
+                        // Connect hasn't finished, but that is fine.
+                        ref err if err.raw_os_error() == Some(libc::EINPROGRESS) => Ok(0),
+                        err => Err(err),
+                    })
+                    .map(|_| socket)
+            })
+            .map(|socket| TcpStream {
+                inner: unsafe { net::TcpStream::from_raw_fd(socket) },
+            })
+    }
+
+    pub fn connect_stream(stream: net::TcpStream, addr: SocketAddr) -> io::Result<TcpStream> {
         set_nonblock(stream.as_raw_fd())?;
 
         match stream.connect(addr) {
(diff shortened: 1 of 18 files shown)


src/sys/unix/tcp.rs
```diff
@@ -254,3 +307,19 @@ impl AsRawFd for TcpListener {
         self.inner.as_raw_fd()
     }
 }
+
+// Implementation taken from the Rust standard library.
+// Copyright 2015 The Rust Project Developers.
```
carllerche: So.... licensing / copyright fun :) where did you get the copyright snippet from? Looking [here](https://github.com/rust-lang/rust/blob/master/LICENSE-MIT), I didn't see anything.

Since mio is MIT and rust is MIT, there is no problems copying as long as we respect the [MIT license](https://github.com/rust-lang/rust/blob/master/LICENSE-MIT).

Key bit:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

Except, Rust removed the copyright notice [here](https://github.com/rust-lang/rust/commit/2a8807e889a43c6b89eb6f2736907afa87ae592f#diff-7f0bdbc0a0545831259b66259ac6b604), which feels incorrect (of course, I am not a lawyer).

However, two things:

- You only need to maintain the license text for copies of *substantial* bits, and I don't think this counts (small and "obvious").
- The code is from **before** the copyright line was removed from the license.

So, I would say we should do one of two things:
- Remove this comment
- Copy [this](https://github.com/rust-lang/rust/blob/8c96f37a98a19444e9c8a82b86732c742b65c8ce/LICENSE-MIT) file verbatim at the end of our license file.

Thomasdezeeuw: I got the function from https://github.com/rust-lang/rust/blob/97b1128589fdaa786a7cf65c5a6ff7ed37a1d2f3/src/libstd/net/addr.rs#L588-L599. But I could change the code to drop the function.

carllerche: Copying is fine, i would just use one of the two strategies above.

Thomasdezeeuw: I'll remove the copyright and continue with the other places net2 is used.

Thomasdezeeuw:
Note to self: check for the same problem as in #1031 in the changes I'm going to make.

carllerche:
We should be fine to move forward w/ this if you want to.

Thomasdezeeuw:
@carllerche I was thinking about waiting until the Windows rewrite as I imagine there will a number of merge conflicts. But I'll finish this.

Thomasdezeeuw:
I've force pushed a commit. It is rebased on master and contains the code for Windows as well, still limited to `TcpStream::connect`.

src/sys/windows/tcp.rs
```diff
@@ -76,7 +85,52 @@ macro_rules! wouldblock {
...
+            socket(domain, SOCK_STREAM, 0),
+            PartialEq::eq,
+            INVALID_SOCKET
+        )
+        .and_then(|socket| {
+            syscall!(ioctlsocket(socket, FIONBIO, &mut 1), PartialEq::ne, 0).map(|_| socket)
+        })
+        .and_then(|socket| {
+            // Required for a future `connect_overlapped` operation to be
+            // executed successfully.
```
Thomasdezeeuw: After the rewrite does this still apply?

Thomasdezeeuw:
After some trial and error this now works on Windows. Can some double check the code since I just developed it of the documentation and didn't run anything locally.

Thomasdezeeuw:
I'm getting the `Either the application has not called WSAStartup, or WSAStartup failed.` error on Windows, can someone with some Windows knowledge help me with that?

carllerche:
cc @piscisaureus @PerfectLaugh ^^

We probably are missing a run_once call to WSAStartup?

PerfectLaugh:
The WSAStartup should be already executed by std library.
Of course it would be exception of you don't import the std::net

PerfectLaugh:
https://github.com/rust-lang/rust/blob/00859e3e653973120006aaf3227823062dde1ba7/src/libstd/sys/windows/net.rs#L31

carllerche:
@PerfectLaugh How can we ensure it gets called?

carllerche:
@Thomasdezeeuw In case you didn't know, you can run `cargo check --target x86_64-pc-windows-msvc` while *not* on windows to at least make sure things will compile.

PerfectLaugh:
What is the test that throw error btw?

Thomasdezeeuw:
> @Thomasdezeeuw In case you didn't know, you can run `cargo check --target x86_64-pc-windows-msvc` while _not_ on windows to at least make sure things will compile.

I know, but I used `build` which doesn't check tests (I just found out) :).

carllerche:
@PerfectLaugh it looks like `init()` is only called when using a TCP or UDP type, so it is possible for us to hit cases where we need `WSAStartup()` called.

Thomasdezeeuw:
Seems that we need basically *a* function in `std::net` for init to be called.

Thomasdezeeuw:
Oh `TcpStream` and `TcpListener` don't implement `to/as/from_raw_socket`, want me to add that? Since I've already used it in a test.

PerfectLaugh:
I think it is good to add that but notice the comment I added on #1041
https://github.com/tokio-rs/mio/issues/1041#issuecomment-513488546

PerfectLaugh:
Maybe we can do WSAStartup by ourselves? It is not bad to multiple call
this function to make sure the poll can work without tcp/udp

Thomasdezeeuw:
I've disabled `connection_reset_by_peer` for now as it requires net2.

carllerche:
@Thomasdezeeuw it seems ok to maintain a `dev-dependency` on net2.

Thomasdezeeuw:
I've added a dummy listener to `test_close_on_drop` as a workaround and opened #1046.

Thomasdezeeuw:
So I've an `init` function to the tests that will create a `TcpListener` (from standard library) to trigger it to call `WSAStartup`, so the tests run fine now. However now the same problem occurs in the examples. I think we need to resolve #1046 before this can be merged.

Thomasdezeeuw:
I've added an `init` function to Windows' TCP types, matching what libstd does. I think this pr is ready, I haven't gotten around to `UdpSocket` but I want to get this merged to prevent further delays (and merge conflicts).

Thomasdezeeuw:
@PerfectLaugh I can't request your review, but can you look at the Windows implementation?

PerfectLaugh:
Looks good to me and have no issue of test in my computer.
Btw. You should look my pull request and try to make changes into your commit. #1042 so I can close my pull request then :)

Thomasdezeeuw:
@asomers, @carllerche any chance you have some time to review this?

PR #1030, opened 2019-07-11
Test refactor continued

Thomasdezeeuw:


Thomasdezeeuw:
Creates the following output (for the functional tests):

```log

     Running target/debug/deps/functional-740432e6e33d57cd

running 1 test
test tests::poll::test_poll_closes_fd ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

The we would have three categories; unit, functional and end to end, and of course the doc tests. Seems pretty good to me.

Thomasdezeeuw:
Cargo fmt fails with `Error writing files: io error: Failed to find module poll in "functional" None`. Rustfmt rustfmt doesn't understand the `path` attribute: https://github.com/rust-lang/rustfmt/issues/1208.

asomers:
Cool!  What happens if you remove the `path` attribute?

Thomasdezeeuw:
Then we can't import poll, as it will look for `tests/poll.rs` or `tests/poll/mod.rs`.

asomers:
Well, what would happen if you eliminated both `#[path]` and the `mod tests`?

Thomasdezeeuw:
I'm stupid. I started with just having `mod poll` (not in a module) in the file and that didn't work. But just adding `mod functional { mod poll }` works, no need for `path`.

Thomasdezeeuw:
@carllerche you also ok with this layout? If so I'll update #1030 with this structure.

Thomasdezeeuw:
Ok, rustfmt failed again but this this it's just a bug I think.

Thomasdezeeuw:
I think it's https://github.com/rust-lang/rustfmt/issues/3572.

carllerche:
@Thomasdezeeuw I'm fine w/ the structure, but how do you plan to deal w/ the rustfmt bug?

Thomasdezeeuw:
I don't yet, I've just asked in https://github.com/rust-lang/rustfmt/issues/3572 for a work around. Maybe we should just merge #1019 as is and move this structure if/when we find a work around or a fix.

carllerche:
@Thomasdezeeuw Yeah, lets just merge #1019 and wait for a fix. Thanks for trying out this option :+1:

Thomasdezeeuw:
I've force pushed some changes, these use a (too complicated) macro to set the path correctly for each file as a workaround for the rustfmt bugs. A downside is that I'm pretty sure that rustfmt doesn't reach and formats the files anymore, but at least it won't return an error.

Thomasdezeeuw:
What you think about the macro approach? It isn't great, but I don't really see another way to get it to work currently.

carllerche:
Up to you, but IMO it isn't worth going against the grain and I would just stick with #1019.

Thomasdezeeuw:
Rustfmt fails with yet anther issue, but I wanted to get some feedback first. @carllerche, @asomers what do you think of the new output.

carllerche:
My vote is still to follow `cargo` idioms, but I will go with whatever :)

asomers:
I like it.  It's just what I had in mind.  Too bad about that annoying rustfmt bug, though.

Thomasdezeeuw:
The last commit works around the last rustfmt bug, however Azure doesn't seem to want to run. Any way to start it manually?

Thomasdezeeuw:
I'm closing this for now, to many rebase conflicts.

Issue #1032, opened 2019-07-11
Eliminate usage of unsafe code in `eventfd::Waker::wake`


cynecx:
https://github.com/tokio-rs/mio/blob/eadafc2bed0c8d8f10e527f8cf25fac606e1fdfc/src/sys/unix/waker.rs#L33

Can be replaced with:

```rust
let buf: [u8; 8] = 1u64.to_ne_bytes(); 
```

https://doc.rust-lang.org/std/primitive.u64.html#method.to_ne_bytes

EDIT:

Also this line could be simplified (no issue with unsafe code here):

https://github.com/tokio-rs/mio/blob/eadafc2bed0c8d8f10e527f8cf25fac606e1fdfc/src/sys/unix/waker.rs#L48

with

```rust
let buf: [u8; 8] = 0u64.to_ne_bytes(); 
```

Thomasdezeeuw:
Fixed in #1033. Thanks for the review @cynecx, if you any more comments let us know.

PR #1033, opened 2019-07-11
 Don't leak file descriptors if registering Waker fails

### LINKED ISSUE #1031: Possible fd-leak in `eventfd::Waker::new`
https://github.com/tokio-rs/mio/blob/eadafc2bed0c8d8f10e527f8cf25fac606e1fdfc/src/sys/unix/waker.rs#L24-L26

A call to `selector.register` can actually fail, hence the `?`, but in an error case we would not destroy the `eventfd` handle which results in a rare-case fd-leak.

EDIT:

Same thing here:

https://github.com/tokio-rs/mio/blob/eadafc2bed0c8d8f10e527f8cf25fac606e1fdfc/src/sys/unix/waker.rs#L129

Thomasdezeeuw:
Thanks to @cynecx for finding these problems.

Fixes #1031 and #1032.

### DIFF
--- a/src/sys/unix/waker.rs
+++ b/src/sys/unix/waker.rs
@@ -21,11 +21,13 @@ mod eventfd {
 
     impl Waker {
         pub fn new(selector: &Selector, token: Token) -> io::Result<Waker> {
-            let fd = syscall!(eventfd(0, libc::EFD_CLOEXEC | libc::EFD_NONBLOCK))?;
-
-            selector.register(fd, token, Interests::READABLE)?;
-            Ok(Waker {
-                fd: unsafe { File::from_raw_fd(fd) },
+            syscall!(eventfd(0, libc::EFD_CLOEXEC | libc::EFD_NONBLOCK)).and_then(|fd| {
+                // Turn the file descriptor into a file first so we're ensured
+                // it's closed when dropped, e.g. when register below fails.
+                let file = unsafe { File::from_raw_fd(fd) };
+                selector
+                    .register(fd, token, Interests::READABLE)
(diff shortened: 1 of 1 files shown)


src/sys/unix/waker.rs
```diff
@@ -21,16 +20,18 @@ mod eventfd {
...
         pub fn new(selector: &Selector, token: Token) -> io::Result<Waker> {
-            let fd = syscall!(eventfd(0, libc::EFD_CLOEXEC | libc::EFD_NONBLOCK))?;
-
-            selector.register(fd, token, Interests::READABLE)?;
-            Ok(Waker {
-                fd: unsafe { File::from_raw_fd(fd) },
+            syscall!(eventfd(0, libc::EFD_CLOEXEC | libc::EFD_NONBLOCK)).and_then(|fd| {
+                // Turn the file descriptor into a file first so we're ensured
+                // it's closed when dropped, e.g. when register below fails.
+                let file = unsafe { File::from_raw_fd(fd) };
```
carllerche: Do we want to define our own type that takes an `Fd` and calls `close` on drop? That might be a bit more reliable?

Thomasdezeeuw: On Unix that is exactly what `File` is and since it's not exposed anywhere I would argue that the extra code doesn't outweigh the benefits.

carllerche: The primary danger, though low probability, is the drop impl starts adding file specific syscalls. Up to you though, I approved.

Thomasdezeeuw: I guess that could be the case, let's tackle that when it comes up.

PR #1034, opened 2019-07-15
New windows implementation strategies from wepoll

PerfectLaugh:
Currently this passed most of the test suite. Except for a test in Token documentation.

https://gist.github.com/PerfectLaugh/07fb4766be01e797a678357706828f1d


src/sys/windows/selector.rs
```diff
@@ -1,596 +1,632 @@
...
+    }
+}
 
-        Ok(Selector {
-            inner,
-            readiness_queue,
-        })
+impl fmt::Debug for IoStatusBlock {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "IO_STATUS_BLOCK")
```
Thomasdezeeuw: Can you change this to `IoStatusBlock`, to match the Rust name?

src/sys/windows/afd.rs
```diff
@@ -0,0 +1,237 @@
...
+pub const AFD_POLL_RECEIVE: u32 = 0x0001;
+pub const AFD_POLL_RECEIVE_EXPEDITED: u32 = 0x0002;
+pub const AFD_POLL_SEND: u32 = 0x0004;
+pub const AFD_POLL_DISCONNECT: u32 = 0x0008;
+pub const AFD_POLL_ABORT: u32 = 0x0010;
+pub const AFD_POLL_LOCAL_CLOSE: u32 = 0x0020;
+pub const AFD_POLL_ACCEPT: u32 = 0x0080;
+pub const AFD_POLL_CONNECT_FAIL: u32 = 0x0100;
+
+pub fn eventflags_to_afd_events(epoll_events: u32) -> u32 {
```
Thomasdezeeuw: I don't know if it makes sense for this code, but you don't have to convert IOCP early anymore, this can now be done in `Event`.

src/token.rs
```diff
@@ -61,6 +61,7 @@
 ///     for _ in 0..(MAX_SOCKETS+1) {
 ///         // Connect then drop the socket
 ///         let _ = TcpStream::connect(addr).unwrap();
+///         std::thread::sleep(std::time::Duration::from_millis(10));
```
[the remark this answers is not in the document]
PerfectLaugh: Yeah. Windows fails with double registered socket somehow if it does without sleep

src/net/tcp.rs
```diff
@@ -355,33 +355,17 @@ fn inaddr_any(other: SocketAddr) -> SocketAddr {
 
 impl Read for TcpStream {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        (&self.sys).read(buf)
-    }
-}
-
-impl<'a> Read for &'a TcpStream {
```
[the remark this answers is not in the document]
PerfectLaugh: That is also my question.

Cargo.toml
```diff
@@ -35,8 +35,11 @@ slab  = "0.4.2"
 libc = "0.2.58"
 
 [target.'cfg(windows)'.dependencies]
```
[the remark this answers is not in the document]
PerfectLaugh: I forgot to remove the previous garbage.

src/lib.rs
```diff
@@ -130,58 +130,3 @@ pub mod unix {
...
-///
-/// When put together these two types enable custom Windows handles to be
-/// registered with mio's event loops. The `Binding` type is used to associate
-/// handles and the `Overlapped` type is used to execute I/O operations. When
-/// the I/O operations are completed a custom function pointer is called which
-/// typically modifies a `SetReadiness` set by `Source` methods which will get
-/// later hooked into the mio event loop.
-#[cfg(windows)]
-pub mod windows {
-    pub use crate::sys::{Binding, Overlapped};
```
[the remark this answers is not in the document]
PerfectLaugh: Added SourceRawSocket

src/sys/windows/afd.rs
PerfectLaugh: Removed all epoll flags.

carllerche:
cc @piscisaureus

src/token.rs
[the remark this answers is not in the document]
PerfectLaugh: Resolved with removing HashMap handling

PerfectLaugh:
Now all the test suite is passed.

src/sys/windows/selector.rs
```diff
@@ -1,596 +1,543 @@
...
-            trace!("select; -> got overlapped");
-            callback(status.entry());
+        if afd_events == 0 {
+            return None;
         }
 
-        trace!("returning");
-        Ok(())
-    }
+        // Codes to emulate ET in mio
```
piscisaureus: @PerfectLaugh 
Can you elaborate a bit on how this works?
In wepoll there's no way I can emulate edge-triggered, because I don't know when the user has attempted to read/write from a socket, so I don't know when to re-submit the AFD_POLL overlappped operation. I think in mio's case it might be possible because we can know when a socket read/write results in WSAEWOULDBLOCK. 
But I don't see how that works here.

src/sys/windows/selector.rs
```diff
@@ -1,596 +1,543 @@
...
-    ///
-    /// Won't actually do anything until associated with an `Selector` loop.
-    pub fn new() -> ReadyBinding {
-        ReadyBinding {
-            binding: Binding::new(),
-            readiness: None,
-        }
+impl SelectorInner {
+    pub fn new() -> io::Result<SelectorInner> {
+        CompletionPort::new(0).map(|cp| SelectorInner {
```
piscisaureus: Is there a way to share the completion port with non-socket users that can't use the AFD_POLL.
In other words, how will mio-named-pipes work?

src/sys/windows/selector.rs
PerfectLaugh: This work is in tcp.rs udp.rs. where I hooked all possible calls the WouldBlock could appear.

src/sys/windows/selector.rs
PerfectLaugh: In this scheme, selector cannot share with others its completion port, since the overlapped pointer is used for Arc::from_raw and NULL for wake

src/sys/windows/selector.rs
PerfectLaugh: And the codes here is to eliminate those events that have been read. So we can simulate proper ET behavior (probably?

src/sys/windows/selector.rs
piscisaureus: So basically we need a more general solution for this issue. After all, not everything is a socket.
I think that mio needs to provide a way for non-socket I/O to use the same completion port (like named pipes, but also [asynchronous DNS resolution](https://docs.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-getaddrinfoexw)).

cc @carllerche 
I stubbed out some not-quite-complete idea I have for this in https://github.com/piscisaureus/iocp-traits

PerfectLaugh: Then Selector will need to have a workaround for detecting inappropriate overlapped to skip. Any idea?

SergejJurecko:
Strangely this compiles in debug but not release mode (x86_64-pc-windows-msvc)

PerfectLaugh:
> Strangely this compiles in debug but not release mode (x86_64-pc-windows-msvc)

Fixed.

src/sys/mod.rs
```diff
@@ -15,8 +15,8 @@ pub mod unix;
 
 #[cfg(windows)]
 pub use self::windows::{
-    event, Binding, Event, Events, Overlapped, Selector, TcpListener, TcpStream, UdpSocket, Waker,
+    event, Event, Events, Selector, SourceRawSocket, TcpListener, TcpStream, UdpSocket, Waker,
 };
 
 #[cfg(windows)]
-mod windows;
+pub mod windows;
```
Thomasdezeeuw: Small thing but I don't think this needs to be public, could be mistaken though.

src/sys/windows/sourcerawsocket.rs
```diff
@@ -0,0 +1,128 @@
...
+/// # #[allow(dead_code)]
+/// pub struct MyIo {
+///     raw_socket: RawSocket,
+/// }
+///
+/// impl event::Source for MyIo {
+///     fn register(&self, registry: &Registry, token: Token, interests: Interests)
+///         -> io::Result<()>
+///     {
+///         SourceRawSocket::new(&self.raw_socket).register(registry, token, interests)
```
Thomasdezeeuw: This usage is rather expensive if we're creating a new `Arc<Mutex<..>>` each call to register. If we can't do without the `SockState` we might need to go for a different design.

src/sys/windows/tcp.rs
```diff
@@ -1,602 +1,271 @@
...
-    inner: FromRawArc<ListenerIo>,
-}
+use super::selector::{Selector, SockState};
 
-struct StreamIo {
-    inner: Mutex<StreamInner>,
-    read: Overlapped, // also used for connect
-    write: Overlapped,
-    socket: net::TcpStream,
+struct RegistryInternalStruct {
```
Thomasdezeeuw: No need to call it `...Struct`, the type already shows that for us.

src/sys/windows/udp.rs
```diff
@@ -1,360 +1,211 @@
...
-pub struct UdpSocket {
-    imp: Imp,
-    registration: Mutex<Option<Registration>>,
-}
+use super::selector::{Selector, SockState};
 
-#[derive(Clone)]
-struct Imp {
-    inner: FromRawArc<Io>,
+struct RegistryInternalStruct {
```
Thomasdezeeuw: Same as for tcp.

test/test_tcp.rs
```diff
@@ -418,6 +418,13 @@ fn write_bufs() {
             match s.write_bufs(&b) {
                 Ok(n) => so_far += n,
                 Err(e) => {
+                    #[cfg(windows)]
```
Thomasdezeeuw: You do mean to check this only on Windows with debug assertions enabled? If so you can use `#[cfg(and(windows, not(debug_assertions)]`. I don't think any Unix platforms needs this check.

src/sys/windows/sourcerawsocket.rs
PerfectLaugh: The problem is we need to simulate ET behavior. Then we do need SockState to maintain what interests have been read

Thomasdezeeuw: That is fine, but would the example still work then? It throws about the same after each call to (re)register doesn't it?

PerfectLaugh: Indeed. You are right. I cannot hold all the actions user took on the RawSocket.

PerfectLaugh: Then the library will need to take level trigger approach rather than edge trigger

PerfectLaugh: or just remove SourceRawSocket

Thomasdezeeuw: Let's remove it for now then and we'll find another approach later.

PerfectLaugh: Removed

test/test_tcp.rs
PerfectLaugh: Same thing happened on linux.
The common point is error ConnectionReset or ConnectionAbort will occur on Release channel's test

src/net/tcp.rs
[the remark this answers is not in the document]
carllerche: I'm pretty sure we need it for Tokio... I will dig in a bit more shortly to see what we can do about it.

src/net/tcp.rs
```diff
@@ -333,7 +333,7 @@ impl TcpStream {
     /// "would block" error is returned. This operation does not block.
     ///
     /// On Unix this corresponds to the `writev` syscall.
-    pub fn write_bufs(&self, bufs: &[&IoVec]) -> io::Result<usize> {
+    pub fn write_bufs(&mut self, bufs: &[&IoVec]) -> io::Result<usize> {
```
carllerche: We are going to want to keep this as `&self`.

src/sys/windows/selector.rs
carllerche: @piscisaureus your strategy is fine to add. Maybe we can tackle it as a follow up after this PR lands?

src/net/tcp.rs
Thomasdezeeuw: This method is gone on master and is replaced by the `io::Write` trait.

src/sys/windows/selector.rs
```diff
@@ -1,596 +1,545 @@
...
-    /// This is later used to fill out and respond to requests to `poll`. Note
-    /// that this is all implemented through the `SetReadiness` structure in the
-    /// `poll` module.
-    pub fn set_readiness(&self, set: Ready) {
-        if let Some(ref i) = self.readiness {
-            trace!("set readiness to {:?}", set);
-            i.set_readiness(set).expect("event loop disappeared?");
-        }
-    }
+        let mut iocp_events = vec![CompletionStatus::zero(); events.capacity()];
```
carllerche: Ideally we would not allocate a new vector in each call to poll.

Can we use the provided `events: &mut Events` for storage?

src/sys/windows/selector.rs
```diff
@@ -1,596 +1,545 @@
...
+        }
+        match self.afd_group.lock().unwrap().last() {
+            Some(rc) => Ok(rc.clone()),
+            None => unreachable!(),
+        }
     }
 
-    pub fn push_event(&mut self, event: Event) {
-        self.events.push(event);
+    fn _release_unused_afd(&self) {
```
carllerche: @piscisaureus is this the strategy you used in `wepoll`?

src/sys/windows/selector.rs
PerfectLaugh: Done

src/sys/windows/afd.rs
```diff
@@ -0,0 +1,184 @@
...
+                _ => Ok(afd),
+            }
+        }
+    }
+
+    pub fn poll(
+        &self,
+        info: &mut AfdPollInfo,
+        iosb: &mut IO_STATUS_BLOCK,
+        apccontext: PVOID,
```
carllerche: Could we get some comments on the `apccontext` argument here?

Is this passing ownership of the data into the function? If it is a reference, what is the lifetime of the pointer?

cc @piscisaureus

src/sys/windows/afd.rs
```diff
@@ -0,0 +1,184 @@
...
+    ) -> io::Result<bool> {
+        unsafe {
+            let info_ptr: PVOID = transmute(info);
```
carllerche: We shouldn't need to use `transmute` to cast pointer types. You should be able to do something like:

```rust
let info_ptr = info as *mut _ as PVOID;
```

`transmute` is **very** heavy and dangerous. We should only use it when absolutely necessary.

PerfectLaugh: Removed all transmutes except the one in overlapped

src/sys/windows/afd.rs
PerfectLaugh: As I observed from other snippets in Google, Apccontext is probably passed directly to CompletionStatus as Overlapped value.

carllerche:
@piscisaureus in general, it would be **super** helpful if we could include some high level docs on what AFD is and how it works in context of mio... i know its hard to track down details online.

src/sys/windows/afd.rs
carllerche: It looks like the only place this is called is in `selector.rs`. Instead of passing in the raw pointer, we could pass in an `&Arc<SockState>` and do the `clone` -> `into_raw` bit in here.

src/sys/windows/afd.rs
```diff
@@ -0,0 +1,184 @@
...
+                None,
+                apccontext,
+                iosb,
+                IOCTL_AFD_POLL,
+                info_ptr,
+                size_of::<AfdPollInfo>() as u32,
+                info_ptr,
+                size_of::<AfdPollInfo>() as u32,
+            );
+            match status {
```
carllerche: In what cases here does `NtDeviceIoControlFile` **not** take ownership of `apccontext`? My guess is, at the very least, it does not take ownership when there is an error.

We need to take care to not leak memory here.

src/sys/windows/selector.rs
```diff
@@ -1,596 +1,552 @@
...
+        /* AFD_POLL_CONNECT_FAIL and AFD_POLL_ABORT are always reported, even when not requested by the caller. */
+        let events = ev.flags | AFD_POLL_CONNECT_FAIL | AFD_POLL_ABORT;
+
+        self.user_evts = events;
+        self.user_data = ev.data;
+
+        (events & !self.pending_evts) != 0
+    }
+
+    fn update(&mut self, self_ptr: *const Mutex<SockState>) -> io::Result<()> {
```
carllerche: This could take in `&Arc<Mutex<SockState>>`.

src/sys/windows/afd.rs
PerfectLaugh: It looks like it just copy paste the apccontext to iocp as a Overlapped pointer. Not using it at all.

src/sys/windows/afd.rs
PerfectLaugh: I thought it would be general case as Afd copied only pointer not using the internal data.

src/sys/windows/afd.rs
PerfectLaugh: And I confirmed we aren't leaking memory using Dr.Memory.

PerfectLaugh: main.rs
```
use std::io;
use std::net::{SocketAddr, Ipv4Addr};
use std::thread::sleep;
use std::time::Duration;

use env_logger;
use mio::{Poll, Token, Interests};
use mio::net::UdpSocket;

fn main() -> io::Result<()> {
    env_logger::init();

    let target = SocketAddr::new(Ipv4Addr::new(192,168,1,1).into(), 53);
    let buf = &[0x00; 1400];
    let socket = UdpSocket::bind(SocketAddr::new(Ipv4Addr::UNSPECIFIED.into(), 0))?;

    let poll = Poll::new()?;
    poll.registry().register(&socket, Token(0), Interests::READABLE)?;

    drop(socket.send_to(buf, target));
    sleep(Duration::from_millis(1000));

    Ok(())
}
```
result.txt
```
Dr. Memory version 2.2.0 build 1 built on Jul  1 2019 00:40:18
Windows version: WinVer=105;Rel=1903;Build=18936;Edition=Professional
Dr. Memory results for pid 31512: "mio-mao.exe"
Application cmdline: ".\target\debug\mio-mao.exe"
Recorded 117 suppression(s) from default C:\Program Files (x86)\Dr. Memory\bin64\suppress-default.txt

Error #1: UNINITIALIZED READ: reading register eax
# 0 WS2_32.dll!WSALookupServiceNextW                                          +0x3031   (0x00007ffdca6eedd1 <WS2_32.dll+0xedd1>)
# 1 WS2_32.dll!WSALookupServiceNextW                                          +0x3c3d   (0x00007ffdca6ef9de <WS2_32.dll+0xf9de>)
# 2 WS2_32.dll!WSAStartup                                                     +0x33c    (0x00007ffdca6efd4d <WS2_32.dll+0xfd4d>)
# 3 std::sync::once::{{impl}}::call_once::{{closure}}<>                        [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\sync\once.rs:220]
# 4 std::sync::once::Once::call_inner                                          [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\sync\once.rs:387]
# 5 std::sys_common::net::UdpSocket::bind                                      [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\sys_common\net.rs:468]
# 6 core::ops::function::FnMut::call_mut<> core::result::Result<>,(core::result::Result<>)> [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libcore\ops\function.rs:148]
# 7 std::net::each_addr<> core::result::Result<>,std::sys_common::net::UdpSocket> [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\net\mod.rs:113]
# 8 std::net::udp::UdpSocket::bind<>                                           [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\net\udp.rs:94]
# 9 mio::net::udp::UdpSocket::bind                                             [C:\Users\dennis\.cargo\git\checkouts\mio-76192d34136afce1\7358ab7\src\net\udp.rs:122]
#10 mio_mao::main                                                              [G:\Dev\test-mio-mao\src\main.rs:15]
#11 std::rt::lang_start::{{closure}}<>                                         [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
#12 std::rt::lang_start_internal                                               [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\rt.rs:48]
#13 std::rt::lang_start<>                                                      [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
#14 main       
Note: @0:00:01.069 in thread 31520
Note: instruction: cmp    %eax %r12d

Error #2: UNINITIALIZED READ: reading register eflags
# 0 WS2_32.dll!WSALookupServiceNextW                                          +0x304f   (0x00007ffdca6eedef <WS2_32.dll+0xedef>)
# 1 WS2_32.dll!WSALookupServiceNextW                                          +0x3c3d   (0x00007ffdca6ef9de <WS2_32.dll+0xf9de>)
# 2 WS2_32.dll!WSAStartup                                                     +0x33c    (0x00007ffdca6efd4d <WS2_32.dll+0xfd4d>)
# 3 std::sync::once::{{impl}}::call_once::{{closure}}<>                        [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\sync\once.rs:220]
# 4 std::sync::once::Once::call_inner                                          [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\sync\once.rs:387]
# 5 std::sys_common::net::UdpSocket::bind                                      [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\sys_common\net.rs:468]
# 6 core::ops::function::FnMut::call_mut<> core::result::Result<>,(core::result::Result<>)> [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libcore\ops\function.rs:148]
# 7 std::net::each_addr<> core::result::Result<>,std::sys_common::net::UdpSocket> [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\net\mod.rs:113]
# 8 std::net::udp::UdpSocket::bind<>                                           [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\net\udp.rs:94]
# 9 mio::net::udp::UdpSocket::bind                                             [C:\Users\dennis\.cargo\git\checkouts\mio-76192d34136afce1\7358ab7\src\net\udp.rs:122]
#10 mio_mao::main                                                              [G:\Dev\test-mio-mao\src\main.rs:15]
#11 std::rt::lang_start::{{closure}}<>                                         [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
#12 std::rt::lang_start_internal                                               [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\rt.rs:48]
#13 std::rt::lang_start<>                                                      [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
#14 main       
Note: @0:00:01.083 in thread 31520
Note: instruction: cmovo  %r13 -> %rax

Error #3: UNINITIALIZED READ: reading 0x0000001dbcefe8c0-0x0000001dbcefe8c4 4 byte(s) within 0x0000001dbcefe8c0-0x0000001dbcefe8c8
# 0 replace_memset                                                             [d:\drmemory_package\drmemory\replace.c:200]
# 1 WS2_32.dll!WSALookupServiceNextW                                          +0x3088   (0x00007ffdca6eee29 <WS2_32.dll+0xee29>)
# 2 WS2_32.dll!WSALookupServiceNextW                                          +0x3c3d   (0x00007ffdca6ef9de <WS2_32.dll+0xf9de>)
# 3 WS2_32.dll!WSAStartup                                                     +0x33c    (0x00007ffdca6efd4d <WS2_32.dll+0xfd4d>)
# 4 std::sync::once::{{impl}}::call_once::{{closure}}<>                        [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\sync\once.rs:220]
# 5 std::sync::once::Once::call_inner                                          [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\sync\once.rs:387]
# 6 std::sys_common::net::UdpSocket::bind                                      [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\sys_common\net.rs:468]
# 7 core::ops::function::FnMut::call_mut<> core::result::Result<>,(core::result::Result<>)> [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libcore\ops\function.rs:148]
# 8 std::net::each_addr<> core::result::Result<>,std::sys_common::net::UdpSocket> [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\net\mod.rs:113]
# 9 std::net::udp::UdpSocket::bind<>                                           [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\net\udp.rs:94]
#10 mio::net::udp::UdpSocket::bind                                             [C:\Users\dennis\.cargo\git\checkouts\mio-76192d34136afce1\7358ab7\src\net\udp.rs:122]
#11 mio_mao::main                                                              [G:\Dev\test-mio-mao\src\main.rs:15]
#12 std::rt::lang_start::{{closure}}<>                                         [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
#13 std::rt::lang_start_internal                                               [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\rt.rs:48]
#14 std::rt::lang_start<>                                                      [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
#15 main       
Note: @0:00:01.105 in thread 31520
Note: instruction: cmp    0x30(%rsp) $0x0000000000000003

Error #4: UNINITIALIZED READ: reading 0x0000001dbcefea4c-0x0000001dbcefea50 4 byte(s) within 0x0000001dbcefe9d0-0x0000001dbcefea54
# 0 system call NtDeviceIoControlFile SOCKET_CONTEXT pre-Padding
# 1 MSWSOCK.dll!NSPStartup                                                    +0x595c   (0x00007ffdc8617afd <MSWSOCK.dll+0x7afd>)
# 2 MSWSOCK.dll!NSPStartup                                                    +0x3d4b   (0x00007ffdc8615eec <MSWSOCK.dll+0x5eec>)
# 3 WS2_32.dll!WSASocketW                                                     +0x115    (0x00007ffdca6e9e16 <WS2_32.dll+0x9e16>)
# 4 std::sys::windows::net::Socket::new                                        [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\sys\windows\net.rs:99]
# 5 std::sys_common::net::UdpSocket::bind                                      [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\sys_common\net.rs:465]
# 6 core::ops::function::FnMut::call_mut<> core::result::Result<>,(core::result::Result<>)> [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libcore\ops\function.rs:148]
# 7 std::net::each_addr<> core::result::Result<>,std::sys_common::net::UdpSocket> [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\net\mod.rs:113]
# 8 std::net::udp::UdpSocket::bind<>                                           [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\net\udp.rs:94]
# 9 mio::net::udp::UdpSocket::bind                                             [C:\Users\dennis\.cargo\git\checkouts\mio-76192d34136afce1\7358ab7\src\net\udp.rs:122]
#10 mio_mao::main                                                              [G:\Dev\test-mio-mao\src\main.rs:15]
#11 std::rt::lang_start::{{closure}}<>                                         [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
#12 std::rt::lang_start_internal                                               [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\rt.rs:48]
#13 std::rt::lang_start<>                                                      [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
#14 main       
Note: @0:00:01.435 in thread 31520

Error #5: UNINITIALIZED READ: reading 0x0000001dbcefeabc-0x0000001dbcefeac0 4 byte(s) within 0x0000001dbcefea40-0x0000001dbcefeac4
# 0 system call NtDeviceIoControlFile SOCKET_CONTEXT pre-Padding
# 1 MSWSOCK.dll!NSPStartup                                                    +0x595c   (0x00007ffdc8617afd <MSWSOCK.dll+0x7afd>)
# 2 MSWSOCK.dll!NSPStartup                                                    +0x5d45   (0x00007ffdc8617ee6 <MSWSOCK.dll+0x7ee6>)
# 3 WS2_32.dll!bind                                                           +0x6e     (0x00007ffdca6f1ddf <WS2_32.dll+0x11ddf>)
# 4 std::sys_common::net::UdpSocket::bind                                      [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\sys_common\net.rs:467]
# 5 core::ops::function::FnMut::call_mut<> core::result::Result<>,(core::result::Result<>)> [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libcore\ops\function.rs:148]
# 6 std::net::each_addr<> core::result::Result<>,std::sys_common::net::UdpSocket> [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\net\mod.rs:113]
# 7 std::net::udp::UdpSocket::bind<>                                           [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\net\udp.rs:94]
# 8 mio::net::udp::UdpSocket::bind                                             [C:\Users\dennis\.cargo\git\checkouts\mio-76192d34136afce1\7358ab7\src\net\udp.rs:122]
# 9 mio_mao::main                                                              [G:\Dev\test-mio-mao\src\main.rs:15]
#10 std::rt::lang_start::{{closure}}<>                                         [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
#11 std::rt::lang_start_internal                                               [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\rt.rs:48]
#12 std::rt::lang_start<>                                                      [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
#13 main       
Note: @0:00:01.463 in thread 31520

Error #6: UNINITIALIZED READ: reading 0x0000001dbcefef3c-0x0000001dbcefef40 4 byte(s) within 0x0000001dbcefef38-0x0000001dbcefef48
# 0 system call NtDeviceIoControlFile AFD_SEND_INFO_UDP.BufferArray
# 1 MSWSOCK.dll!Tcpip6_WSHGetWildcardSockaddr                                 +0x417    (0x00007ffdc861f9f8 <MSWSOCK.dll+0xf9f8>)
# 2 WS2_32.dll!sendto                                                         +0xe9     (0x00007ffdca6f16ca <WS2_32.dll+0x116ca>)
# 3 std::sys_common::net::UdpSocket::send_to                                   [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\sys_common\net.rs:499]
# 4 std::net::udp::UdpSocket::send_to<>                                        [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\net\udp.rs:177]
# 5 mio::sys::windows::udp::UdpSocket::send_to                                 [C:\Users\dennis\.cargo\git\checkouts\mio-76192d34136afce1\7358ab7\src\sys\windows\udp.rs:78]
# 6 mio::net::udp::UdpSocket::send_to                                          [C:\Users\dennis\.cargo\git\checkouts\mio-76192d34136afce1\7358ab7\src\net\udp.rs:224]
# 7 mio_mao::main                                                              [G:\Dev\test-mio-mao\src\main.rs:20]
# 8 std::rt::lang_start::{{closure}}<>                                         [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
# 9 std::rt::lang_start_internal                                               [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\rt.rs:48]
#10 std::rt::lang_start<>                                                      [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
#11 main       
Note: @0:00:01.721 in thread 31520

Error #7: UNINITIALIZED READ: reading 0x0000001dbcefee14-0x0000001dbcefee18 4 byte(s) within 0x0000001dbcefee14-0x0000001dbcefee18
# 0 system call NtDeviceIoControlFile AFD_SEND_INFO_UDP.SizeOfRemoteAddress
# 1 MSWSOCK.dll!Tcpip6_WSHGetWildcardSockaddr                                 +0x417    (0x00007ffdc861f9f8 <MSWSOCK.dll+0xf9f8>)
# 2 WS2_32.dll!sendto                                                         +0xe9     (0x00007ffdca6f16ca <WS2_32.dll+0x116ca>)
# 3 std::sys_common::net::UdpSocket::send_to                                   [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\sys_common\net.rs:499]
# 4 std::net::udp::UdpSocket::send_to<>                                        [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\net\udp.rs:177]
# 5 mio::sys::windows::udp::UdpSocket::send_to                                 [C:\Users\dennis\.cargo\git\checkouts\mio-76192d34136afce1\7358ab7\src\sys\windows\udp.rs:78]
# 6 mio::net::udp::UdpSocket::send_to                                          [C:\Users\dennis\.cargo\git\checkouts\mio-76192d34136afce1\7358ab7\src\net\udp.rs:224]
# 7 mio_mao::main                                                              [G:\Dev\test-mio-mao\src\main.rs:20]
# 8 std::rt::lang_start::{{closure}}<>                                         [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
# 9 std::rt::lang_start_internal                                               [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\rt.rs:48]
#10 std::rt::lang_start<>                                                      [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
#11 main       
Note: @0:00:01.735 in thread 31520

Error #8: UNINITIALIZED READ: reading 0x0000001dbcefee18-0x0000001dbcefee20 8 byte(s) within 0x0000001dbcefee18-0x0000001dbcefee20
# 0 system call NtDeviceIoControlFile AFD_SEND_INFO_UDP.RemoteAddress
# 1 MSWSOCK.dll!Tcpip6_WSHGetWildcardSockaddr                                 +0x417    (0x00007ffdc861f9f8 <MSWSOCK.dll+0xf9f8>)
# 2 WS2_32.dll!sendto                                                         +0xe9     (0x00007ffdca6f16ca <WS2_32.dll+0x116ca>)
# 3 std::sys_common::net::UdpSocket::send_to                                   [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\sys_common\net.rs:499]
# 4 std::net::udp::UdpSocket::send_to<>                                        [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\net\udp.rs:177]
# 5 mio::sys::windows::udp::UdpSocket::send_to                                 [C:\Users\dennis\.cargo\git\checkouts\mio-76192d34136afce1\7358ab7\src\sys\windows\udp.rs:78]
# 6 mio::net::udp::UdpSocket::send_to                                          [C:\Users\dennis\.cargo\git\checkouts\mio-76192d34136afce1\7358ab7\src\net\udp.rs:224]
# 7 mio_mao::main                                                              [G:\Dev\test-mio-mao\src\main.rs:20]
# 8 std::rt::lang_start::{{closure}}<>                                         [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
# 9 std::rt::lang_start_internal                                               [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\rt.rs:48]
#10 std::rt::lang_start<>                                                      [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
#11 main       
Note: @0:00:01.738 in thread 31520

Error #9: LEAK 40 direct bytes 0x00000198a8ea07c0-0x00000198a8ea07e8 + 8 indirect bytes
# 0 replace_RtlAllocateHeap                    [d:\drmemory_package\common\alloc_replace.c:3771]
# 1 ntdll.dll!RtlpAddVectoredHandler
# 2 std::rt::lang_start_internal               [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\/src\libstd\rt.rs:34]
# 3 std::rt::lang_start<>                      [/rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c\src\libstd\rt.rs:64]
# 4 main       

===========================================================================
FINAL SUMMARY:

DUPLICATE ERROR COUNTS:

SUPPRESSIONS USED:

ERRORS FOUND:
      0 unique,     0 total unaddressable access(es)
      8 unique,     8 total uninitialized access(es)
      0 unique,     0 total invalid heap argument(s)
      0 unique,     0 total GDI usage error(s)
      0 unique,     0 total handle leak(s)
      0 unique,     0 total warning(s)
      1 unique,     1 total,     48 byte(s) of leak(s)
      0 unique,     0 total,      0 byte(s) of possible leak(s)
ERRORS IGNORED:
     24 potential error(s) (suspected false positives)
         (details: C:\Users\dennis\AppData\Roaming\Dr. Memory\DrMemory-mio-mao.exe.31512.000\potential_errors.txt)
     22 unique,    22 total,   3585 byte(s) of still-reachable allocation(s)
         (re-run with "-show_reachable" for details)
Details: C:\Users\dennis\AppData\Roaming\Dr. Memory\DrMemory-mio-mao.exe.31512.000\results.txt
```

piscisaureus: The "apccontext" pointer as such is treated as an opaque value by the kernel.

(The actual value of the apccontext here is a pointer to an ntstatus block embedded in an overlapped. The kernel does take ownership of the ntstatus until the operation completes)

src/sys/windows/selector.rs
piscisaureus: This is better, in wepoll I never close afd driver handles.
(That said, you would want to avoid repeatedly closing one and then immediately creating a new one, which I didn't check)

src/sys/windows/afd.rs
piscisaureus: See https://github.com/tokio-rs/mio/pull/1034#discussion_r304438939

src/sys/windows/afd.rs
```diff
@@ -0,0 +1,184 @@
...
+                _ => Err(io::Error::from_raw_os_error(
+                    RtlNtStatusToDosError(status) as i32
+                )),
+            }
+        }
+    }
+
+    pub fn cancel(&self, iosb: &mut IO_STATUS_BLOCK) -> io::Result<()> {
+        unsafe {
+            if iosb.u.Status != STATUS_PENDING {
```
piscisaureus: The IO_STATUS_BLOCK can be updated by the kernel any time when the operation is pending.
So it should probably be wrapped in an UnsafeCell and be pinned.

src/sys/windows/selector.rs
piscisaureus: I agree this can land first.

src/sys/windows/afd.rs
```diff
@@ -0,0 +1,184 @@
...
+
+impl fmt::Debug for AfdPollInfo {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("AfdPollInfo").finish()
+    }
+}
+
+impl Afd {
+    pub fn new(cp: &CompletionPort) -> io::Result<Afd> {
+        let mut afd_helper_name = OsStr::new(AFD_HELPER_NAME)
```
piscisaureus: It's unfortunate that this is done every time a handle to the afd driver is created.
It'd be nicer to make afd_helper_attributes a static global, maybe with the help of lazy_static.

piscisaureus: But doing so in a follow up patch would be okay.

src/sys/windows/afd.rs
PerfectLaugh: UnsafeCell has no Deref trait implemented therefore cannot be pinned. I used the Box instead.

Thomasdezeeuw: If you can guarantee that certain memory doesn't move it can be pinned, then you don't need to use `Box`. But these kind of optimisations can be done later.

PerfectLaugh: Uploaded new code. Please check

src/sys/windows/afd.rs
PerfectLaugh: Done.

src/sys/windows/selector.rs
PerfectLaugh: The afd poll size is same as your wepoll, so the behavior should be the same.

PerfectLaugh:
Please check the documentation I added also. My English is the worst of the world and I don't want any syntax/grammar error.

src/sys/windows/selector.rs
```diff
@@ -1,596 +1,541 @@
...
+    flags
 }
 
-impl fmt::Debug for Overlapped {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("Overlapped").finish()
+fn get_base_socket(raw_socket: RawSocket) -> io::Result<RawSocket> {
+    let mut base_socket: RawSocket = 0;
+    let mut bytes: u32 = 0;
+    const SIO_BASE_HANDLE: u32 = 0x48000022;
```
Thomasdezeeuw: This is also exported by `winapi` in `winapi::um::mswsock::SIO_BASE_HANDLE`, please use that.

src/sys/windows/selector.rs
```diff
@@ -1,596 +1,541 @@
...
-        for status in events.statuses[..n].iter() {
-            // This should only ever happen from the waker.
-            if status.overlapped() as usize == 0 {
-                let token = Token(status.token());
-                if token == WAKE {
-                    continue;
-                }
-                events.events.push(Event::new(Ready::READABLE, token));
-                continue;
+        unsafe {
```
Thomasdezeeuw: Can you reduce this unsafe block and add a comment why it is safe?

src/sys/windows/mod.rs
```diff
@@ -1,197 +1,20 @@
...
+pub use udp::UdpSocket;
+pub use waker::Waker;
 
-    let r = winapi::um::winbase::SetFileCompletionNotificationModes(handle, flags);
-    if r == winapi::shared::minwindef::TRUE {
-        Ok(())
-    } else {
-        Err(io::Error::last_os_error())
-    }
+pub trait WindowsSocketState {
```
Thomasdezeeuw: You can drop the Windows from this name.

src/sys/windows/afd.rs
```diff
@@ -0,0 +1,222 @@
...
+        if status == STATUS_SUCCESS || status == STATUS_NOT_FOUND {
+            return Ok(());
+        }
+        Err(io::Error::from_raw_os_error(
+            RtlNtStatusToDosError(status) as i32
+        ))
+    }
+}
+
+pub const AFD_POLL_RECEIVE: u32 = 0x0001;
```
Thomasdezeeuw: Don't any Windows API crate export these? If fine with defining them here, but I would rather some other crate that care of them, e.g. `winapi`.

src/sys/windows/io_status_block.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    pub fn as_mut_ptr(&self) -> *mut IO_STATUS_BLOCK {
+        self.0.get()
+    }
+}
+
+impl Deref for IoStatusBlock {
+    type Target = IO_STATUS_BLOCK;
+    fn deref(&self) -> &Self::Target {
+        unsafe { &*self.as_ptr() }
```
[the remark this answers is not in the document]
PerfectLaugh: I think this usage is ambiguous. I removed that.

src/sys/windows/io_status_block.rs
```diff
@@ -0,0 +1,44 @@
+use std::cell::UnsafeCell;
+use std::fmt;
+use std::mem;
+use std::ops::{Deref, DerefMut};
+
+use ntapi::ntioapi::IO_STATUS_BLOCK;
+
+pub struct IoStatusBlock(UnsafeCell<IO_STATUS_BLOCK>);
+
+unsafe impl Send for IoStatusBlock {}
```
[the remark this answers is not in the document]
PerfectLaugh: Left the comment here.
```
// There is a pointer field in `IO_STATUS_BLOCK_u`, which we don't use that. Thus it is safe to implement Send here.
```

src/sys/windows/selector.rs
```diff
@@ -1,596 +1,541 @@
...
-    /// The actual completion port that's used to manage all I/O
-    port: CompletionPort,
-
-    /// A pool of buffers usable by this selector.
-    ///
-    /// Primitives will take buffers from this pool to perform I/O operations,
-    /// and once complete they'll be put back in.
-    buffers: Mutex<BufferPool>,
+pub struct SockState {
+    iosb: Pin<IoStatusBlock>,
```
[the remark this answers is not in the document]
PerfectLaugh: SockState is wrapped by Arc<Mutex<SockState>> all the time. I think it is actually safe not using Pin though.
I will do Pin<Box<IoStatusBlock>> for now.

src/sys/windows/selector.rs
PerfectLaugh: I cannot because the logic is connected. It is better to put this still.
```
// We use the status info in IO_STATUS_BLOCK to determine the socket poll status. It is unsafe to use a pointer of IO_STATUS_BLOCK.
```

src/sys/windows/afd.rs
PerfectLaugh: This is internal component in AFD which is not public API of Windows.

PerfectLaugh: I think this cannot be in winapi.

src/sys/windows/io_status_block.rs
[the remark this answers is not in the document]
Thomasdezeeuw: I mean in the code itself.

Thomasdezeeuw:
@PerfectLaugh can you not mention me via @ in commits, now I'm going to get an email from GitHub every time something happens with that commit, thanks.

PerfectLaugh:
> @PerfectLaugh can you not mention me via @ in commits, now I'm going to get an email from GitHub every time something happens with that commit, thanks.

sorry

src/sys/windows/io_status_block.rs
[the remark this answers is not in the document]
PerfectLaugh: I left here as a backup

src/sys/windows/selector.rs
carllerche: Could you explain how this code eliminates events that have been read?

src/sys/windows/afd.rs
carllerche: @piscisaureus right, but what I mean is that **we** are passing a pointer in... unless we never deref it again (which would be surprising), we need to make sure the lifetimes are correct. I would assume we get the pointer back when the afd event completes?

src/sys/windows/selector.rs
PerfectLaugh: This code simply rely on assumption that mio can control and intercept every aspects that user will do to this socket. With this assumption (As long as rawsocket is not being manipulated by external library or raw api calls), we can wipe down the interests that user hasn't done it, simulating the ET behavior.

src/sys/windows/afd.rs
carllerche: @Thomasdezeeuw Yeah, AFD isn't **technically** a public API, so the `winapi` crate doesn't include it.

carllerche:
I would like to get this PR merged sooner than later. Any non critical follow up change can be tracked in a follow up issue.

@Thomasdezeeuw @piscisaureus could you list out any changes that are blocking merging this?

src/sys/windows/afd.rs
PerfectLaugh: There are only two operations that use the pointer. update and cancel in the socket state. The socket state is being included by Mutex. So the operation can only be happening one time. Update can be used at start of the poll. Cancel can be used while the polling. And the pointer isn't being freed if the operation is pending.
Of course my code is just copying the implementation of wepoll source code. So there's maybe a better refactor structure we can do.

Thomasdezeeuw:
I'm fine with merging now, all my comments can be fixed later.

carllerche:
I opened #1041 to track changes. Once we get the :+1: from @piscisaureus, I will merge this and extract all remaining items & questions into #1041.

Thomasdezeeuw:
Thank you @PerfectLaugh you've helped us with one of the major blockers for Mio v0.7!

Issue #941, opened 2019-05-14
Reconsider providing `hup` as a readiness value for all *nix platforms.


carllerche:
For `epoll`, `HUP` indicates that both the read *and* the write directions of a socket are closed. With `kqueue`, read and write directions are tracked separately. Each direction can have an optional `EOF` filter indicating that the direction is closed. There is no way to know that **both** directions are closed without tracking additional state.

However, for TCP, `EOF` for the write direction is equivalent to HUP. This does not apply to other types of evented handles.

The question is, how should this be normalized?

One option is to have `hup` be a platform specific readiness value. Another option would be to translate `EPOLLHUP` -> "write EOF" and document that for TCP, this indicates a HUP.

carllerche:
This is a continuation of discussion starting [here](https://github.com/tokio-rs/mio/pull/933#issuecomment-490605534).

cc @Thomasdezeeuw @asomers

Thomasdezeeuw:
I'm going to think about this a bit more over the weekend so it might take me a while to respond to this and #939. However I'm looking for some examples of how HUP is used currently, the only usage I could find is from Tokio.

If anyone uses HUP readiness could you share how you're using it, or point us to the code? I think that would help make a decision on how to handle this.

carllerche:
@Thomasdezeeuw Ideally we can get #939 merged & released sooner than later as it is causing issues in production (cloudflare hit this recently).

Thomasdezeeuw:
We could merge #939 and determine how to handle HUP cross-platform for 0.7.

carllerche:
@Thomasdezeeuw I think that is the best strategy.

Thomasdezeeuw:
#983 makes `is_hup` available on all platforms. Anything else that needs to be done?

carllerche:
@Thomasdezeeuw The main question is if HUP will be triggered on only linux or on all *nix platforms. I believe we have to stick with Linux only here. I believe that all that is left to close this issue is verify that all HUP related code is deleted from kqueue.

Thomasdezeeuw:
Shouldn't we document the different meaning for HUP on epoll/kqueue platforms and support both?

carllerche:
[...]

PR #1037, opened 2019-07-16
Split Event::is_hup into is_hup and is_read_hup and refer to OS selector docs

### LINKED ISSUE #1040: Add a `rdhup` event.
This can be supported by all OSes.

Thomasdezeeuw:
This doesn't mention Windows, I think the Windows rewrite should be merged first.

Updates #941.
Closes #1040.

### DIFF
--- a/src/event/event.rs
+++ b/src/event/event.rs
@@ -71,9 +71,8 @@ impl Event {
     ///
     /// Because of the above be cautions when using this in cross-platform
     /// applications, Mio makes no attempt at normalising this indicator and
-    /// only provides a convenience method to read it. We advice looking at the
-    /// documentation provided for the selectors (see below) when using this
-    /// indicator.
+    /// only provides a convenience method to read it. Refer to the selector
+    /// documentation (below) when using this indicator.
     ///
     /// The table below shows what flags are checked on what OS.
     ///
--- a/src/event/event.rs
+++ b/src/event/event.rs
@@ -69,19 +69,11 @@ impl Event {
     /// Method is available on all platforms, but not all platforms (can) use
     /// this indicator.
     ///
-    /// When HUP events are received is different per platform. For example on
-    /// epoll platforms HUP will be received when a TCP socket reached EOF,
-    /// however on kqueue platforms a HUP event will be received when `shutdown`
-    /// is called on the connection (both when shutting down the reading
-    /// **and/or** writing side). Meaning that even though this function might
-    /// return true on one platform in a given situation, it might return false
-    /// on a different platform in the same situation. Furthermore even if true
-    /// was returned on both platforms it could simply mean something different
-    /// on the two platforms.
-    ///
     /// Because of the above be cautions when using this in cross-platform
     /// applications, Mio makes no attempt at normalising this indicator and
-    /// only provides a convenience method to read it.
+    /// only provides a convenience method to read it. We advice looking at the
+    /// documentation provided for the selectors (see below) when using this
+    /// indicator.
     ///
     /// The table below shows what flags are checked on what OS.
     ///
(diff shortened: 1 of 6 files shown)


src/event/event.rs
```diff
@@ -71,6 +82,17 @@ impl Event {
     /// Because of the above be cautions when using this in cross-platform
     /// applications, Mio makes no attempt at normalising this indicator and
     /// only provides a convenience method to read it.
+    ///
+    /// The table below shows what flags are checked on what OS.
+    ///
+    /// | [OS selector] | Flag(s) checked |
+    /// |---------------|-----------------|
+    /// | [epoll]       | `EPOLLHUP` and `EPOLLRDHUP` |
```
carllerche: for `hup` **only** `EPOLLHUP` can be checked (nothing for kqueue).

We need to add a separate `rdhup` event, which can be supported by all OSes (tracking issue #1040).

src/event/event.rs
```diff
@@ -58,30 +69,72 @@ impl Event {
     /// Method is available on all platforms, but not all platforms (can) use
     /// this indicator.
     ///
-    /// When HUP events are received is different per platform. For example on
```
Thomasdezeeuw: I'm not sure this documentation is accurate anymore, so I decide to refer to the selector documentation instead. I think it's the best solution as the indicator is too different per platforms. @carllerche do you agree?

Thomasdezeeuw:
Rebased on master.

src/event/event.rs
carllerche: Yes, I think that is the right decision :+1:

Some wordsmithing, instead of:

> Method is available on all platforms, but not all platforms (can) use this indicator.

Maybe:


> Method is available on all platforms, but not all platforms trigger the HUP event.

src/event/event.rs
```diff
@@ -58,30 +69,72 @@ impl Event {
...
-    /// **and/or** writing side). Meaning that even though this function might
-    /// return true on one platform in a given situation, it might return false
-    /// on a different platform in the same situation. Furthermore even if true
-    /// was returned on both platforms it could simply mean something different
-    /// on the two platforms.
-    ///
     /// Because of the above be cautions when using this in cross-platform
     /// applications, Mio makes no attempt at normalising this indicator and
-    /// only provides a convenience method to read it.
+    /// only provides a convenience method to read it. We advice looking at the
```
carllerche: Instead of:

> We advice looking at the documentation provided for the selectors (see below) when using this indicator.

Maybe:

> Refer to the selector documentation (below) when using this indicator.

src/event/event.rs
```diff
@@ -58,30 +69,72 @@ impl Event {
...
+    /// documentation provided for the selectors (see below) when using this
+    /// indicator.
+    ///
+    /// The table below shows what flags are checked on what OS.
+    ///
+    /// | [OS selector] | Flag(s) checked |
+    /// |---------------|-----------------|
+    /// | [epoll]       | `EPOLLRDHUP`    |
+    /// | [kqueue]      | `EV_EOF`        |
```
carllerche: I think it is: `EVFILT_READ` with `EV_EOF`.

Thomasdezeeuw: You're right but doesn't the read part of the method name imply that? I'm fine with adding it though.

src/sys/unix/kqueue.rs
```diff
@@ -343,7 +343,12 @@ pub mod event {
...
     }
 
-    pub fn is_hup(event: &Event) -> bool {
+    pub fn is_hup(_: &Event) -> bool {
+        // Not supported.
+        false
+    }
+
+    pub fn is_read_hup(event: &Event) -> bool {
         (event.flags & libc::EV_EOF) != 0
```
carllerche: This needs to also check, `event.filter == libc::EVFILT_READ`. Sorry about that, I should have seen it.

PR #1048, opened 2019-08-03
Drop support for LIO

### LINKED ISSUE #1036: Remove Event::is_lio?
It's only supported on FreeBSD and event the manual of kqueue doesn't mention it: https://www.freebsd.org/cgi/man.cgi?query=kqueue&manpath=FreeBSD+12.0-RELEASE+and+Ports.

/cc @asomers

Thomasdezeeuw:
list directed I/O (LIO) was only sort of supported on FreeBSD. The
kqueue manual doesn't mention it and we didn't have tests for it making
it all round poorly supported.

For v0.7, in an effort to reduce the API, LIO is removed. It can be
added back later (with proper support and testing).

Closes #1036.

### DIFF
--- a/src/event/event.rs
+++ b/src/event/event.rs
@@ -162,17 +162,6 @@ impl Event {
         sys::event::is_aio(&self.inner)
     }
 
-    /// Returns true if the event contains LIO readiness.
-    ///
-    /// # Notes
-    ///
-    /// Method is available on all platforms, but only FreeBSD supports LIO. On
-    /// FreeBSD this method checks the `EVFILT_LIO` flag.
-    #[inline]
-    pub fn is_lio(&self) -> bool {
-        sys::event::is_lio(&self.inner)
-    }
-
     /// Create a reference to an `Event` from a platform specific event.
     pub(crate) fn from_sys_event_ref(sys_event: &sys::Event) -> &Event {
         unsafe {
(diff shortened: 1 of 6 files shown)


asomers:
Whoa!  I'm still using lio (in the mio-aio and tokio-file) crates.  I don't want to drop it from mio.  What should we do to improve the testing within mio itself?

carllerche:
@Thomasdezeeuw I guess we need a general policy of what we support and how.

Thomasdezeeuw:
@asomers 

> Whoa! I'm still using lio (in the mio-aio and tokio-file) crates. I don't want to drop it from mio. What should we do to improve the testing within mio itself?

Well any tests would be nice. As you can see from the removed code currently LIO isn't tested at all. To start:

* If the options (`Interests`) are passed correctly to kqueue.
* Any events are generated and check if we correctly detect the flag (in `Event::is_lio`).

Thomasdezeeuw:
@carllerche 

> @Thomasdezeeuw I guess we need a general policy of what we support and how.

Yes, most of the arguments for removal also apply to AIO.

asomers:
Well, the only way to test them would pretty much be to copy a simplified version of mio-aio.  I'll look into doing that.

Thomasdezeeuw:
Closing in favour of #1054.

PR #1050, opened 2019-08-05
 Add Interests::and 

Thomasdezeeuw:
This allows adding two interests together, much like the BitOr
implementation, but is a constant function.

I'm not 100% sure on the name, as people might conflict it with `BitAnd`, however `or` didn't make sense either.

### DIFF
--- a/src/interests.rs
+++ b/src/interests.rs
@@ -71,11 +71,12 @@ impl Interests {
     /// ```
     /// use mio::Interests;
     ///
-    /// const INTERESTS: Interests = Interests::READABLE.and(Interests::WRITABLE);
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
I've been thinking but haven't found any good name for the method. My current favourite is `add`.

carllerche:
Let's go w/ `add` then 👍

Thomasdezeeuw:
Clippy is complaining that we should consider implementing `std::ops::Add` instead of adding an `add` method. Maybe `add` isn't the best option then.

carllerche:
We can silence clippy here imo

Issue #1041, opened 2019-07-19
Remaining `wepoll` items.


carllerche:
Meta issue tracking remaining work to do on the `wepoll` windows implementation. This description will be updated over time.

Original PR: #1034. Original issue #1024.

- [x] Dealing with `WSAStartup`, #1046. (solved in #1029)
- [ ] benchmark new implementation vs. old.
- [x] ensure wepoll licensing requirements are met (copy license file).
- [ ] code comment ownership of pointers being passed around the system.
- [ ] ensure memory is not leaked on error [here](https://github.com/tokio-rs/mio/blob/65f19ddcb3ac554c23b0c3fa0665e64ff80141f1/src/sys/windows/afd.rs#L159).
- [ ] Ensure all old Windows related issues are fixed with the new implementation.
- [ ] Ability to add custom IOCP handler #1047.
- [ ] Document interaction with `RawSocket` (do not read / write directly).
- [ ] Concurrent polling & registering, see #1053.
- [ ] Find source of possible polling delay, see https://github.com/tokio-rs/mio/pull/1051#issuecomment-522232178.
- [ ] Fix issues found by additional tests added in #1070:
  - [x] #1073.
  - [ ] #1074.
  - [x] #1078.
  - [x] #1079.
  - [x] #1080.

Thomasdezeeuw:
- [ ] Ensure all old Windows related issues are fixed with the new implementation.

carllerche:
@Thomasdezeeuw I added your item to the OP. Do you have permission to edit the OP directly?

Thomasdezeeuw:
Yes, but didn't know that.

carllerche:
@Thomasdezeeuw no worries 👍

PerfectLaugh:
about memory leak on #1042

PerfectLaugh:
And we probably need to make a notice that the socket can NEVER be used in other place simultaneously with Mio. Because we simulated ET behavior by intercepting all operations that users could do.
If they use the socket in the place where Mio cannot observe (like RawSocket), the polling behavior will be undefined.

piscisaureus:
I would like to add: the current locking scheme is too granular.

It makes little sense to use separate locks for the update_queue and the iocp HANDLE; it's likely to hurt performance and introduce subtle bugs. One example I could find:

https://github.com/tokio-rs/mio/blob/211223e92865e5ae276333cc91673d764bbc386d/src/sys/windows/selector.rs#L368-L370
```rust
[...]

PR #1051, opened 2019-08-06
Remove Arc from Poll and Registry

### LINKED ISSUE #1025: Merging Registry and Poll (again)
In v0.7 `Poll` on unix is just an file descriptor, on the master branch it's split into `Registry` and `Poll` both using `Arc` to share the file descriptor. In my use case I don't need this split as I only use `Poll` from a single location, making the `Arc` unneeded overhead. I propose to remove `Registry` and move back to the structure that was present in v0.7, i.e. a single `Poll`. Making all methods on `Poll` take a reference to self so that the user can use `Arc` and make the split if needed.

Thomasdezeeuw:
This removes the `Arc` from `Poll` and `Registry` and instead used `dup(2)` on Unix to create the `Registry`, on Windows it's clones the inner `Arc<SelectorInner>`. The change forces a single poller, as `Poll::poll` takes `&mut self`, but allows for multiple registrations as `Registry::({re, de})register` takes `&self`.

`Poll` now implements `Deref<Target = Registry>` so that `Poll::register` also works. We however might want to change this into a method, e.g. `Poll::as_register`, or implement it as `AsRef` instead.

@carllerche I think this satisfies all requirements as discussed in #1025, also see the test I've added in 8687ec33663a95aa5d7c29980938f1f78da01c59.

Closes #1025.

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -1,4 +1,3 @@
-use std::ops::Deref;
 #[cfg(unix)]
 use std::os::unix::io::{AsRawFd, RawFd};
 #[cfg(debug_assertions)]
@@ -55,7 +54,7 @@ use crate::{event, sys, Events, Interests, Token};
 /// let stream = TcpStream::connect(server.local_addr()?)?;
 ///
 /// // Register the stream with `Poll`
-/// poll.register(&stream, Token(0), Interests::READABLE | Interests::WRITABLE)?;
+/// poll.registry().register(&stream, Token(0), Interests::READABLE | Interests::WRITABLE)?;
 ///
 /// // Wait for the socket to become ready. This has to happens in a loop to
 /// // handle spurious wakeups.
@@ -149,7 +148,7 @@ use crate::{event, sys, Events, Interests, Token};
 ///
 /// // The connect is not guaranteed to have started until it is registered at
 /// // this point
-/// poll.register(&sock, Token(0), Interests::READABLE | Interests::WRITABLE)?;
+/// poll.registry().register(&sock, Token(0), Interests::READABLE | Interests::WRITABLE)?;
 /// #     Ok(())
 /// # }
 /// ```
@@ -199,15 +198,12 @@ use crate::{event, sys, Events, Interests, Token};
 /// [`Poll::poll`]: struct.Poll.html#method.poll
 #[repr(transparent)]
 pub struct Poll {
-    // `Poll` must have the same memory layout as `Registry`, see
-    // the `Deref` implementation.
-    selector: sys::Selector,
+    registry: Registry,
 }
 
 /// Registers I/O resources.
 #[repr(transparent)]
 pub struct Registry {
-    // `Registry` must have the same memory layout as `Poll`, see `Poll`.
     selector: sys::Selector,
 }
 
@@ -259,15 +255,15 @@ impl Poll {
     /// # }
     /// ```
     pub fn new() -> io::Result<Poll> {
-        sys::Selector::new().map(|selector| Poll { selector })
+        sys::Selector::new().map(|selector| Poll {
+            registry: Registry { selector },
+        })
     }
 
     /// Create a separate `Registry` which can be used to register
     /// `event::Source`s.
-    pub fn registry(&self) -> io::Result<Registry> {
-        self.selector
-            .registry()
-            .map(|selector| Registry { selector })
+    pub fn registry(&self) -> &Registry {
+        &self.registry
     }
 
     /// Wait for readiness events
@@ -341,7 +337,7 @@ impl Poll {
     /// let stream = TcpStream::connect(addr)?;
     ///
     /// // Register the stream with `Poll`
-    /// poll.register(
+    /// poll.registry().register(
     ///     &stream,
     ///     Token(0),
     ///     Interests::READABLE | Interests::WRITABLE)?;
@@ -389,7 +385,7 @@ impl Poll {
         loop {
             let now = Instant::now();
             // First get selector events
-            let res = self.selector.select(events.sys(), timeout);
+            let res = self.registry.selector.select(events.sys(), timeout);
 
             match res {
                 Ok(()) => break,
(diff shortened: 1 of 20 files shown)


Thomasdezeeuw:
Hmmm... Windows doesn't seem to get an event when registering a `TcpListener` and then connecting a `TcpStream` to it in the test I've added. As all other tests still pass I don't think I broke something. Can someone with Windows help me figure out what wrong?

Thomasdezeeuw:
After thinking it over I think polling holds some lock inside `Selector` on Windows that registering also needs, I'll look into the code a bit.

Thomasdezeeuw:
I've created pr #1053 to run the test I've added on Windows. Conclusion is that the test also fails on current master, which means that on master it isn't possible to correctly `poll` and `register` concurrently on Windows.

carllerche:
@Thomasdezeeuw are you saying that the test in #1053 fails to run because it blocks in `poll` preventing the registration from happening?

Is this related to the concurrency setting on the IOCP handle?

src/poll.rs
```diff
@@ -257,19 +259,15 @@ impl Poll {
...
-        Ok(Poll { registry })
+        sys::Selector::new().map(|selector| Poll { selector })
     }
 
-    /// Return a reference to the associated `Registry`.
-    pub fn registry(&self) -> &Registry {
-        &self.registry
+    /// Create a separate `Registry` which can be used to register
+    /// `event::Source`s.
+    pub fn registry(&self) -> io::Result<Registry> {
```
carllerche: It should be possible to keep this function unchanged. `Poll` is a wrapper around `Registry`. Then, we can add a `try_clone(&self) -> io::Result<Registry>` function on `Registry`.

That would allow us to easily switch back to an internal `Arc` if we need to w/o a breaking change (add `Clone` impl to `Registry` and deprecate `try_clone()`).

src/poll.rs
```diff
@@ -418,6 +413,19 @@ impl Poll {
     }
 }
 
+impl Deref for Poll {
```
carllerche: I prefer to reserve `Deref` implementations for "smart pointer" types. I would opt to keep a `fn registry(&self) -> `&Registry` function.

IMO, a `Deref` impl is a pretty heavy guarantee to maintain for forwards compat.

Thomasdezeeuw:
@carllerche I don't know, maybe. As I don't have Windows I really can't tell. I was mainly looking at the code around mutexes, but I haven't seen anything that was clearly locking each other out but I could easily overlook something.

carllerche:
I guess I will have to get my windows VM working again... I have no idea how to do debugging on windows. MSVC?

src/poll.rs
Thomasdezeeuw: I do like that we can do `Poll::register` with this but I'll change it back.

Thomasdezeeuw:
I've just disabled the test on Windows and added an item to #1041.

src/poll.rs
carllerche: You can always add `register`, `reregister`, `deregister` fns directly on `Poll`.

Thomasdezeeuw:
@carllerche it could be the case the IOCP supports concurrent polling & registering and that the problem lays in our use of locks.

src/poll.rs
Thomasdezeeuw: Let's merge this as is and we can always add it in the future.

carllerche:
fine with me

mati865:
> I have no idea how to do debugging on windows. MSVC?

Windows VM should have preinstalled MSVC so you only have to install rust. I have no idea about debugging this target.

Rust Windows GNU has various issues (most of them can be worked around) but you can use GDB and is faster to setup on clean Windows install. You can ping me somewhere if you need help with this target.

PerfectLaugh:
image
Strangely I have un-ignored the failure section you described and there's no error for me.

Thomasdezeeuw:
@PerfectLaugh what happens when you run the tests in a loop? There is a race between the thread that polls and the thread that registers, it could be that in you case the registering threads just happened to win.

PerfectLaugh:
```

**Comment by PerfectLaugh** [comment]
```
#[test]
fn test_registry_behind_arc() {
    // `Registry` should work behind an `Arc`, being `Sync` and `Send`.
    init();

    for _ in 0..1000 {
        let registry = Arc::new(poll.registry().try_clone().unwrap());
        let mut events = Events::with_capacity(128);

        let addr = localhost();
        let barrier = Arc::new(Barrier::new(3));

        let registry2 = Arc::clone(&registry);
        let registry3 = Arc::clone(&registry);
        let barrier2 = Arc::clone(&barrier);
        let barrier3 = Arc::clone(&barrier);

        let handle1 = thread::spawn(move || {
            let listener = TcpListener::bind(addr).unwrap();
            registry2
                .register(&listener, Token(0), Interests::READABLE)
                .unwrap();
            barrier2.wait();
        });
        let handle2 = thread::spawn(move || {
            let stream = TcpStream::connect(addr).unwrap();
            registry3
                .register(&stream, Token(1), Interests::READABLE | Interests::WRITABLE)
                .unwrap();
            barrier3.wait();
        });

        poll.poll(&mut events, None).unwrap();
        assert!(events.iter().count() >= 1);

        // Let the threads return.
        barrier.wait();

        handle1.join().unwrap();
        handle2.join().unwrap();
    }
}
```
image
I ran it 1000 times and there's no error in my code

**Comment by PerfectLaugh** [comment]
However the error only disappear without timeout. If I set the value below around 600ms there will be error at some point.

**Comment by PerfectLaugh** [comment]
```
#[test]
fn test_registry_behind_arc() {
    use std::time::Instant;
    // `Registry` should work behind an `Arc`, being `Sync` and `Send`.
    init();

    for n in 0..1000 {
        let registry = Arc::new(poll.registry().try_clone().unwrap());
        let mut events = Events::with_capacity(128);

        let addr = localhost();
        let barrier = Arc::new(Barrier::new(3));

        let registry2 = Arc::clone(&registry);
        let registry3 = Arc::clone(&registry);
        let barrier2 = Arc::clone(&barrier);
        let barrier3 = Arc::clone(&barrier);

        let handle1 = thread::spawn(move || {
            let listener = TcpListener::bind(addr).unwrap();
            registry2
                .register(&listener, Token(0), Interests::READABLE)
                .unwrap();
            barrier2.wait();
        });
        let handle2 = thread::spawn(move || {
            let stream = TcpStream::connect(addr).unwrap();
            registry3
                .register(&stream, Token(1), Interests::READABLE | Interests::WRITABLE)
                .unwrap();
            barrier3.wait();
        });

        println!("{}", n);

        let now = Instant::now();
        poll.poll(&mut events, None).unwrap();
        println!("poll: {}", now.elapsed().as_millis());
        assert!(events.iter().count() >= 1);

        // Let the threads return.
        barrier.wait();

        handle1.join().unwrap();
        handle2.join().unwrap();
    }
}
```
With this code I found some strange behavior in Windows.
image
There are strange periodical 500ms delays randomly somehow.

carllerche:
Could discussion move to a tracking issue (and reference this) Otherwise, we will lose the convo 👍

Thomasdezeeuw:
I've added it to #1041.

PR #1055, opened 2019-08-17
Fix for windows selector locking mechanism

PerfectLaugh:
Refer to https://github.com/tokio-rs/mio/issues/1041#issuecomment-519674615

### DIFF
--- a/src/sys/windows/event.rs
+++ b/src/sys/windows/event.rs
@@ -14,10 +14,10 @@ pub fn token(event: &Event) -> Token {
 }
 
 pub fn is_readable(event: &Event) -> bool {
-    if is_error(event) {
+    if is_error(event) || is_read_hup(event) {
         return true;
     }
-    event.flags & (afd::AFD_POLL_RECEIVE | afd::AFD_POLL_ACCEPT | afd::AFD_POLL_DISCONNECT) != 0
+    event.flags & (afd::AFD_POLL_RECEIVE | afd::AFD_POLL_ACCEPT) != 0
 }
 
 pub fn is_writable(event: &Event) -> bool {
@@ -35,9 +35,8 @@ pub fn is_hup(event: &Event) -> bool {
     event.flags & afd::AFD_POLL_ABORT != 0
 }
 
-pub fn is_read_hup(_: &Event) -> bool {
-    // Not supported.
-    false
+pub fn is_read_hup(event: &Event) -> bool {
+    event.flags & afd::AFD_POLL_DISCONNECT != 0
 }
 
 pub fn is_priority(event: &Event) -> bool {
--- a/src/sys/windows/event.rs
+++ b/src/sys/windows/event.rs
@@ -14,25 +14,25 @@ pub fn token(event: &Event) -> Token {
 }
 
 pub fn is_readable(event: &Event) -> bool {
-    if is_hup(event) {
+    if is_error(event) {
         return true;
     }
-    (event.flags & (afd::KNOWN_AFD_EVENTS & !afd::AFD_POLL_SEND)) != 0
+    event.flags & (afd::AFD_POLL_RECEIVE | afd::AFD_POLL_ACCEPT | afd::AFD_POLL_DISCONNECT) != 0
 }
 
 pub fn is_writable(event: &Event) -> bool {
-    if is_hup(event) {
+    if is_error(event) {
         return true;
     }
-    (event.flags & (afd::AFD_POLL_SEND)) != 0
+    event.flags & afd::AFD_POLL_SEND != 0
 }
 
 pub fn is_error(event: &Event) -> bool {
-    event.flags == afd::AFD_POLL_CONNECT_FAIL
+    event.flags & afd::AFD_POLL_CONNECT_FAIL != 0
 }
 
 pub fn is_hup(event: &Event) -> bool {
-    (event.flags & (afd::AFD_POLL_ABORT | afd::AFD_POLL_CONNECT_FAIL)) != 0
+    event.flags & afd::AFD_POLL_ABORT != 0
 }
 
 pub fn is_read_hup(_: &Event) -> bool {
@@ -41,7 +41,7 @@ pub fn is_read_hup(_: &Event) -> bool {
 }
 
 pub fn is_priority(event: &Event) -> bool {
-    (event.flags & afd::AFD_POLL_RECEIVE_EXPEDITED) != 0
+    event.flags & afd::AFD_POLL_RECEIVE_EXPEDITED != 0
 }
 
 pub fn is_aio(_: &Event) -> bool {
--- a/src/sys/windows/selector.rs
+++ b/src/sys/windows/selector.rs
@@ -1,11 +1,14 @@
+use std::cell::RefCell;
 use std::collections::VecDeque;
 use std::io;
 use std::mem::size_of;
 use std::pin::Pin;
 use std::ptr::null_mut;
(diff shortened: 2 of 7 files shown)


PerfectLaugh:
Sorry I was too hurry.

src/sys/windows/selector.rs
```diff
@@ -351,42 +354,55 @@ impl Selector {
...
     cp: CompletionPort,
-    active_poll_count: AtomicUsize,
-    update_queue: Mutex<VecDeque<Arc<Mutex<SockState>>>>,
+    active_poll_count: RefCell<usize>,
+    update_queue: RefCell<VecDeque<Arc<Mutex<SockState>>>>,
     afd_group: Mutex<Vec<Arc<Afd>>>,
 }
 
+// We have ensured thread safety by introducing lock manually.
+unsafe impl Sync for SelectorInner {}
```
Thomasdezeeuw: You can't implement `Sync` when using `RefCell`, you must use `UnsafeCell` and ensure that only a single thread can access it any given time.

src/sys/windows/event.rs
```diff
@@ -14,25 +14,25 @@ pub fn token(event: &Event) -> Token {
 }
 
 pub fn is_readable(event: &Event) -> bool {
-    if is_hup(event) {
+    if is_error(event) {
```
Thomasdezeeuw: Why are you returning `true` if there is an error? Same for `is_writable`.

src/sys/windows/event.rs
```diff
@@ -14,25 +14,25 @@ pub fn token(event: &Event) -> Token {
...
         return true;
     }
-    (event.flags & (afd::KNOWN_AFD_EVENTS & !afd::AFD_POLL_SEND)) != 0
+    event.flags & (afd::AFD_POLL_RECEIVE | afd::AFD_POLL_ACCEPT | afd::AFD_POLL_DISCONNECT) != 0
```
Thomasdezeeuw: When does `AFD_POLL_DISCONNECT` trigger? Because libuv seems to think this match `EV_EOF`/`EPOLLRDHUP` (see https://github.com/libuv/libuv/commit/5ad6a9104e3a81934176bcda780945fae7eb9550) and we're using those in `is_read_hup`.

Thomasdezeeuw: I also have some more questions about the other flags, but I can't find any documentation on Windows' AFD.

tests/poll.rs
```diff
@@ -112,8 +111,7 @@ fn test_registry_behind_arc() {
         barrier3.wait();
     });
 
-    poll.poll(&mut events, Some(Duration::from_millis(100)))
-        .unwrap();
+    poll.poll(&mut events, None).unwrap();
```
Thomasdezeeuw: Can keep the timeout, if needed you can increase the timeout (max 1 second or so), I don't tests to get stuck.

src/sys/windows/selector.rs
```diff
@@ -449,129 +569,92 @@ impl SelectorInner {
...
-        let mut need_alloc = false;
-        {
-            let afd_group = self.afd_group.lock().unwrap();
-            if afd_group.len() == 0 {
-                need_alloc = true;
-            } else {
-                // + 1 reference in Vec
-                if Arc::strong_count(afd_group.last().unwrap()) >= POLL_GROUP__MAX_GROUP_SIZE + 1 {
-                    need_alloc = true;
+    // The ret value here means processed iocp_events not events itself.
```
Thomasdezeeuw: Can you replace this comment with it returns, i.e. Returns number of IOCP events processed (if that is correct).

Thomasdezeeuw: Also I don't see any path that doesn't process all `iocp_events`, so I don't really the point of the return value (I could be wrong though).

src/sys/windows/selector.rs
```diff
@@ -351,43 +400,104 @@ impl Selector {
...
+
+        {
+            let _guard = self.lock.lock().unwrap();
+
+            unsafe {
+                *self.active_poll_count.get() -= 1;
+            }
+
+            if let Err(e) = result {
+                use winapi::shared::winerror::WAIT_TIMEOUT;
```
Thomasdezeeuw: Can you move this to the other imports at the top of the file.

src/sys/windows/selector.rs
```diff
@@ -351,43 +400,104 @@ impl Selector {
...
+                }
+                return Ok(());
+            } else {
+                if n >= events.statuses.len() {
+                    return Ok(());
+                }
+                let deadline = deadline.unwrap();
+                let now = Instant::now();
+                if now >= deadline {
+                    return Ok(());
```
Thomasdezeeuw: This behaviour is different compared to the Unix selectors. Here we return immediately if a timeout `Duration::from_secs(0)` would be passed, on unix we still do the system call. @carllerche you ok with the difference? I think it could prevent a use case were you don't want to block but still want to poll events and thus pass a zero duration (that is not none).

src/sys/windows/event.rs
carllerche: I would expect this to be correct. An error would imply that the socket is readable / writable and will return an error on that operation.

src/sys/windows/event.rs
carllerche: summoning @piscisaureus, the AFD expert.

src/sys/windows/selector.rs
carllerche: We still need to do the syscall, just not block. We often call `poll` w/ duration 0 from Tokio because we want to fetch any pending IO events but don't want to block (we still have other work to do).

carllerche: PS, Nice catch @Thomasdezeeuw :+1:

carllerche:
Thanks for doing all this work :+1:. @Thomasdezeeuw, whenever you are happy w/ this, feel free to merge.

src/sys/windows/selector.rs
PerfectLaugh: Since Windows expected iocp events arrays are initialized with zeros, I think we can't just use the previous element.

src/sys/windows/selector.rs
Thomasdezeeuw: I think this is solved, right @PerfectLaugh?

PerfectLaugh: I didn't make the test for Some(Duration::from_secs(0)), so I don't know whether the behavior will be same as unix either.

Thomasdezeeuw: I've created a test in #1065.

Thomasdezeeuw:
Please don't include the last two commit, this review has been long enough lets not any more code to it.

PerfectLaugh:
> Please don't include the last two commit, this review has been long enough lets not any more code to it.

Sorry I didn't use git quite well. I just make the first branch in my life :(

src/sys/windows/event.rs
PerfectLaugh: I think libuv is correct in this case, since libuv is battle-tested library.

Thomasdezeeuw:
> Sorry I didn't use git quite well. I just make the first branch in my life :(

That's alright, git isn't the easiest thing to use and we all needed to learn it at some point. For a quick into you can look at https://learnxinyminutes.com/docs/git/, or if you have some more time & interest you can read [Pro git](https://git-scm.com/book/en/v2).

Back to the code, I'm going to merge this.

PR #1057, opened 2019-08-19
Add more CI targets for cross compilation

Thomasdezeeuw:
Continuing of #952.

### DIFF
--- a/ci/azure-cross-compile.yml
+++ b/ci/azure-cross-compile.yml
@@ -22,9 +22,13 @@ jobs:
         vmImage: ubuntu-16.04
         target: arm-linux-androideabi
 
-      Android_64:
+      Android_ARM64:
         vmImage: ubuntu-16.04
         target: aarch64-linux-android
+
+      Android_32:
+        vmImage: ubuntu-16.04
+        target: i686-unknown-linux-gnu
 
       NetBSD:
         vmImage: ubuntu-16.04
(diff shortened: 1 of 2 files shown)


ci/azure-cross-compile.yml
```diff
@@ -34,14 +38,6 @@ jobs:
         vmImage: ubuntu-16.04
         target: x86_64-sun-solaris
 
-      Linux_ARM:
-        vmImage: ubuntu-16.04
-        target: arm-linux-androideabi
```
Thomasdezeeuw: This was the same as the `Android` target (see above) so I removed it.

ci/azure-cross-compile.yml
```diff
@@ -37,7 +53,7 @@ jobs:
     - script: rustup target add $(target)
       displayName: "Add target"
 
-    - script: cargo check --target $(target)
+    - script: cargo build --target $(target)
```
asomers: Why change `cargo check` to `cargo build`?  Won't that just slow things down for little benefit?

Thomasdezeeuw: That was in the original pr, see https://github.com/tokio-rs/mio/pull/952/files#diff-424edf9074d0455636ef5bf26a0c9cd4R56. @carllerche any reason why you changed it?

carllerche: I'm fine either way. So up to y'all.

carllerche: I don't remember why I changed it

Thomasdezeeuw: I'll change it back, because I do think its faster.

PR #878, opened 2018-10-05
Add minimal versions test to CI

cramertj:

PR #1058, opened 2019-08-19
Test with minimal versions on CI

Thomasdezeeuw:
Continuing #878.

/cc @cramertj

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -36,6 +36,10 @@ jobs:
     cmd: check
     cross: true
 
+- template: ci/azure-minimal-versions.yml
+  parameters:
+    name: minimal_versions
+
 - template: ci/azure-clippy.yml
   parameters:
     name: clippy
--- a/ci/azure-minimal-versions.yml
+++ b/ci/azure-minimal-versions.yml
@@ -1,0 +1,37 @@
+parameters:
+  rust_version: nightly
+
+jobs:
+- job: ${{ parameters.name }}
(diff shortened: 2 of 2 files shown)


Thomasdezeeuw:
Well it worked and we found the first problem.

asomers:
Sadly `-Zminimal-versions` is nearly useless unless all of your dependencies (and all of their dependencies) also use it in CI.  It probably won't be possible for most crates to use it until a `-Zminimal-versions-for-me-but-not-my-dependencies` gets added.  See https://github.com/rust-lang/cargo/issues/5657 .

If you merely do `cargo check` instead of `cargo test` with minimal versions, does that pass?

Thomasdezeeuw:
I found the problem, it's `bytes` that uses `iovec` v0.1.0 which uses `winapi` v0.2. But we only use `bytes` for testing, so `cargo check` works fine (assuming you don't add `--tests`).

Thomasdezeeuw:
It seems `bytes` v0.5 (unreleased) doesn't have `iovec` as a dependency any more, which would solve this problem, but I don't know when that would be released.

carllerche:
@Thomasdezeeuw simultaneously w/ mio probably... we can switch to a git dependency on bytes master.

Thomasdezeeuw:
Ok, will do.

Thomasdezeeuw:
Created #1059.

Thomasdezeeuw:
CI is green.

Issue #1035, opened 2019-07-16
Testing in release mode on CI?


Thomasdezeeuw:
Following https://github.com/tokio-rs/mio/pull/1034#issuecomment-511660944 and #1004 maybe it's worth to test in release mode on the CI.

Thomasdezeeuw:
Done in #1067.

PR #1067, opened 2019-08-27
misc cleanup

carllerche:
* split up TCP files into stream / listener
* organize import order
* split up misc other files

### DIFF
--- a/tests/size.rs
+++ b/tests/size.rs
@@ -1,9 +1,8 @@
-use mio::net::*;
-
 #[test]
 #[cfg(unix)]
 #[cfg(not(debug_assertions))]
 fn assert_size() {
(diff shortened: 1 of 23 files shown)


tests/size.rs
```diff
@@ -0,0 +1,13 @@
+#[test]
+#[cfg(not(debug_assertions))]
+fn assert_size() {
+    use std::mem::size_of;
+
+    use crate::sys;
```
Thomasdezeeuw: This shouldn't be possible as `sys` is not public.

Thomasdezeeuw: This doesn't get tested on the CI since it always has debug assertions, so can you can check with `cargo test --release`.

Related #1035.

PR #1068, opened 2019-08-29
Add TcpListener::bind_with_options

Thomasdezeeuw:
The allows the user set options before the socket is bound.

I would like some feedback on whether this should be in Mio, or inot. I know could create the socket myself (outside of Mio) and then convert into a `TcpListener`, but I only need to make a single external call I don't really want to do all of it myself using `libc` or use `net2`/`socket2` for a essentially a single call.

### DIFF
--- a/src/net/tcp_listener.rs
+++ b/src/net/tcp_listener.rs
@@ -8,6 +8,8 @@ use std::io;
 use std::net::SocketAddr;
 #[cfg(unix)]
 use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
+#[cfg(windows)]
+use std::os::windows::raw::SOCKET as RawFd;
 
 /// A structure representing a socket server
 ///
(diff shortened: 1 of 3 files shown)


carllerche:
I feel your pain. I agree the situation with custom socket options is not ideal. I also hesitate to add APIs that could possibly be external. Of course, where do we draw the line between what is included and what isn't?

My rule of thumb has been:

- Minimum set of APIs to expose OS selector (epoll, kqueue, ...) functionality in a portable way.
- Socket impls that exist in `std` (tcp, udp, uds now) and to just mirror `std` for the APIs.

IMO, any function on our socket types that is not in `std` should, at this point, have a high bar for addition.

There also needs to be a good way to customize sockets while reducing syscalls.

I see a few way forward:

- We provide a `mio-util` crate that includes APIs for customizing socket options.
- We add a `util` module in mio that is behind a feature flag w/ these customization APIs and we have a big warning saying that this module is _not_ stable and could break in `1.x` releases.
- We work on stabilizing an API for customizing socket options, which seems difficult given the history of attempts (net2, socket2) and no clear consensus.

carllerche:
We could add a `TcpSocket` struct that lets you configure the socket, then have `bind` and `connect` fns on it....

carllerche:
@Thomasdezeeuw can you describe the specific use case you are trying to solve? What options are you trying to set?

Thomasdezeeuw:
I basically need what is in the example, setting `SO_REUSEPORT` before binding the socket. I rather not add a `TcpSocket` type because at that point we might as well point people to `socket2`/`net2`, no point in duplicating the effort.

Thomasdezeeuw:
Closing this for now, I haven't really found a good alternative but I currently don't think this is the way to go.

Thomasdezeeuw:
I have though about `TcpSocket` but I don't think is a good idea to put this into Mio. I think once we add `TcpSocket` people might also want `UdpSocket`, and then a bunch of options and we basically have a copy of socket2/net2. I don't its worth it to duplicate the effort, so instead we can use socket2/net2 for this without adding more code that isn't necessary to Mio.

PR #1070, opened 2019-08-29
Port gaea tests

Thomasdezeeuw:
Still have to port the tests for `TcpStream` and `TcpListener`.

### DIFF
--- a/tests/poll.rs
+++ b/tests/poll.rs
@@ -1,9 +1,11 @@
 use mio::net::{TcpListener, TcpStream};
-use mio::*;
+use mio::{event, Events, Interests, Poll, Registry, Token};
+
 use std::net;
-use std::sync::{Arc, Barrier};
+use std::sync::{Arc, Barrier, Mutex};
 use std::thread::{self, sleep};
 use std::time::Duration;
+use std::{fmt, io};
 
 mod util;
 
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw:
Note that there is some overlap in the tests for `UdpSocket` now, if I have some time I'll remove the unneeded tests.

tests/tcp_listener.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    // Expect a single connection.
+    let (mut stream, peer_address) = listener.accept().expect("unable to accept connection");
+    assert!(peer_address.ip().is_loopback());
+    assert_eq!(stream.peer_addr().unwrap(), peer_address);
+    assert_eq!(stream.local_addr().unwrap(), address);
+
+    // Expect the stream to be non-blocking.
+    let mut buf = [0; 20];
+    assert_would_block(stream.read(&mut buf));
```
Thomasdezeeuw: @carllerche here's a test that ensure the stream to be non-blocking, per what you mentioned on Gitter.

Thomasdezeeuw:
It seems Linux and macOS handles shutting down the reading side differently. Linux still allows the remaining buffer to be read, while macOS doesn't.

carllerche:
Shutting down the reading side is pretty much unspecified and has no meaning at the TCP level. I wouldn’t bother testing it.

carllerche:
Ideally we can get this merged sooner than later as it has the non-blocking fixes.

Thomasdezeeuw:
@carllerche sorry for the delay, I haven't had much time recently to work on this. As for the failures; there are a number of network issues which should resolves itself. The only actual failures are on Windows the `tcp_stream_ipv4` and `tcp_stream_ipv6` tests fails because they don't receive an expected event here: https://github.com/tokio-rs/mio/pull/1070/files#diff-c448a2bd877d650a0a8a8c0cdb0dba95R50. This looks like a bug in the Windows implementation to me.

I could ignore the tests for now on Windows so we can merge this and open an issue for it to fix it later.

Thomasdezeeuw:
I think something strange is going in the `nodelay` test failure: https://dev.azure.com/tokio-rs/Tokio/_build/results?buildId=2295&view=logs&jobId=429a383f-e486-57ac-3982-9ac65641a32a&taskId=1f926ffb-0ec7-5f9b-ab05-76fbb04597a8&lineStart=168&lineEnd=168&colStart=1&colEnd=7, error message: `panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 10022, kind: InvalidInput, message: "An invalid argument was supplied." }'`. It passed in the previous commits and no changes were made, further more the argument can't be invalid as only `true` or `false` are possible and std lib takes cares of the rest.

Thomasdezeeuw:
I've opened some issues because some tests are failing on Windows, #1073, #1074, #1078 and #1079, in an attempt to get this merged.

Thomasdezeeuw:
@carllerche CI finally green, I've opened a number of issue to investigate why the tests fail on Windows, but I currently don't have the time to investigate. In any case the is ready for a review.

PR #1087, opened 2019-09-16
Check EV_EOF in Event::is_hup on kqueue platforms

Thomasdezeeuw:
This matches checking for `EPOLLHUP` done on `epoll(2)` platforms for Unix pipes. I don't know if this needs a tests, but I got one in another crate.

### DIFF
--- a/src/sys/unix/kqueue.rs
+++ b/src/sys/unix/kqueue.rs
@@ -341,9 +341,8 @@ pub mod event {
             (event.flags & libc::EV_EOF) != 0 && event.fflags != 0
     }
 
-    pub fn is_hup(_: &Event) -> bool {
-        // Not supported.
-        false
+    pub fn is_hup(event: &Event) -> bool {
+        event.filter == libc::EVFILT_WRITE && (event.flags & libc::EV_EOF) != 0
     }
 
     pub fn is_read_hup(event: &Event) -> bool {
--- a/src/event/event.rs
+++ b/src/event/event.rs
@@ -79,7 +79,7 @@ impl Event {
     /// | [OS selector] | Flag(s) checked |
     /// |---------------|-----------------|
     /// | [epoll]       | `EPOLLHUP`      |
-    /// | [kqueue]      | Not supported   |
+    /// | [kqueue]      | `EV_EOF`        |
     ///
     /// [OS selector]: ../struct.Poll.html#implementation-notes
     /// [epoll]: http://man7.org/linux/man-pages/man7/epoll.7.html


Thomasdezeeuw:
The CI failure seems unrelated. I've got a test here https://github.com/Thomasdezeeuw/mio-pipe/blob/4596b93d43d47d1b27d278244991ba8efa0a7fc4/tests/pipe.rs#L84, `is_error` is true on epoll and `is_hup` is true on kqueue (with this change).

src/sys/unix/kqueue.rs
```diff
@@ -341,9 +341,8 @@ pub mod event {
             (event.flags & libc::EV_EOF) != 0 && event.fflags != 0
     }
 
-    pub fn is_hup(_: &Event) -> bool {
-        // Not supported.
-        false
+    pub fn is_hup(event: &Event) -> bool {
+        event.filter == libc::EVFILT_WRITE && (event.flags & libc::EV_EOF) != 0
```
asomers: Why restrict it to `EVFILT_WRITE`?  It's relevant for reading, too.

Thomasdezeeuw: We have `is_read_hup` for that, which is the same check but using `EVFILT_READ`.

Thomasdezeeuw:
@asomers true I think only `is_writable` and `is_readable` are currently properly tested.

Thomasdezeeuw:
@carllerche you ok with this change?

carllerche:
Sorry for the delay (was on all day meetings), but this must be reverted. There is no correct way to detect “hup” with kqueue the way epoll does.

We had this issue in 0.6. I can find the relevant PRs but am on mobile atm. 

Kqueue can only reliably detect read_hup

Thomasdezeeuw:
We already have documentation that `Event::is_hup` provided on a best-effort bases ([here](https://github.com/tokio-rs/mio/blob/ebc85b153b63f3d31c24d9b4402206fa1d7398f4/src/event/event.rs#L69-L70)): 

> Method is available on all platforms, but not all platforms trigger the HUP event.

This method now at least works for Unix pipes.

src/sys/unix/kqueue.rs
carllerche: This must be reverted. The problem is that kqueue will fire this event in unexpected scenarios. 

There is an older PR to 0.6 with more info, but IIRC, kqueue will fire this event when some APIs are called in process. For example, if the user calls shutdown(write) this event is triggered even though the remote could send more data. 

To correctly simulate epoll HUP we would need to track state indicating that the socket has previously seen a read hup and only fire HUP when both the read and the write filters have gotten EOF

carllerche:
is_hup only has value if there is a guarantee of no false positives. Otherwise, the caller can take no action based on it.

carllerche:
If we want to expose this capability, we would need to explicitly add is_write_hup or something.

carllerche:
is_hup signifies closure in both directions. 

Another thought, we may want to use another name than “hup”. EOF, done, complete, ... unsure.

Thomasdezeeuw:
The pr that removed in v0.6 is https://github.com/tokio-rs/mio/pull/958 I think, but I don't see any references to the issues. I'm ok with dropping hup from the name and move to something different.

But I do really need this check in some way. For Unix pipe closing the receiving end will cause an `EPOLLERR` on Linux (which makes `is_error` return true) and on kqueue platforms `EV_EOF` is set.

Thomasdezeeuw:
Maybe we should make some list of scenarios in which what flags (of `kqueue`/`epoll`) are triggered and determine how we can (somewhat) uniformly we can expose that via `Event`.

Thomasdezeeuw:
I've started on this in #1091.

PR #1095, opened 2019-09-25
Use FD_CLOEXEC in call to fcntl over

### LINKED ISSUE #1092: Waker pipe sets CLOEXEC with F_SETFL when pipe2 is unavailable.
Version: 0.6.19

sys/unix/mod.rs
```rust
    // Use pipe2 for atomically setting O_CLOEXEC if we can, but otherwise
    // just fall back to using `pipe`.
    dlsym!(fn pipe2(*mut c_int, c_int) -> c_int);
```

However then mio sets flags using F_SETFL, which result in CLOEXEC missed.

```
epoll_create1(0)                        = 3
fcntl(3, F_GETFD)                       = 0
fcntl(3, F_SETFD, FD_CLOEXEC)           = 0
pipe([4, 5])                            = 0
fcntl(4, F_SETFL, O_RDONLY|O_NONBLOCK|O_CLOEXEC) = 0
fcntl(5, F_SETFL, O_RDONLY|O_NONBLOCK|O_CLOEXEC) = 0
...
fcntl(3, F_GETFD)                       = 0x1 (flags FD_CLOEXEC)
fcntl(4, F_GETFD)                       = 0
fcntl(5, F_GETFD)                       = 0
```
epoll fd has correct flag, pipe fds will be inherited after exec

(seems to be not an issue in 0.7 due to use of eventfd)

Thomasdezeeuw:
O_CLOEXEC is not accepted in fnctl, but no OS complains about it.

Closes #1092 

/cc @reiver-dev

### DIFF
--- a/src/sys/unix/mod.rs
+++ b/src/sys/unix/mod.rs
@@ -55,19 +55,21 @@ pub fn pipe() -> ::io::Result<(Io, Io)> {
             Some(pipe2_fn) => {
                 let flags = libc::O_NONBLOCK | libc::O_CLOEXEC;
                 cvt(pipe2_fn(pipes.as_mut_ptr(), flags))?;
+                Ok((Io::from_raw_fd(pipes[0]), Io::from_raw_fd(pipes[1])))
             }
             None => {
                 cvt(libc::pipe(pipes.as_mut_ptr()))?;
+                // Ensure the pipe are closed if any of the system calls below
+                // fail.
+                let r = Io::from_raw_fd(pipes[0]);
+                let w = Io::from_raw_fd(pipes[1]);
                 cvt(libc::fcntl(pipes[0], libc::F_SETFD, libc::FD_CLOEXEC))?;
                 cvt(libc::fcntl(pipes[1], libc::F_SETFD, libc::FD_CLOEXEC))?;
                 cvt(libc::fcntl(pipes[0], libc::F_SETFL, libc::O_NONBLOCK))?;
                 cvt(libc::fcntl(pipes[1], libc::F_SETFL, libc::O_NONBLOCK))?;
+                Ok((r, w))
             }
         }
-    }
-
-    unsafe {
-        Ok((Io::from_raw_fd(pipes[0]), Io::from_raw_fd(pipes[1])))
     }
 }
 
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
CI fails due to warnings on newer versions of Rust.

reiver-dev:
This is what musl is doing:
[https://github.com/esmil/musl/blob/master/src/unistd/pipe2.c](pipe2.c)

Here FD_CLOEXEC is used with F_SETFD and O_NONBLOCK with F_SETFL. I think we need to mimic musl implementation here and call fcntl twice.

Thomasdezeeuw:
@reiver-dev I think you're right I've updated the code, can you check that everything works now?

reiver-dev:
Checked you branch:
```
epoll_create1(0)                        = 3
fcntl(3, F_GETFD)                       = 0
fcntl(3, F_SETFD, FD_CLOEXEC)           = 0
pipe([4, 5])                            = 0
fcntl(4, F_SETFD, FD_CLOEXEC)           = 0
fcntl(5, F_SETFD, FD_CLOEXEC)           = 0
fcntl(4, F_SETFL, O_RDONLY|O_NONBLOCK)  = 0
fcntl(5, F_SETFL, O_RDONLY|O_NONBLOCK)  = 0

...

fcntl(3, F_GETFD)                       = 0x1 (flags FD_CLOEXEC)
fcntl(4, F_GETFD)                       = 0x1 (flags FD_CLOEXEC)
fcntl(5, F_GETFD)                       = 0x1 (flags FD_CLOEXEC)
```
Checking with ```ls -la /proc/self/fd```

Pipes not inherited
```
0 -> /dev/pts/0
1 -> /dev/pts/0
2 -> /dev/pts/0
3 -> /proc/18147/fd
```

Prevously the result was
```
0 -> /dev/pts/0
1 -> /dev/pts/0
2 -> /dev/pts/0
3 -> /proc/18784/fd
4 -> 'pipe:[61348]'
5 -> 'pipe:[61348]'
```

Thomasdezeeuw:
Then I think this is ready to merged (despite the CI failures), cc @carllerche. Or should I send another pr silencing the warnings?

reiver-dev:
It seems that if fcntl fails, pipe fds will be leaked. This should never happen, however. I think we should either ignore the fcntl error or close fds.

carllerche:
Ideally we could fix in a separate PR before force merging this.

Thomasdezeeuw:
I've created #1097 to silence the warnings on the CI and pushed a commit to fix the leaking of file descriptors.

Thomasdezeeuw:
@carllerche CI is green now.

PR #1097, opened 2019-10-01
Silence warnings on newer Rustc versions

Thomasdezeeuw:
This is a minimal commit to silence warnings on newer compiler versions,
while we're still supporting old compilers in v0.6. This does not
include any function changes.

For #1095.

### DIFF
--- a/benches/bench_poll.rs
+++ b/benches/bench_poll.rs
@@ -1,4 +1,5 @@
 #![feature(test)]
+#![allow(deprecated)]
 
 extern crate mio;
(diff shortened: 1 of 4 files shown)


Thomasdezeeuw:
The CI is failing on `test_tcp_shutdown::test_abrupt_shutdown` and `test_tcp_shutdown::test_graceful_shutdown`, both seem like legit failures. But I don't have the time to look into them right now.

Thomasdezeeuw:
Superseded by #1099.

PR #1098, opened 2019-10-01
Add UDS support for Unix systems

kleimkuhler:
### Summary

This change adds UDS support for Unix systems. The networking primitives
`src/net/uds`. These modules provide the structs for `UnixDatagram`s,
`UnixListener`s, and `UnixStream`s. Each of the methods on these structs
dispatch to the Unix system structs in `src/sys/unix/uds.rs`.

Use of these new structs should feel fairly similar to the `Tcp[...]`
equivalents. The tests added also reflect the similar behavior that would be
expected.

### Details

Since this change only adds UDS support, there should be no behavior changes
anywhere else. For reviewing, I recommend starting from the networking
primitives and seeing what methods they dispatch to.

The tests added in `tests/uds.rs` should feel fairly similar to
`tcp_stream.rs`--the main difference coming from supporting `PathBuf`s as the
socket address to connect to.

I originally pulled in some testing infrastructure added to `v0.6.x` in #944,
but ultimately found the functionality to not be as helpful due to a lot of
changes that have since taken place in `master`. I think #944 could still be
helpful, but as a separate change that is more about improving testing
functionality and not included in a change that introduces additional
functionality.


src/net/uds/datagram.rs
```diff
@@ -0,0 +1,153 @@
...
+        self.std.connect(path)
+    }
+
+    /// Creates a new independently owned handle to the underlying socket.
+    ///
+    /// The returned `UnixListener` is a reference to the same socket that this
+    /// object references. Both handles can be used to accept incoming
+    /// connections and options set on one listener will affect the other.
+    pub fn try_clone(&self) -> io::Result<UnixDatagram> {
+        self.std.try_clone().map(|std| UnixDatagram { std })
```
hawkw: Nit/TIOLI: wdyt of 
```suggestion
        self.std.try_clone().map(Self::from_std)
```

src/sys/unix/uds.rs
```diff
@@ -0,0 +1,298 @@
...
+use std::cmp::Ordering;
+use std::ffi::OsStr;
+use std::fmt;
+use std::io;
+use std::mem;
+use std::os::unix::net::{UnixDatagram, UnixListener, UnixStream};
+use std::os::unix::prelude::*;
+use std::path::Path;
+
+/// An address associated with a mio specific unix socket.
```
hawkw: nit:
```suggestion
/// An address associated with a `mio`-specific Unix socket.
```

asomers:
How does this PR compare to alexchrichton/mio-uds ?

src/sys/unix/uds.rs
```diff
@@ -0,0 +1,298 @@
...
+enum AddressKind<'a> {
+    Unnamed,
+    Pathname(&'a Path),
+    Abstract(&'a [u8]),
+}
+
+pub fn connect_stream(path: &Path) -> io::Result<UnixStream> {
+    let socket = new_socket(libc::AF_UNIX, libc::SOCK_STREAM)?;
+    let (raw_addr, raw_addr_length) = socket_addr(path)?;
+    let raw_addr = &raw_addr as *const _ as *const _;
```
hawkw: Why is the second cast to `*const _` necessary?

src/sys/unix/uds.rs
```diff
@@ -0,0 +1,298 @@
...
+            UnixStream::from_raw_fd(fds[0]),
+            UnixStream::from_raw_fd(fds[1]),
+        )
+    })
+}
+
+pub fn bind_datagram(path: &Path) -> io::Result<UnixDatagram> {
+    let socket = new_socket(libc::AF_UNIX, libc::SOCK_DGRAM)?;
+    let (raw_addr, raw_addr_length) = socket_addr(path)?;
+    let raw_addr = &raw_addr as *const _ as *const _;
```
hawkw: Again, if the second `as *const _` is required, it would be good to have a comment explaining why; this seems weird.

src/sys/unix/uds.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    syscall!(bind(socket, raw_addr, raw_addr_length))?;
+    Ok(unsafe { UnixDatagram::from_raw_fd(socket) })
+}
+
+// pub fn accept(listener: &UnixListener) -> io::Result<Option<(UnixStream, SocketAddr)>> {
```
hawkw: Nit: rm commented out code.

src/sys/unix/uds.rs
```diff
@@ -0,0 +1,298 @@
...
+    #[cfg(not(any(
+        target_os = "ios",
+        target_os = "macos",
+        target_os = "netbsd",
+        target_os = "solaris"
+    )))]
+    let sock_addr = {
+        let flags = libc::SOCK_NONBLOCK | libc::SOCK_CLOEXEC;
+
+        match syscall!(accept4(listener.as_raw_fd(), raw_storage, &mut len, flags)) {
```
hawkw: nit: why not just 
```rust
let sock_addr = {
    let flags =...;
    syscall!(accept4(listener.as_raw_fd(), raw_storage, &mut len, flags))?
};
```
rather than the match?

src/sys/unix/uds.rs
```diff
@@ -0,0 +1,298 @@
...
+        }
+    };
+
+    #[cfg(any(
+        target_os = "ios",
+        target_os = "macos",
+        target_os = "netbsd",
+        target_os = "solaris"
+    ))]
+    let sock_addr = match syscall!(accept(listener.as_raw_fd(), raw_storage, &mut len)) {
```
hawkw: similarly, can probably use `?` here?

src/sys/unix/uds.rs
```diff
@@ -0,0 +1,298 @@
...
+            // Close the socket if we hit an error, ignoring the error from
+            // closing since we can't pass back two errors.
+            let _ = unsafe { libc::close(socket) };
+            err
+        })
+        .map(|_| unsafe { UnixListener::from_raw_fd(socket) })
+}
+
+pub fn socket_addr(path: &Path) -> io::Result<(libc::sockaddr_un, libc::socklen_t)> {
+    unsafe {
```
hawkw: this is a pretty big wad of `unsafe`; can the unsafe block possibly be made more granular?

src/net/uds/listener.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub fn from_std(std: std::os::unix::net::UnixListener) -> UnixListener {
+        UnixListener { std }
+    }
+
+    /// Accepts a new incoming connection to this listener.
+    ///
+    /// The call is responsible for ensuring that the listening socket is in
+    /// non-blocking mode.
+    pub fn accept(&self) -> io::Result<(UnixStream, SocketAddr)> {
+        match sys::uds::accept(&self.std) {
```
hawkw: TIOLI:  seems like this could be:
```rust
let (stream, addr) = sys::uds::accept(&self.std)?;
Ok((UnixStream::from_std(stream), addr))
```

src/sys/unix/uds.rs
kleimkuhler: The casts are `&` -> `*const libc::sockaddr_un` -> `*const libc::sockaddr`. The `*const` casts can be elided, but I could put them back in so that it is more explicit.

I don't think a direct cast can happen from `&` to `*const libc::sockaddr`.

src/sys/unix/uds.rs
kleimkuhler: See response above!

src/sys/unix/uds.rs
hawkw: hmm, take it or leave it, but i might just put the actual types in:

```rust
    let raw_addr = &raw_addr as *const libc::sockadder_un as *const libc::sockaddr;
```
imo this is clearer.

carllerche:
@asomers this is updating the UDS implementation to the latest mio. It also moves it in tree because `std` now supports UDS and moving it in tree reduces the number of dependencies.

src/sys/unix/uds.rs
kleimkuhler: Yea I like it too. It's added to the rest of the elided casts as well.

src/sys/unix/uds.rs
kleimkuhler: So I reduced a lot of the unsafe here. It turns out that a `libc::sockaddr_un` zeroed out is a properly initialized value. I explained why in the comment I added, but the rest of the expressions that set the field values will not be UB. Let me know if that makes sense.

I also reduced the unsafe usage in `sun_path_offset` and added a comment for why that is needed in the first place.

src/net/uds/stream.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub fn connect<P: AsRef<Path>>(p: P) -> io::Result<UnixStream> {
+        let std = sys::uds::connect_stream(p.as_ref())?;
+        Ok(UnixStream { std })
+    }
+
+    /// Converts a `std` `UnixStream` to a `mio` `UnixStream`.
+    ///
+    /// The caller is responsible for ensuring that the socket is in
+    /// non-blocking mode.
+    pub fn from_std(std: std::os::unix::net::UnixStream) -> UnixStream {
```
Thomasdezeeuw: Same as for `UnixDatagram` and `UnixListener` please remove this method and move the docs to `FromRawFd`.

src/net/uds/listener.rs
```diff
@@ -0,0 +1,98 @@
...
+
+use std::io;
+use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
+use std::os::unix::net;
+use std::path::Path;
+
+/// A non-blocking Unix domain socket server.
+#[derive(Debug)]
+pub struct UnixListener {
+    std: std::os::unix::net::UnixListener,
```
Thomasdezeeuw: Same as for `UnixDatagram`:
* `std` -> `sys`,
* `std::os::unix::net::UnixListener` -> `net::UnixListener`, and
* add `SelectorId`.

src/net/uds/stream.rs
```diff
@@ -0,0 +1,159 @@
...
+use std::io::{self, IoSlice, IoSliceMut};
+use std::net::Shutdown;
+use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
+use std::os::unix::net::SocketAddr;
+use std::path::Path;
+
+/// A non-blocking Unix stream socket.
+#[derive(Debug)]
+pub struct UnixStream {
+    std: std::os::unix::net::UnixStream,
```
Thomasdezeeuw: Same as for `UnixDatagram` and `UnixListener`:
* `std` -> `sys`,
* `std::os::unix::net::UnixListener` -> `net::UnixListener`, and
* add `SelectorId`.

src/lib.rs
```diff
@@ -126,6 +126,6 @@ pub use waker::Waker;
 #[cfg(unix)]
 pub mod unix {
     //! Unix only extensions.
-
+    pub use crate::sys::SocketAddr;
```
Thomasdezeeuw: I'm not too happy this, although I completely understand. Has there been any effort to add a way to create a `SocketAddr` to std lib?

src/net/uds/datagram.rs
```diff
@@ -0,0 +1,153 @@
...
+    pub fn bind<P: AsRef<Path>>(path: P) -> io::Result<UnixDatagram> {
+        let std = sys::uds::bind_datagram(path.as_ref())?;
+        Ok(UnixDatagram { std })
+    }
+
+    /// Converts a `std` `UnixDatagram` to a `mio` `UnixDatagram`.
+    ///
+    /// The caller is responsible for ensuring that the socket is in
+    /// non-blocking mode.
+    pub fn from_std(std: std::os::unix::net::UnixDatagram) -> UnixDatagram {
```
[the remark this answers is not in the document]
carllerche: @Thomasdezeeuw I don't remember the history behind this? I know that we will need some conversions from `std` for Tokio. I can't think of a reason why a conversion from std would not be possible?

src/net/uds/datagram.rs
```diff
@@ -0,0 +1,153 @@
...
+use std::io;
+use std::net::Shutdown;
+use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
+use std::os::unix::net::SocketAddr;
+use std::path::Path;
+
+/// A Unix datagram socket.
+#[derive(Debug)]
+pub struct UnixDatagram {
+    std: std::os::unix::net::UnixDatagram,
```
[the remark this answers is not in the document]
carllerche: Not a critical change to me. We can add the sys type when needed :+1:

src/sys/unix/uds.rs
```diff
@@ -0,0 +1,296 @@
...
+    let socket = new_socket(libc::AF_UNIX, libc::SOCK_DGRAM)?;
+    let (sockaddr, socklen) = socket_addr(path)?;
+    let sockaddr = &sockaddr as *const libc::sockaddr_un as *const libc::sockaddr;
+
+    syscall!(bind(socket, sockaddr, socklen))?;
+    Ok(unsafe { UnixDatagram::from_raw_fd(socket) })
+}
+
+pub fn accept(listener: &UnixListener) -> io::Result<(UnixStream, SocketAddr)> {
+    let mut storage: libc::sockaddr_un = unsafe { mem::zeroed() };
```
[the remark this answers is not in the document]
carllerche: Now that you mention it, I think you are correct. We probably should use `MaybeUninit` in this case

src/sys/unix/uds.rs
```diff
@@ -0,0 +1,296 @@
...
+    }
+    Ok(())
+}
+
+// On Linux, this funtion equates to the same value as
+// `size_of::<sun_path>()`, but some other implementations include other
+// fields before `sun_path`, so the expression more portably describes the
+// size of the address structure.
+fn sun_path_offset() -> usize {
+    let sockaddr = unsafe { mem::MaybeUninit::<libc::sockaddr_un>::uninit().assume_init() };
```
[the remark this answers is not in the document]
carllerche: Yes, this is undefined behavior. It is explicitly called out in https://doc.rust-lang.org/std/mem/union.MaybeUninit.html.

src/lib.rs
carllerche: We probably should open an issue. If we match the API of `std`, it should not be a breaking change to switch this export to be the std version once we can use it.

src/sys/unix/uds.rs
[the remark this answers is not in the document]
kleimkuhler: Ah yep thanks for catching this. Meant to change it over in my last commit and missed it!

src/sys/unix/uds.rs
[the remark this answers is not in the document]
kleimkuhler: Yep good call. I originally thought this was okay because `MaybeUninit<T>` guarantees size and alignment, but it specifies later:

>Rust does not in general guarantee that the fields of a Foo<T> have the same order as a Foo<U> even if T and U have the same size and alignment

Calculating an offset from this is therefore UB because the fields may not be ordered!

That being said-this can be removed by initializing the struct with what you pointed out above:

```
let mut sockaddr = libc::sockaddr_un {
    sun_family: libc::AF_UNIX as libc::sa_family_t,
    sun_path: [0; 108],
};
```

src/sys/unix/uds.rs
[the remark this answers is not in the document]
kleimkuhler: It seems like this is possible now?

src/sys/unix/uds.rs
```diff
@@ -0,0 +1,296 @@
...
+    // bytes is properly initialized.
+    //
+    // `0` is a valid value for `sockaddr_un::sun_family`; it is
+    // `libc::AF_UNSPEC`.
+    //
+    // `[0; 108]` is a valid value for `sockaddr_un::sun_path`; it begins an
+    // abstract path.
+    let mut sockaddr = unsafe { sockaddr.assume_init() };
+
+    sockaddr.sun_family = libc::AF_UNIX as libc::sa_family_t;
```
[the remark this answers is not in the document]
kleimkuhler: The `MaybeUninit` has been removed.

src/lib.rs
kleimkuhler: Sure I can open an issue. I don't see any reason that would not be okay? It should be pretty easy to follow up with the fix for it as well.

src/sys/unix/uds.rs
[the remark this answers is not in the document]
kleimkuhler: Never mind it introduced issues cross-platform. I have it changed to `MaybeUninit::zeroed()` followed by a use a `assume_init()` with why that is safe.

src/sys/unix/uds.rs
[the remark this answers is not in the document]
kleimkuhler: As I mentioned above, this `MaybeUninit` has been added back.

I do not think we move this above the `assume_init` however. The `assume_init` gives us a `libc::sockaddr_un` and that is why we can set the `sun_family` field here without a pointer indirection through `MaybeUninit`. It is a properly initialized value at this point, so there should not be any UB.. is this what you were suggesting?

kleimkuhler:
> Same as for `UnixDatagram` and `UnixListener`:
> * `std` -> `sys`,
> * `std::os::unix::net::UnixListener` -> `net::UnixListener`, and
> * add `SelectorId`.


@Thomasdezeeuw I'm not sure I follow all the parts of this comment in `UnixStream`, `UnixDatagram`, and `UnixListener`.

I can rename the `std` field to `sys`, and I have added the `selector_id` field for debug assertions.

I'm not sure I follow `std::os::unix::net::UnixListener -> net::UnixListener`. These types are not available in `std::net`, they are only available through these paths and must be imported directly.

src/net/uds/datagram.rs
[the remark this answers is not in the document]
kleimkuhler: Answered below in a PR comment

src/net/uds/listener.rs
kleimkuhler: Answered below in a PR comment

src/net/uds/stream.rs
kleimkuhler: Answered below in a PR comment

kleimkuhler:
@Thomasdezeeuw I think most of your comments should be addressed from the last commit. Below is my commit message:

> This commit introduces a significant restructure from the original UDS
> support proposal. Originally, the UDS types dispatched the the `std` UDS
> types from the `net` module. This was different from other `net` module
> types because it did not use the `sys` modules types for that OS.
> 
> This was okay because currently UDS is only supported by Unix, but when
> windows support is added this would have been a problem.
> 
> The new structure sets up most of the work for Windows UDS support as
> the `net` module UDS types now all dispatch to their inner `sys` module
> types.

I have left all your comments from your last review that had to do with this unresolved, but I think they should all be addressed at this point.

src/net/uds/datagram.rs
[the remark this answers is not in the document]
Thomasdezeeuw: We did this to match the API of the std lib more closely, see #1045.

tests/uds.rs
```diff
@@ -0,0 +1,479 @@
+#![cfg(target_os = "unix")]
+#[macro_use]
+mod util;
+
+use mio::net::{UnixListener, UnixStream};
+use mio::{Interests, Token};
+use std::io::{self, IoSlice, IoSliceMut, Read, Write};
+use std::net::Shutdown;
+#[cfg(unix)]
```
Thomasdezeeuw: Not needed as we have target os above.

tests/uds.rs
```diff
@@ -0,0 +1,479 @@
...
+    assert_ok!(handle.join());
+}
+
+#[test]
+fn is_send_and_sync() {
+    assert_send::<UnixListener>();
+    assert_sync::<UnixListener>();
+
+    assert_send::<UnixStream>();
+    assert_sync::<UnixStream>();
```
Thomasdezeeuw: Missing `UnixDatagram`.

Thomasdezeeuw:
Since this change is so large, can some double check this and merge if acceptable?

src/sys/unix/uds/listener.rs
```diff
@@ -0,0 +1,251 @@
...
+
+/// An address associated with a `mio` specific Unix socket.
+///
+/// This is implemented instead of imported from [`net::SocketAddr`] because
+/// there is no way to create a [`net::SocketAddr`]. One must be returned by
+/// [`accept`], so this is returned instead.
+///
+/// [`net::SocketAddr`]: std::os::unix::net::SocketAddr
+/// [`accept`]: #method.accept
+pub struct SocketAddr {
```
carllerche: Let's **not** do it in this PR, but as a follow up it would be nice to move this type to a separate file.

src/net/uds/datagram.rs
```diff
@@ -0,0 +1,164 @@
+use crate::event::Source;
+#[cfg(debug_assertions)]
+use crate::poll::SelectorId;
+use crate::{sys, Interests, Registry, Token};
+
+use std::io;
+use std::net::Shutdown;
+use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
+use std::os::unix::net::SocketAddr;
```
carllerche: It looks like we are using the `std` version of `SocketAddr` here. Should we focus on only using `mio`s version?

src/net/uds/listener.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// The returned `UnixListener` is a reference to the same socket that this
+    /// object references. Both handles can be used to accept incoming
+    /// connections and options set on one listener will affect the other.
+    pub fn try_clone(&self) -> io::Result<UnixListener> {
+        let sys = self.sys.try_clone()?;
+        Ok(UnixListener::new(sys))
+    }
+
+    /// Returns the local socket address of this listener.
+    pub fn local_addr(&self) -> io::Result<net::SocketAddr> {
```
carllerche: Same: should we focus on only using mio's version of `SocketAddr`?

src/net/uds/stream.rs
```diff
@@ -0,0 +1,170 @@
+use crate::event::Source;
+#[cfg(debug_assertions)]
+use crate::poll::SelectorId;
+use crate::{sys, Interests, Registry, Token};
+
+use std::io::{self, IoSlice, IoSliceMut};
+use std::net::Shutdown;
+use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
+use std::os::unix::net::SocketAddr;
```
carllerche: Same: should we focus on only using mio's version of `SocketAddr`?

kleimkuhler:
@carllerche I had limited the use of mio's `SocketAddr` because it had felt more like an inconvenience that libstd's `SocketAddr` was not public; I leaned towards using it in the least amount of places possible.

That's a good point though about using it everywhere for now and changing over all at once when we can. I'll make that change!

kleimkuhler:
@carllerche Had to introduce some further `mio::unix::SocketAddr` functionality to return it everywhere it is used, but `net::SocketAddr` is no longer used now.

Also fixed a `cfg` in the UDS tests that was not actually running them correctly.

PR #1099, opened 2019-10-03
Test PR: Silence warnings on different Rust versions

kleimkuhler:
#1097 is currently failing and this is test PR that adds some additional changes. This works local with stable and 1.18.0


kleimkuhler:
Triggering HUP events was removed in 85a90f483d53abc7609a8467fde6176a494113a9. The additional TCP shutdown tests added checks for that back, and the HUP checks were failing on kqueue platforms.

"Min Rust MacOS" seems to fail because a macro path in a dependency.

"Nightly Linux" fails because `#![feature] may not be used on the stable release channel`. I've tried the pinned nightly, but it appears to still think it's running on stable. I may be setting it incorrectly in `azure_pipelines.yml`?

kleimkuhler:
`cfg-if` released version `0.1.10` a few days ago and both the dependencies `log` and `net2` pull in the latest version. Version `0.1.10` however does not work with Rust version `1.18.0`, so the build fails.

This change now version controls the lock file so that `cfg-if` is locked to `0.1.9`.

The install scripts were also fixed; they were not installing different Rust versions correctly.

Thomasdezeeuw:
Would adding `cfg = "=0.1.9"` to our own `Cargo.toml` file prevent from updating to `0.1.10`?

kleimkuhler:
@Thomasdezeeuw Ah thanks yea looks like that should work--it does locally. Adds a tracked dependency, but nicer than having to VC the lock file.

Thomasdezeeuw:
@kleimkuhler thanks! Now the other pr can start moving again.

cuviper:
> Would adding `cfg = "=0.1.9"` to our own `Cargo.toml` file prevent from updating to `0.1.10`?

Please don't do this -- it makes it impossible for your users to current `cfg-if` and current `mio` at the same time. In fact, if I test a new project with loose dependencies:

```toml
[dependencies]
cfg-if = "0.1"
mio = "0.6"
```

Then `cargo update` resolves to `cfg-if v0.1.10` and `mio v0.6.19`. I can force `cargo update -p mio --precise 0.6.20`, and it downgrades to `cfg-if v0.1.9` to make it work, but a plain `cargo update` reverts back to current `cfg-if` and older `mio`.

---

Controlling your own lockfile for CI is a better approach. Either downgrade dependencies with `--precise` when testing older compilers, or use a known-good lockfile. I do that in rayon [here](https://github.com/rayon-rs/rayon/blob/401678ee554f90ab11abe70eb23737e26b489ddd/.travis.yml#L15-L19).

Thomasdezeeuw:
Honestly I don't really want to spend too much time on this. This is not the only crates that ran into issues because `cfg-if` bumped the required Rust version with a major alpha version (i.e. `0.2`). @cuviper if you want to make a pr I would accept it, but otherwise I prefer to focus on Mio 0.7 which doesn't have `cfg-if` as a dependency.

cuviper:
I feel your pain, as I'm conservative with minimum Rust too. Pinning versions forces your decision on others though, which is also not great. I'll send a PR if I get a chance.

PR #1102, opened 2019-10-16
Fix socket_addr() to not drop chars in path

### LINKED ISSUE #1101: UnixListener::bind drops last letter in file name
UnixListener::bind drops last letter in file name, as in the code below:

```rust
use mio::net::UnixListener;

fn main() {
  let listener = UnixListener::bind("abcde").unwrap();
  println!("{:#?}", listener); // local pathname is "abcd"
}
```

I suspect it happens in `socket_addr` function: 
https://github.com/tokio-rs/mio/blob/9fe594eae1b92ad1d045c1456cc3d608b0ca7689/src/sys/unix/uds/mod.rs#L16

glelk:
Fixes #1101

### DIFF
--- a/src/sys/unix/uds/listener.rs
+++ b/src/sys/unix/uds/listener.rs
@@ -242,7 +242,7 @@ impl SocketAddr {
     // describes the size of the address structure.
     fn path_offset(&self) -> usize {
         let base = &self.sockaddr as *const _ as usize;
-        let path = &self.sockaddr as *const _ as usize;
+        let path = &self.sockaddr.sun_path as *const _ as usize;
         path - base
     }
 }
--- a/src/sys/unix/uds/mod.rs
+++ b/src/sys/unix/uds/mod.rs
@@ -53,7 +53,7 @@ pub fn socket_addr(path: &Path) -> io::Result<(libc::sockaddr_un, libc::socklen_
     // View `SocketAddr::path_offset` documentation for why this is necessary.
     let offset = {
         let base = &sockaddr as *const _ as usize;
-        let path = &sockaddr as *const _ as usize;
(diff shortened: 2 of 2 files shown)


glelk:
Good catch!
Change added.
It's great that you have tests ready for the changes. I'm curious what is the smart way to test functions which create files on disk - and potentially leave them behind if test have failed.

kleimkuhler:
@glelk Yes so that is something I'm wondering about too... I'm not sure why the [tests](https://github.com/tokio-rs/mio/blob/9fe594eae1b92ad1d045c1456cc3d608b0ca7689/tests/uds.rs#L439) did not originally catch this.

The tests I have added are only unit tests for the `socket_addr` function to make sure it returns expected `socklen` values for known values.

kleimkuhler:
@glelk `connect` and `bind` should return `EINVAL` errors if `addrlen` is wrong. The fact that it is 2 bytes short and still successfully connects is interesting. The `smoke_test` _should_ have caught this, but it seems to connect/bind okay even with the wrong `addrlen`.

```
running 1 test
path: "/tmp/uds.6SiU3OV6QtyD/foo", socklen: 26
successfully bound socket
successfully connected
test smoke_test ... ok
```

```
running 1 test
path: "/tmp/uds.3m77UYivQ67E/foo", socklen: 28
successfully bound socket
successfully connected
test smoke_test ... ok
```

Just curious what OS you ran into the issue on? Maybe it has something to do with the `libc::bind` impl and it just doesn't error on some OSs--however the mio CI runs on quite a few platforms.

Either way this fix looks good and I'll follow up with some additional unit tests.

glelk:
@kleimkuhler I ran the code on x68_64 Arch Linux

PR #1103, opened 2019-10-17
Add unit tests for sock_addr and path_offset functions

kleimkuhler:
## Motivation

The `sock_addr` function dispatches to a fair amount of `libc` calls and has
some additional complexity that is really only explained by man pages. It does
not have any unit tests and was incorrectly calculating `addrlen`.

## Solution

A new `path_offset` function has been separated out from `sock_addr` that
returns values dependent on the platform. It is a single function that can be
unit tested and now also used within the `sys::unix::uds` module.

Basic unit tests have also been added that check correct values for pathname
socket addresses and abstract socket addresses.

### DIFF
--- a/src/sys/unix/uds/listener.rs
+++ b/src/sys/unix/uds/listener.rs
@@ -1,4 +1,4 @@
-use super::socket_addr;
+use super::{path_offset, socket_addr};
 use crate::event::Source;
 use crate::sys::unix::net::new_socket;
 use crate::sys::unix::UnixStream;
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
LGTM pending the CI fix.

kleimkuhler:
@Thomasdezeeuw Looks like a spurious failure in CI. Tested locally on a Mac as well to double check.

kleimkuhler:
@glelk These are unit tests and refactor I was talking about these functions benefitting from!

PR #1104, opened 2019-10-18
Fix fcntl calls and ensure fds are closed on error

kleimkuhler:
## Motivation

The primary focus of this PR was to fix the `FD_CLOEXEC` flag not being set on
sockets. On platforms where the flag cannot be set on the `socket` sys call, it
is attempted to be set via the `F_SETFL` command in the `fcntl` sys call.

`FD_CLOEXEC` can only be set with the `F_SETFD` command in a separate `fcntl`
sys call.

This was recently fixed on `v0.6.x` via #1095.

In addition to the primary motivation, if `fcntl` sys calls fail they leave the
newly created sockets open which leads to file descriptor leaks.

## Solution

The `new_socket` function now has two separate `fctnl` sys calls to set the
`O_NONBLOCK` and `FD_CLOEXEC` flags.

The blocks of platform specific code where `fcntl` sys calls are required now
properly cleanup newly created sockets when there are errors.

### DIFF
--- a/src/sys/unix/net.rs
+++ b/src/sys/unix/net.rs
@@ -13,26 +13,38 @@ pub fn new_ip_socket(addr: SocketAddr, socket_type: libc::c_int) -> io::Result<l
 
 /// Create a new non-blocking socket.
 pub fn new_socket(domain: libc::c_int, socket_type: libc::c_int) -> io::Result<libc::c_int> {
-    #[cfg(not(any(target_os = "ios", target_os = "macos", target_os = "solaris")))]
-    {
-        let socket_type = socket_type | libc::SOCK_NONBLOCK | libc::SOCK_CLOEXEC;
-        syscall!(socket(domain, socket_type, 0))
-    }
+    #[cfg(any(
+        target_os = "android",
+        target_os = "bitrig",
+        target_os = "dragonfly",
+        target_os = "freebsd",
+        target_os = "linux",
+        target_os = "netbsd",
+        target_os = "openbsd"
+    ))]
+    let socket_type = socket_type | libc::SOCK_NONBLOCK | libc::SOCK_CLOEXEC;
 
+    // Gives a warning for platforms without SOCK_NONBLOCK.
+    #[allow(clippy::let_and_return)]
+    let socket = syscall!(socket(domain, socket_type, 0));
+
+    // Darwin doesn't have SOCK_NONBLOCK or SOCK_CLOEXEC. Not sure about
+    // Solaris, couldn't find anything online.
     #[cfg(any(target_os = "ios", target_os = "macos", target_os = "solaris"))]
-    {
-        let socket = syscall!(socket(domain, socket_type, 0))?;
-        if let Err(e) = (|| {
-            syscall!(fcntl(socket, libc::F_SETFL, libc::O_NONBLOCK))?;
-            syscall!(fcntl(socket, libc::F_SETFD, libc::FD_CLOEXEC))
-        })() {
-            // If either of the `fcntl` calls failed, ensure the socket is
-            // closed and return the error.
-            syscall!(close(socket))?;
-            return Err(e);
-        }
-        Ok(socket)
-    }
+    let socket = socket.and_then(|socket| {
+        // For platforms that don't support flags in socket, we need to
+        // set the flags ourselves.
+        syscall!(fcntl(socket, libc::F_SETFL, libc::O_NONBLOCK,))
+            .and_then(|_| syscall!(fcntl(socket, libc::F_SETFD, libc::FD_CLOEXEC,)).map(|_| socket))
+            .map_err(|e| {
+                // If either of the `fcntl` calls failed, ensure the socket is
+                // closed and return the error.
+                let _ = syscall!(close(socket));
+                e
+            })
+    });
+
+    socket
 }
 
 pub fn socket_addr(addr: &SocketAddr) -> (*const libc::sockaddr, libc::socklen_t) {
--- a/src/sys/unix/net.rs
+++ b/src/sys/unix/net.rs
@@ -13,36 +13,26 @@ pub fn new_ip_socket(addr: SocketAddr, socket_type: libc::c_int) -> io::Result<l
 
 /// Create a new non-blocking socket.
 pub fn new_socket(domain: libc::c_int, socket_type: libc::c_int) -> io::Result<libc::c_int> {
-    #[cfg(any(
-        target_os = "android",
-        target_os = "bitrig",
-        target_os = "dragonfly",
-        target_os = "freebsd",
-        target_os = "linux",
-        target_os = "netbsd",
-        target_os = "openbsd"
-    ))]
-    let socket_type = socket_type | libc::SOCK_NONBLOCK | libc::SOCK_CLOEXEC;
+    #[cfg(not(any(target_os = "ios", target_os = "macos", target_os = "solaris")))]
+    {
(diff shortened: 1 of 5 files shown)


src/sys/unix/net.rs
```diff
@@ -13,36 +13,26 @@ pub fn new_ip_socket(addr: SocketAddr, socket_type: libc::c_int) -> io::Result<l
...
-        target_os = "linux",
-        target_os = "netbsd",
-        target_os = "openbsd"
-    ))]
-    let socket_type = socket_type | libc::SOCK_NONBLOCK | libc::SOCK_CLOEXEC;
-
-    // Gives a warning for platforms without SOCK_NONBLOCK.
-    #[allow(clippy::let_and_return)]
-    let socket = syscall!(socket(domain, socket_type, 0));
+    #[cfg(not(any(target_os = "ios", target_os = "macos", target_os = "solaris")))]
```
Thomasdezeeuw: I don't think we want to do cfg this way. If we add a new target in the old code it would explode indicating that we need to look if this would work on the new platform, for the new code this is not the case.

src/sys/unix/net.rs
```diff
@@ -13,36 +13,26 @@ pub fn new_ip_socket(addr: SocketAddr, socket_type: libc::c_int) -> io::Result<l
...
-            libc::F_SETFL,
-            libc::O_NONBLOCK | libc::O_CLOEXEC
-        ))
-        .map(|_| socket)
-    });
-
-    socket
+    {
+        let socket = syscall!(socket(domain, socket_type, 0))?;
+        if let Err(e) = (|| {
```
Thomasdezeeuw: Not really a fan of this closure, can you can change this to `syscall!(..).and_then(|_| syscal!(..))`.

src/sys/unix/net.rs
```diff
@@ -13,36 +13,26 @@ pub fn new_ip_socket(addr: SocketAddr, socket_type: libc::c_int) -> io::Result<l
...
-    // Gives a warning for platforms without SOCK_NONBLOCK.
-    #[allow(clippy::let_and_return)]
-    let socket = syscall!(socket(domain, socket_type, 0));
+    #[cfg(not(any(target_os = "ios", target_os = "macos", target_os = "solaris")))]
+    {
+        let socket_type = socket_type | libc::SOCK_NONBLOCK | libc::SOCK_CLOEXEC;
+        syscall!(socket(domain, socket_type, 0))
+    }
 
-    // Darwin doesn't have SOCK_NONBLOCK or SOCK_CLOEXEC. Not sure about
```
Thomasdezeeuw: Can we keep this comment.

src/sys/unix/uds/datagram.rs
```diff
@@ -37,14 +37,31 @@ impl UnixDatagram {
...
+
+        // Darwin and Solaris do not have SOCK_NONBLOCK or SOCK_CLOEXEC.
+        //
+        // In order to set those flags, additional `fcntl` sys calls must be
+        // made in `pair_descriptors` that are fallible. If a `fnctl` fails
+        // after the sockets have been created, the file descriptors will
+        // leak. Creating `s1` and `s2` below ensure that if there is an
+        // error, the file descriptors are closed.
+        #[cfg(any(target_os = "ios", target_os = "macos", target_os = "solaris"))]
+        {
```
Thomasdezeeuw: Why can't we use this code on other platforms as well?

src/sys/unix/uds/stream.rs
```diff
@@ -44,14 +44,31 @@ impl UnixStream {
...
-        })
+        // Darwin and Solaris do not have SOCK_NONBLOCK or SOCK_CLOEXEC.
+        //
+        // In order to set those flags, additional `fcntl` sys calls must be
+        // made in `pair_descriptors` that are fallible. If a `fnctl` fails
+        // after the sockets have been created, the file descriptors will
+        // leak. Creating `s1` and `s2` below ensure that if there is an
+        // error, the file descriptors are closed.
+        #[cfg(any(target_os = "ios", target_os = "macos", target_os = "solaris"))]
+        {
```
Thomasdezeeuw: Same question as for the datagram.

src/sys/unix/uds/mod.rs
```diff
@@ -70,21 +70,22 @@ pub fn socket_addr(path: &Path) -> io::Result<(libc::sockaddr_un, libc::socklen_
 
 fn pair_descriptors(mut fds: [RawFd; 2], flags: i32) -> io::Result<()> {
     #[cfg(not(any(target_os = "ios", target_os = "macos", target_os = "solaris")))]
-    let flags = flags | libc::SOCK_NONBLOCK | libc::SOCK_CLOEXEC;
-
-    syscall!(socketpair(libc::AF_UNIX, flags, 0, fds.as_mut_ptr()))?;
+    {
```
Thomasdezeeuw: Why do we need these changes? I don't think anything changed.

src/sys/unix/uds/datagram.rs
kleimkuhler: On other platforms, we'll know right away if `socket` succeeds or fails. That is the more common path of execution, and it saves on allocating extra `UnixDatagram` (or `UnixStreams`) that are not strictly necessary. I may just get rid of `pair_descriptors`. It's similar to why #1095 you had to allocate [here](https://github.com/Thomasdezeeuw/mio/blob/2ac1db79d9ddda7255e7edaf0be2bf1fbc31916b/src/sys/unix/mod.rs#L64) in the `fcntl` calls block, but not in the one [above](https://github.com/Thomasdezeeuw/mio/blob/2ac1db79d9ddda7255e7edaf0be2bf1fbc31916b/src/sys/unix/mod.rs#L58).

I may just remove `pair_descriptors` and have some code duplication between `UnixDatagram::pair` and `UnixStream::pair`.

kleimkuhler:
@Thomasdezeeuw Thanks for the review and the concern about the initial changes. I changed the functions back to sharing logic when able to (including returns).

The majority of the changes are now in `datagram.rs` and `stream.rs` where the call to `pair_descriptors` is in a `cfg` block. I would still like to keep this, as the common case platform doesn't need to allocate the `UnixDatagram`/`UnixStream`s before `pair_descriptors` as there is no cleanup to do if `libc::socket` fails.

src/sys/unix/net.rs
```diff
@@ -34,12 +34,14 @@ pub fn new_socket(domain: libc::c_int, socket_type: libc::c_int) -> io::Result<l
     let socket = socket.and_then(|socket| {
         // For platforms that don't support flags in socket, we need to
         // set the flags ourselves.
-        syscall!(fcntl(
-            socket,
-            libc::F_SETFL,
-            libc::O_NONBLOCK | libc::O_CLOEXEC
-        ))
-        .map(|_| socket)
+        syscall!(fcntl(socket, libc::F_SETFL, libc::O_NONBLOCK,))
```
Thomasdezeeuw: Weird that rustfmt doesn't remove the last `,`, but no need to change it.

Thomasdezeeuw:
Thank you!

kleimkuhler:
@Thomasdezeeuw That is a good point--I should have recognized that when you originally asked about the need for the `cfg` there. [Here](https://godbolt.org/z/Ki0RVs) it looks like any level of optimization leads to `control` and `wrap` resulting in the same assembly. I'll make sure to simplify that method next time I'm working in that area of the code.

src/sys/unix/net.rs
hawkw: rustfmt doesn't generally touch code inside macro invocations.

Issue #1079, opened 2019-09-07
Investigate TcpStream nodelay test failures


Thomasdezeeuw:
Pr #1070 is going to add (or added, depending on when you read this) a tests for  `TcpStream::(set)_nodelay` `nodelay`, which only fails on Windows. Figure out why.

/cc @PerfectLaugh

JuSiZeLa:
The test is failing when the `set_nodelay` function is called on a socket which is not connected yet. All calls to `setsockopt` while the socket is not connected yet are going to fail.

Thomasdezeeuw:
Fixed in #1105.

PR #1105, opened 2019-10-19
Fix nodelay test on Windows

JuSiZeLa:
This patch fixes issue #1079.
The issue was happening because the set_nodelay
function, that internally uses the setsockopt,
was called on a socket that was not connected yet.


Thomasdezeeuw:
Thank you!

Issue #1078, opened 2019-09-07
Investigate TcpStream smoke test failures


Thomasdezeeuw:
Pr #1070 is going to add (or added, depending on when you read this) two smoke tests for  `TcpStream` `tcp_stream_ipv4` and `tcp_stream_ipv6`, which only fails on Windows. Figure out why.

/cc @PerfectLaugh

dtacalau:
@Thomasdezeeuw could you please assign this issue to me. I've started working on it.

dtacalau:
Tests fail because of issue described here #1131, we need to reregister the interests after TcpStream::read which emptied the sock bufffer.

PR #1107, opened 2019-10-20
fix tcpstream smoke test on Windows

JuSiZeLa:
This patch fixes issue #1078.
The issue was happening because there was
no pending read before waiting for READABLE
event.

### DIFF
--- a/tests/tcp_stream.rs
+++ b/tests/tcp_stream.rs
@@ -32,13 +32,11 @@ fn is_send_and_sync() {
 }
 
 #[test]
-#[cfg_attr(windows, ignore = "fails on Windows, see #1078")]
 fn tcp_stream_ipv4() {
     smoke_test_tcp_stream(any_local_address());
 }
 
 #[test]
-#[cfg_attr(windows, ignore = "fails on Windows, see #1078")]
 fn tcp_stream_ipv6() {
     smoke_test_tcp_stream(any_local_ipv6_address());
 }
@@ -88,6 +86,10 @@ fn smoke_test_tcp_stream(addr: SocketAddr) {
     assert_eq!(&buf[..n], DATA1);
 
     assert!(stream.take_error().unwrap().is_none());
+
+    let mut buf = [0; 32];
+    assert_would_block(stream.peek(&mut buf));
+    assert_would_block(stream.read(&mut buf));
 
     let bufs = [IoSlice::new(&DATA1), IoSlice::new(&DATA2)];
     let n = stream


Thomasdezeeuw:
I don't think this is the right fix. If the problem is that we need to register readable interests again (done in this change by peeking/reading), then perhaps this should be done in the Windows implementation not in the test. To me the test as is (without the change) should just work.

JuSiZeLa:
According to the documentation of the `Poll` function in https://tokio-rs.github.io/mio/doc/mio/struct.Poll.html#draining-readiness, there is no guarantee that another readiness event will be delivered if the last corresponding operation did not return `WouldBlock`

Thomasdezeeuw:
Out of curiosity would adding `READABLE` to the interests when registering before calling `peer_addr` fix the test on Windows? Here: https://github.com/tokio-rs/mio/blob/1d1e73268c1563123911ee20f25f4269e6dae6dc/tests/tcp_stream.rs#L57

JuSiZeLa:
No, it will not.
From what I understood from the code, you are trying to use Windows' I/O completion ports like epoll on unix but they have a different approach: IOCP tells you when a requested operation is completed, or failed to complete while epoll tells you when a file descriptor is ready to perform a requested operation. To use IOCP like epoll for the READABLE interest, for example, implies to always have a pending read that returned `WouldBlock`.
I think almost all test failures on Windows are related to this problem

JuSiZeLa:
I did some other tests and I understood better how the code works.
I also found that reregistering the READABLE interest before each poll makes the test to succeed.

Thomasdezeeuw:
*re*registering works but registering with `READABLE` doesn't? If so that is very confusing and should really be fixed in the implementation, not in the tests.

JuSiZeLa:
It is already register for `READABLE` here https://github.com/tokio-rs/mio/blob/1d1e73268c1563123911ee20f25f4269e6dae6dc/tests/tcp_stream.rs#L51
Adding a `READABLE` where you suggested makes the test failing on that line because that event will not be available.

Thomasdezeeuw:
In any case I don't think is the way to "fix" this, the problem is in the implementation. It should be perfectly ok to just write to connection and never read from it, but as the test shows that is broken on Windows.

Thomasdezeeuw:
Closing this in favour of #1137.

Thomasdezeeuw:
@JuSiZeLa thank you for looking into this, due to your discovery of needing to register a number of issues were fixed.

PR #1091, opened 2019-09-19
Add test to list OS event details

Thomasdezeeuw:
This adds a simple test that print the system events in a specific scenario. This pr only include a couple scenarios for Unix pipes.

For further discussion in #1087.

/cc @carllerche @asomers

PR #1110, opened 2019-10-25
Add read_close and write_close readiness events

kleimkuhler:
## Motivation

The existing `is_hup` method on `Events` was difficult to add additional checks
to that were both cross-platform and guaranteed no false positives.

For example, if the write half of a socket is closed in certain scenarios, this
causes a set of expected events to be triggered. With `is_read_hup` and
`is_hup`, there was no way to check for for this set of events that did not also
result in false positives on other platforms.

## Solution

To solve this, `is_read_closed` and `is_write_closed` are being introduced as
cross-platform methods on `Events` that are best effort implementations--meaning
false negatives are possible, but false positives will not occur.

Finding the expected event behavior for each of these platforms has not been
very easy, so below is best-effort summary of what can be expected. #1091
provides a lot of the visibility into where this data came from, but a lot of
the scenarios checked for change when they occur independently.

### epoll

A helpful thing to keep in mind is that `EPOLLHUP` is a mask for a socket's
`RCV_SHUTDOWN` and `SEND_SHUTDOWN` flags. Once both of these flags are set,
`EPOLLHUP` will trigger.

#### is_read_closed

`EPOLLRDHUP` is now a registered event when there is interest in
`Interest::READABLE`.

`EPOLLRDHUP` occurs after *either* of the following events that sets the
`RCV_SHTUDOWN` flag:
  1. The local socket has called `shutdown(Shutdown::Read)`
  2. The peer stream has called `shutdown(Shutdown::Write)` which sends a `FIN`
     packet; that packet has been acknowledged by the local socket.

`EPOLLHUP` occurs after *both* of the following:
  1. `EPOLLRDHUP` by the previous point
  2. The local socket has called `shutdown(Shutdown::Write)` sending a `FIN` to
     the peer and marking setting its own `SEND_SHUTDOWN` flag.

`is_read_closed` checks for *either* of the following:
  1. `EPOLLHUP`
  2. `EPOLLIN` and `EPOLLRDHUP`

#### is_write_closed

`is_write_closed` is a hard event to check for with `epoll` without returning
false positives. The only special-case set of events that can be checked is both
`EPOLLOUT` and `EPOLLERR` which occurs when a Unix pipe sender closes.

Therefore, `is_write_closed` checks for *either* of the following:
  1. `EPOLLHUP` by the previous point
  2. `EPOLLOUT` and `EPOLLERR`

#### Things to note (or at least I found interesting)

A peer calling `shutdown(Shutdown::Both)` does not result in `EPOLLHUP` on the
local socket. It results in a `EPOLLIN` | `EPOLLOUT` | `EPOLLRDHUP` because the
peer has sent a `FIN` packet that has been acknowledged. Keep in mind however,
the local socket has not *also* sent a `FIN`, therefore it's `SEND_SHUTDOWN`
flag is not set.

Except for Unix pipes, a local socket calling `shutdown(Shutdown::Write)` only
results in a `EPOLLOUT`, which is not very helpful! Because there is no
`EPOLLRDHUP` equivalent event for when the `SEND_SHUTDOWN` flag is set, it is
hard to say whether or not the write half is _actually_ closed.

### kqueue

`kqueue` is has fairly predictable event behavior and therefore read and write
closed events are easier check. Note that some of the event checks in the tests
will only compile for `kqueue` platforms because `is_write_closed` will not
return false negatives.

#### is_read_closed

`is_read_closed` checks for the `EVFILT_READ` filter *and* `EV_EOF` flag

#### is_write_closed

`is_write_closed` checks for the `EVFILT_WRITE` filter *and* `EV_EOF` flag.

#### Things to note (or at least I found interesting)

A peer calling `shutdown(Shutdown::Both)` has the same result as it does in
`epoll`. `is_read_closed` will return `true`, but nothing has happend on the
local socket to indicate the write half will error.

### IOCP

`IOCP` is the least tested selector in this change, and therefore has few
situations where `is_read_closed` and `is_write_closed` do not return false
negatives. It checks for the equivalent behavior as before, mainly that
`is_write_closed` is equivalent to the current `is_hup`.

`is_read_closed` checks for the `AFD_POLL_DISCONNECT` flag.

`is_write_closed` checks for the `AFD_POLL_ABORT` flag.


src/interests.rs
```diff
@@ -57,11 +60,17 @@ impl Interests {
...
-    pub const AIO: Interests = Interests(unsafe { NonZeroU8::new_unchecked(AIO) });
+    pub const AIO: Interests = Interests(unsafe { NonZeroU16::new_unchecked(AIO) });
 
     /// Returns a `Interests` set representing LIO completion interests.
     #[cfg(target_os = "freebsd")]
-    pub const LIO: Interests = Interests(unsafe { NonZeroU8::new_unchecked(LIO) });
+    pub const LIO: Interests = Interests(unsafe { NonZeroU16::new_unchecked(LIO) });
+
+    /// TODO
+    pub const READ_CLOSE: Interests = Interests(unsafe { NonZeroU16::new_unchecked(READ_CLOSE) });
```
Thomasdezeeuw: I don't know if we want to do this, I believe that kqueue has this enabled (or tracks this) by default, we might just want to replicate that in epoll/IOCP as well.

Thomasdezeeuw: Such that `READABLE` interests also register read close interests.

src/event/event.rs
```diff
@@ -62,59 +62,51 @@ impl Event {
         sys::event::is_error(&self.inner)
     }
 
-    /// Returns true if the event contains HUP readiness.
+    /// Returns true if the event contains read close readiness.
```
Thomasdezeeuw: We'll need to expand this documentation to explain when "read close" readiness can be expected, can be a follow up pr. Same for write docs of course.

src/event/event.rs
```diff
@@ -62,59 +62,51 @@ impl Event {
...
+    /// |               | `EPOLLIN` and `EPOLLRDHUP` |
     /// | [kqueue]      | `EV_EOF`        |
     ///
     /// [OS selector]: ../struct.Poll.html#implementation-notes
     /// [epoll]: http://man7.org/linux/man-pages/man7/epoll.7.html
     /// [kqueue]: https://www.freebsd.org/cgi/man.cgi?query=kqueue&sektion=2
     #[inline]
-    pub fn is_hup(&self) -> bool {
-        sys::event::is_hup(&self.inner)
+    pub fn is_read_close(&self) -> bool {
```
Thomasdezeeuw: Should we go with is close or close*d*?

src/event/event.rs
```diff
@@ -62,59 +62,51 @@ impl Event {
...
+    /// Method is a best effort implementation. While some platforms may not
+    /// return readiness when read half is closed, it is guaranteed that
+    /// false-positives will not occur.
     ///
     /// The table below shows what flags are checked on what OS.
     ///
     /// | [OS selector] | Flag(s) checked |
     /// |---------------|-----------------|
-    /// | [epoll]       | `EPOLLHUP`      |
+    /// | [epoll]       | `EPOLLHUP`
```
Thomasdezeeuw: Can you change this to "`EPOLLHUP`, or" to indicate not all three flags have be set. Same for write table.

tests/close_events.rs
```diff
@@ -0,0 +1,268 @@
...
+use std::net::{Shutdown, SocketAddr};
+use std::sync::{mpsc, Arc, Barrier};
+use std::thread::{self, JoinHandle};
+use std::time::Duration;
+use util::{any_local_address, any_local_ipv6_address, expect_events, init_with_poll, ExpectEvent};
+
+const DATA: &[u8] = b"hello";
+const TOK: Token = Token(0);
+
+macro_rules! generate_tests {
```
Thomasdezeeuw: Do we really need a macro for this? Can't we use the same strategy we use for the other net tests, e.g: https://github.com/tokio-rs/mio/blob/162cdd3868742ed494ba03e7c843708eebd6849a/tests/tcp_listener.rs#L28-L37

tests/close_events.rs
```diff
@@ -0,0 +1,268 @@
+#[macro_use]
```
Thomasdezeeuw: I don't think we should create another file for this, each type (e.g. `TcpStream`, `UdpSocket` etc.) should just have another test in there module/file.

src/event/event.rs
kleimkuhler: @Thomasdezeeuw I had this same thought and thought "close" seemed better because it's the same tense as `is_error` and `is_hup`. Also since these are methods on `Events`, they are used to check if an `Event` is a _close_ event, not if the socket half is _closed_. Do you feel differently about that? I could go either way!

tests/close_events.rs
kleimkuhler: Yea I initially took this approach and had some issues with errors getting buried by the call through `smoke_test_tcp_listener`. I may have had it set up wrong I can try changing it back.

src/interests.rs
kleimkuhler: @carllerche and I briefly talked about this in Discord--short back and forth but basically was we may want to avoid registering interest in events that the user doesn't ask for.

I realize now that on `kqueue` platforms we may need to `cfg` these methods since this is tracked by default, in which case I may prefer having `EPOLLRDHUP` automatically register if `READABLE` interest is registered.

src/event/event.rs
Thomasdezeeuw: I've always preferred readable and writable, reading it as "the event indicates the socket is readable", so it that case closed ("the event indicates the socket is closed") made more sense to me. But I don't really feel strongly either way.

kleimkuhler: I changed the method names as well as constants to be past tense

src/event/event.rs
kleimkuhler: Docs have been updated to included the events when these methods _should_ return `true`

tests/tcp_stream.rs
```diff
@@ -494,8 +614,9 @@ fn echo_listener(addr: SocketAddr, n_connections: usize) -> (thread::JoinHandle<
 fn start_listener(
     n_connections: usize,
     barrier: Option<Arc<Barrier>>,
+    drop_write: bool,
```
Thomasdezeeuw: `shutdown_write`?

tests/uds.rs
```diff
@@ -2,17 +2,26 @@
...
 
-use mio::net::{UnixDatagram, UnixListener, UnixStream};
-use mio::unix::SocketAddr;
-use mio::{Interests, Token};
-use std::io::{self, IoSlice, IoSliceMut, Read, Write};
-use std::net::Shutdown;
-use std::sync::mpsc::{channel, Receiver};
-use std::thread;
+use log::warn;
+use mio::{
```
Thomasdezeeuw: Same for these imports as for the TCP tests.

tests/tcp_stream.rs
```diff
@@ -452,11 +452,131 @@ fn deregistering() {
...
+    barrier.wait();
+    handle.join().expect("failed to join thread");
+}
+
 /// Start a listener that accepts `n_connections` connections on the returned
 /// address. It echos back any data it reads from the connection before
 /// accepting another one.
 fn echo_listener(addr: SocketAddr, n_connections: usize) -> (thread::JoinHandle<()>, SocketAddr) {
-    let (sender, receiver) = channel();
+    let (sender, receiver) = mpsc::channel();
```
Thomasdezeeuw: Change not needed, also in the UDS tests.

tests/util/mod.rs
```diff
@@ -12,6 +12,8 @@ use log::{error, warn};
 use mio::event::Event;
 use mio::{Events, Interests, Poll, Token};
 
+pub const TIMEOUT: Option<Duration> = Some(Duration::from_millis(500));
```
Thomasdezeeuw: Why do we need such a large timeout? This makes `expect_no_events` take 10 times longer.

I'm also not the biggest fan of one timeout fits all, it's often not the case.

tests/tcp_stream.rs
```diff
@@ -1,19 +1,19 @@
...
-use mio::{Interests, Token};
-
+#[macro_use]
 mod util;
 
+use log::warn;
+use mio::{net::TcpStream, Interests, Token};
+#[cfg(unix)]
+use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd};
+use std::{
```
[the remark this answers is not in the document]
kleimkuhler: Okay sounds good. Yea not a major reason to include this in the PR, but I feel like the new lines and deciding when to nest imports vs split them can make it less readable sometimes.

tests/tcp_stream.rs
```diff
@@ -1,19 +1,19 @@
...
-#[cfg(unix)]
-use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd};
-use std::sync::mpsc::channel;
-use std::sync::{Arc, Barrier};
-use std::thread;
-
-use mio::net::TcpStream;
-use mio::{Interests, Token};
-
+#[macro_use]
```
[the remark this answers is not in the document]
kleimkuhler: I believe `#[macro_use]` is still required for macros defined within the the crate that a module is trying to use it in. Imports only work for macros defined in external crates.

tests/util/mod.rs
kleimkuhler: Good point. It will be changed back

kleimkuhler:
@Thomasdezeeuw @carllerche This is ready for review now. I updated the PR comment to include a summary of what these methods check for. Most of this was through the data that #1091 exposes, but these events are fairly independent of each other.

tests/util/mod.rs
```diff
@@ -56,6 +56,32 @@ macro_rules! assert_err {
...
+                    found = true;
+                    break 'outer;
+                } else {
+                    // Accept sporadic events.
+                    warn!("got unexpected event: {:?}", event);
+                }
+            }
+        }
+        if !found {
+            panic!("failed to find ")
```
Thomasdezeeuw: It think something is missing here.

tests/util/mod.rs
```diff
@@ -56,6 +56,32 @@ macro_rules! assert_err {
     }};
 }
 
+/// Expect an event that may not be represented by an Interest, such as
+/// `is_read_closed` and `is_write_closed` on some platforms.
+#[macro_export]
+macro_rules! expect_flaky_event {
```
Thomasdezeeuw: We're really abusing `Interests` in `ExpectEvent`, so in the future we should merge this with `expect_event` and just make a readiness type for testing.

Thomasdezeeuw:
If you can rebase on master and fix the small issue with the panic message we can merge this.

src/sys/unix/epoll.rs
```diff
@@ -139,12 +139,20 @@ pub mod event {
         (event.events as libc::c_int & libc::EPOLLERR) != 0
     }
 
-    pub fn is_hup(event: &Event) -> bool {
-        (event.events as libc::c_int & libc::EPOLLHUP) != 0
+    pub fn is_read_closed(event: &Event) -> bool {
+        // Both halves of the socket have closed
+        event.events as libc::c_int & libc::EPOLLHUP != 0
+            // Socket has received FIN or called shutdown(SHUT_RD)
+            || (event.events as libc::c_int & libc::EPOLLIN != 0
```
carllerche: Out of curiosity, is it possible to receive `EPOLLRDHUP` w/o receiving `EPOLLIN`?

kleimkuhler:
@Thomasdezeeuw I merged master instead of rebasing as to not reset the diff. It should be good to go

Thomasdezeeuw:
@kleimkuhler GitHub says "This branch cannot be rebased due to conflicts". Also we don't usually squash commits but rebased them on the master branch.

Can you locally run:
```bash
git fetch upstream # assuming upstream is tokio-rs/mio
git rebase upstream/master
git push
```

Thomasdezeeuw:
Great! Thanks for all the work.

PR #1111, opened 2019-10-25
Unlink bound path on uds drop

jean-airoldie:
This allows a user to reuse a bound path after the uds was dropped
without having to do any cleanup.

### DIFF
--- a/src/net/uds/datagram.rs
+++ b/src/net/uds/datagram.rs
@@ -26,6 +26,11 @@ impl UnixDatagram {
     }
 
     /// Creates a Unix datagram socket bound to the given path.
+    ///
+    /// On success, this creates a socket file at the bound path as a side effect.
+    /// The [`unlink`] method can be used to prevent leaking this file.
+    ///
+    /// [`unlink`]: #method.unlink
     pub fn bind<P: AsRef<Path>>(path: P) -> io::Result<UnixDatagram> {
         let sys = sys::UnixDatagram::bind(path.as_ref())?;
         Ok(UnixDatagram::new(sys))
@@ -116,6 +121,14 @@ impl UnixDatagram {
     /// (see the documentation of `Shutdown`).
     pub fn shutdown(&self, how: Shutdown) -> io::Result<()> {
         self.sys.shutdown(how)
+    }
+
+    /// Consumes the socket, unlinking the bound socket file, if there is one.
+    ///
+    /// Note that a socket file is only removed once all references to it are
+    /// closed.
+    pub fn unlink(self) {
+        self.sys.unlink()
     }
 }
 
--- a/src/net/uds/listener.rs
+++ b/src/net/uds/listener.rs
@@ -27,6 +27,11 @@ impl UnixListener {
     }
 
     /// Creates a new `UnixListener` bound to the specified socket.
+    ///
+    /// On success, this creates a socket file at the bound path as a side effect.
+    /// The [`unlink`] method can be used to prevent leaking this file.
+    ///
+    /// [`unlink`]: #method.unlink
     pub fn bind<P: AsRef<Path>>(path: P) -> io::Result<UnixListener> {
         let sys = sys::UnixListener::bind(path.as_ref())?;
         Ok(UnixListener::new(sys))
@@ -59,6 +64,14 @@ impl UnixListener {
     /// Returns the value of the `SO_ERROR` option.
     pub fn take_error(&self) -> io::Result<Option<io::Error>> {
         self.sys.take_error()
+    }
+
+    /// Consumes the socket, unlinking the bound socket file, if there is one.
+    ///
+    /// Note that a socket file is only removed once all references to it are
+    /// closed.
+    pub fn unlink(self) {
+        self.sys.unlink()
     }
 }
 
--- a/src/sys/unix/uds/datagram.rs
+++ b/src/sys/unix/uds/datagram.rs
@@ -4,11 +4,11 @@ use crate::sys::unix::net::new_socket;
 use crate::unix::SourceFd;
 use crate::{Interests, Registry, Token};
 
-use std::io;
 use std::net::Shutdown;
 use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
 use std::os::unix::net;
 use std::path::Path;
+use std::{io, mem};
 
 #[derive(Debug)]
 pub struct UnixDatagram {
@@ -128,6 +128,29 @@ impl UnixDatagram {
 
     pub(crate) fn shutdown(&self, how: Shutdown) -> io::Result<()> {
         self.inner.shutdown(how)
+    }
+
+    pub(crate) fn unlink(mut self) {
+        // Retrieve the local bound and drop the socket.
+        let maybe_path = {
+            let socket = mem::replace(&mut self.inner, unsafe {
+                net::UnixDatagram::from_raw_fd(-1)
+            });
+            match socket.local_addr() {
+                Ok(addr) => match addr.as_pathname() {
+                    Some(path) => Some(path.to_owned()),
+                    None => None,
+                },
+                Err(_) => None,
+            }
+        };
+
+        // If the socket was bound to a named path, unlink it.
+        if let Some(path) = maybe_path {
+            let _ = std::fs::remove_file(path);
+        }
+
+        mem::forget(self);
     }
 }
 
--- a/src/sys/unix/uds/listener.rs
+++ b/src/sys/unix/uds/listener.rs
@@ -127,6 +127,29 @@ impl UnixListener {
     pub(crate) fn take_error(&self) -> io::Result<Option<io::Error>> {
         self.inner.take_error()
     }
+
+    pub(crate) fn unlink(mut self) {
+        // Retrieve the local bound and drop the socket.
+        let maybe_path = {
+            let socket = mem::replace(&mut self.inner, unsafe {
+                net::UnixListener::from_raw_fd(-1)
+            });
+            match socket.local_addr() {
+                Ok(addr) => match addr.as_pathname() {
+                    Some(path) => Some(path.as_os_str().to_owned()),
+                    None => None,
+                },
+                Err(_) => None,
+            }
+        };
+
+        // If the socket was bound to a named path, unlink it.
+        if let Some(path) = maybe_path {
+            let _ = std::fs::remove_file(path);
+        }
+
+        mem::forget(self);
+    }
 }
 
 impl Source for UnixListener {
--- a/tests/uds.rs
+++ b/tests/uds.rs
@@ -107,6 +107,30 @@ fn is_send_and_sync() {
 }
 
 #[test]
+fn datagram_unlink() {
+    let path = "bar";
+
+    let datagram = assert_ok!(UnixDatagram::bind(path));
+    assert_err!(UnixDatagram::bind(path));
+    datagram.unlink();
+
+    let datagram = assert_ok!(UnixDatagram::bind(path));
+    datagram.unlink();
+}
+
+#[test]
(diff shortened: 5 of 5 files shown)


Thomasdezeeuw:
Is this something we want to do? In other words are there any use cases in which we don't always want to unlink?

Also this adds an allocation and makes the types larger, which wouldn't make this zero cost anymore.

Thomasdezeeuw:
@kleimkuhler what do you think?

jean-airoldie:
Note that unlink doesnt do anything if another open socket refers to the same fd. So I couldn't think of a use case where you wouldn't want such RAII semantics.

There might be a way to avoid the allocation by retrieving the local_address of the bound socket in the Drop implementation. But the socket needs to be closed before unlink is called, and I could not figure out how to do that.

Thomasdezeeuw:
> So I couldn't think of a use case where you wouldn't want such RAII semantics.

I think we should think the other way around, i.e. why would we want such semantics, as doing nothing is always faster. In Mio we try to match the standard library went it comes to API and as far as I can tell it doesn't unlink anything either, does it?

jean-airoldie:
> it doesn't unlink anything either, does it

No, otherwise this wouldn't be needed.

I can understand the concern for the additional allocation, but AFAIK dropping a socket will always be part of a cold path, so paying an additional syscall for convenience seems like a decent trade.

jean-airoldie:
> I think we should think the other way around, i.e. why would we want such semantics, as doing nothing is always faster.

One could argue that calling close on a socket falls in the same category. Not calling close on a socket would leak the fd, but its always faster. In this situation, no calling unlink of the bound path leaks the file, but its always faster.

jean-airoldie:
Ideally this should be done in the drop impl of the std UnixDatagram, but I think this is considered a breaking change, so I'm not confident it would be merged.

kleimkuhler:
I think the fact that `mio` tries to match `std` is in an important thing to keep in mind here. Also as @Thomasdezeeuw mentioned I think we should keep these types zero-cost.

That being said, I wasn't aware that this tends to be such a difficult problem when using UDS. After looking around online for how others have dealt with this, it seems like there is no great solution.

@jean-airoldie I like that we have the ability to use a `Drop` impl here and it should be doable after some quick playing around with it. The only issue that comes up is the `IntoRawFd` impl which becomes something along the lines of:

```
impl IntoRawFd for UnixListener {
    fn into_raw_fd(mut self) -> RawFd {
        let fd = mem::replace(&mut self.inner, unsafe { mem::uninitialized() });
        mem::forget(self);
        fd.into_raw_fd()
    }
}
```

That was just to make my changes compile; there may still need to be path cleanup there?

I also wouldn't worry about this being a breaking change, because `0.7` already has a host of breaking changes!

So I think this is something we should consider _if_ it can be a zero-cost solution. I think it should also be documented very clearly because this is not standard behavior.

carllerche:
@jean-airoldie Thanks for the PR. I see how this could arguably be a good thing. Could you provide an argument for why Mio should deviate from `std` in this case? Perhaps, the first step would be to propose the change to `std`.

jean-airoldie:
For my personal use case, I'm not using `mio` directly so I solved the problem from my end using a similar approach. So I don't depend on this getting merged.

I tend to agree that the right solution would be a change in the `std`. I've never contributed directly to the `std` but I assume that it would take quite a while for the change to merged and then deployed. However, I'm wondering whether this is enough of a usability problem that it would be worth it to fix it here also so it propagates to `tokio` faster (assuming such a changed would get merged by std).

My general argument for this that the current behavior of leaking the file makes the lifecycle of a bound UnixDatagram non-idempotent. If you look at the tests in mio involving unix domain sockets, they all involve using a temp dir so that its not an issue. Since this leaked file cannot be reused by a socket, Its completely useless from the user's perspective. I would also argue that its rather unexpected behavior.

I didn't think about `mem::replace` with uninitialized memory, it seems very obvious now that I see it @kleimkuhler. I'll rewrite the PR.

src/sys/unix/uds/datagram.rs
```diff
@@ -26,7 +27,9 @@ impl UnixDatagram {
         let sockaddr = &sockaddr as *const libc::sockaddr_un as *const libc::sockaddr;
 
         syscall!(bind(socket, sockaddr, socklen))?;
-        Ok(unsafe { UnixDatagram::from_raw_fd(socket) })
+        let inner = unsafe { net::UnixDatagram::from_raw_fd(socket) };
```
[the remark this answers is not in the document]
jean-airoldie: Not sure what i was thinking.

jean-airoldie:
I'd like also point out the lack of `SO_REUSEADDR` for `AF_UNIX` sockets, so its not like you could do anything with that leaked file anyway.

src/net/uds/datagram.rs
```diff
@@ -26,6 +26,10 @@ impl UnixDatagram {
     }
 
     /// Creates a Unix datagram socket bound to the given path.
+    ///
+    /// This operation create a socket file at the given path which will
```
Thomasdezeeuw: create*s*.

src/net/uds/listener.rs
```diff
@@ -27,6 +27,10 @@ impl UnixListener {
     }
 
     /// Creates a new `UnixListener` bound to the specified socket.
+    ///
+    /// This operation create a socket file at the given path which will
```
Thomasdezeeuw: Also create*s*.

src/sys/unix/uds/datagram.rs
```diff
@@ -163,7 +164,23 @@ impl FromRawFd for UnixDatagram {
 }
 
 impl IntoRawFd for UnixDatagram {
-    fn into_raw_fd(self) -> RawFd {
-        self.inner.into_raw_fd()
+    fn into_raw_fd(mut self) -> RawFd {
+        let fd = mem::replace(&mut self.inner, unsafe {
+            mem::MaybeUninit::uninit().assume_init()
```
Thomasdezeeuw: We could use`net::UnixDatagram::from_raw_fd(-1)`, then at least we don't have any memory unsafety.

jean-airoldie:
> Have any issues been opened to include the behaviour in std lib?

Fair point, I'll open an issue.

jean-airoldie:
The `net::UnixDatagram::from_raw_fd(-1)` approach doesn't seem to be working for macos.

jean-airoldie:
So I opened an issue in the std lib and the issue was closed with the conclusion is that:

> We generally want to adhere to the underlying system primitives as closely as possible so that things behave as expected to people who know how the underlying system primitives behave.

So in my understanding, the reasoning was more to leave the posix API more or less untouched in order to maximize its familiarity.

Thomasdezeeuw:
Personally I think we should follow what the standard library is doing, adding documentation (and perhaps a method) that manual removal of the socket is required.

jean-airoldie:
How would such a method work? The only reason this is an issue in the first place is that you have to close the socket before being able to unlink it. I guess that method could return a guard that unlinks the path when drop, but that's pretty weird. You would still have to make sure that the socket is dropped before the guard.

From what I understand, in the kernel, people deal with this problem by calling unlink on each path before binding it, which seems like a huge pain for the users.

Let's say there was a problem in the posix spec, which I would argue that this is, should `mio` fix it for the users? Or should we let every user fix the issue themselves so that we maintain familiarity? We still don't have any example where this behavior would be desirable.

Thomasdezeeuw:
> How would such a method work? The only reason this is an issue in the first place is that you have to close the socket before being able to unlink it. I guess that method could return a guard that unlinks the path when drop, but that's pretty weird. You would still have to make sure that the socket is dropped before the guard.

I was thinking a method like: `fn unlink(self) { }` which would close itself and then unlink, essentially what this change proposes to do by default when dropped.

> Let's say there was a problem in the posix spec, which I would argue that this is, should `mio` fix it for the users? Or should we let every user fix the issue themselves so that we maintain familiarity? We still don't have any example where this behavior would be desirable.

I think Mio is complex enough as is, I don't we should deviate from the behaviour from std lib to make it even more complex. But that is my opinion, I don't know if everyone agrees with that.

jean-airoldie:
> I was thinking a method like: fn unlink(self) { } which would close itself and then unlink, essentially what this change proposes to do by default when dropped.

I would be satisfied by that. Its kinda weird but it would do the trick.

src/sys/unix/uds/datagram.rs
```diff
@@ -129,6 +129,29 @@ impl UnixDatagram {
     pub(crate) fn shutdown(&self, how: Shutdown) -> io::Result<()> {
         self.inner.shutdown(how)
     }
+
+    pub(crate) fn unlink(mut self) {
+        // Retrieve the local bound and drop the socket.
+        let maybe_path = {
```
kleimkuhler: Nit: I''m not a fan of the name `maybe_path`. It's `maybe_path` because it's an `Option<Path>`, but the naming of `Option<T>` types don't usually follow a `maybe_...` naming convention. How about just `path` instead?

src/sys/unix/uds/datagram.rs
```diff
@@ -129,6 +129,29 @@ impl UnixDatagram {
     pub(crate) fn shutdown(&self, how: Shutdown) -> io::Result<()> {
         self.inner.shutdown(how)
     }
+
+    pub(crate) fn unlink(mut self) {
```
kleimkuhler: I have a comment about naming in this method, but I think it could be structured differently. In situations where there is a `None => None` and (or) `Err(_) => None`, `map` is probably going to making the branching read a little nicer.

How about something like:
```
let socket = mem::replace(&mut self.inner, unsafe {
    net::UnixListener::from_raw_fd(-1)
});
let _ = socket.local_addr().map(|addr| {
    addr.as_pathname().map(|path| {
        let path = path.as_os_str().to_owned();
        let _ = std::fs::remove_file(path);
    })
});
mem::forget(self);
```

I think `match` and `if let` can certainly improve readability, but in this case there is no action to take on the `None` or `Err` variants.

This applies to the `UnixListener` method as well.

jean-airoldie: The problem is that the socket has to be dropped first before we can unlink. Also I realized that `remove_file` doesn't work because it remove unconditionally the socket file, which is not what we want. I'll add a test case.

jean-airoldie:
Alright so turns out that you can straight up delete the socket file of a bound socket which will silently delete the socket without returning an error. This means this mechanism can't work because its possible that you duplicated the socket.

jean-airoldie:
The only solution I can think of would involve reference counting the number of socket dups but its not a zero-cost abstraction, so it wouldn't fit in mio.

jean-airoldie:
This behavior isn't consistent with the `unlink` method man:
>  unlink() deletes a name from the filesystem.  If that name was the last link to a file and no processes have the file open, the file is deleted and the space it was using is made available for reuse.

src/sys/unix/uds/datagram.rs
jean-airoldie: Nevermind, see my comment below.

kleimkuhler:
Yea I see cloning the listener in `echo_remote` and `unlink`ing the original one leads to a `NotFound` error. That is unfortunate.. I imagine the man is lacking additional detail on what exactly "no processes have the file open" means. I assume since `try_clone` is just `dup` under the hood, duplicating the file descriptor is not enough to keep the file from being deleted. Maybe opening another file _description_ would do the trick, but definitely out of the scope of what `mio` should do at this point.

jean-airoldie:
Yeah that's clearly out of the scope of mio, I'll close the issue. Thanks for your time.

I'll probably keep track of the refcount via a atomic counter living in shared memory, then unlink the path when the counter would reach 0.

Thomasdezeeuw:
Is there any documentation we should add based on this?

PR #1086, opened 2019-09-16
Windows cleanup

Thomasdezeeuw:

PR #1112, opened 2019-10-29
Windows cleanup

Thomasdezeeuw:
Various cleanups/improvements in the Windows implementation. I would recommend review this on a commit by commit based, the commits themselves are fairly small, but adding them together it becomes a large change.


Thomasdezeeuw:
This is sort of a follow up of #1086, but this code actually works (according to the CI as well). :)

src/sys/windows/event.rs
```diff
@@ -6,7 +6,10 @@ pub struct Event {
 
 use miow::iocp::CompletionStatus;
 
-use super::afd;
+use super::afd::{
+    AFD_POLL_ABORT, AFD_POLL_ACCEPT, AFD_POLL_CONNECT_FAIL, AFD_POLL_DISCONNECT, AFD_POLL_RECEIVE,
```
carllerche: take it or leave it (not a bit deal either way): You could drop the `AFD_` prefix from the constants and then reference them as `afd::POLL_ABORT`.

src/sys/windows/udp.rs
```diff
@@ -13,7 +13,7 @@ use std::{fmt, io};
 use winapi::um::winsock2::{bind, closesocket, SOCKET_ERROR, SOCK_DGRAM};
 
 pub struct UdpSocket {
-    internal: Arc<Mutex<Option<InternalState>>>,
+    internal: Mutex<Option<InternalState>>,
```
carllerche: What is the size of the struct now?

Longer term, I think we can optimize `InternalState` to be a single arc, but that can happen later.

src/sys/windows/selector.rs
```diff
@@ -405,10 +404,8 @@ impl Selector {
 
 #[derive(Debug)]
 pub struct SelectorInner {
-    lock: Mutex<()>,
     cp: Arc<CompletionPort>,
-    active_poll_count: UnsafeCell<usize>,
```
carllerche: Isn't this required for correctness?

src/sys/windows/event.rs
Thomasdezeeuw: I like that, I'll change it.

src/sys/windows/udp.rs
Thomasdezeeuw: Currently it's 56 bytes (https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2058e09e5c44b46dca119ca7165dec16). In the future it might change to use parking lot (https://github.com/rust-lang/rust/pull/56410) in which it's a single bit/byte.

Want to change to a `Box`? That brings it down to 16 (on the stack).

src/sys/windows/selector.rs
Thomasdezeeuw: No tests failed, so if it is we don't have sufficient test coverage to show that.

src/sys/windows/event.rs
```diff
@@ -17,27 +14,28 @@ pub fn token(event: &Event) -> Token {
 }
 
 pub fn is_readable(event: &Event) -> bool {
-    event.flags & (AFD_POLL_RECEIVE | AFD_POLL_DISCONNECT | AFD_POLL_CONNECT | AFD_POLL_ACCEPT) != 0
+    event.flags & (afd::POLL_RECEIVE | afd::POLL_DISCONNECT | afd::POLL_CONNECT | afd::POLL_ACCEPT)
+        != 0
```
Thomasdezeeuw: Rustfmt :(

src/sys/windows/mod.rs
```diff
@@ -22,6 +22,20 @@ macro_rules! syscall {
...
+    ($self: ident, $method: ident $(, $args: expr)*)  => {{
+        let result = (&$self.inner).$method($($args),*);
+        if let Err(ref e) = result {
+            if e.kind() == io::ErrorKind::WouldBlock {
+                $self.io_blocked_reregister()?;
+            }
+        }
+        result
+    }};
+}
```
PerfectLaugh: I think it can also reregister on successful operation.

```
macro_rules! try_io {
    ($self: ident, $method: ident $(, $args: expr)*)  => {{
        let result = (&$self.inner).$method($($args),*);
        if let Err(ref e) = result {
            if e.kind() == io::ErrorKind::WouldBlock {
                $self.io_blocked_reregister()?;
            }
        } else {
            $self.io_blocked_reregister()?;
        }
        result
    }};
}
```

src/sys/windows/event.rs
```diff
@@ -1,46 +1,41 @@
...
 pub fn token(event: &Event) -> Token {
     Token(event.data as usize)
 }
 
 pub fn is_readable(event: &Event) -> bool {
-    if is_error(event) || is_read_hup(event) {
-        return true;
-    }
-    event.flags & (afd::AFD_POLL_RECEIVE | afd::AFD_POLL_ACCEPT) != 0
+    event.flags & (afd::POLL_RECEIVE | afd::POLL_DISCONNECT | afd::POLL_CONNECT | afd::POLL_ACCEPT)
```
piscisaureus: I don't think ``POLL_CONNECT` should be in here, for two reasons:

* `POLL_CONNECT` indicates that an outbound stream socket has been successfully established. However this does not imply that the server actually sent any data that can be read from the socket. It's very common for this not to be the case (e.g. http servers will not send any data to a client until they've received the request header). Note that select and epoll also don't indicate readability on a connected socket (they will indicate *writability* only).
* The `POLL_CONNECT` flag will go up once a connection has been made, and from there on it *stays up* (remember this is an edge triggered interface). This means that after this patch a connected socket will report that it is *always readable*. I've made this mistake myself in the past (see https://github.com/piscisaureus/wepoll/commit/8b7b340610f88af3d83f40fb728e7b850b090ece).
* Since this bug is apparently not caught by the test, it might be worthwhile to port [this test](https://github.com/piscisaureus/wepoll/blob/d5f8f5f1b1be1a4ba8adb51eb4ee4de7a305a9c8/test/test-connect-success-events.c) to mio.

piscisaureus: The tried-and-true mappings that wepoll uses: https://github.com/piscisaureus/wepoll/blob/d5f8f5f1b1be1a4ba8adb51eb4ee4de7a305a9c8/src/sock.c#L160-L200

src/sys/windows/mod.rs
Thomasdezeeuw: Do you mean we have to register on successful operations? Because I've kept the meaning of the original code.

src/sys/windows/event.rs
Thomasdezeeuw: I've updated the mappings. Is there are (publicly available) documentation on AFD? I've been searching but couldn't find anything beyond libuv and wepoll.

src/sys/windows/mod.rs
PerfectLaugh: Because the events are deregistered on event received. We should reregister on every successful operation.

Thomasdezeeuw: Maybe we want to always reregister then? I'm not sure if we get more events on kqueue/epoll after an I/O operation returned an error.

Thomasdezeeuw: Are there are cases in which we can't reregister?

PerfectLaugh: I don't think notification will come up if we failed to perform these operations, so I want to let the reregister happened on sucessful registeration.

Thomasdezeeuw: Done.

PerfectLaugh: There are some place to modify to pass the test suite (except try_clone on TcpListener and several minors):
mark delete the sockstate in deregister.
wait for writable event on ttl test in tcp_stream.rs

I will send pr after you done this

PerfectLaugh:
I'm not quite sure. But accept will fail if you successfully accepted one
connection and continue to listen for accept event

src/sys/windows/mod.rs
PerfectLaugh: I'm not quite sure. But accept will fail if you successfully accepted one connection and continue to listen for accept event

src/sys/windows/event.rs
piscisaureus: > I've updated the mappings. Is there are (publicly available) documentation on AFD? I've been searching but couldn't find anything beyond libuv and wepoll.

There isn't, but the events listed in the [WSAAsyncSelect](https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-wsaasyncselect) docs map 1:1 to `IOCTL_AFD_POLL` events. The only event missing are `AFD_POLL_ABORT` and `AFD_POLL_DISCONNECT`, which are used to indicate that, respectively, a `RST` or `FIN` packet has been received.

piscisaureus:
> Looks great 👍, but I did have a question about removing `active_poll_count` as, my understanding, is that it is required for correctness. I pinged @piscisaureus as he has the best understanding of the AFD stuff.

The `active_poll_count` counter is used in wepoll to ensure that when *multiple* threads are blocked on the same completion port, and then an event happens, they *all* wake up (instead of just one thread receiving the event). This is to match epoll "thundering herd" behavior. However it's not super efficient so I think that mio might be better off without that logic. 

But TBH I don't know enough about mio internals to know for sure.

Thomasdezeeuw:
> The `active_poll_count` counter is used in wepoll to ensure that when _multiple_ threads are blocked on the same completion port, and then an event happens, they _all_ wake up (instead of just one thread receiving the event). This is to match epoll "thundering herd" behavior. However it's not super efficient so I think that mio might be better off without that logic.

Well we don't set `EPOLLEXCLUSIVE` on `epoll(2)` either so we have the same problem there.

Thomasdezeeuw:
Since the review comments slowed down I'm going to rebase this on master.

src/sys/windows/mod.rs
Thomasdezeeuw: I've dropped the commit from the rebased commits I've pushed to see if the CI is green.

piscisaureus:
> Well we don't set EPOLLEXCLUSIVE on epoll(2) either so we have the same problem there.

EPOLLEXCLUSIVE is not quite the same thing. It controls whether multiple threads wake up when they're waiting on *different* epoll FDs. The counter is used to wake up all threads that wait on the *same* port.

Thomasdezeeuw:
> EPOLLEXCLUSIVE is not quite the same thing. It controls whether multiple threads wake up when they're waiting on _different_ epoll FDs. The counter is used to wake up all threads that wait on the _same_ port.

You're right, but since `Poll::poll` takes `&mut self` the only way to poll concurrently is to duplicate the underlying file descriptor and then you're using two different epoll fds. Or using the fd directly in a call to `epoll_wait(2)`, but I'm not sure we if we currently really support that.

In any case concurrently polling the same fd/port shouldn't be possible when only using the safe api of Mio. So I don't think this is a problem, or am I missing something else?

src/sys/windows/event.rs
dtacalau: > I've updated the mappings. Is there are (publicly available) documentation on AFD? I've been searching but couldn't find anything beyond libuv and wepoll.

@Thomasdezeeuw not sure if you found this before: https://github.com/DeDf/afd . Seems to be the afd implementation.

piscisaureus:
> `Poll::poll` takes `&mut self` 

That's great. You can't duplicate a IOCP on windows, so that at least tells us that there's no reason to keep the `active_poll_count` on Windows.

> the only way to poll concurrently is to duplicate the underlying file descriptor and then you're using two different epoll fds.

I think that if you duplicate an epoll fd on linux, you're still polling the same epoll set and therefore the EPOLLEXCLUSIVE flag would still have no effect. But luckily this all doesn't matter, because this isn't about linux.

src/sys/windows/selector.rs
Thomasdezeeuw: According to @piscisaureus this is ok since we use `&mut self` in `Poll::poll`, see https://github.com/tokio-rs/mio/pull/1112#issuecomment-549620142

src/sys/windows/udp.rs
Thomasdezeeuw: @carllerche what do you want to here? I don't know how much the net types are moved around normally.

Thomasdezeeuw:
@piscisaureus I've added a note about the safe usage of Windows' `Selector::select`.

Thomasdezeeuw:
@dtacalau I can't seem to request your re-review via GitHub, but can you take another look at this?

src/sys/windows/selector.rs
```diff
@@ -591,7 +575,8 @@ impl SelectorInner {
     }
 
     unsafe fn update_sockets_events_if_polling(&self) -> io::Result<()> {
```
piscisaureus: Since active_poll_count has been removed, this function and all its call sites can be removed too.

Thomasdezeeuw: Done.

Thomasdezeeuw:
@piscisaureus I've created #1128.

PR #1113, opened 2019-10-29
Fix issues pointed out by Clippy

Thomasdezeeuw:
This makes clippy happy again. Unneeded clone and an exhaustive match of
a return value.

### DIFF
--- a/src/sys/unix/uds/datagram.rs
+++ b/src/sys/unix/uds/datagram.rs
@@ -4,7 +4,6 @@ use crate::sys::unix::net::new_socket;
 use crate::unix::SourceFd;
 use crate::{Interests, Registry, Token};
 
-use std::cmp::Ordering;
 use std::io;
 use std::net::Shutdown;
 use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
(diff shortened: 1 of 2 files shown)


src/sys/unix/uds/datagram.rs
```diff
@@ -99,12 +100,10 @@ impl UnixDatagram {
...
-            if count > 0 {
-                Ok(1)
-            } else if count == 0 {
-                Ok(0)
-            } else {
-                Err(io::Error::last_os_error())
+            match count.cmp(&0) {
+                Ordering::Greater => Ok(1),
+                Ordering::Equal => Ok(0),
+                Ordering::Less => Err(io::Error::last_os_error()),
```
kleimkuhler: Thoughts on this?

```suggestion
if count == -1 {
    Err(io::Error::last_os_error())
} else {
    Ok(count as i32)
}
```

It's similar to what `syscall!` does under the hood so it may be a more expected handling of the result value (at least when reading the code).

Thomasdezeeuw: I would have done that but the original code returned `Ok(1)` if `count > 0`, that is why I asked on Discord.

Thomasdezeeuw: I've changed it to use the `syscall!` macro if the result doesn't matter.

kleimkuhler:
@Thomasdezeeuw I'm looking into this UDS test. I don't think it should be flakey, but it's making my PR fail as well. Sorry about that

kleimkuhler:
Should be good after merging master!

Thomasdezeeuw:
I've rebased on master, will merge once CI is green.

PR #1115, opened 2019-10-29
Set Rustup profile to minimal in CI

### LINKED ISSUE #1100: Use Rustup profiles in CI
Rustup 1.20 was released (https://blog.rust-lang.org/2019/10/15/Rustup-1.20.0.html) which includes support for profiles. It includes a minimal profile which should be used in CI environments to reduce the amount of stuff downloaded by rustup. Could possibly fix #1044.

Thomasdezeeuw:
Closes #1100.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -9,7 +9,7 @@ task:
   setup_script:
     - pkg install -y curl
     - curl https://sh.rustup.rs -sSf --output rustup.sh
-    - sh rustup.sh -y
+    - sh rustup.sh -y --profile minimal
   cargo_cache:
     folder: $HOME/.cargo/registry
   build_script:
--- a/ci/azure-install-rust.yml
+++ b/ci/azure-install-rust.yml
@@ -2,7 +2,7 @@ steps:
   # Linux and macOS.
   - script: |
       set -e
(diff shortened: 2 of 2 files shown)


Thomasdezeeuw:
Doesn't seem to speed up the build that much, but it couldn't hurt I guess.

.cirrus.yml
```diff
@@ -10,6 +10,8 @@ task:
     - pkg install -y curl
     - curl https://sh.rustup.rs -sSf --output rustup.sh
     - sh rustup.sh -y
```
asomers: By that point it's too late, because line 12 already installs the default profile.  Do `sh rustup.sh -y --profile=minimal` instead.

ci/azure-install-rust.yml
```diff
@@ -4,6 +4,7 @@ steps:
       set -e
       curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain none
```
asomers: Similarly here.  Add `--profile=minimal` to this line.

.cirrus.yml
Thomasdezeeuw: Ah, didn't know that. I've force pushed since the change is so small. Let's see if this does have an impact.

Thomasdezeeuw:
Still not really a big change in build times as far as I can see.

PR #1116, opened 2019-10-29
Once the UDS test listener reads 0, do not try to write

kleimkuhler:
This also removes the expectation that a new READABLE event is generated
which should not be the case.


kleimkuhler:
@Thomasdezeeuw I think this should fix the flakiness that is pretty much failing three of the PRs open right now. Passing once is a little tough to interpret, but I ran it on a Mac a lot locally without failure. The error happening would make sense as well without this check.

kleimkuhler:
#1110 Just merged this branch in and passed as well

PR #1119, opened 2019-11-01
chore: better group files by component

carllerche:
Files related to a single component are grouped together in a module.
This was done already for some components, such as UDP.


src/net/tcp/mod.rs
```diff
@@ -0,0 +1,6 @@
+mod listener;
+pub use self::listener::TcpListener;
+
+mod stream;
+pub use self::stream::TcpStream;
+
```
Thomasdezeeuw: Extra empty line.

src/sys/unix/selector/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[cfg(any(
+    target_os = "dragonfly",
+    target_os = "freebsd",
+    target_os = "ios",
+    target_os = "macos",
+    target_os = "netbsd",
+    target_os = "openbsd"
+))]
+pub use self::kqueue::{event, Event, Selector};
```
Thomasdezeeuw: Maybe `Events` should be move here as well? I know that would make a somewhat useless re-export, but then we have it all in one place.

Issue #1117, opened 2019-10-30
tcp_stream test fails with TcpStream::set_ttl crash


dtacalau:
Problem:
When running the unit tests on Windows 8.1 Pro, tcp_stream ttl test crash in TcpStrea::set_ttl :  
```
RUST_BACKTRACE=1 cargo test --test tcp_stream
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
     Running target\debug\deps\tcp_stream-7340f8b6dbab4dde.exe

running 12 tests
test is_send_and_sync ... ok
test shutdown_write ... ignored
test tcp_stream_ipv4 ... ignored
test tcp_stream_ipv6 ... ignored
test ttl ... FAILED
test registering ... ok
test deregistering ... ok
test nodelay ... ok
test reregistering ... ok
test shutdown_both ... ok
test shutdown_read ... ok
test try_clone ... ok

failures:

---- ttl stdout ----
thread 'ttl' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 10022, kind: InvalidInput, message: "An invalid argument was supplied." }', src\libcore\result.rs:1084:5
stack backtrace:
   0: backtrace::backtrace::trace_unsynchronized
             at C:\Users\VssAdministrator\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.34\src\backtrace\mod.rs:66
   1: std::sys_common::backtrace::_print
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54\/src\libstd\sys_common\backtrace.rs:47
   2: std::sys_common::backtrace::print
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54\/src\libstd\sys_common\backtrace.rs:36
   3: std::panicking::default_hook::{{closure}}
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54\/src\libstd\panicking.rs:200
   4: std::panicking::default_hook
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54\/src\libstd\panicking.rs:211
   5: std::panicking::rust_panic_with_hook
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54\/src\libstd\panicking.rs:477
   6: std::panicking::continue_panic_fmt
[...]

PR #1120, opened 2019-11-04
Fix for issue #1117, temporary fix for set_ttl test on Windows, this …

dtacalau:
…test was crashing on Windows because, on Windows calling setsockopt on a

not connected socket is undefined behavior. More details can be found in issue #1117.


Thomasdezeeuw:
This change looks good to me, two questions:
* Does this apply to more methods on `TcpStream`?
* Can you add some documentation to `TcpStream::set_ttl` (and perhaps other methods) that it's required to receive an before using the method?

dtacalau:
> * Does this apply to more methods on `TcpStream`?

@Thomasdezeeuw you mean other TcpStream methods which are using setsockopt? It so, there's only set_nodelay which suffers from the same issue, but its test has been already fixed.

> * Can you add some documentation to `TcpStream::set_ttl` (and perhaps other methods) that it's required to receive an before using the method?

Absolutely, ATM there are two affected: set_ttl and set_nodelay, others?

Thomasdezeeuw:
If the problem is calling `setsockopt` on an unconnected socket then I think only `set_ttl` and `set_nodelay` are a problem. So only those two methods would need docs.

But does that same hold for `getsockopt`? Then it would also apply to `ttl` and `nodelay` (the getter methods).

dtacalau:
> But does that same hold for `getsockopt`? Then it would also apply to `ttl` and `nodelay` (the getter methods).

Could be, will check all 4 of them.

dtacalau:
Latest changes:
- on Windows, all 4 methods: set ttl, get ttl, set nodelay, get nodelay require a connected stream to work. I've added tests for all of them, also added documentation notes.
- added set/get ttl tests for TcpListener;
- added set/get ttl tests for UdpSocket;

src/net/tcp_stream.rs
```diff
@@ -105,6 +105,9 @@ impl TcpStream {
     /// small amount of data. When not set, data is buffered until there is a
     /// sufficient amount to send out, thereby avoiding the frequent sending of
     /// small packets.
+    ///
+    /// Note On Windows: Make sure the stream is connected before calling this method.
```
Thomasdezeeuw: Can you changes this to:

```rust
    /// # Notes
    /// 
    /// On Windows make sure the stream is connected before calling this method,
    /// by receiving an (writable) event. Trying to set `nodelay` on an
    /// unconnected `TcpStream` is undefined behaviour.
```

Same for the others.

tests/udp_socket.rs
```diff
@@ -111,6 +111,26 @@ fn smoke_test_unconnected_udp_socket(socket1: UdpSocket, socket2: UdpSocket) {
...
+
+#[test]
+fn get_ttl_without_previous_set() {
+    let socket1 = UdpSocket::bind(any_local_address()).unwrap();
+
+    // expect a get TTL to work w/o any previous set_ttl
+    socket1.ttl().expect("unable to get TTL for UDP socket");
+}
+
+
```
Thomasdezeeuw: Double empty line.

dtacalau: Done, removed.

src/net/tcp_stream.rs
dtacalau: Done, reworded.

dtacalau:
> Can you rebase on master and fix the last couple things, then we can merge this.

Done

src/net/tcp_stream.rs
Thomasdezeeuw: You missed the Notes header.

dtacalau: Sorry for that, I've added it.

PR #1121, opened 2019-11-05
Add no_inline doc attribute to Events

Thomasdezeeuw:
This makes it clear that its a re-export from the event module.


kleimkuhler:
Is there any reason we don't re-export `Event`? It seems like if we re-export `Events`, it would make sense for `Event` as well

Thomasdezeeuw:
Mainly because it is usually not imported, while `Events` almost always is.

PR #1123, opened 2019-11-05
split up crate using feature flags

carllerche:
Individual socket types are only implemented when they are requested via
the feature flag. Also, unless the `os-poll` feature is specified, the
crate is a shell.

This is just a spike to illustrate the idea. A lot of the feature flag separation could be cleaner (probably). Also, the "shell" versions of UDS and UDP are not yet done.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,11 @@ exclude       = [
 publish = false
 
 [features]
+os-poll = []
+os-ext = []
+tcp = []
+udp = []
+uds = []
 
 [dependencies]
 log = "0.4.8"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -126,6 +126,9 @@ pub use waker::Waker;
 #[cfg(unix)]
 pub mod unix {
     //! Unix only extensions.
+    #[cfg(feature = "uds")]
     pub use crate::sys::SocketAddr;
+
+    #[cfg(feature = "os-ext")]
     pub use crate::sys::SourceFd;
 }
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -5,17 +5,19 @@
 //! `event`: a module with various helper functions for `Event`, see
 //!          `crate::event::Event` for the required functions.
 
-#[cfg(unix)]
-pub use self::unix::{
-    event, Event, Events, Selector, SocketAddr, SourceFd, TcpListener, TcpStream, UdpSocket,
-    UnixDatagram, UnixListener, UnixStream, Waker,
-};
+#[cfg(all(unix, feature = "os-poll"))]
+pub use self::unix::*;
 
-#[cfg(unix)]
+#[cfg(all(unix, feature = "os-poll"))]
 mod unix;
 
-#[cfg(windows)]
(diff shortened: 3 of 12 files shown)


carllerche:
The `cargo-hack` crate is pretty useful for testing feature flags. I used it w/ the following command:

```
RUSTFLAGS="-Dwarnings" cargo hack check --each-feature
```

I also don't expect CI to necessarily pass at this point.

Cargo.toml
```diff
@@ -24,6 +24,11 @@ exclude       = [
 publish = false
 
 [features]
+os-poll = []
```
Thomasdezeeuw: Do what different features for epoll/kqueue/iocp? In the future we might add other selector, e.g. io_uring on Linux, do we provide any guarantee what selector is used?

src/sys/mod.rs
```diff
@@ -5,17 +5,19 @@
...
+pub use self::windows::*;
 
-#[cfg(windows)]
+#[cfg(all(windows, feature = "os-poll"))]
 mod windows;
+
+#[cfg(not(feature = "os-poll"))]
+mod shell;
+#[cfg(not(feature = "os-poll"))]
+pub(crate) use self::shell::*;
```
Thomasdezeeuw: Just a question; can we use the `compile_error` macro here? I don't if that work when a crate depends on Mio without features though.

src/sys/unix/mod.rs
```diff
@@ -5,7 +5,7 @@ macro_rules! syscall {
     ($fn: ident ( $($arg: expr),* $(,)* ) ) => {{
         let res = unsafe { libc::$fn($($arg, )*) };
         if res == -1 {
-            Err(io::Error::last_os_error())
+            Err(std::io::Error::last_os_error())
```
Thomasdezeeuw: This isn't needed.

src/sys/unix/net.rs
```diff
@@ -1,8 +1,8 @@
-use std::io;
-use std::mem::size_of_val;
+#[cfg(any(feature = "tcp", feature = "udp"))]
```
Thomasdezeeuw: Should this entire module be conditionally used? By adding `cfg` above `mod net` in `mod.rs`?

src/lib.rs
```diff
@@ -126,6 +126,9 @@ pub use waker::Waker;
 #[cfg(unix)]
 pub mod unix {
```
Thomasdezeeuw: Maybe add a conditional in this mod and `mod net` as well, because without any feature these are now two empty modules, so might as well not have them.

src/sys/mod.rs
carllerche: Well. That is a good question. The goal is that libs **dont** pull in the os-poll feature but end executables do. If it is a compile error, running cargo check in the lib would fail. Is there a way to handle that?

Thomasdezeeuw: After a quick test: it fails to compile/check with `cargo check` on a crate the doesn't use `os-poll`, using `compile_error`.

Thomasdezeeuw: Maybe we can use something like https://github.com/japaric/panic-never, that fails a link time. But that can be done later.

carllerche:
Superseded by #1192

PR #1124, opened 2019-11-06
Add from_std methods to TCP/Unix Stream/Listener

kleimkuhler:
## Motivation

#1029 removed the `from_std` methods on the TCP (Unix) Stream (Listener)
resources. These methods are still needed in upstream libraries where creating
wrapped resources from standard types is needed.

Tokio: [TcpListener::from_std](https://github.com/tokio-rs/tokio/blob/master/tokio/src/net/tcp/listener.rs#L195-L199)

### DIFF
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -5,6 +5,7 @@ use crate::{event, sys, Interests, Registry, Token};
 
 use std::fmt;
 use std::io;
+use std::net;
 use std::net::SocketAddr;
 #[cfg(unix)]
 use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
@@ -41,6 +42,20 @@ pub struct TcpListener {
 }
 
 impl TcpListener {
+    /// Creates a new `TcpListener` from a standard `net::TcpListener`.
+    ///
+    /// This function is intended to be used to wrap a TCP listener from the
+    /// standard library in the Mio equivalent. The conversion assumes nothing
+    /// about the underlying listener.
+    pub fn from_std(listener: net::TcpListener) -> Self {
+        let sys = sys::TcpListener::from_std(listener);
+        Self {
+            sys,
+            #[cfg(debug_assertions)]
+            selector_id: SelectorId::new(),
+        }
+    }
+
     /// Convenience method to bind a new TCP listener to the specified address
     /// to receive new connections.
     ///
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -44,20 +44,6 @@ pub struct TcpListener {
 }
 
 impl TcpListener {
-    /// Creates a new `TcpListener` from a standard `net::TcpListener`.
-    ///
-    /// This function is intended to be used to wrap a TCP listener from the
-    /// standard library in the Mio equivalent. The conversion assumes nothing
-    /// about the underlying listener.
-    pub fn from_std(listener: net::TcpListener) -> Self {
-        let sys = sys::TcpListener::from_std(listener);
-        Self {
-            sys,
-            #[cfg(debug_assertions)]
-            selector_id: SelectorId::new(),
-        }
-    }
-
     /// Convenience method to bind a new TCP listener to the specified address
     /// to receive new connections.
     ///
@@ -73,6 +59,21 @@ impl TcpListener {
             #[cfg(debug_assertions)]
             selector_id: SelectorId::new(),
         })
+    }
+
+    /// Creates a new `TcpListener` from a standard `net::TcpListener`.
+    ///
+    /// This function is intended to be used to wrap a TCP listener from the
+    /// standard library in the Mio equivalent. The conversion assumes nothing
+    /// about the underlying listener; ; it is left up to the user to set it
+    /// in non-blocking mode.
+    pub fn from_std(listener: net::TcpListener) -> TcpListener {
+        let sys = sys::TcpListener::from_std(listener);
+        TcpListener {
+            sys,
+            #[cfg(debug_assertions)]
+            selector_id: SelectorId::new(),
+        }
     }
 
     /// Accepts a new `TcpStream`.
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -9,6 +9,8 @@ use std::net;
 use std::net::SocketAddr;
 #[cfg(unix)]
 use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
+#[cfg(windows)]
+use std::os::windows::io::{AsRawSocket, FromRawSocket, IntoRawSocket, RawSocket};
 
 /// A structure representing a socket server
 ///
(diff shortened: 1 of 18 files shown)


Thomasdezeeuw:
Can't this be implemented using `FromRawFd` in Tokio? I know the Windows trait `FromRawSocket` is missing, but is something we need to add.

carllerche:
@Thomasdezeeuw It's not _critical_ to merge this, but IMO it is valuable to have an API for going from std -> mio in a portable way. Requiring going via raw conversions would require platform specific code.

src/net/tcp/listener.rs
```diff
@@ -41,6 +44,20 @@ pub struct TcpListener {
 }
 
 impl TcpListener {
+    /// Creates a new `TcpListener` from a standard `net::TcpListener`.
+    ///
+    /// This function is intended to be used to wrap a TCP listener from the
+    /// standard library in the Mio equivalent. The conversion assumes nothing
+    /// about the underlying listener.
```
Thomasdezeeuw: Maybe add a note saying that the listener is not put in non-blocking mode as that is up to the user.

src/net/tcp/listener.rs
```diff
@@ -41,6 +44,20 @@ pub struct TcpListener {
...
+    pub fn from_std(listener: net::TcpListener) -> Self {
```
Thomasdezeeuw: Can you use `TcpListener` instead of `Self`. I personally use `Self` in build patterns to indicate it returns the same object, but other people might use it differently.

src/net/tcp/stream.rs
```diff
@@ -55,6 +58,24 @@ impl TcpStream {
         }
     }
 
+    /// Creates a new `TcpStream` from a standard `net::TcpStream`.
+    ///
+    /// This function is intended to be used to wrap a TCP stream from the
+    /// standard library in the Mio equivalent. The conversion assumes nothing
+    /// about the underlying stream.
+    ///
+    /// Note: The TCP stream here will not have `connect` called on it, so it
```
Thomasdezeeuw: Can you change this "Note:" to a header, i.e. `# Notes`.

src/net/tcp/listener.rs
Thomasdezeeuw: Same for the other types.

src/net/tcp/listener.rs
Thomasdezeeuw: Same for the other types.

src/net/tcp/stream.rs
Thomasdezeeuw: Same for the other types.

src/sys/unix/tcp/listener.rs
```diff
@@ -13,6 +13,12 @@ pub struct TcpListener {
 }
 
 impl TcpListener {
+    pub(crate) fn from_std(inner: net::TcpListener) -> Self {
```
Thomasdezeeuw: You can just use `pub`, the type isn't exported from the crate.

Thomasdezeeuw: Same for the other types.

src/sys/unix/tcp/listener.rs
```diff
@@ -13,6 +13,12 @@ pub struct TcpListener {
...
+        let raw_fd = inner.into_raw_fd();
```
Thomasdezeeuw: We are you used `into_raw` and then `from_raw` here? You just return `TcpStream { inner }`.

src/net/tcp/listener.rs
```diff
@@ -41,6 +44,20 @@ pub struct TcpListener {
 }
 
 impl TcpListener {
+    /// Creates a new `TcpListener` from a standard `net::TcpListener`.
```
Thomasdezeeuw: Can you move the method below `bind`, to indicate in the documentation that `bind` is the preferred way to create a listener.

tests/tcp_listener.rs
```diff
@@ -72,6 +72,12 @@ fn smoke_test_tcp_listener(addr: SocketAddr) {
     thread_handle.join().expect("unable to join thread");
 }
 
+// #[test]
+// fn tcp_listener_from_std() {
```
Thomasdezeeuw: Maybe you can create a `net::TcpListener` (from std), set it to non-blocking convert it to a mio listener and reuse the smoke_test function (after a small rewrite)?

Thomasdezeeuw: Same for the other types.

tests/tcp_stream.rs
```diff
@@ -120,6 +120,14 @@ fn smoke_test_tcp_stream(addr: SocketAddr) {
     thread_handle.join().expect("unable to join thread");
 }
 
+// #[test]
```
carllerche: There a reason this is still commented out?

src/sys/unix/tcp/listener.rs
carllerche: I believe they using pub for pub(crate) is falling out of favor.  There is a lint (enabled in tokio) that warns on this.

Thomasdezeeuw: I'm fine with changing to `pub(crate)` to, as long as its consistent.

carllerche: The lint is `unreachable_pub`. I'm fine either leaving this here and fixing everything else in a follow up PR or switch to just `pub` in the PR here and then add the lint & fix.

kleimkuhler: I'll remove from these fns as part of this PR. An addition of the lint and subsequent cleanup can be done in a separate PR!

kleimkuhler:
@Thomasdezeeuw Thanks for the review. All your comments have been addressed. I added a lot more around tests; each I/O resource now have tests that test the `from_std` functions.

tests/tcp_stream.rs
```diff
@@ -36,21 +36,47 @@ fn is_send_and_sync() {
 #[test]
 #[cfg_attr(windows, ignore = "fails on Windows, see #1078")]
 fn tcp_stream_ipv4() {
-    smoke_test_tcp_stream(any_local_address());
+    let (poll, events) = init_with_poll();
+    let (handle, addr) = echo_listener(any_local_address(), 1);
+    let stream = TcpStream::connect(addr).unwrap();
```
Thomasdezeeuw: This is more setup then I would like. Why not change `smoke_test_tcp_stream` to accept an `FnOnce(SocketAddr) -> io::Result<TcpStream>` and use it like:

```rust
smoke_test_tcp_stream(any_local_address(), |addr| TcpStream::connect(addr));
```

This reduces the overhead of the setup, which is now in four functions.

tests/udp_socket.rs
```diff
@@ -49,6 +49,21 @@ fn unconnected_udp_socket_ipv6() {
...
+    // `std::net::UdpSocket`s are blocking by default, so make sure they are
+    // in non-blocking mode before wrapping in a Mio equivalent.
+    assert_ok!(socket1.set_nonblocking(true));
+    assert_ok!(socket2.set_nonblocking(true));
+
+    let socket1 = UdpSocket::from_std(socket1);
+    let socket2 = UdpSocket::from_std(socket2);
+    smoke_test_unconnected_udp_socket(socket1, socket2);
+}
+
```
Thomasdezeeuw: Can you also add it for connected `UdpSocket`s.

tests/uds.rs
```diff
@@ -30,14 +30,39 @@ const LOCAL: Token = Token(0);
...
-    let (mut poll, mut events) = init_with_poll();
+fn uds_stream() {
+    let (poll, events) = init_with_poll();
 
     let (sync_sender, sync_receiver) = channel();
     let (handle, remote_addr) = echo_remote(1, sync_receiver);
 
     let path = remote_addr.as_pathname().expect("not a pathname");
-    let mut local = assert_ok!(UnixStream::connect(path));
+    let local = assert_ok!(UnixStream::connect(path));
```
Thomasdezeeuw: Same here with using a setup function to create an `UnixStream`.

Thomasdezeeuw:
Thanks!

PR #1125, opened 2019-11-06
Added support to print as alternate platform specific event data

dtacalau:
This has helped me during debugging windows issues. Was thinking it's a good idea to have it on master too.


Thomasdezeeuw:
I don't think any types in libc implement `fmt::Debug` so that is why the CI is failing on Unix platforms.

dtacalau:
> I don't think any types in libc implement `fmt::Debug` so that is why the CI is failing on Unix platforms.

@Thomasdezeeuw Got it, I limited this change to Windows only.

Thomasdezeeuw:
I was thinking about this a bit and let's do this only when providing `{:#?}`, normally this means the output is pretty printed, but we can (ab)use to also add additional platforms specific information. For Unix you can look at #1091 on how do add the even details.

dtacalau:
> I was thinking about this a bit and let's do this only when providing `{:#?}`, normally this means the >output is pretty printed, but we can (ab)use to also add additional platforms specific information. >For Unix you can look at #1091 on how do add the even details.

@Thomasdezeeuw could you please give more details, ATM is not clear to me what would you like pretty-printed for each platform. Thanks!

Thomasdezeeuw:
@dtacalau sorry I wasn't really clear. What I meant was checking the [`fmt.alternate()` flag](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.alternate) and to only show the platform specific details then. This way regular users of Mio won't be surprised/confused by them.

#1091, specifically https://github.com/tokio-rs/mio/pull/1091/files#diff-8be43826e3d9aa67792b01b56899deb9R331-R640, has the code on how to format the platform specific event data.

That would result in something like:

```rust
// `event` is readable & writable.
println!("{:?}", event); // Prints: Event { token: Token(0), readable: true, writable: true ... etc. } (no change).
println!("{:#?}", event); // Prints: Event { (same fields as above) .. platform: EVFILT_READ | EVFILT_WRITE }
```

dtacalau:

src/event/event.rs
```diff
@@ -190,16 +190,24 @@ impl Event {
 
 impl fmt::Debug for Event {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("Event")
-            .field("token", &self.token())
+        let alternate = f.alternate();
+        let d = &mut f.debug_struct("Event");
```
Thomasdezeeuw: Why `&mut f.debug_struct("Event")`?

src/sys/unix/selector/epoll.rs
```diff
@@ -121,6 +121,7 @@ pub type Events = Vec<Event>;
 pub mod event {
     use crate::sys::Event;
     use crate::Token;
+    use std::fmt::Write as FmtWrite;
```
Thomasdezeeuw: You're just import the trait right, so there is no need for the `as FmtWrite`.

Thomasdezeeuw: Same for the other places.

src/event/events.rs
```diff
@@ -210,8 +210,13 @@ impl<'a> Iterator for Iter<'a> {
...
 impl fmt::Debug for Events {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("Events")
-            .field("capacity", &self.capacity())
-            .finish()
+        let mut d = f.debug_struct("Events");
+        d.field("capacity", &self.capacity());
+
+        for event in self.iter() {
+            d.field("event", event);
```
Thomasdezeeuw: Now we get a bunch of "event" which seems odd. Not sure about a better alternative though.

src/event/events.rs
```diff
@@ -210,8 +210,13 @@ impl<'a> Iterator for Iter<'a> {
 
 impl fmt::Debug for Events {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("Events")
-            .field("capacity", &self.capacity())
-            .finish()
+        let mut d = f.debug_struct("Events");
+        d.field("capacity", &self.capacity());
+
+        for event in self.iter() {
```
Thomasdezeeuw: Let's only do this when alternative is enabled, just like with `Event`.

src/event/event.rs
```diff
@@ -190,16 +190,24 @@ impl Event {
...
             .field("read_closed", &self.is_read_closed())
             .field("write_closed", &self.is_write_closed())
             .field("priority", &self.is_priority())
             .field("aio", &self.is_aio())
-            .field("lio", &self.is_lio())
-            .finish()
+            .field("lio", &self.is_lio());
+
+        if alternate {
+            let mut data_str = String::new();
```
Thomasdezeeuw: Can we find a better way to do this then allocating a string?

src/event/event.rs
```diff
@@ -190,16 +190,24 @@ impl Event {
...
+            sys::event::get_platform_event_data_str(&mut data_str, &self.inner);
```
Thomasdezeeuw: Java level function name ;)

src/sys/windows/event.rs
```diff
@@ -49,6 +51,37 @@ pub fn is_lio(_: &Event) -> bool {
...
+        POLL_SEND,
+        POLL_DISCONNECT,
+        POLL_ABORT,
+        POLL_LOCAL_CLOSE,
+        POLL_ACCEPT,
+        POLL_CONNECT_FAIL,
+    );
+
+    write!(data_str, ")").unwrap();
+    ()
```
Thomasdezeeuw: Not needed.

tests/tcp_listener.rs
```diff
@@ -343,6 +343,34 @@ fn try_clone_different_poll() {
     thread_handle2.join().expect("unable to join thread");
 }
 
+#[test]
+fn platform_specific_data_print() {
```
Thomasdezeeuw: I don't really think there is a need for a test as the output is not stable.

If you do want a test I would suggest making one per platform the there `event` module and manually creating the `Event`s.

src/event/event.rs
Thomasdezeeuw: Would be better if we can somehow pass `fmt::Formatter`.

src/event/event.rs
Thomasdezeeuw: Maybe call it `details`?

src/sys/unix/selector/epoll.rs
```diff
@@ -168,6 +169,46 @@ pub mod event {
         // Not supported.
         false
     }
+
+    pub fn get_platform_event_data_str(data_str: &mut String, event: &Event) {
+        write!(data_str, "events (").unwrap();
```
Thomasdezeeuw: This is not needed in the `Debug` impl the field name is already added.

src/sys/unix/selector/epoll.rs
dtacalau: ok, I removed "as FmtWrite"

src/event/event.rs
dtacalau: Removed &

src/event/events.rs
dtacalau: Removed field name "event", was printed twice.

src/event/events.rs
dtacalau: done

src/event/event.rs
dtacalau: Redone by passing the fmt::Formatter.

src/event/event.rs
dtacalau: renamed to write_details, fwiw I like long explanatory names :)

src/sys/windows/event.rs
dtacalau: probably, removed!

src/sys/unix/selector/epoll.rs
dtacalau: Removed

tests/tcp_listener.rs
dtacalau: Removed the test, not worth the trouble atm.

dtacalau:
Here's how details are printed after changes:

- w/o alternate
```
poll events: Events { capacity: 16 }
```
- w/ alternate
```
poll events: Events {
    capacity: 16,
    : Event { token: Token(1), readable: false, writable: true, error: false, read_closed: false, write_closed: false, priority: false, aio: false, lio: false, platform: POLL_SEND },
}
```

Issue #1126, opened 2019-11-06
TcpListener::accept won't generate READABLE w/o reregistering the interests before poll


dtacalau:
This is about reregistering on successful operations on Windows. So far it has been mentioned here https://github.com/tokio-rs/mio/issues/1074#issuecomment-548501031 and here https://github.com/tokio-rs/mio/pull/1112#discussion_r340302964 .

I'd like to have this tracked separately, because it fixes some tests but not all. 

Problem:
The test below fails on the second expect_events assert. If I enable the "reregister", after the second stream is started, the second expect_events assert will pass. I've created the test below by modifying test try_clone_different_poll test to its simplest form that is still failing: one poll, one listener, two streams.

```rust
#[test]
//#[cfg_attr(windows, ignore = "fails on Windows, see #1073")]
fn tcp_listener_two_streams() {
    let (mut poll1, mut events) = init_with_poll();

    let listener1 = TcpListener::bind(any_local_address()).unwrap();
    let address = listener1.local_addr().unwrap();

    let barrier = Arc::new(Barrier::new(3));
    let thread_handle1 = start_connections(address, 1, barrier.clone());

    poll1
        .registry()
        .register(&listener1, ID1, Interests::READABLE)
        .unwrap();

    expect_events(
        &mut poll1,
        &mut events,
        vec![ExpectEvent::new(ID1, Interests::READABLE)],
    );
   
    let (stream, peer_address) = listener1.accept().expect("unable to accept connection");
    assert!(peer_address.ip().is_loopback());
    assert_eq!(stream.peer_addr().unwrap(), peer_address);
    assert_eq!(stream.local_addr().unwrap(), address);

    let thread_handle2 = start_connections(address, 1, barrier.clone());
/*
    poll1
        .registry()
        .reregister(&listener1, ID1, Interests::READABLE)
        .unwrap();*/

    expect_events(
        &mut poll1,
        &mut events,
        vec![ExpectEvent::new(ID1, Interests::READABLE)],
    );

    assert_would_block(listener1.accept());
    assert!(listener1.take_error().unwrap().is_none());

    barrier.wait();
    thread_handle1.join().expect("unable to join thread");
    thread_handle2.join().expect("unable to join thread");
} 
```

Cause:
The test fails because, on Windows, mio simulates Edge-triggered behavior, the interests are removed after first poll call, subsequent polls will need to have interests set again, one way being reregister the socket.

```rust
// In selector.rs.
...
[...]

Issue #1074, opened 2019-08-31
Investigate TcpListener::try_clone test failures


Thomasdezeeuw:
Pr #1070 is going to add (or added, depending on when you read this) two testing involving `TcpListener::try_clone` `try_clone_same_poll` and `try_clone_different_poll`, which only fails on Windows. Figure out why.

/cc @PerfectLaugh

dtacalau:
I've modified try_clone_different_poll test to the simplest form that is still failing: one poll, one listener, two streams.

Problem:
The test below fails on the second  expect_events assert. If I enable the "reregister", after the second stream is started, the test won't fail.
```
#[test]
//#[cfg_attr(windows, ignore = "fails on Windows, see #1073")]
fn tcp_listener_two_streams() {
    let (mut poll1, mut events) = init_with_poll();

    let listener1 = TcpListener::bind(any_local_address()).unwrap();
    let address = listener1.local_addr().unwrap();

    let barrier = Arc::new(Barrier::new(3));
    let thread_handle1 = start_connections(address, 1, barrier.clone());

    poll1
        .registry()
        .register(&listener1, ID1, Interests::READABLE)
        .unwrap();

    expect_events(
        &mut poll1,
        &mut events,
        vec![ExpectEvent::new(ID1, Interests::READABLE)],
    );
   
    let (stream, peer_address) = listener1.accept().expect("unable to accept connection");
    assert!(peer_address.ip().is_loopback());
    assert_eq!(stream.peer_addr().unwrap(), peer_address);
    assert_eq!(stream.local_addr().unwrap(), address);

    let thread_handle2 = start_connections(address, 1, barrier.clone());
/*
    poll1
        .registry()
        .reregister(&listener1, ID1, Interests::READABLE)
        .unwrap();*/

    expect_events(
        &mut poll1,
        &mut events,
        vec![ExpectEvent::new(ID1, Interests::READABLE)],
    );

    assert_would_block(listener1.accept());
    assert!(listener1.take_error().unwrap().is_none());

    barrier.wait();
    thread_handle1.join().expect("unable to join thread");
    thread_handle2.join().expect("unable to join thread");
} 
```

Cause: 
The test fails because, on Windows, mio simulates Edge-triggered behavior, the interests are removed after first poll call, subsequent polls will need to have interests set again, one way being reregister the socket.

```
selector.rs
...
        afd_events &= self.user_evts;
...
[...]

PR #1127, opened 2019-11-06
Fix issue #1126. Make sure to reregister on successful accept.

dtacalau:
Reregister on every successful operation. Re-enabled some tests which were previously failing on windows. Added a new test with subsequent poll which fails w/o reregistering.

### DIFF
--- a/src/sys/windows/tcp.rs
+++ b/src/sys/windows/tcp.rs
@@ -365,6 +365,7 @@ impl TcpListener {
 
     pub fn accept(&self) -> io::Result<(TcpStream, SocketAddr)> {
         try_io!(self, accept).and_then(|(inner, addr)| {
+            self.io_blocked_reregister()?;
             inner.set_nonblocking(true).map(|()| {
                 (
                     TcpStream {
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
This conflicts with #1112, lets try to get that merged first.

As for this patch, I don't this would correctly emulate edge triggers. I think we only need to register if we hit `WouldBlock` or if we read/write `n <= buf.len()` bytes (i.e. filled/emptied the entire buffer). I happen to think about this today too.

dtacalau:
@Thomasdezeeuw reregistering is not enough for try_clone, have a look at my latest findings here https://github.com/tokio-rs/mio/issues/1074#issuecomment-550371293.

Thomasdezeeuw:
@dtacalau let's try to fix one problem per pr/issue, because I'm losing oversight of it now. I thought that reregistering on successful operations was needed for the smoke tests for `TcpStream`, which pass with this change. Let's move the discussion for `try_clone` to #1074.

dtacalau:

src/sys/windows/tcp.rs
```diff
@@ -365,6 +365,7 @@ impl TcpListener {
 
     pub fn accept(&self) -> io::Result<(TcpStream, SocketAddr)> {
         try_io!(self, accept).and_then(|(inner, addr)| {
+            self.io_blocked_reregister()?;
```
Thomasdezeeuw: Maybe we should ignore the error here and return the stream even if we error?

dtacalau: Makes sense, will ignore result.

dtacalau:
@Thomasdezeeuw I think this PR can be merged, thanks!

Issue #1073, opened 2019-08-31
Deregistering TcpListener on Windows doesn't work


Thomasdezeeuw:
See #1070, relevant CI logs is available here: https://dev.azure.com/tokio-rs/Tokio/_build/results?buildId=2274&view=logs&jobId=429a383f-e486-57ac-3982-9ac65641a32a&taskId=1f926ffb-0ec7-5f9b-ab05-76fbb04597a8&lineStart=150&lineEnd=151&colStart=1&colEnd=1. Relevant test is available here: https://github.com/Thomasdezeeuw/mio/blob/add262520f777d5564c26b7f2fc8596a07c4e83f/tests/tcp_listener.rs#L156-L185.

/cc @PerfectLaugh 

----------

Copy of the log and code, for when the links above don't work anymore.

```log
---- deregister stdout ----
[ERROR tcp_listener::util] unexpected event: Event { token: Token(0), readable: true, writable: false, error: false, hup: false, read_hup: false, priority: false, aio: false, lio: false }
thread 'deregister' panicked at 'received events, but didn't expect any, see above', tests\util\mod.rs:180:9
stack backtrace:
   0:     0x7ff75223807c - backtrace::backtrace::trace_unsynchronized
                               at C:\Users\VssAdministrator\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.29\src\backtrace\mod.rs:66
   1:     0x7ff75223807c - std::sys_common::backtrace::_print
                               at /rustc/eae3437dfe991621e8afdc82734f4a172d7ddf9b\/src\libstd\sys_common\backtrace.rs:47
   2:     0x7ff75223b52e - std::sys_common::backtrace::print
                               at /rustc/eae3437dfe991621e8afdc82734f4a172d7ddf9b\/src\libstd\sys_common\backtrace.rs:36
   3:     0x7ff75223b52e - std::panicking::default_hook::{{closure}}
                               at /rustc/eae3437dfe991621e8afdc82734f4a172d7ddf9b\/src\libstd\panicking.rs:200
   4:     0x7ff75223b13b - std::panicking::default_hook
                               at /rustc/eae3437dfe991621e8afdc82734f4a172d7ddf9b\/src\libstd\panicking.rs:211
   5:     0x7ff75223bdc0 - std::panicking::rust_panic_with_hook
                               at /rustc/eae3437dfe991621e8afdc82734f4a172d7ddf9b\/src\libstd\panicking.rs:477
[...]

PR #1129, opened 2019-11-07
fix for issue #1073, deregistering not working on windows. 

dtacalau:
When deregistering a socket make sure it is marked for deletion so it won't be used in poll updates.

### DIFF
--- a/tests/tcp.rs
+++ b/tests/tcp.rs
@@ -1207,8 +1207,11 @@ fn tcp_no_events_after_deregister() {
         .register(&stream, Token(3), Interests::READABLE)
         .unwrap();
 
-    // Wait a moment for stream to connect before trying accept
-    thread::sleep(Duration::from_millis(100));
+    while events.is_empty() {
+        poll.poll(&mut events, None).unwrap();
+    }
+    assert_eq!(events.iter().count(), 1);
+    assert_eq!(events.iter().next().unwrap().token(), Token(1));
 
     let mut stream2 = listener.accept().unwrap().0;
     poll.registry()
@@ -1221,8 +1224,6 @@ fn tcp_no_events_after_deregister() {
     poll.registry().deregister(&stream).unwrap();
     poll.registry().deregister(&stream2).unwrap();
 
-    // note: without deregister, poll would have retrieved 3 events:
-    // listener-READABLE, stream2-WRITABLE, stream-READABLE
     expect_no_events(&mut poll, &mut events);
 
     let mut buf = [0; 10];
--- a/tests/tcp.rs
+++ b/tests/tcp.rs
@@ -17,7 +17,10 @@ use mio::{Events, Interests, Poll, Registry, Token};
 
 mod util;
 
-use util::{any_local_address, assert_send, assert_sync, init, TryRead, TryWrite};
+use util::{
+    any_local_address, assert_send, assert_sync, expect_no_events, init, init_with_poll, TryRead,
+    TryWrite,
+};
 
 const LISTEN: Token = Token(0);
 const CLIENT: Token = Token(1);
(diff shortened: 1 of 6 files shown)


Thomasdezeeuw:
Shouldn't this happen in the `deregister` implementation?

dtacalau:
> Shouldn't this happen in the `deregister` implementation?

@Thomasdezeeuw the key here is calling mark_delete() for a sock_state once it's deregistered. The call to mark_delete() could be done from other places, including deregister, but, I've chosen set_sock_state to keep things specific to TcpListener sock_state. Maybe we want other behavior for sock_state of other objects. What do you think?

Thomasdezeeuw:
I think you're right this is the best place. Does this need to happen for `UdpSocket` and `TcpStream` as well?

dtacalau:
> I think you're right this is the best place. Does this need to happen for `UdpSocket` and `TcpStream` as well?

Good point, will check.

dtacalau:
@Thomasdezeeuw Both, UdpSocket and TcpStream, were affected by the same deregistering issue. I've made fixes for them too.

tests/tcp.rs
```diff
@@ -1188,3 +1191,41 @@ fn write_then_deregister() {
...
+
+    poll.registry()
+        .register(&listener, Token(1), Interests::READABLE)
+        .unwrap();
+    poll.registry()
+        .register(&stream, Token(3), Interests::READABLE)
+        .unwrap();
+
+    // Wait a moment for stream to connect before trying accept
+    thread::sleep(Duration::from_millis(100));
```
Thomasdezeeuw: Why aren't you polling here?

tests/tcp.rs
```diff
@@ -1188,3 +1191,41 @@ fn write_then_deregister() {
...
+        .register(&stream2, Token(2), Interests::WRITABLE)
+        .unwrap();
+
+    stream2.write_all(&[1, 2, 3, 4]).unwrap();
+
+    poll.registry().deregister(&listener).unwrap();
+    poll.registry().deregister(&stream).unwrap();
+    poll.registry().deregister(&stream2).unwrap();
+
+    // note: without deregister, poll would have retrieved 3 events:
```
Thomasdezeeuw: Don't think we need this note.

tests/tcp_stream.rs
```diff
@@ -524,26 +524,39 @@ fn reregistering() {
...
+    // Also, write should work
+    let mut buf = [0; 16];
+    assert_would_block(stream.peek(&mut buf));
+    assert_would_block(stream.read(&mut buf));
+
+    let n = stream.write(&DATA1).expect("unable to write to stream");
+    assert_eq!(n, DATA1.len());
+    stream.flush().unwrap();
+
+    // note: without deregistering at this point poll would retrieve Interests::READABLE
```
Thomasdezeeuw: No need for the note.

tests/tcp_stream.rs
```diff
@@ -524,26 +524,39 @@ fn reregistering() {
...
     poll.registry()
-        .register(&stream, ID1, Interests::READABLE)
+        .register(&stream, ID1, Interests::WRITABLE.add(Interests::READABLE))
         .expect("unable to register TCP stream");
 
     poll.registry()
         .deregister(&stream)
         .expect("unable to deregister TCP stream");
 
+    // note: without deregistering at this point poll would retrieve Interests::WRITABLE
```
Thomasdezeeuw: No need for the note.

tests/tcp.rs
dtacalau: Wanted to not exhaust the event by polling it, and keep it for later check after deregister, but I guess there are enough checks, will do polling.

tests/tcp.rs
dtacalau: ok

tests/tcp_stream.rs
dtacalau: ok

tests/tcp_stream.rs
dtacalau: ok

Thomasdezeeuw:
Thanks!

PR #1130, opened 2019-11-08
Wrap kqueue::Events so that it can be Send and Sync

kleimkuhler:
## Motivation

Both `Poll` and `Registry` are `Send` and `Sync`. `Events` is passed as an
argument to `Poll::poll`, and is intended to be created once--at the same time
as `Poll`--and reused on each call. It would be expected that `Events` is
therefore `Send` and `Sync` but that is currently not the case in kqueue.

#983 changed the "events" type (`KeventList` -> `Events`) to have a named field,
but also [dropped](https://github.com/tokio-rs/mio/pull/983/files#diff-1b9651542d08c6eca04e6025b1c6fd53L330) the `Send` and `Sync` impls.

For epoll events this has been okay because these marker traits can be derived
from the fields within `epoll_event`, but `kevent` has a `*mut T` field making
this not possible.

## Solution

`kqueue::Events` is now a smart pointer wrapping a `Vec<libc::kevent>` and impls
`Send` and `Sync`. `Deref` and `DerefMut` traits have also been implemented
which reduce changes in the rest of `sys`.

Assertions for `Events` have been added to the `assert_send`/`assert_sync` test
as well.

### DIFF
--- a/src/sys/unix/selector/kqueue.rs
+++ b/src/sys/unix/selector/kqueue.rs
@@ -327,8 +327,10 @@ impl DerefMut for Events {
     }
 }
 
+// `Events` cannot derive `Send` because of the `udata: *mut ::c_void` field
+// in `libc::kevent`. However, `Events`'s public API treats the `udata` field as
+// a `uintptr_t` which is `Send`.
 unsafe impl Send for Events {}
-unsafe impl Sync for Events {}
 
 pub mod event {
     use crate::sys::Event;
--- a/src/sys/unix/selector/kqueue.rs
+++ b/src/sys/unix/selector/kqueue.rs
@@ -2,6 +2,7 @@ use crate::{Interests, Token};
 
 use log::error;
 use std::mem::MaybeUninit;
+use std::ops::{Deref, DerefMut};
 use std::os::unix::io::{AsRawFd, RawFd};
 #[cfg(debug_assertions)]
 use std::sync::atomic::{AtomicUsize, Ordering};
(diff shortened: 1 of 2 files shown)


src/sys/unix/selector/kqueue.rs
```diff
@@ -304,7 +305,30 @@ impl Drop for Selector {
...
+    }
+}
+
+impl Deref for Events {
+    type Target = Vec<libc::kevent>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
hawkw: Should this perhaps be an impl of `AsRef<[libc::kevent]>`?

src/sys/unix/selector/kqueue.rs
```diff
@@ -304,7 +305,30 @@ impl Drop for Selector {
...
+
+impl DerefMut for Events {
```
hawkw: And this `AsMut<[libc::kevent]>`?

hawkw: Nevermind, I'm wrong: as per an offline conversation, it seems like a lot more `Vec` methods _need_ to be called through `Events`, `Deref` is probably the right choice. Carry on.

src/sys/unix/selector/kqueue.rs
```diff
@@ -304,7 +305,30 @@ impl Drop for Selector {
...
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+unsafe impl Send for Events {}
```
Thomasdezeeuw: Can you add a comment saying why this is safe? For example @asomers comment from https://github.com/nix-rust/nix/pull/442/files#diff-7a42d5ac10efafbf1371c20fe9247662R195-R197.

PR #1132, opened 2019-11-08
Various documentation cleanups

Thomasdezeeuw:


src/poll.rs
```diff
@@ -434,11 +431,10 @@ impl Registry {
...
     /// [`event::Source`]: crate::event::Source
-    /// [`struct`]: #
-    /// [`reregister`]: #method.reregister
-    /// [`deregister`]: #method.deregister
-    /// [`poll`]: #method.poll
-    /// [`Token`]: struct.Token.html
+    /// [`poll`]: Poll::poll
+    /// [`reregister`]: Registry::reregister
+    /// [`deregister`]: Registry::deregister
+    /// [`Token`]: Token
```
kleimkuhler: It looks like a lot of the links don't work properly in the documentation, but since this PR is changing these they should be:

```suggested
/// [`poll`]: struct.Poll.html#method.poll
/// [`reregister`]: #method.reregister
/// [`deregister`]: #method.deregister
/// [`Token`]: struct.Token.html
```

src/event/events.rs
```diff
@@ -154,6 +151,10 @@ impl Events {
 
     /// Clearing all `Event` values from container explicitly.
     ///
+    /// # Notes
+    ///
+    /// Events are cleared before every `poll`, it not needed to this.
```
kleimkuhler: Looks like there is a typo here. How about: "... every `poll`, but this is not required."

src/poll.rs
Thomasdezeeuw: These are intra docs links. They link e.g. the `Poll::poll` method. These should work, or at least don't give any warning on my machine.

src/event/events.rs
```diff
@@ -25,15 +25,15 @@ use std::fmt;
 ///
 /// let mut events = Events::with_capacity(1024);
 /// let mut poll = Poll::new()?;
-///
-/// assert_eq!(0, events.iter().count());
+/// #
+/// # assert!(events.is_empty());
```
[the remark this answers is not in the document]
Thomasdezeeuw: The assertion doesn't add much to the documentation does it? It still runs as a part of the test though.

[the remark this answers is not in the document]
kleimkuhler: Oh yes you are correct. I didn't realize the assertion was still run when commented out in the doc test. I see that now locally!

src/poll.rs
kleimkuhler: So running `cargo doc --open` locally and clicking these links in the `Registry` documentation works? It does not me; it loads a blank page.

Thomasdezeeuw: As discussed on Discord; this is nightly only (https://github.com/rust-lang/rust/issues/63305).

Thomasdezeeuw:
@kleimkuhler I'm keeping the intra-docs as is for now. Once we're ready to release a v0.7 we'll change all links I've opened #1136 for it.

PR #1133, opened 2019-11-08
Move most of the crate documentation to README

Thomasdezeeuw:
Thing like the features, non-goals and platforms are moved to the README
in an effort to reduce the scrolling when opening the API documentation.
Furthermore once you decide on using Mio most of these thing don't
really add any value anymore. So instead these things are moved to the
README of the repository where people (most likely) will land when first
encountering Mio on GitHub or crates.io.

The example is also moved to the README as the README was missing one
and most types now have examples in there documentation, including Poll.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Mio – Metal IO
 
-Mio is a lightweight I/O library for Rust with a focus on adding as little
+Mio is a fast, low-level I/O library for Rust focusing on non-blocking APIs and
+event notification for building high performance I/O apps with as little
 overhead as possible over the OS abstractions.
 
 [![Crates.io][crates-badge]][crates-url]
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
Also see #1134.

carllerche:
IMO we should investigate strategies that allow having the doc in both places without duplicating.

carllerche:
https://github.com/livioribeiro/cargo-readme/blob/master/README.md

kleimkuhler:
I'm not sure moving examples to the README is the more accessible route for documenting how one may use Mio. I think maybe a smaller example would be okay. Speaking from my perspective, I definitely default to [docs.rs](https://docs.rs) for the click-ability through documentation more than a README. Therefore, I would prefer seeing example usage there, rather than splitting it between crate documentation and a README.

Thomasdezeeuw:
@carllerche, @kleimkuhler have you looked at #1134? It adds a guide currently containing mostly the example move to the readme here. I don't really like that I have to scroll down when opening the Mio API docs as an experienced user. That why I moved them to the README and a new guide module. New users can still find the example, while experienced users are inconvenienced by it. What do you think?

Thomasdezeeuw:
If #1134 is merged the example could also be moved into the user guide. Something like `mio::guide::examples`.

Alternately we can just remove it completely from the readme and main crate docs and point to the examples directory and the getting started guide in both the readme and main crate docs.

kleimkuhler:
> .. and point to the examples directory and the getting started guide in both the readme and main crate docs

I would be okay with this change if we still point to it from the crate docs! As I mentioned above, my only concern is moving a link (or the actual example) only to the README. If we also link to it in the crate doc and have a submodule in the guide I'd be okay with that.

Thomasdezeeuw:
@kleimkuhler I've added section to refer to the examples and getting started guide. What do you think?

PR #1135, opened 2019-11-08
Remove unneeded calls to update_sockets_events

### LINKED ISSUE #1128: Remove unneeded calls to `update_sockets_events`
Just so it does get lost: https://github.com/tokio-rs/mio/commit/78f36941ea3c8287417bbbb788c155a5b85e1c61#commitcomment-35854690.

Thomasdezeeuw:
Closes #1128.

### DIFF
--- a/src/sys/windows/selector.rs
+++ b/src/sys/windows/selector.rs
@@ -515,7 +515,6 @@ impl SelectorInner {
         socket.set_sock_state(Some(sock));
         unsafe {
             self.add_socket_to_update_queue(socket);
-            self.update_sockets_events()?;
         }
 
         Ok(())
@@ -543,7 +542,6 @@ impl SelectorInner {
         }
         unsafe {
             self.add_socket_to_update_queue(socket);
-            self.update_sockets_events()?;
         }
 
         Ok(())


src/sys/windows/selector.rs
```diff
@@ -515,7 +515,6 @@ impl SelectorInner {
         socket.set_sock_state(Some(sock));
         unsafe {
             self.add_socket_to_update_queue(socket);
-            self.update_sockets_events()?;
```
Thomasdezeeuw: @piscisaureus are you sure this call isn't needed? It think this breaks registering on one thread, while polling on another; see https://dev.azure.com/tokio-rs/Tokio/_build/results?buildId=3275&view=logs&jobId=429a383f-e486-57ac-3982-9ac65641a32a&taskId=1f926ffb-0ec7-5f9b-ab05-76fbb04597a8&lineStart=93&lineEnd=94&colStart=1&colEnd=1.

piscisaureus: @Thomasdezeeuw

Yes, that's expected. But it shouldn't be possible to poll one thread while registering in another, because [Poll:: poll takes an &mut](https://github.com/tokio-rs/mio/pull/1112#issuecomment-548735109).

If it is possible I overlooked some interior mutable struct and the above *is* possible. In that case the previous active_poll_count counter needs to be restored, as well as the update_socket_events_if_polling() method.

Yes, the tests will pass if you unconditionally call update_socket_events(), but doing that is *very* inefficient and will likely lead to busyloops.

piscisaureus: Actually let me look at this myself.

Thomasdezeeuw:
Closing in favour of #1140.

PR #1139, opened 2019-11-09
Fix UnixListener registration

mlang:
Before, `Registry::register()` on a `UnixListener` would return
`Err(io::ErrorKind::NotFound)`.
Calling `sys::UnixListener::register()` instead of `sys::UnixListener::reregister()`
fixes this.


Thomasdezeeuw:
Shouldn't we have a failing test on Linux for this? Normally epoll doesn't accept being reregistered without ever registering first.

Thomasdezeeuw:
@mlang thanks!

@kleimkuhler you'll follow up with some additional testing?

PR #1140, opened 2019-11-10
Submit socket events to kernel immediately only if another thread is polling

piscisaureus:
Since it is not possible to modify or synchronously cancel an
IOCTL_AFD_POLL operation, and there can be only one active AFD_POLL
operation per (socket, completion port) at any time, it is expensive
to change a socket's event registration after it has been submitted to
the kernel.

Therefore, if no other threads are polling when interest in a socket
event is (re)registered, the socket is added to the 'update queue', but
the actual syscall to start the IOCTL_AFD_POLL operation is deferred
until just before the GetQueuedCompletionStatusEx() call is made.

However, when another thread is already blocked on
GetQueuedCompletionStatusEx() we tell the kernel about the registered
socket event(s) immediately.

Closes: #1128
Closes: #1135

### DIFF
--- a/src/sys/windows/selector.rs
+++ b/src/sys/windows/selector.rs
@@ -13,7 +13,8 @@ use std::os::windows::io::{AsRawSocket, RawSocket};
 use std::pin::Pin;
 use std::ptr::null_mut;
 #[cfg(debug_assertions)]
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 use std::{io, ptr};
@@ -407,6 +408,7 @@ pub struct SelectorInner {
     cp: Arc<CompletionPort>,
     update_queue: Mutex<VecDeque<Arc<Mutex<SockState>>>>,
     afd_group: AfdGroup,
+    is_polling: AtomicBool,
 }
 
 // We have ensured thread safety by introducing lock manually.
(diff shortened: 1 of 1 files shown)


src/sys/windows/selector.rs
```diff
@@ -407,6 +406,7 @@ pub struct SelectorInner {
     cp: Arc<CompletionPort>,
     update_queue: Mutex<VecDeque<Arc<Mutex<SockState>>>>,
     afd_group: AfdGroup,
+    active_poll_count: AtomicUsize,
```
Thomasdezeeuw: Since this we can only have a single `poll` call at a time can you make this: `active_poll: AtomicBool`? And add a comment explaining why this is needed. Your commit message has all the needed details, but its nice to have it in the code as well.

src/sys/windows/selector.rs
```diff
@@ -515,7 +516,7 @@ impl SelectorInner {
         socket.set_sock_state(Some(sock));
         unsafe {
             self.add_socket_to_update_queue(socket);
-            self.update_sockets_events()?;
+            self.update_sockets_events_if_polling()?;
```
dtacalau: Maybe I missed something, but I don't understand the need for calling self.update_sockets_events_if_polling()?; during register/reregister, why can't we just defer the "update_sockets_events" until the next select2 call?

Thomasdezeeuw: If one thread is polling and another is registering a new I/O source we want to make sure the polling thread wakes up for events on the new source. Otherwise the polling thread might block forever, which is happening in the `registry_behind_arc` (also see #1135 and https://dev.azure.com/tokio-rs/Tokio/_build/results?buildId=3275&view=logs&jobId=429a383f-e486-57ac-3982-9ac65641a32a&taskId=1f926ffb-0ec7-5f9b-ab05-76fbb04597a8&lineStart=93&lineEnd=94&colStart=1&colEnd=1)

src/sys/windows/selector.rs
piscisaureus: Done! Thanks for the suggestions.

PR #1142, opened 2019-11-11
Fix UnixStream::pair and UnixDatagram::pair on all platforms

kleimkuhler:
#1141 Addressed an issue with `UnixStream::pair` and `UnixDatagram::pair`
where the file descriptor array passed in to `pair_descriptors` was copied in
instead of passed by reference. Even if `pair_descriptors` succeded, the file
descriptor array used to create the socket pairs was the original array
initialized to `[-1; 2]`.

The issue still remains for target platforms where additional `fcntl` sys
calls must be made. The sockets are created before the call to
`pair_descriptors`, but a successful return does not actually affect the
sockets.

Ultimately, both of the `pair` methods and `pair_descriptors` function were
incorrect. I feel `pair_descriptors` role in trying to reduce code duplication
while being generic over `UnixStream` and `UnixDatagram` adds some uncessary
complexity without actually resulting in less code.

### Solution

The `pair` methods are now separate and no longer dispatch to a generic
function. The sockets are created after the required sys call. The `fcntl`
calls are made after socket creation to allow for cleanup.

Tests have been added for these specific methods.

A final PR separate from this will expand all UDS tests so that it is more
heavily tested.

### DIFF
--- a/tests/uds.rs
+++ b/tests/uds.rs
@@ -149,11 +149,13 @@ fn stream_pair() {
     );
 
     let mut buf = [0; DEFAULT_BUF_SIZE];
+    assert_would_block(s1.read(&mut buf));
     let wrote = assert_ok!(s1.write(&DATA1));
     assert_eq!(wrote, DATA1_LEN);
     assert_ok!(s1.flush());
 
     let read = assert_ok!(s2.read(&mut buf));
+    assert_would_block(s2.read(&mut buf));
     assert_eq!(read, DATA1_LEN);
     assert_eq!(&buf[..read], DATA1);
     assert_eq!(read, wrote, "unequal reads and writes");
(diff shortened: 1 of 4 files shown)


tests/uds.rs
```diff
@@ -130,6 +130,84 @@ fn is_send_and_sync() {
...
+        &mut events,
+        vec![ExpectEvent::new(Token(0), Interests::WRITABLE)],
+    );
+
+    let mut buf = [0; DEFAULT_BUF_SIZE];
+    assert_would_block(s1.read(&mut buf));
+    let wrote = assert_ok!(s1.write(&DATA1));
+    assert_eq!(wrote, DATA1_LEN);
+    assert_ok!(s1.flush());
+
```
Thomasdezeeuw: We should check if we receive a readable event here, same for the next read.

tests/uds.rs
```diff
@@ -130,6 +130,84 @@ fn is_send_and_sync() {
...
+        &mut poll,
+        &mut events,
+        vec![ExpectEvent::new(Token(0), Interests::WRITABLE)],
+    );
+
+    let mut buf = [0; DEFAULT_BUF_SIZE];
+    assert_would_block(s1.recv(&mut buf));
+    let wrote = assert_ok!(s1.send(&DATA1));
+    assert_eq!(wrote, DATA1_LEN);
+
```
Thomasdezeeuw: Same as for `stream_pair`, needs needs a check for a readable event, same for next recv.

kleimkuhler:
@Thomasdezeeuw Thanks for the review. I added the readable event checks!

PR #1145, opened 2019-11-12
Various test clean ups

Thomasdezeeuw:
Various test clean ups. Easiest to review by commit.

Updates #995

### DIFF
--- a/tests/util/mod.rs
+++ b/tests/util/mod.rs
@@ -1,5 +1,5 @@
 // Not all functions are used by all tests.
-#![allow(dead_code)]
+#![allow(dead_code, unused_macros)]
 
 use std::net::SocketAddr;
 use std::ops::BitOr;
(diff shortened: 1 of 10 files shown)


Thomasdezeeuw:
It seems that Azure pipelines rebases the pr on master, which had an additional commit compare to my local branch. That is why the CI is failing.

Thomasdezeeuw:
After a rebase on master the CI is green now.

Thomasdezeeuw:
@kleimkuhler I agree that when using `unwrap` having an indirect/incorrect line number is a bit annoying. I always run with `RUST_BACKTRACE=full` (or `1`), so the problem is less pressing. I don't really agree that having two macros for is worth it. Also hopefully something like https://github.com/rust-lang/rfcs/pull/2091 would fix the problem in future.

I also fixed the ` expect_read` docs.

tests/util/mod.rs
```diff
@@ -146,23 +147,90 @@ impl<T> MapNonBlock<T> for io::Result<T> {
...
+
+impl BitOr for Readiness {
+    type Output = Self;
+
+    fn bitor(self, other: Self) -> Self {
+        Readiness(self.0 | other.0)
+    }
+}
+
+impl From<Interests> for Readiness {
```
kleimkuhler: I like this! The Mio update in Tokio could probably use something similar instead of the explicit [checks](https://github.com/tokio-rs/tokio/pull/1767/files#diff-a5dfd4753df93e043d7629b7067ebd23R235) that are there.

kleimkuhler:
@Thomasdezeeuw Even with `assert_ok!`, I do usually end up switching to `RUST_BACKTRACE=1` once I need to actually see why a `Result` is not the expected value, so that is a good point.

There are going to be some conflicts with the UDS tests getting split up, but I think most will just be replacing the macros with `unwrap`. Once that is taken care of I'll be good with this change!

Thomasdezeeuw:
Rebased this on master @kleimkuhler can you take another look? I'll to merge this quickly as rebasing this is a pain in the.. you know what :)

Issue #1106, opened 2019-10-20
Poll function on Windows always returns after timeout expiration


JuSiZeLa:
Version: master
The polling function on Windows always returns after the timeout expiration even if some events have been received

Thomasdezeeuw:
@PerfectLaugh can you look at this?

dtacalau:
@Thomasdezeeuw please assign this to me, thanks!

PR #1153, opened 2019-11-15
fix for issue #1106. Poll should not wait the full timeout

dtacalau:
According to documentation the poll function will block until either at least one readiness event has been received or timeout has elapsed.

### DIFF
--- a/tests/tcp.rs
+++ b/tests/tcp.rs
@@ -18,8 +18,8 @@ use mio::{Events, Interests, Poll, Registry, Token};
 mod util;
 
 use util::{
-    any_local_address, assert_send, assert_sync, expect_no_events, init, init_with_poll, TryRead,
-    TryWrite,
+    any_local_address, assert_send, assert_sync, expect_events, expect_no_events, init,
+    init_with_poll, ExpectEvent, TryRead, TryWrite,
 };
 
 const LISTEN: Token = Token(0);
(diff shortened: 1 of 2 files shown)


tests/tcp.rs
```diff
@@ -1229,4 +1248,13 @@ fn tcp_no_events_after_deregister() {
     let mut buf = [0; 10];
     assert_eq!(stream.read(&mut buf).unwrap(), 4);
     assert_eq!(&buf[0..4], &[1, 2, 3, 4]);
+
+    stream2.write_all(&[1, 2, 3, 4]).unwrap();
+    expect_no_events(&mut poll, &mut events);
+
+    std::thread::sleep(std::time::Duration::from_millis(200));
```
Thomasdezeeuw: Why don't we wait for a readable event for stream?

dtacalau: At this point the stream has benn deregitered, goal of the test is to check there are no events returned after deregister. But I understand what you're saying, and implemented expect_no_event_with_timeout instead of doing a plain sleep. I wanted to make sure there's stuff to read before calling read.

tests/tcp.rs
```diff
@@ -1229,4 +1248,13 @@ fn tcp_no_events_after_deregister() {
     let mut buf = [0; 10];
     assert_eq!(stream.read(&mut buf).unwrap(), 4);
     assert_eq!(&buf[0..4], &[1, 2, 3, 4]);
+
+    stream2.write_all(&[1, 2, 3, 4]).unwrap();
+
+    expect_no_events_with_timeout(&mut poll, &mut events, Some(Duration::from_millis(200)));
```
Thomasdezeeuw: Why does this need a special timeout value? I rather not have an additional function for this.

dtacalau: I want to make sure there's enough delay between the previous write and this read, so the read will complete successfuly. That's why I had the timeout in the first place. I think having two versions of expect_no_events, one with timeout exposed will be usefull for future tests. It's weird that both expect_no_events and expect_events do poll with timeout inside, we might want to expose them both once the tests took of and we write many more tests. That's my thinking atm.

Thomasdezeeuw: On second thought I think you had it right the first time round. Lets change this back to `sleep(Duration::from_millis(200))`, you can import `std::thread::sleep` at the top. Sorry for this distraction.

dtacalau: np, I've reverted the last change

tests/tcp.rs
kleimkuhler: `expect_no_events` already polls with a `50ms` timeout, so I think this explicit sleep following a call to `expect_no_events` may not be necessary--especially the fact that there is another call to it at the end of the test

dtacalau: Could be, but I don't want the test result to depend on how fast expect_no_events is, or what timeout it uses internally. Doesn't hurt having an explicit timeout, it is safer imo.

PR #1154, opened 2019-11-16
Simplify windows SockState reference counting

piscisaureus:


src/sys/windows/mod.rs
```diff
@@ -51,8 +52,8 @@ pub use udp::UdpSocket;
 pub use waker::Waker;
 
 pub trait SocketState {
-    fn get_sock_state(&self) -> Option<Arc<Mutex<SockState>>>;
-    fn set_sock_state(&self, sock_state: Option<Arc<Mutex<SockState>>>);
+    fn get_sock_state(&self) -> Option<Pin<Arc<Mutex<SockState>>>>;
```
Thomasdezeeuw: Could you add a comment saying why this needs to be `Pin`?

src/sys/windows/selector.rs
```diff
@@ -192,11 +161,12 @@ impl SockState {
...
 
-            let wrapped_overlapped = OverlappedArcWrapper::new(self_arc);
-            let overlapped = wrapped_overlapped.get_ptr() as *const _ as PVOID;
             let result = unsafe {
-                self.afd
-                    .poll(&mut self.poll_info, (*self.iosb).as_mut_ptr(), overlapped)
+                self.afd.poll(
+                    &mut self.poll_info,
+                    &mut *self.iosb,
+                    transmute_copy(self_arc),
```
Thomasdezeeuw: Can you move this `transmute_copy` out the call and use explicit types, transmute is generally a very unsafe function.

src/sys/windows/selector.rs
```diff
@@ -211,13 +181,9 @@ impl SockState {
...
             }
 
-            if self.self_wrapped.is_some() {
-                // This shouldn't be happening. We cannot deallocate already pending overlapped before feed_event so we need to stand out here to declare unreachable.
-                unreachable!();
-            }
             self.poll_status = SockPollStatus::Pending;
-            self.self_wrapped = Some(wrapped_overlapped);
             self.pending_evts = self.user_evts;
+            forget(self_arc.clone());
```
Thomasdezeeuw: Can you add a comment saying why we need to increase the ref count by one?

src/sys/windows/selector.rs
```diff
@@ -8,7 +8,8 @@ use crate::{Interests, Token};
 use miow::iocp::{CompletionPort, CompletionStatus};
 use miow::Overlapped;
 use std::collections::VecDeque;
-use std::mem::size_of;
+use std::marker::PhantomPinned;
+use std::mem::{forget, size_of, transmute_copy};
```
Thomasdezeeuw: Is there any way we can not use `transmute_copy`?

piscisaureus: I don't agree that `transmute_copy` is specifically "very unsafe", it is in fact safer than performing a raw pointer cast.
E.g. it could be done as follows:

```
fn update(&mut self, self_arc: &Pin<Arc<Mutex<SockState>>>) -> io::Result<()> {
  // <snip>
  let overlapped = unsafe { *(self_arc as *const _ as *const PVOID) };
```

... but that does not even validate that `Arc<T>` and `void*` have the same size.
So, unless there are other reasons to switch, I'm sticking with transmute_copy.

piscisaureus: I forgot to mention:

The more readable solution would have been to use `Arc::into_raw(self_arc.clone())` and `Arc::from_raw()`.
Unfortunately that doesn't work in this case because we need to ensure the reference count is increased only if the overlapped AFD_POLL operation starts successfully, but we need the pointer to try starting it. 

🐔/🥚, so to speak.

Thomasdezeeuw: > ... but that does not even validate that `Arc<T>` and `void*` have the same size.

Neither does `transmute_copy`, in the docs it states "It is not a compile-time error if `T` and `U` have different sizes, but it is highly encouraged to only invoke this function where `T` and `U` have the same size. This function triggers undefined behavior if `U` is larger than `T`." (https://doc.rust-lang.org/std/mem/fn.transmute_copy.html)

Can't we increase the ref counter using `Arc::into_raw(self_arc.clone())` and then if we fail reduce it again? And if we succeed just continue?

piscisaureus: > Neither does transmute_copy, in the docs it states "It is not a compile-time error if T and U have different sizes, but it is highly encouraged to only invoke this function where T and U have the same size. This function triggers undefined behavior if U is larger than T." (https://doc.rust-lang.org/std/mem/fn.transmute_copy.html)

I didn't realize transmute_copy behaves different than transmute() in this regard.

> Can't we increase the ref counter using Arc::into_raw(self_arc.clone()) and then if we fail reduce it again? And if we succeed just continue?

It is possible, but is it better?

I have no interest in churning this PR, I have other things to do.
Feel free to push to this branch and land it once it you think it is acceptable.

dtacalau: @Thomasdezeeuw here's another idea, why not ensuring token is unique for every registered socket and use the token as overlapped instead of playing with pointers and reference counting. Maybe I'm missing something but, Pin and transmute_copy doesn't seem to bring any benefit over the previous solution with overlapped_wrapepr. 

If token is uniqe, we could easily id a socket when you get a completion callback, no more playing with reference, and unsafe pointers, also it is going to work nicely with deleting sockets on selector drop (the fix for mem leaks).

carllerche: In idiomatic rust, it is far preferred to do the raw pointer deref / cast than using `transmute`. Generally, one should avoid all transmute unless there is no other way to achieve the goal.

carllerche:
I believe that `Arc` implies `Pin` (since you cannot get a `&mut`), so having both would be redundant?

Thomasdezeeuw:
> I believe that `Arc` implies `Pin` (since you cannot get a `&mut`), so having both would be redundant?

This not correct. `Pin` implies that the object it points to (`P`) must be pinned in memory and thus may not be moved, unless it implements `Unpin`. Since the memory is used by the kernel it must not be moved or the kernel might write into memory we don't own (any more). So `Pin<Arc<T>>` is completely valid.

Come to thing of it, we might want to add `PhantomPinned` to the memory that may not be moved to ensure this.

piscisaureus:
> Come to thing of it, we might want to add PhantomPinned to the memory that may not be moved to ensure this.

It is already there: https://github.com/tokio-rs/mio/pull/1154/files#diff-8e89b097d3e3119d177930239584d29eR103

carllerche:
@Thomasdezeeuw you can't move memory out of an `Arc` w/o `unsafe` anyway? Or am i missing something?

dtacalau:
@carllerche , @Thomasdezeeuw Not sure where this PR will end up, if it's going to get merged or not, I'm trying to get a resolution with the mem leak fix . I've also pinged @piscisaureus . Could you take a look at what I'm proposing here 
https://github.com/tokio-rs/mio/pull/1147#issuecomment-555686890 and let me know what you think, yay or nay?

Thomasdezeeuw:
> @Thomasdezeeuw you can't move memory out of an `Arc` w/o `unsafe` anyway? Or am i missing something?

You're right, but it more about showing the intent of the code through the type system. There is for example: [`Arc::get_mut`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut), which would allow you to move the memory, `Pin` would disallow this.

Thomasdezeeuw:
@piscisaureus I've created https://github.com/piscisaureus/mio/pull/1, if that is OK & merged then this can be merged.

dtacalau:
@Thomasdezeeuw, @piscisaureus  I've created https://github.com/piscisaureus/mio/pull/2, I think I finally nailed the leak fix as simple as possible :).

Thomasdezeeuw:
@dtacalau If I understand the problem correctly I don't think your fix will work when a socket has been registered (i.e. the ref count increased), but has no events for it when the `Selector` is being dropped.

Also I'm not a fan of allocating `Events` in the `Drop` impl.

dtacalau:
> @dtacalau If I understand the problem correctly I don't think your fix will work when a socket has been registered (i.e. the ref count increased), but has no events for it when the `Selector` is being dropped.

@Thomasdezeeuw it works, do you mean this kind of test? Because this was the test I used to discover the leak initially. Also it's been my main method of testing the leak fix worked. Let me know if you have other test in mind for which you think the leak is not fixed.
```
#[test]
fn no_leak_registered_listener_no_events()
{
    let (mut poll, mut events) = init_with_poll();

    let listener = TcpListener::bind(any_local_address()).unwrap();
    let address = listener.local_addr().unwrap();

    poll.registry()
        .register(&listener, ID1, Interests::READABLE)
        .expect("unable to register TCP listener");

      
    poll.poll(&mut events, Some(std::time::Duration::from_millis(500))).expect("unable to poll");
}

```

> 
> Also I'm not a fan of allocating `Events` in the `Drop` impl.

@Thomasdezeeuw what's wrong with that?

Thomasdezeeuw:
> > Also I'm not a fan of allocating `Events` in the `Drop` impl.
> 
> @Thomasdezeeuw what's wrong with that?

Allocating memory using it once and then dropping it again, especially in the `Drop` impl, is an unneeded waste. We can use an stack allocated array of raw events and the raw system call and do minimal amount of processing.

Also for the test: I'm a little surprised that it leaks memory. Because I wouldn't think that draining the events would resolve it, as I wouldn't expect any events for the `TcpListener`.

dtacalau:
> Allocating memory using it once and then dropping it again, especially in the `Drop` impl, is an unneeded waste. We can use an stack allocated array of raw events and the raw system call and do minimal amount of processing.

@Thomasdezeeuw I can do it optimal if that's the issue.

> Also for the test: I'm a little surprised that it leaks memory. Because I wouldn't think that draining the events would resolve it, as I wouldn't expect any events for the `TcpListener`.

It works because draining the events will also release the SockState Arc reference which is blocked during afd poll call.

dtacalau:
> Allocating memory using it once and then dropping it again, especially in the `Drop` impl, is an 
> unneeded waste. We can use an stack allocated array of raw events and the raw system call and do > minimal amount of processing.
 
@Thomasdezeeuw simplified, please have a look https://github.com/piscisaureus/mio/pull/2/commits/92e037e6a1db8f7a9ce6914b1ad7d390ec301655

dtacalau:
@Thomasdezeeuw , @carllerche please review so we can merge this ASAP.

Thomasdezeeuw:
I still don't entirely agree with the last commit, but I don't have enough time to properly review it. So let's move that to another pr so we can this merged.

src/sys/windows/selector.rs
```diff
@@ -274,6 +272,21 @@ impl SockState {
     }
 }
 
+/// Converts the pointer to a `SockState` into a raw pointer.
+/// To revert see `from_overlapped`.
+fn into_overlapped(sock_state: Pin<Arc<Mutex<SockState>>>) -> PVOID {
+    let overlapped_ptr: *const Mutex<SockState> =
+        unsafe { Arc::into_raw(Pin::into_inner_unchecked(sock_state)) };
```
dtacalau: This breaks CI tests on Windows. If this gets merged as it is, it will block testing on Windows side because CI will always fail here. 

error[E0658]: use of unstable library feature 'pin_into_inner'
   --> src\sys\windows\selector.rs:279:32
    |
279 |         unsafe { Arc::into_raw(Pin::into_inner_unchecked(sock_state)) };
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: for more information, see https://github.com/rust-lang/rust/issues/60245

For more information about this error, try `rustc --explain E0658`.
error: Could not compile `mio`.

To learn more, run the command again with --verbose.

dtacalau:
> I still don't entirely agree with the last commit, but I don't have enough time to properly review it. So let's move that to another pr so we can this merged.

@Thomasdezeeuw I would need the leak fix commit merged, so if you don't have time to review it now, I'd suggest merging it as it is and review it later. It fixed the leaks, and that's good enough for now IMO.

dtacalau:
@Thomasdezeeuw what's the plan with this PR? When do you plan to merge it? The CI error seems blocking ```279 | unsafe { Arc::into_raw(Pin::into_inner_unchecked(sock_state)) };``` , at the same time the leak fix depends on your comit from this PR.

Thomasdezeeuw:
I've created #1172 to fix the CI. @piscisaureus can you drop the last commit, then we can merge this. @dtacalau can you crate a new pr with the last commit? So we can discuss it further without stalling this pr.

Going forward I would prefer to focus the prs on single issues, prs like these take way too long to get merged even without all the additional commits added on. In the future let's focus on a single (smaller) change in prs and try to get them merged faster. Note that I'm also far from perfect in this regard with my own prs.

dtacalau:
> I've created #1172 to fix the CI. @piscisaureus can you drop the last commit, then we can merge >this. @dtacalau can you crate a new pr with the last commit? So we can discuss it further without >stalling this pr.

@Thomasdezeeuw ok, new PR it is. Would appreciate it if you'd review it sooner than later.

> Going forward I would prefer to focus the prs on single issues, prs like these take way too long to >get merged even without all the additional commits added on. In the future let's focus on a single
>(smaller) change in prs and try to get them merged faster. Note that I'm also far from perfect in >this regard with my own prs.

I agree 100% with what you've said. Also, on another note, seems to me we're spending too much time trying to get the code perfect. Doesn't make much sense atm,  because, at least on Windows, the code base is still young, things gets changed often. I think we'll be more efficient with our time doing good enough changes, and leave the perfect for later when the code base is stable.

carllerche:
@Thomasdezeeuw I merged master, which should fix CI. I think we should move forward w/ this now and iterate in follow up PRs. wdyt?

Thomasdezeeuw:
I've merged this to get things moving, but I've reopened #1146 to continue there.

PR #1162, opened 2019-11-22
chore: bump CI nighlty + remove `=` dependency

carllerche:


Cargo.toml
```diff
@@ -31,7 +31,7 @@ log      = "0.4"
 slab     = "0.4.0"
 net2     = "0.2.29"
 iovec    = "0.1.1"
-cfg-if   = "=0.1.9"
```
hawkw: was there a reason for the “=“ dep, or is this just lost to the ages?

carllerche: no idea.

carllerche: Oh, to match the min rust version  mio targets.

carllerche: I reverted to keep it simple.

PR #1165, opened 2019-11-23
Update UDP example

Thomasdezeeuw:
Changes the structure a bit to match the TCP server example. Enables
logging and prints how the server can be accessed, also matching the TCP
server example

### DIFF
--- a/examples/udp_server.rs
+++ b/examples/udp_server.rs
@@ -1,58 +1,71 @@
-use mio::net::UdpSocket;
-use mio::{Events, Interests, Poll, Token};
 use std::io;
 
-const IN: Token = Token(0);
+use mio::net::UdpSocket;
+use mio::{Events, Interests, Poll, Token};
+
+// A token to allow us to identify which event is for the `UdpSocket`.
+const UDP_SOCKET: Token = Token(0);
 
 fn main() -> io::Result<()> {
-    // Set up a new socket on port 9000 to listen on.
-    let socket = UdpSocket::bind("0.0.0.0:9000".parse().unwrap())?;
-    // Initialize poller.
-    let mut poll = Poll::new()?;
-    // Register our socket with the token IN (defined above) and an interest
-    // in being `READABLE`.
-    poll.registry().register(&socket, IN, Interests::READABLE)?;
+    env_logger::init();
 
-    // Prepare a buffer for the number of events we can handle at a time.
-    // Someone might wat to echo really fast so lets give it some size.
-    let mut events = Events::with_capacity(1024);
-    // Initialize a buffer for the UDP datagram
-    let mut buf = [0; 65535];
-    // Main loop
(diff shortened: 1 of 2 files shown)


examples/udp_server.rs
```diff
@@ -1,58 +1,75 @@
...
                         }
                     }
                 },
-                // We only have IN as a token, so this should never ever be hit
-                _ => unreachable!(),
+                _ => {
+                    // This should never happen as we only registered our
+                    // `UdpSocket` using the `UDP_SOCKET` token, but if it ever
+                    // does we'll log it.
+                    warn!("Got event for unexpected token: {:?}", event);
```
shirshak55: shouldn't it throw error?

Thomasdezeeuw: I wouldn't advice to so. Although it should never happen, you probably don't want your production application to go down just because it received an event it can easily ignore.

shirshak55: Hmm it should be error! instead of warn!?

Thomasdezeeuw: That depends on your philosophy on logging I guess. Personally I use `error!` for things that need intimidate attention, for me this doesn't fall in that category.

examples/udp_server.rs
```diff
@@ -1,58 +1,75 @@
...
-    // in being `READABLE`.
-    poll.registry().register(&socket, IN, Interests::READABLE)?;
-
-    // Prepare a buffer for the number of events we can handle at a time.
-    // Someone might wat to echo really fast so lets give it some size.
-    let mut events = Events::with_capacity(1024);
-    // Initialize a buffer for the UDP datagram
-    let mut buf = [0; 65535];
-    // Main loop
+    // Create storage for events, since we will only register a single socket a
```
kleimkuhler: TIOLI: "Create storage for events. Since we will only register a single socket, a capacity of 1 will do."

examples/udp_server.rs
```diff
@@ -1,58 +1,75 @@
...
+    let socket = UdpSocket::bind(addr)?;
+
+    // Register our socket with the token defined above and an interest in being
+    // `READABLE`.
+    poll.registry()
+        .register(&socket, UDP_SOCKET, Interests::READABLE)?;
+
+    println!("You can connect to the server using `nc`:");
+    println!(" $ nc -u 127.0.0.1 9000");
+    println!("Anything you type we'll be echoed back to you.");
```
kleimkuhler: typo: "... will be echoed back ..."

examples/udp_server.rs
```diff
@@ -1,58 +1,75 @@
...
+
+    // Initialize a buffer for the UDP packet. We use the maximum size of a UDP
+    // packet, which is the maximum value of 16 bit integer.
```
kleimkuhler: typo: "... of a 16 bit integer."

Issue #1080, opened 2019-09-08
Investigate UdpSocket test failures


Thomasdezeeuw:
See #1070 tests that are ignored on Windows.

dtacalau:
@Thomasdezeeuw could you please assign this issue to me. I've started working on it.

dtacalau:
Tests fail because of issue described here #1131, we need to reregister the interests after a read which emptied the sock bufffer.

dtacalau:
Fixed by @1167. @Thomasdezeeuw  this can be closed.

Issue #1131, opened 2019-11-08
Edge-triggered behaviour is broken for Windows poll


dtacalau:
Mio tests expect Windows poll to be ET, but it's not, this  is the cause of many test failures on Windows. I want to have a place to discuss ways to fix this away from other issues which may cause specific tests to fail.

So far we have the following data:
1. a TcpListener::accept won't trigger READABLE event, issue #1126 
2. a TcpStream::write done after a TcpStream::read which emptied the sock buffer, won't trigger READABLE event, issue #1078
3. UDP sockets are also affected by no.2, issue #1082 

All failures above are fixed if I add reregister(READABLE) before the actual poll call.

Previous discussions on reregistering on successful call here 
https://github.com/tokio-rs/mio/pull/1112/commits/3acd892d27cd76b37f5f4660a0e0cf7f56bef2c1#r340302964 and here https://github.com/tokio-rs/mio/pull/1127#issuecomment-550386326

I think the reregistering should be handled per each specific call, so, for each call affected by this issue we need to detect the condition when the next call would block and do reregister:
1. accept -> do reregister on succesful accept
```
TcpListener::accept 
    pub fn accept(&self) -> io::Result<(TcpStream, SocketAddr)> {
        try_io!(self, accept)
        if (OK) reregister(READABLE)
```
2. read -> do peek after each successful read to see if next read would block
```
TcpStream::read 
    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
        try_io!(self, read, buf)
        if (peek() = wouldblock) reregister(READABLE) 

    }
```
3. write ??? not sure how to detect next write would block?

@Thomasdezeeuw let me know what you think, I could come up with a patch, but would like to be on the same page with this before coding it :).

Thomasdezeeuw:
I think I mentioned this already on one of the issues; I think we way to fix this is to reregister after an I/O operation if:
* We hit a would-block error (already happens).
* We exhausted the I/O operations, e.g. in case of a read: `bytes_read <= buf.len()`.

Thomasdezeeuw:
@dtacalau what do you think? I might same some time to try this later tonight.

dtacalau:
> I think I mentioned this already on one of the issues; I think we way to fix this is to reregister after an I/O operation if:
> 
> * We hit a would-block error (already happens).
> * We exhausted the I/O operations, e.g. in case of a read: `bytes_read <= buf.len()`.

[...]

PR #1167, opened 2019-11-25
fixed issue #1080, re-enable failing udp tests on windows

dtacalau:
Fixes issues #1080 & #1131.

Couple of fixes two re-enable two udp tests which were failing on windows.
Added two new tests to check ET for UDP sockets, tests rely on a WouldBlock recv(_from) to reregister the socket and reset the interests.
Added  a note in the read/peek(_from) docs that on Windows reading with a too small a buffer will return an error. 

No need to emulate edge-triggers for UDP sockets by doing reregister on partial reads/write.

### DIFF
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -228,6 +228,14 @@ impl UdpSocket {
     /// Receives data from the socket. On success, returns the number of bytes
     /// read and the address from whence the data came.
     ///
+    /// # Notes
+    ///
+    /// On Windows, if the data is larger than the buffer specified, the buffer
+    /// is filled with the first part of the data, and recv_from returns the error
+    /// WSAEMSGSIZE(10040). The excess data is lost.
+    /// Make sure to always use a sufficiently large buffer to hold the
+    /// maximum UDP packet size, which can be up to 64 kilobytes in size.
+    ///
     /// # Examples
     ///
     /// ```no_run
@@ -255,6 +263,14 @@ impl UdpSocket {
     /// Receives data from the socket, without removing it from the input queue.
     /// On success, returns the number of bytes read and the address from whence
     /// the data came.
+    ///
+    /// # Notes
+    ///
+    /// On Windows, if the data is larger than the buffer specified, the buffer
+    /// is filled with the first part of the data, and peek_from returns the error
+    /// WSAEMSGSIZE(10040). The excess data is lost.
+    /// Make sure to always use a sufficiently large buffer to hold the
+    /// maximum UDP packet size, which can be up to 64 kilobytes in size.
     ///
     /// # Examples
     ///
@@ -288,12 +304,28 @@ impl UdpSocket {
 
     /// Receives data from the socket previously bound with connect(). On success, returns
     /// the number of bytes read.
+    ///
+    /// # Notes
+    ///
+    /// On Windows, if the data is larger than the buffer specified, the buffer
+    /// is filled with the first part of the data, and recv returns the error
+    /// WSAEMSGSIZE(10040). The excess data is lost.
+    /// Make sure to always use a sufficiently large buffer to hold the
+    /// maximum UDP packet size, which can be up to 64 kilobytes in size.
     pub fn recv(&self, buf: &mut [u8]) -> io::Result<usize> {
         self.sys.recv(buf)
     }
 
     /// Receives data from the socket, without removing it from the input queue.
     /// On success, returns the number of bytes read.
+    ///
+    /// # Notes
+    ///
+    /// On Windows, if the data is larger than the buffer specified, the buffer
+    /// is filled with the first part of the data, and peek returns the error
+    /// WSAEMSGSIZE(10040). The excess data is lost.
+    /// Make sure to always use a sufficiently large buffer to hold the
+    /// maximum UDP packet size, which can be up to 64 kilobytes in size.
     pub fn peek(&self, buf: &mut [u8]) -> io::Result<usize> {
         self.sys.peek(buf)
     }
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -234,7 +234,7 @@ impl UdpSocket {
     /// is filled with the first part of the data, and recv_from returns the error
     /// WSAEMSGSIZE(10040). The excess data is lost.
     /// Make sure to always use a sufficiently large buffer to hold the
-    /// maximum UDP packet size, which can be up to 64 kilobytes in size.
+    /// maximum UDP packet size, which can be up to 65536 bytes in size.
     ///
     /// # Examples
     ///
@@ -270,7 +270,7 @@ impl UdpSocket {
     /// is filled with the first part of the data, and peek_from returns the error
     /// WSAEMSGSIZE(10040). The excess data is lost.
     /// Make sure to always use a sufficiently large buffer to hold the
-    /// maximum UDP packet size, which can be up to 64 kilobytes in size.
+    /// maximum UDP packet size, which can be up to 65536 bytes in size.
     ///
     /// # Examples
     ///
@@ -311,7 +311,7 @@ impl UdpSocket {
     /// is filled with the first part of the data, and recv returns the error
     /// WSAEMSGSIZE(10040). The excess data is lost.
     /// Make sure to always use a sufficiently large buffer to hold the
-    /// maximum UDP packet size, which can be up to 64 kilobytes in size.
+    /// maximum UDP packet size, which can be up to 65536 bytes in size.
     pub fn recv(&self, buf: &mut [u8]) -> io::Result<usize> {
         self.sys.recv(buf)
     }
@@ -325,7 +325,7 @@ impl UdpSocket {
     /// is filled with the first part of the data, and peek returns the error
     /// WSAEMSGSIZE(10040). The excess data is lost.
     /// Make sure to always use a sufficiently large buffer to hold the
-    /// maximum UDP packet size, which can be up to 64 kilobytes in size.
+    /// maximum UDP packet size, which can be up to 65536 bytes in size.
     pub fn peek(&self, buf: &mut [u8]) -> io::Result<usize> {
         self.sys.peek(buf)
     }
--- a/tests/udp_socket.rs
+++ b/tests/udp_socket.rs
@@ -288,7 +288,6 @@ fn reconnect_udp_socket_sending() {
 }
 
 #[test]
-#[cfg_attr(windows, ignore = "fails on Windows, see #1080")]
 fn reconnect_udp_socket_receiving() {
     let (mut poll, mut events) = init_with_poll();
 
@@ -334,7 +333,11 @@ fn reconnect_udp_socket_receiving() {
     let mut buf = [0; 20];
     expect_read!(socket1.recv(&mut buf), DATA1);
 
+    //this will reregister socket1 resetting the interests
+    assert_would_block(socket1.recv(&mut buf));
+
     socket1.connect(address3).unwrap();
+
     checked_write!(socket3.send(DATA2));
 
     expect_events(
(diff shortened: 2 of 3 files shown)


dtacalau:

src/sys/windows/udp.rs
```diff
@@ -57,23 +57,30 @@ impl UdpSocket {
...
+        try_io!(
+            self.send_to(buf, target),
+            |bytes_written: &usize| *bytes_written <= buf.len()
+        )
     }
 
     pub fn recv_from(&self, buf: &mut [u8]) -> io::Result<(usize, SocketAddr)> {
-        try_io!(self, recv_from, buf)
+        try_io!(self.recv_from(buf), |t: &(usize, SocketAddr)| match t {
+            (bytes_read, _) => *bytes_read <= buf.len(),
```
Thomasdezeeuw: UDP doesn't work the same way TCP does. Calling `recv(_from)` will receive a single packet. Because of this`buf` is usually `1 << 16` bytes (~65k, the maximum size of a single packet) and most read packets will be smaller then that. Which means that it will always re-register, which emulating level triggers.

src/sys/windows/udp.rs
```diff
@@ -57,23 +57,30 @@ impl UdpSocket {
...
+        })
     }
 
     pub fn peek_from(&self, buf: &mut [u8]) -> io::Result<(usize, SocketAddr)> {
         try_io!(self, peek_from, buf)
```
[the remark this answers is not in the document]
dtacalau: > Do we need it on `peek(_from)`?

@Thomasdezeeuw I don't think we need it, peek is usually followed by read which does the reregister thing.

tests/util/mod.rs
```diff
@@ -172,8 +172,8 @@ pub fn assert_error<T, E: fmt::Display>(result: Result<T, E>, expected_msg: &str
         Err(err) => assert!(
             err.to_string().contains(expected_msg),
             "wanted: {}, got: {}",
+            expected_msg,
```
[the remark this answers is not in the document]
dtacalau: @Thomasdezeeuw what am I missing? This is how a triggered assert looks like with the fix. 
```

---- unconnected_udp_socket_connected_methods stdout ----
thread 'unconnected_udp_socket_connected_methods' panicked at 'wanted: blablabla, got: A request to send or receive data was disallowed because the socket is not connected and (when sending on a datagram socket using a sendto call) no address was supplied. (os error 10057)', tests\util\mod.rs:172:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

```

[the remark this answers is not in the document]
dtacalau: The fix seems correct to me, wanted is what you gave as argument to assert, right?

src/sys/windows/udp.rs
dtacalau: In this case we need to always reregister for recv(_from) on successfull op.

src/sys/windows/udp.rs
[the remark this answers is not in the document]
Thomasdezeeuw: But we must also support cases where we don't `read` after a `peek`. That said I guess you could argue that peek don't drain the readiness. So lets leave it for now.

src/sys/windows/udp.rs
```diff
@@ -57,31 +57,39 @@ impl UdpSocket {
...
+        try_io!(self.recv(buf), |_| true)
     }
 
     pub fn peek(&self, buf: &mut [u8]) -> io::Result<usize> {
         try_io!(self, peek, buf)
     }
 
     pub fn connect(&self, addr: SocketAddr) -> io::Result<()> {
-        self.inner.connect(addr)
+        let result = self.inner.connect(addr);
```
Thomasdezeeuw: Are you sure this is required?

tests/udp_socket.rs
```diff
@@ -343,9 +342,15 @@ fn reconnect_udp_socket_receiving() {
...
-    // Read only a part of the data.
-    let max = 4;
-    expect_read!(socket1.recv(&mut buf[..max]), &DATA2[..max]);
+    if cfg!(not(target_os = "windows")) {
+        // On all platform except Windows, read only a part of the data.
+        let max = 4;
+        expect_read!(socket1.recv(&mut buf[..max]), &DATA2[..max]);
+    }
+    if cfg!(target_os = "windows") {
+        // On Windows, read all data, reading part of data returns error WSAEMSGSIZE (10040)
```
Thomasdezeeuw: Let's not split up this test in Windows/others. Lets just read the entire packet on all platform.

Also lets add a note in the `read`/`peek`(`_from`) docs that on Windows reading with a too small a buffer will return an error. Encouraging people to use `1 << 16` sized buffers (the maximum size for a UDP packet).

tests/udp_socket.rs
```diff
@@ -343,9 +342,15 @@ fn reconnect_udp_socket_receiving() {
...
+        expect_read!(socket1.recv(&mut buf), DATA2);
+    }
 
     // Now connect back to socket 2, dropping the unread data.
```
Thomasdezeeuw: This needs to be updated if we read the entire packet.

tests/udp_socket.rs
```diff
@@ -387,7 +391,15 @@ fn unconnected_udp_socket_connected_methods() {
     );
 
     // Socket is unconnected, but we're using an connected method.
-    assert_error(socket1.send(DATA1), "address required");
+    if cfg!(not(target_os = "windows")) {
```
Thomasdezeeuw: Same here as above.

tests/udp_socket.rs
```diff
@@ -848,3 +860,79 @@ pub fn multicast() {
...
+        vec![
+            ExpectEvent::new(ID1, Interests::WRITABLE),
+            ExpectEvent::new(ID2, Interests::WRITABLE),
+        ],
+    );
+
+    {
+        socket2.connect(address3).unwrap();
+
+        expect_events(
```
Thomasdezeeuw: Connect doesn't generate event on most platforms. So I don't think this behaviour is required.

tests/util/mod.rs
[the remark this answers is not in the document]
Thomasdezeeuw: Sorry you're right. Weird that nobody noticed.

src/sys/windows/udp.rs
Thomasdezeeuw: Don't think its required, see comment in `udp_socket_edge_triggered_behavior` test.

tests/udp_socket.rs
Thomasdezeeuw: This is also why the CI fails.

dtacalau:
@Thomasdezeeuw I found some interesting things related to ET while playing with tests. This PR depends on the conclusion from here https://github.com/tokio-rs/mio/issues/1131#issuecomment-559073004

src/sys/windows/udp.rs
dtacalau: you're right, not required.

tests/udp_socket.rs
dtacalau: ok, will read all

src/sys/windows/udp.rs
dtacalau: reverted this change, not needed.

tests/udp_socket.rs
dtacalau: In which way needs to be updated?

tests/udp_socket.rs
dtacalau: Not sure if possible, there are different error messages on Windows vs other platforms. How would you have one assert checking two different expected messages?

tests/udp_socket.rs
dtacalau: reverted, not needed. Actually connect generates events on all platforms except Linux. CI failed on Linux only.

tests/udp_socket.rs
dtacalau: I've added note for all read/peek(_from)

dtacalau:
@carllerche please review, thanks!

dtacalau:
https://github.com/tokio-rs/mio/issues/1174 opened to investigate *nix behavior.

tests/udp_socket.rs
carllerche: Mio does not guarantee to generate events unless you get a WouldBlock :)

tests/udp_socket.rs
carllerche: The comment implies data is dropped, but you updated the test to read all the data (no data is dropped). I guess it should say:

```suggestion
    // Now connect back to socket 2.
```

tests/udp_socket.rs
carllerche: I don't think mio makes any guarantees as to error messages. It doesn't matter terribly to me if the test **only** check that an error is received **or** it adds cfg flags to test messages per platform.

tests/udp_socket.rs
dtacalau: Yep, love this Mio guarantee, makes things much clearer than before.

src/net/udp.rs
```diff
@@ -228,6 +228,14 @@ impl UdpSocket {
     /// Receives data from the socket. On success, returns the number of bytes
     /// read and the address from whence the data came.
     ///
+    /// # Notes
+    ///
+    /// On Windows, if the data is larger than the buffer specified, the buffer
+    /// is filled with the first part of the data, and recv_from returns the error
+    /// WSAEMSGSIZE(10040). The excess data is lost.
+    /// Make sure to always use a sufficiently large buffer to hold the
+    /// maximum UDP packet size, which can be up to 64 kilobytes in size.
```
Thomasdezeeuw: Its actually 65536 bytes, lets just use `1 << 16`.

Thomasdezeeuw: Same for the other notes.

tests/udp_socket.rs
Thomasdezeeuw: You can just change this to `assert_error(socket1.send(DATA1), "address")` and add a comment saying it must say something about missing an address. Then we don't have this complex `cfg!` stuff.

tests/udp_socket.rs
```diff
@@ -848,3 +862,101 @@ pub fn multicast() {
...
+        &mut poll,
+        &mut events,
+        vec![ExpectEvent::new(ID2, Interests::WRITABLE)],
+    );
+
+    {
+        socket1.connect(address2).unwrap();
+
+        let mut buf = [0; 20];
+        checked_write!(socket1.send(DATA1));
```
Thomasdezeeuw: Writing without receiving an writable event first.

tests/udp_socket.rs
```diff
@@ -848,3 +862,101 @@ pub fn multicast() {
...
+        .register(&socket2, ID2, Interests::READABLE.add(Interests::WRITABLE))
+        .expect("unable to register UDP socket");
+
+    expect_events(
+        &mut poll,
+        &mut events,
+        vec![ExpectEvent::new(ID2, Interests::WRITABLE)],
+    );
+
+    {
```
Thomasdezeeuw: Don't think this extra block is needed.

dtacalau: ok, removed

tests/udp_socket.rs
dtacalau: ok, will expect writable before

src/net/udp.rs
dtacalau: 64 kilobytes is 65536 bytes. I will replace with 65536 bytes, which is the most explicit, I dont find 1 << 16 better.

tests/udp_socket.rs
dtacalau: I'd rather not change this, unless you have a strong reason against it. Having cfg! in tests doesn't feel complex at all, actually it's weird we don't have more cfg! in them. Having just an assert_error(socket1.send(DATA1), "address") doesn't say much about the actual error you are expecting there. "address required" is enough to get an idea,  "no address was supplied. (os error 10057)" is even better. IMO tests should be as explicit as possible.

dtacalau:
> Some small nits, but looks good overall.

@Thomasdezeeuw fixed some of them.

PR #1168, opened 2019-11-25
More changes for the leak fix, issue 1146

dtacalau:


dtacalau:
@Thomasdezeeuw this is the new PR for the leak fix you asked me to open. Don't mind the other two comits, I'll rebase it once you merge the other two comits, it can be reviewed, nonetheless.

Thomasdezeeuw:
Does this have changes not in #1154?

dtacalau:
> Does this have changes not in #1154?

Yes, this comit is new.
https://github.com/tokio-rs/mio/pull/1168/commits/7fa44c02e9329e8bd7c5d44da255a25c097ff68b

dtacalau:
I rebased on master.

Thomasdezeeuw:
Can you easily create a test for this? Or would it be hard to reproduce such a scenario?

dtacalau:
> Can you easily create a test for this? Or would it be hard to reproduce such a scenario?

@Thomasdezeeuw  What kind of test do you have in mind? 

I accidentally noticed while testing that you might not get all the events in one call. For instance if you have readable and writable registered for the same socket, you might get readable event in the first poll call, and writable in the 2nd poll call. It happened once during debugging though, was not able to reproduce it.

Thomasdezeeuw:
@dtacalau thanks.  I've merged this as is, if you can think of a test to ensure we don't regress I would appreciate it.

PR #1170, opened 2019-11-27
Remove = requirement and use cargo update --precise instead

### LINKED ISSUE #1169: Locked cfg-if version conflicts with log 0.4.8 on mio 0.6.20
```
$ cargo fetch

error: failed to select a version for `cfg-if`.
    ... required by package `mio v0.6.20`
    ... which is depended on by `tokio v0.2.0`
    ... which is depended on by (my project)
versions that meet the requirements `= 0.1.9` are: 0.1.9

all possible versions conflict with previously selected packages.

  previously selected package `cfg-if v0.1.10`
    ... which is depended on by `log v0.4.8`
    ... which is depended on by `fern v0.5.9`
    ... which is depended on by (my project)

failed to select a version for `cfg-if` which could resolve this conflict
```

est31:
Adopting the changes suggested by @cuviper in https://github.com/tokio-rs/mio/pull/1099#issuecomment-557664365 .

Fixes #1169

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ log      = "0.4"
 slab     = "0.4.0"
 net2     = "0.2.29"
 iovec    = "0.1.1"
-cfg-if   = "=0.1.9"
+cfg-if   = "0.1.9"
 
 [target.'cfg(target_os = "fuchsia")'.dependencies]
 fuchsia-zircon = "0.3.2"
--- a/ci/azure-test-stable.yml
+++ b/ci/azure-test-stable.yml
@@ -23,7 +23,10 @@ jobs:
     parameters:
       rust_version: ${{ parameters.rust_version }}
 
-  - script: cargo ${{ parameters.cmd }}
+  - script: |
+      cargo update
+      cargo update -p cfg-if --precise 0.1.9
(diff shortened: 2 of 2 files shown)


Thomasdezeeuw:
Thanks!

@carllerche this might require a new 0.6.x release.

carllerche:
If we are going to make this change, we should increase the minimum supported rust version.

est31:
@carllerche why? There is no need to increase it on the 0.6.x branch too, and it is useful to have a large MSRV range.

Most people probably won't, but anyone who wants to actually use Rust 1.18 can just run`cargo update --precise` themselves, or if they want to, they can add a `=` requirement for cfg-if to their own Cargo.toml as well. Whatever mio does, it's *already* the case that you can't run `cargo update` blindly if you need to support an older Rust version, but you have to update crates one by one. So people just do `cargo update -p mio`.

There are further alternatives like a clone of the entire registry that contains only crates supported by a specific Rust version, or custom resolvers entirely (there exist prototypes for both). Managing MSRVs via `=` requirements is not good though and causes problems for everyone. Nor is increasing MSRV just for this :).

carllerche:
sounds good

PR #1172, opened 2019-11-27
Update minimum Rust version to 1.39

Thomasdezeeuw:


Thomasdezeeuw:
This could also use some documentation in the README but that can be done later.

PR #1175, opened 2019-11-27
Reverted reregistering on partial read/write for TcpStream. 

dtacalau:
Mio guarantee is that, after you get a WouldBlock, you will receive a readiness notification, so, you always have to perform the operation you care about, receive would block, then wait for the readiness event.

A WouldBlock read doing reregister is enough to fix the TcpStream test failing on Windows. More details in this issue https://github.com/tokio-rs/mio/issues/1131#issuecomment-559179472

### DIFF
--- a/src/sys/windows/mod.rs
+++ b/src/sys/windows/mod.rs
@@ -35,19 +35,6 @@ macro_rules! try_io {
         }
         result
     }};
-    ($self: ident.$method: ident ( $($arg: expr),* $(,)* ), $reregister_check: expr)  => {{
-        let result = (&$self.inner).$method($($arg),*);
-        let should_reregister = match &result {
-            // See if we need reregister interest to emulate edge-triggers.
-            Ok(ok) => $reregister_check(ok),
-            Err(err) if err.kind() == io::ErrorKind::WouldBlock => true,
-            Err(_) => false,
-        };
-        if should_reregister {
-            $self.io_blocked_reregister()?;
-        }
-        result
-    }};
 }
 
 mod afd;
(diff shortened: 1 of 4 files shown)


Thomasdezeeuw:
I don't intend on accepting this, see https://github.com/tokio-rs/mio/issues/1131#issuecomment-559489244.

Thomasdezeeuw:
Is the extended variant of `try_io!` (that adds a check to see if `reregister` needs to be called) still used somewhere?

Thomasdezeeuw:
This should also be done for `TcpListener::accept` I guess?

dtacalau:
@Thomasdezeeuw I will double check try_io and accept!

dtacalau:
> @Thomasdezeeuw I will double check try_io and accept!

@Thomasdezeeuw I've removed the reregister done on successfull TcpListener::accept, and commented-out the extended try_io! macro. Could be usefull in the future, I'd not remove it completely.

src/sys/windows/mod.rs
```diff
@@ -35,7 +35,9 @@ macro_rules! try_io {
         }
         result
     }};
-    ($self: ident.$method: ident ( $($arg: expr),* $(,)* ), $reregister_check: expr)  => {{
+   /* // This try_io variant does reregister if a certain condition is met
```
Thomasdezeeuw: You can just remove this, git history hold this code anyway.

dtacalau: ok, removed.

Thomasdezeeuw:
@dtacalau thanks!

PR #1178, opened 2019-11-30
added notes about limitations of TcpListener try_clone

dtacalau:
For start I've added notes only for the TcpListener. After review I'll add them for TcpStream and UdpSocket also.

Initial discussion: https://github.com/tokio-rs/mio/issues/1074#issuecomment-552478325

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -135,12 +135,13 @@ use std::{fmt, io};
 /// # fn main() -> Result<(), Box<dyn Error>> {
 /// use mio::{Poll, Interest, Token};
 /// use mio::net::TcpStream;
+/// use std::net::SocketAddr;
 /// use std::time::Duration;
 /// use std::thread;
 ///
-/// let address = "127.0.0.1:9001".parse()?;
-/// # let _listener = net::TcpListener::bind(address)?;
-/// let mut sock = TcpStream::connect(address)?;
+/// let address: SocketAddr = "127.0.0.1:0".parse()?;
+/// let listener = net::TcpListener::bind(address)?;
+/// let mut sock = TcpStream::connect(listener.local_addr()?)?;
 ///
 /// thread::sleep(Duration::from_secs(1));
 ///
@@ -438,13 +439,14 @@ impl Registry {
     /// # fn main() -> Result<(), Box<dyn Error>> {
     /// use mio::{Events, Poll, Interest, Token};
     /// use mio::net::TcpStream;
+    /// use std::net::SocketAddr;
     /// use std::time::{Duration, Instant};
     ///
     /// let mut poll = Poll::new()?;
     ///
-    /// let address = "127.0.0.1:9002".parse()?;
-    /// # let _listener = net::TcpListener::bind(address)?;
-    /// let mut socket = TcpStream::connect(address)?;
+    /// let address: SocketAddr = "127.0.0.1:0".parse()?;
+    /// let listener = net::TcpListener::bind(address)?;
+    /// let mut socket = TcpStream::connect(listener.local_addr()?)?;
     ///
     /// // Register the socket with `poll`
     /// poll.registry().register(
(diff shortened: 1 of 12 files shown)


Thomasdezeeuw:
@carllerche what do you think? I do want a `ty_clone` method, but we're basically saying it's broken by design.

carllerche:
Yes we should remove `try_clone` if it is broken to that extent.

Thomasdezeeuw:
@dtacalau can you prepare a commit removing `TcpListener::try_clone` with as commit message the documentation provided in this pr? If not I can do it.

dtacalau:
@Thomasdezeeuw was about to ask that. I can do it, of course. Remove it on windows only?

Thomasdezeeuw:
@dtacalau No please remove it on all platforms. `epoll(2)` has its own set of problems with `dup`ed file descriptors.

dtacalau:
@Thomasdezeeuw I've prepared a commit with the removal of TcpListener::try_clone. Not sure if it's a good idea to remove the tests also, could be useful later on. WDYT?

I think we should remove UdpSocket::try_clone and TcpStream::try_clone too, because on Windows they suffer from the same issues as TcpListener::try_clone. I'd expect to be the same on *nices too.

Thomasdezeeuw:
> @Thomasdezeeuw I've prepared a commit with the removal of TcpListener::try_clone. Not sure if it's a good idea to remove the tests also, could be useful later on. WDYT?

Remove the tests as well, git history keeps everything.

> I think we should remove UdpSocket::try_clone and TcpStream::try_clone too, because on Windows they suffer from the same issues as TcpListener::try_clone. I'd expect to be the same on *nices too.

Let's remove all  `try_clone` methods, but can you put in a single commit with the previously mentioned message. This way when people ask why its removed we can point them to the commit.

dtacalau:
Rebased to latest master.

carllerche:
Looks like a doc test (net::tcp::stream::TcpStream ) is specifying an explicit port instead of using `0`. That makes for a Ci race condition.

dtacalau:
> Looks like a doc test (net::tcp::stream::TcpStream ) is specifying an explicit port instead of using `0`. That makes for a Ci race condition.

@carllerche changed the doc tests to always use 0 port.

src/poll.rs
```diff
@@ -135,12 +135,13 @@ use std::{fmt, io};
...
 /// use mio::{Poll, Interest, Token};
 /// use mio::net::TcpStream;
+/// use std::net::SocketAddr;
 /// use std::time::Duration;
 /// use std::thread;
 ///
-/// let address = "127.0.0.1:9001".parse()?;
-/// # let _listener = net::TcpListener::bind(address)?;
-/// let sock = TcpStream::connect(address)?;
+/// let address:SocketAddr = "127.0.0.1:0".parse()?;
```
kleimkuhler: Missing a space here

src/poll.rs
```diff
@@ -582,13 +585,14 @@ impl Registry {
...
     /// use mio::net::TcpStream;
+    /// use std::net::SocketAddr;
     /// use std::time::Duration;
     ///
     /// let mut poll = Poll::new()?;
     ///
-    /// let address = "127.0.0.1:9004".parse()?;
-    /// # let _listener = net::TcpListener::bind(address)?;
-    /// let socket = TcpStream::connect(address)?;
+    /// let address:SocketAddr = "127.0.0.1:0".parse()?;
```
kleimkuhler: Missing a space here

src/poll.rs
dtacalau: removed

src/poll.rs
dtacalau: removed

kleimkuhler:
@dtacalau Could you resolve the merge conflicts? With #1180 merging before this, there were a few changes that should be pretty quick to fix!

dtacalau:
> @dtacalau Could you resolve the merge conflicts? With #1180 merging before this, there were a few changes that should be pretty quick to fix!

@kleimkuhler done

dtacalau:
@carllerche I think this can be merged.

kleimkuhler:
@dtacalau I'd like to get #1184 good to go before merging in this. They are not dependent on each other, but if they are going to remain separate lets land them when both are ready.

Thomasdezeeuw:
Thanks @dtacalau.

PR #1179, opened 2019-12-04
Rename Interests to Interest

### LINKED ISSUE #993: Consider renaming `Interests` -> `Interest`
I'm not 100% sure about this. I believe naming the type `Interest` would be more correct, but I could be wrong.

Thomasdezeeuw:
Closes #993


Thomasdezeeuw:
Since this touches (almost) all files a quick review would be nice.

Issue #1064, opened 2019-08-20
Change event::Source to take &mut (instead of &self)


Thomasdezeeuw:
This would allow future implementations to forgo any locking/synchronisation if it requires mutable access to any of its internals. Any types that don't need this, for example most types currently exported, can also implement the trait for `&Self` (e.g. `impl event::Source for &TcpStream`) to allow concurrent access. We do something similar with the `io::Read` and `io::Write` traits for `TcpStream`.

`Registry::(re,de)register` would also take `&mut S` (`S: event::Source`).

/cc @carllerche @asomers

carllerche:
We can go this route, but that would mean we aren’t merging Poll and Registry?

carllerche:
Basically, if we keep the Poll / Registry split, this is a good next step. If not, we probably want to keep &self and say Source impls should be sync

Thomasdezeeuw:
Now that `Poll` and `Registry` no longer require an `Arc` I'm fine with keeping them split apart. How does Tokio use (or would use) the new `Registry`? Really I'm fine with either option at this point.

carllerche:
Ah, Tokio would want to store a `Registry` handle in an `Arc`, so I guess that would mean `Registry` can't take `&mut self`.

asomers:
This would be helpful for mio-aio, which needs to modify the `Evented` during `register` and `deregister`.

carllerche:
The downside would be it would make tokio's `PollEvented` helper more complicated. Right now, it takes `T: Source`. Given that `PollEvented` requires `Sync`, i'm not 100% sure how it would be updated yet.

carllerche:
It would probably require adding a new trait (not necessarily in mio): `SyncSource`

carllerche:
I actually don't think ^^ is bad. I probably should probably not export mio from tokio anyway and make it an implementation detail

Thomasdezeeuw:
@carllerche maybe something like this would work; https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=20a13845d951953a9bf8eb90f76e251b.

carllerche:
@Thomasdezeeuw but what does PollEvented<T> use as a bound for T?

I think it would require GAT... which could be simulated with where for but that is ugly

Thomasdezeeuw:
[...]

PR #1182, opened 2019-12-04
Remove some mutexes from Windows implementation

Thomasdezeeuw:
After #1064 we don't need as many mutexes in the Windows implementation.

### DIFF
--- a/src/sys/windows/selector.rs
+++ b/src/sys/windows/selector.rs
@@ -483,12 +483,10 @@ impl SelectorInner {
         token: Token,
         interests: Interest,
     ) -> io::Result<InternalState> {
-        let flags = interests_to_afd_flags(interests);
-
         let sock = {
             let sock = this._alloc_sock_for_rawsocket(socket)?;
             let event = Event {
-                flags,
+                flags: interests_to_afd_flags(interests),
                 data: token.0 as u64,
             };
             sock.lock().unwrap().set_event(event);
@@ -502,16 +500,12 @@ impl SelectorInner {
             sock_state: sock.clone(),
         };
 
-        unsafe {
-            let mut update_queue = this.update_queue.lock().unwrap();
-            update_queue.push_back(sock);
-            this.update_sockets_events_if_polling()?;
-        }
-
-        Ok(state)
+        let mut update_queue = this.update_queue.lock().unwrap();
+        update_queue.push_back(sock);
+        unsafe { this.update_sockets_events_if_polling().map(|()| state) }
     }
 
-    pub fn reregister(
+    pub(super) fn reregister(
         &self,
         state: &Pin<Arc<Mutex<SockState>>>,
         token: Token,
@@ -522,7 +516,6 @@ impl SelectorInner {
                 flags: interests_to_afd_flags(interests),
                 data: token.0 as u64,
             };
-
             state.lock().unwrap().set_event(event);
         }
 
--- a/src/sys/windows/selector.rs
+++ b/src/sys/windows/selector.rs
@@ -499,7 +499,7 @@ impl SelectorInner {
             selector: this.clone(),
             token,
             interests,
-            sock_state: Some(sock.clone()),
+            sock_state: sock.clone(),
         };
 
         unsafe {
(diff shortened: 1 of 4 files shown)


Thomasdezeeuw:
@dtacalau I don't like doing this but I know the `registry_behind_arc` test fails. I think there is a problem with locking going on, but I don't really have the tools on Windows to properly debug it. Could you take a quick look? If you can't find it quickly let me know and I'll continue debugging.

dtacalau:
@Thomasdezeeuw no problem, let me have a look.

dtacalau:
@Thomasdezeeuw problem is that when the test ```registry_behind_arc``` reached this line ``` poll.poll(&mut events, Some(Duration::from_millis(1000)))``` the update queue is empty, so there's no sockstate to retrieve events for.  ```    unsafe fn update_sockets_events(&self) -> io::Result<()> {
        let mut update_queue = self.update_queue.lock().unwrap();```  . The update queue should have had two sock states, one for listener and another one for stream. That's all I could find for now.

dtacalau:
@Thomasdezeeuw on another note I thought we agreed not to have large PRs with multiple changes at once.  Maybe I'm missing something but these commits look like major refactoring to me, not just ```remove some mutexes```. Is this the right thing to do before the release when we have just stabilized the tests? Sorry if I don't have the full picture with what we're trying to achieve here.

https://github.com/tokio-rs/mio/pull/1154#issuecomment-559043943
```
Going forward I would prefer to focus the prs on single issues, prs like these take way too long to get merged even without all the additional commits added on. In the future let's focus on a single (smaller) change in prs and try to get them merged faster
```

src/sys/windows/selector.rs
```diff
@@ -357,6 +357,16 @@ impl Selector {
         self.inner.register(socket, token, interests)
     }
 
+    pub(super) fn register2(
```
dtacalau: what's the purpose of (re)register2?

src/sys/windows/selector.rs
```diff
@@ -348,6 +347,7 @@ impl Selector {
         self.inner.select(events, timeout)
     }
 
+    // FIXME: replace with `register2`.
```
dtacalau: why replace it with register2?

Thomasdezeeuw:
> @Thomasdezeeuw on another note I thought we agreed not to have large PRs with multiple changes at once. Maybe I'm missing something but these commits look like major refactoring to me, not just `remove some mutexes`.

You're right about the large pr, I've split the parts in three commits (one for each net type), followed by 3 clean-up commits. I could have send each commit as a separate pr.

>  Is this the right thing to do before the release when we have just stabilized the tests? Sorry if I don't have the full picture with what we're trying to achieve here.

Pr #1180 only recently got merged and this allows us to simplify these implementations. Personally I feel this makes the entire Windows implementation simpler. Especially if you look at the last four commits of #1183 (which builds on top of this pr).

dtacalau:
@Thomasdezeeuw All I'm saying is this PR is hard to follow and review. I think the whole process would become far mor easier for everyone if you'd replace this large PR with smaller PRs, maybe the root cause of the ```registry_behind_arc ``` bug will also pop out and we won't lose time debugging it. 

What do you say? We got to start from somewhere with the small PRs agreement, right?

Thomasdezeeuw:
@dtacalau can you review by commit, I always try to make nice clean commits. I don't have time to create a pr for each commit.

dtacalau:
> @dtacalau can you review by commit, I always try to make nice clean commits. I don't have time to create a pr for each commit.

@Thomasdezeeuw I'd prefer separate PRs, why not letting these changes for later when you have time for multiple PRs?

Reviewing by commit doesn't seem feasible, in this case there are changes added in one commit and removed in next commits, for instance register2.

Just my 5 cents on the large vs small PRs issue: Reviewing a large PR would cost me more time than needed reviewing a bunch of smaller PRs. So someone will have to spend more time eventually.

Thomasdezeeuw:
I'll try to redo this at some point when I have some time.

Issue #880, opened 2018-10-18
TcpStream & TcpListener rely on internals


damonbarry:
I'm working on UDS support for Windows. I hope to start with my own mio-uds-windows crate and maybe get it merged into mio-uds at some point. I'm new to all this so I'm probably missing something obvious, but I've had a really hard time finding a good starting point for my Windows UnixStream and UnixListener.

The obvious choice was to start with mio's TcpStream and TcpListener. But I can't build them outside of the mio codebase. They make use of SelectorId, which isn't public. I can make a copy of the SelectorId code, but it accesses the private `selector` field on `Poll`. Also, they wrap the public `mio::windows::Binding` with `ReadyBinding`. When I copy that type, it won't build because it accesses the private `selector` field to get/put buffers down in `BufferPool`.

I could start from mio-named-pipe, but it only implements the stream not the listener, it doesn't implement `read_bufs`/`write_bufs` (even if it did, it wouldn't be able to use the internal `BufferPool`), and it preallocates memory and does an extra copy for reads (8K).

Anyway, I'm wondering if it would be possible to extricate TcpStream and TcpListener in such a way that they make a good example for other streams and listeners, but without breaking things?

carllerche:
Sorry for the delay. It looks like you found a work around? I'm not entirely sure I follow the original question.

In general, the windows impl strategy is most likely going to significantly change in the future. I expect we will use a strategy similar to http://github.com/piscisaureus/wepoll.

damonbarry:
@carllerche Yes I found a workaround, although it's not something we want to stick with long-term.

@arsing describes the problem more succinctly in [this comment](https://github.com/alexcrichton/mio-uds/issues/7#issuecomment-473025878) on alexcrichton/mio-uds#7.

arsing:
@carllerche

The problem we're trying to solve is that we want mio-uds and downstream crates to support Unix Domain Sockets on Windows. Currently we have a hacky way of doing this that involves duplicating a lot of code from libstd, mio and mio-uds into https://github.com/Azure/mio-uds-windows and using `[replace]` to swap out mio-uds for that repo. But we want to work out a way to upstream this into mio-uds in the long term.

[...]

PR #1183, opened 2019-12-05
Add SourceSocket

Thomasdezeeuw:
The Windows only counterpart to `SourceFd`.

All commits in #1182 are in the pr as well.

Updates #880.


Thomasdezeeuw:
The CI will fail with the same failure as in #1182.

Thomasdezeeuw:
I'll try to redo this at some point when I have some time.

PR #1192, opened 2019-12-06
Introduce default API shim and feature flags for net/OS resources 

### LINKED ISSUE #1118: RFC: default features is API core only, require feature flags for tcp, udp, uds and OS selector
## Goal

Mio should be as light as possible and be forwards compatible to other non-blocking I/O strategies, such as userland sockets via DPDK.

## Proposal

By default, the Mio crate would be an API only shim. Types such as `Poll`, `Events`, etc would be present, but would be shims only. Attempting to use them would result in a panic. TCP, UDP, and UDS types would not be present by default and would require feature flags to opt-in.

In order to include the epoll, kqueue, iocp implementations of `Poll`, the user of Mio would need to include the `poll-os` feature flag (open to naming suggestions). In order to get the socket implementations, the user would need to include the `tcp`, `udp`, or `uds` feature flags.

In the future, we could include a `dpdk` feature flag, enabling user-land sockets.

Including the `tcp`, `udp`, and `uds` feature flags without `poll-os` would result in "shim" socket APIs.

Including no features by default is so that libraries that depend on mio don't accidentally include "the kitchen sink".

kleimkuhler:
## Motivation

Closes #1118

Mio should be as light as possible and forwards compatible to other non-blocking
I/O strategies such as user-land sockets.

Currently, depending on Mio means that all platform-specific networking
primitives and I/O source readiness polling functionality is present by default.

If a library is interested only in networking primitives but aims to provide an
other strategy for I/O source readiness polling, there is no way to disable the
provided implementation.

If a library is interested in the provided I/O source readiness polling
implementations, but only uses the TCP networking types, the UDP and UDS types
are also still provided.

This change proposes that Mio by default provides an API shim for its common
types such as `Poll`, `Events`, etc. Attempting to use them will result in a
panic. TCP, UDP, and UDS networking types are **not** present by default and
require feature flags to opt-in.

### Overview

Below are the four feature flags that Mio now includes as opt-in features. By
default, no functionality is provided--only an API shim.

- `os-poll`: Provide the cross-platform implementations of `Poll`. This is
  currently epoll, IOCP, or kqueue depending on the platform.

- `os-util`: Provide the cross-platform adapters for the platform's underlying
  I/O source that implements `event::Source`. Currently, only Unix provides it's
  `SourceFd` type. There are proposed solutions for a Windows equivalent being
  worked on.

- `tcp`: Provide the TCP types and implementations

- `udp`: Provide the UDP types and implementations

- `uds`: Provide the UDS types and implementations. Currently, only Unix
  provides types and implementations for this feature.

The above feature flags can be used in any combination. If a library opts in to
`os-poll` and `tcp`, then the `Poll` and TCP implementations will be provided.
Later, the library could provide it's own `Poll` implementation and opt-out of
`os-poll`. It could alternatively opt-in to UDP at a later point without issue.

### Details

Mio now has the following cfg macros: `cfg_os_poll`, `cfg_not_os_poll`,
`cfg_net`, `cfg_tcp`, `cfg_udp`, `cfg_uds`, `cfg_any_os_util` that conditionally
provide the related types and implementations. *(If any of these names are not
obvious about what they provide, please let me know!)*.

The macros--again similar to Tokio--conditionally provide shell types that
**panic upon use**, or hide/expose methods on types an API shim should not
provide (e.g. `Poll::new`).

## Tips for Review

This change touches a lot of files, but there are no intended functional or
behavioral changes. I would recommend first looking at what types are provided
by the root-level `lib.rs` or changes in the `mod.rs` files. *(If any of the
`mod.rs` files are not clear about what they conditionally expose, please let me
know!)*.

Next, I would look at the methods that are now conditionally provided on the
types. These are mostly confined to the OS specific `Poll` implementations, but
there are a few exceptions.

Finally, there are some further changes that reorganize `impl`s or move `use`
statements due to being conditionally used. Because this change is ultimately a
"refactor", I found myself moving some blocks of code around to group with
similar `impl`s and such. I'd like to keep those as part of this change (instead
of a separate PR), but please me know if have questions/concerns about any of
those.


Thomasdezeeuw:
Why are you creating a new `mio` directory?

Cargo.toml
```diff
@@ -24,7 +24,11 @@ exclude       = [
 publish = false
 
 [features]
-default = []
+os-poll = []
```
dtacalau: I think it would be nice to have the purpose/contents of each feature documented. Newcomers, as myself, will appreciate it. Not a strong requirement, just a nice to have, if you have time.

kleimkuhler: Yep that makes sense. I'll add some explanation!

dtacalau: Thanks!

kleimkuhler:
@Thomasdezeeuw `examples` relies on non-default features. In order to depend on specific features of `mio`, it needed to moved outside of the crate into the now `mio` workspace. With this change, `mio` now has a top-level workspace with just two members (`mio` and `examples`).

It's the same thing `tokio` does with having a top-level workspace for reference!

Thomasdezeeuw:
> @Thomasdezeeuw `examples` relies on non-default features. In order to depend on specific features of `mio`, it needed to moved outside of the crate into the now `mio` workspace. With this change, `mio` now has a top-level workspace with just two members (`mio` and `examples`).
> 
> It's the same thing `tokio` does with having a top-level workspace for reference!

I don't know if we want to add more members/directories later, but you can use "." as a workspace member and keep the original structure and just make the examples directory another crate.

kleimkuhler:
@Thomasdezeeuw `examples` is already another crate in this change. It's `Cargo.toml` depends on Mio with `mio = { path = "../mio", features = ["os-poll", "tcp", "udp"] }`.

In my original attempts I tried making it `.. path = "." ..` and similar variants had no success. That's when I saw how Tokio handles it and now have this structure.

Thomasdezeeuw:
Would the same structure as [`libc-test`](https://github.com/rust-lang/libc/blob/4f11029a68040c90acf771976b019c1ef273a8cd/libc-test/Cargo.toml) work?

kleimkuhler:
@Thomasdezeeuw Interesting! I tried that structure at it kind of works. The main issue I have with it is the entire examples still needed to be cfg'd and run with `cargo run --features os-poll,tcp --example tcp_server`. The run command makes sense, but the example has to be cfg'd and provide a `main` function that is compiled when the required features are not passed in order to satisfy RLS/RA.

It complicates the example which I think should be avoided, because the examples shouldn't require understanding why two separate `main`s are provided that are separately cfg'd.

carllerche:
@kleimkuhler you can use the `required-features` attribute in this case: https://doc.rust-lang.org/cargo/reference/manifest.html#the-required-features-field-optional

carllerche:
We will probably want to add feature doc flags:

- Add [this](https://github.com/tokio-rs/tokio/blob/master/tokio/Cargo.toml#L132) to `Cargo.toml`.
- [This](https://github.com/tokio-rs/tokio/blob/master/tokio/src/macros/cfg.rs#L74) to macros

That should get docs started.

kleimkuhler:
@Thomasdezeeuw @carllerche I reverted the restructure and the examples now use `required-features = ..`. Thanks for the pointers from both of you on that!

Documentation is now also cfg'd. The CI scripts have mostly been changed to use `--all-features` for commands like `check` and `test`.

Lastly, a `_require_all.rs` test file has been added to `tests/` to indicate tests should be run with the required features. This is similar to [tokio's](https://github.com/tokio-rs/tokio/blob/master/tokio/tests/_require_full.rs)

src/sys/mod.rs
```diff
@@ -13,52 +13,81 @@
...
 
-#[cfg(windows)]
-pub use self::windows::{event, Event, Events, Selector, TcpListener, TcpStream, UdpSocket, Waker};
+    cfg_udp! {
+        pub(crate) use self::unix::UdpSocket;
+    }
+
+    cfg_uds! {
+        pub(crate) use self::unix::{UnixDatagram, UnixListener, UnixStream};
+        pub use self::unix::SocketAddr;
```
Thomasdezeeuw: Is there are reason for the mix of `pub` and `pub(crate)`, I think can all be `pub(crate)`?

src/sys/mod.rs
```diff
@@ -13,52 +13,81 @@
...
+    }
+}
 
 #[cfg(windows)]
-mod windows;
+cfg_os_poll! {
+    mod windows;
+    pub(crate) use self::windows::*;
```
Thomasdezeeuw: Is there a reason for the difference with the `unix` module imports? I like this method better personally.

src/sys/unix/mod.rs
```diff
@@ -1,33 +1,55 @@
...
+        mod uds;
+        pub use self::uds::SocketAddr;
+        pub(crate) use self::uds::{UnixDatagram, UnixListener, UnixStream};
+    }
+}
+
+cfg_not_os_poll! {
+    cfg_uds! {
+        mod uds;
+        pub use self::uds::SocketAddr;
```
Thomasdezeeuw: `pub(crate)`? Same for util below.

src/sys/unix/selector/epoll.rs
```diff
@@ -72,33 +67,34 @@ impl Selector {
...
         syscall!(epoll_ctl(self.ep, libc::EPOLL_CTL_ADD, fd, &mut event)).map(|_| ())
     }
+}
 
-    pub fn reregister(&self, fd: RawFd, token: Token, interests: Interest) -> io::Result<()> {
-        let mut event = libc::epoll_event {
-            events: interests_to_epoll(interests),
-            u64: usize::from(token) as u64,
-        };
+cfg_os_poll! {
```
Thomasdezeeuw: Is this needed? It seems odd that `Selector::register` for example is not here.

src/sys/unix/selector/kqueue.rs
```diff
@@ -255,47 +151,162 @@ impl Selector {
...
-    .map(|_| ())
-    .or_else(|err| {
-        // According to the manual page of FreeBSD: "When kevent() call fails
-        // with EINTR error, all changes in the changelist have been applied",
-        // so we can safely ignore it.
-        if err.raw_os_error() == Some(libc::EINTR) {
-            Ok(())
-        } else {
-            Err(err)
+cfg_os_poll! {
```
Thomasdezeeuw: Same question as for `epoll`; is this needed?

Cargo.toml
```diff
@@ -25,6 +25,16 @@ publish = false
 
 [features]
 default = []
```
Thomasdezeeuw: Do we want to include all features by default and make libraries set `default-features = false`?

tests/_require_all.rs
```diff
@@ -0,0 +1,8 @@
+#![cfg(not(all(
+    feature = "os-poll",
+    feature = "os-util",
+    feature = "tcp",
+    feature = "udp",
+    feature = "uds"
+)))]
+compile_error!("run main Mio tests with `--all-features`");
```
Thomasdezeeuw: Don't we also want to support tests with partial features enabled? Maybe not now but in the future?

ci/azure-cross-compile.yml
```diff
@@ -52,17 +52,8 @@ jobs:
...
 
-    - script: cargo check --tests --target $(target)
+    - script: cargo check --tests --target $(target) --all-features
       displayName: Check tests
 
-    - script: cargo check --examples --target $(target)
+    - script: cargo check --examples --target $(target) --all-features
       displayName: Check examples
-
-    - script: cargo check --target $(target) --no-default-features
```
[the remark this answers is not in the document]
kleimkuhler: I've been using [cargo-hack](https://github.com/taiki-e/cargo-hack)'s `--feature-powerset` for testing and it has worked well. For now I have it testing all features, but I think a follow-up PR can make some additional changes that focus on mixes of features to test.

Cargo.toml
```diff
@@ -25,6 +25,16 @@ publish = false
...
+# Include implementations of I/O source readiness polling.
+os-poll = []
+# Include adapters for underlying OS I/O sources.
+os-util = []
```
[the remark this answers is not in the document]
kleimkuhler: It's only `SourceFd` on Unix for now as you probably saw. I'll update the comment; I agree that is not obvious right now.

src/sys/mod.rs
kleimkuhler: `SocketAddr` should be `pub` because it's the type returned by `UnixListener::accept`. The other resources should only be used within the crate because they are private fields in the `net` equivalents.

src/macros/mod.rs
```diff
@@ -0,0 +1,103 @@
+#![allow(unused_macros)]
+
+// ===== Poll =====
```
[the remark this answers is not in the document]
kleimkuhler: Removed

src/lib.rs
```diff
@@ -54,7 +59,9 @@ pub use waker::Waker;
 #[cfg(unix)]
 pub mod unix {
     //! Unix only extensions.
-    pub use crate::sys::SourceFd;
+    cfg_any_os_util! {
```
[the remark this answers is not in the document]
kleimkuhler: Ah yes good catch!

tests/_require_all.rs
kleimkuhler: Possibly! This works well for now, but should be easy to change in the future as our support for tests do. I don't want to worry about that now with this change though.

Cargo.toml
kleimkuhler: @carllerche preferred the default being the API shim, but it was a brief question in the Discord channel. Not sure if he has any more thoughts on it now.

src/sys/mod.rs
kleimkuhler: Before, the `mod unix` types were re-exported as `pub use`. In this change as you've picked up on in another comment, I added `pub use(crate)`s where applicable because it helped a lot with properly structuring things visibility wise and making sure warnings appeared on things that were not used (that should have been). Unfortunately since `SourceFd` is `pub use` (and there is no equivalent in Windows), this makes the Unix module a little more explicit.

I know what you mean about preferring the glob imports here, but I really would like to keep the `pub use(crate)`s on things that should only be crate visible. It helped a lot with making sure things were not exported that should not be.

src/sys/unix/mod.rs
kleimkuhler: `SocketAddr` and `SourceFd` are `pub`

ci/azure-cross-compile.yml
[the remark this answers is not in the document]
Thomasdezeeuw: Ok. Should we start making a list of follow-up items?

kleimkuhler:
I left comments unresolved that I didn't address or that I just answered a question. Feel free to resolve if it answers your question.

In regards to these two comments:

> Is this needed? It seems odd that Selector::register for example is not here. (context `cfg_os_poll!`)

> Same question as for epoll; is this needed? (context `cfg_os_poll!`)

Thanks for pointing that out and I removed those cfg's. I didn't really make sense that those would be cfg'd if `Poll::poll` is not cfg'd. Put another way, if you have a `Poll` instance, then it would make sense to be able to use the `Registry` API since it's not actually making a new one or anything.

Cargo.toml
Thomasdezeeuw: For new users it might be a bit annoying to start using Mio, as well as for all people that come from v0.6. I see Tokio also doesn't include anything by default, I'll leave it up to @carllerche then. I don't really have a strong opinion either way.

ci/azure-cross-compile.yml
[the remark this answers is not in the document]
kleimkuhler: Just made #1214

tests/_require_all.rs
Thomasdezeeuw: Tracked in #1214.

Cargo.toml
Thomasdezeeuw: Created #1216.

Issue #1047, opened 2019-07-27
windows: support custom completion port integrations


piscisaureus:
The new "wepoll" strategy is great for sockets but it makes it impossible to use the thread's I/O completion port for other purposes.
I would suggest to add a low-level "plugin" system so that other crates can share the completion port for different purposes.

This could be used for supporting named pipes and files, which can use overlapped I/O but their APIs work differently; more exotic examples are [ReadDirectoryChangesW](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-readdirectorychangesw) (file change notifications), [GetAddrInfoExW](https://docs.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-getaddrinfoexw) (async DNS name lookup), and [Job Objects](https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-_jobobject_associate_completion_port) (notifications about child processes).

Job objects in particular are exotic in that windows will send messages to the completion port where `lpOverlapped` does not actually point to a `struct OVERLAPPED`, but rather contains an indication of the event that happened.

So I would propose to to "switch" based on the `lpCompletionKey` value. The most simple solution would be to store a function pointer in it which receives the raw completion information, e.g. `fn handle_completion(&OVERLAPPED_ENTRY)`.

To illustrate what I mean, here's a crude prototype: https://github.com/piscisaureus/mio/commit/ac40cd71c31a5014592ba232523a4607f07f371a

Notice how the `Selector` is no longer knows anything about `Wake` events.
They work simply because the `lpCompletionKey` contains an appropriate [function pointer](https://github.com/piscisaureus/mio/commit/ac40cd71c31a5014592ba232523a4607f07f371a#diff-bde7ec4aecb3fa131e34b63a2c7cc173R33)

Thomasdezeeuw:
Related #526.

Thomasdezeeuw:
I think this is related to a `SourceSocket` idea as discussed in https://github.com/tokio-rs/mio/issues/880#issuecomment-522720856.

Thomasdezeeuw:
Punting to v1.0.

carllerche:
@piscisaureus I have a question re: this proposal.

My understanding is that, when an OVERLAPPED operation is submitted to the OS, the OS needs "ownership" of the pointer. Ownership is returned back to the caller when the operation is completed and the overlapped is received during a "poll".

[...]

Issue #526, opened 2017-01-23
Supporting Windows Job Objects


alexcrichton:
We've recently added support to handle custom `HANDLE` objects with IOCP on Windows. This relies, however, on all completion notifications coming back with a mio-specific `Overlapped` type to ensure that dispatch happens appropriately.

Job Objects on Windows support IOCP where they can send notifications to a port. There's two major differences, however, from how today's I/O objects interact with IOCP:

* Association with a port is done through a custom job object API. This means that we'll have to expose the raw IOCP handle.
* IOCP notifications do not have an associated `OVERLAPPED` pointer specified by the user. Instead each type of notification has a particular `OVERLAPPED` value which may be some extra data payload.

The former is pretty easy to solve but the latter is much more difficult as it's a big architectural change to mio. It basically means that the only thing we have to dispatch on is a token, not an overlapped pointer like we did before. This means that we'd need to have some sort of token-based dispatch ahead of the overlapped-based dispatch. I'm unsure on how to do this in a performant fashion, however, so I'm opening an issue!

dtacalau:
@Thomasdezeeuw please assign this issue to me, will start working on it.

Thomasdezeeuw:
This is not going to block a v1.0 release.

Thomasdezeeuw:
This has been open for four years without a solution, I think we have to say that using the current AFD based implementation we can't support this. If there is a reasonable solution to support more Window specific API please open a new issue

PR #1195, opened 2019-12-07
Add IoSource

### LINKED ISSUE #880: TcpStream & TcpListener rely on internals

### LINKED ISSUE #1187: Add a SourceFd (EventedFd) equivalent for Windows
Target v1.0-release

Thomasdezeeuw:
`IoSource` is a wrapper around a raw file descriptor or socket that implements `event::Source`.

This will replace `SourceFd` on Unix (not removed in this pr). I'll send some more pr that replace the usage of `SourceFd` on Unix and use them in the net types on Windows.

Updates #880
Updates #1187

/cc @dtacalau

### DIFF
--- a/src/io_source.rs
+++ b/src/io_source.rs
@@ -1,0 +1,202 @@
+#![allow(dead_code)] // Currently unused.
+
+use std::ops::{Deref, DerefMut};
+#[cfg(unix)]
+use std::os::unix::io::AsRawFd;
+#[cfg(windows)]
+use std::os::windows::io::AsRawSocket;
+use std::{fmt, io};
+
+#[cfg(unix)]
+use crate::poll;
+#[cfg(debug_assertions)]
+use crate::poll::SelectorId;
+use crate::sys::IoSourceState;
+use crate::{event, Interest, Registry, Token};
+
+/// Adapter for a [`RawFd`] or [`RawSocket`] providing an [`event::Source`]
+/// implementation.
+///
+/// `IoSource` enables registering any FD or socket wrapper with [`Poll`].
+///
+/// While only implementations for TCP, UDP, and UDS (Unix only) are provided,
+/// Mio supports registering any FD or socket that can be registered with the
+/// underlying OS selector. `IoSource` provides the necessary bridge.
+///
+/// [`RawFd`]: std::os::unix::io::RawFd
+/// [`RawSocket`]: std::os::windows::io::RawSocket
+///
+/// # Notes
+///
+/// To handle the registrations and events properly **all** I/O operations (such
+/// as `read`, `write`, etc.) must go through the [`do_io`] method to ensure the
+/// internal state is updated accordingly.
+///
+/// [`Poll`]: crate::Poll
+/// [`do_io`]: IoSource::do_io
+/*
+///
+/// # Examples
+///
+/// Basic usage.
+///
+/// ```
+/// # use std::error::Error;
+/// # fn main() -> Result<(), Box<dyn Error>> {
+/// use mio::{Interest, Poll, Token};
+/// use mio::IoSource;
+///
+/// use std::net;
+///
+/// let poll = Poll::new()?;
+///
+/// // Bind a std TCP listener.
+/// let listener = net::TcpListener::bind("127.0.0.1:0")?;
+/// // Wrap it in the `IoSource` type.
+/// let mut listener = IoSource::new(listener);
+///
+/// // Register the listener.
+/// poll.registry().register(&mut listener, Token(0), Interest::READABLE)?;
+/// #     Ok(())
+/// # }
+/// ```
+*/
+pub struct IoSource<T> {
+    state: IoSourceState,
+    inner: T,
+    #[cfg(debug_assertions)]
+    selector_id: SelectorId,
+}
+
+impl<T> IoSource<T> {
+    /// Create a new `IoSource`.
+    pub fn new(io: T) -> IoSource<T> {
+        IoSource {
+            state: IoSourceState::new(),
+            inner: io,
+            #[cfg(debug_assertions)]
+            selector_id: SelectorId::new(),
+        }
+    }
+
+    /// Execute an I/O operations ensuring that the socket receives more events
+    /// if it hit a [`WouldBlock`] error.
+    ///
+    /// # Notes
+    ///
+    /// This method is required to be called for **all** I/O operations to
+    /// ensure the user will receive events once the socket is ready again after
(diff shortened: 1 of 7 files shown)


Thomasdezeeuw:
#1196 (the last commit) changes `TcpStream` to use this type, see it for example usage.

carllerche:
Thanks for putting this together. I will need to spend more time going through it.

A quick question: if this is public, in what scenarios would a user use this? In my understanding, the AFD strategy only works for types that mio already supports out of the box.

Thomasdezeeuw:
> A quick question: if this is public, in what scenarios would a user use this? In my understanding, the AFD strategy only works for types that mio already supports out of the box.

Correct, but isn't the same true for `SourceFd`? It depends on the selector what file descriptors are supported.

carllerche:
Right... such a refactor is good for internals, but I hesitate to make the type public due to limitations. I think SourceFd is good for Unix and we can document that it is passed through to the selector so see OS docs for limitations. On windows we would need a custom IOCP handler.

Anyway, I will try to make time to review in depth ASAP

Thomasdezeeuw:
> Right... such a refactor is good for internals, but I hesitate to make the type public due to limitations. I think SourceFd is good for Unix and we can document that it is passed through to the selector so see OS docs for limitations. On windows we would need a custom IOCP handler.

Fair. After the first round of reviews I'll make them private and we can see from there if they can be made public.

dtacalau:
With my current understanding of the code I did not see the point of *nix SourceFd, seems like an extra level which can be dropped w/o losing something.

dtacalau:
> On windows we would need a custom IOCP handler.
@Thomasdezeeuw @carllerche Imo we should not do any refactoring, major or minor, of the Windows implementation internals until custom IOCP handlers are implemented (#1047 and #526). Couple of reasons why I think that: current implementation is stable, no known bugs, it has tests supporting it, provides a good base for implementing a new Internal Selector and a new HandlerState. Any refactoring done now could be duplicate work or just code that won't be needed after the custom IOCP handlers. That's why I think this PR, and the now closed PRs #1182 and #1183, would make much sense afterwards. 

Let me know what you think.

Thomasdezeeuw:
> With my current understanding of the code I did not see the point of *nix SourceFd, seems like an extra level which can be dropped w/o losing something.

`SourceFd` is the only way to use Mio with types defined outside of Mio itself. For example I use for [mio-pipe](https://github.com/Thomasdezeeuw/mio-pipe/blob/096e3f598be31d47216ca088ccfaadbc0a61ca99/src/sys/unix.rs#L13-L30) and [mio-signals](https://github.com/Thomasdezeeuw/mio-signals/blob/66d69243f17221a8ea7ca489a331301ca30282cc/src/sys/signalfd.rs#L104-L121).

> @Thomasdezeeuw @carllerche Imo we should not do any refactoring, major or minor, of the Windows implementation internals until custom IOCP handlers are implemented (#1047 and #526). Couple of reasons why I think that: current implementation is stable, no known bugs, it has tests supporting it, provides a good base for implementing a new Internal Selector and a new HandlerState. We would be in a better position to refactor having the custom iocp up and running along the Afd stuff. Any refactoring done now could be duplicate work or just code that won't be needed after the custom IOCP handlers. That's why I think this PR, and the now closed PRs #1182 and #1183, would make much sense afterwards.
> 
> Let me know what you think.

I strongly disagree with "current implementation is stable, no known bugs", you're one of the people fixing plenty of bugs in the implementation with changes as recent as today. So I wouldn't call it stable or bug free. Furthermore if we get "scared" (for lack of a better word) to change the code it simply means we don't have enough test coverage in my opinion.

If all net types (i.e. `UdpSocket`, `TcpStream` and `TcpListener`) would use `IoSource` the number of lines (and possible places of bugs) and code duplication in the Windows implementation would be greatly reduced as shown in https://github.com/tokio-rs/mio/pull/1196/commits/bb1e93fe1411df208d8133052c79554f11f6d1a0.

dtacalau:
> `SourceFd` is the only way to use Mio with types defined outside of Mio itself. 

@Thomasdezeeuw thanks for explaining usage of SourceFd, I've reviewed the docs, makes more sense now.

As for my other opinion expressed above, "Scared" might not be the proper word, I'd say "cautious" if you want a word to describe it :).

src/io_source.rs
```diff
@@ -0,0 +1,198 @@
...
+use crate::poll;
+#[cfg(debug_assertions)]
+use crate::poll::SelectorId;
+use crate::sys::IoSourceState;
+use crate::{event, Interest, Registry, Token};
+
+/// Adapter for a [`RawFd`] or [`RawSocket`] providing an [`event::Source`]
+/// implementation.
+///
+/// `IoSource` enables registering any type with an FD or socket with [`Poll`].
```
kleimkuhler: Suggestion: "`IoSource` enables registering any FD or socket wrapper with [`Poll`]"

I suggest this because the current wording of "with an FD or socket" may imply it could be a type with multiple fields; one of which is an FD or socket.

src/io_source.rs
```diff
@@ -0,0 +1,198 @@
...
+///
+/// While only implementations for TCP and UDP are provided, Mio supports
```
kleimkuhler: Suggestion: ".. for TCP, UDP, and UDS (Unix only) are provided .."

Thomasdezeeuw:
Postponing until next week, see https://github.com/tokio-rs/mio/pull/1196#issuecomment-564957220.

Thomasdezeeuw:
Closing in favour of #1236.

PR #1184, opened 2019-12-05
Removed `try_clone` Unix UDS impl 

dtacalau:
(associated tests were also removed) This api had several limitations . More details in issue #1074.

`try_clone` was just a dup, so the behavior depended on the underlying epoll
and kqueue implementations. This meant there was no defined behavior because of
the platform dependent differences between epoll and kqueue implementations.

Signed-off-by: Daniel Tacalau <dst4096@gmail.com>

PR #1196, opened 2019-12-07
Change TcpStream to use IoSource

Thomasdezeeuw:
Changes `TcpStream` to use `IoSource`.

~~Depends on #1195 (included), #1178 and #1184.~~
~~Depends on #1236.~~


src/sys/unix/net.rs
```diff
@@ -1,6 +1,28 @@
 use std::io;
 use std::mem::size_of_val;
-use std::net::SocketAddr;
+use std::net::{self, SocketAddr};
+use std::os::unix::io::FromRawFd;
+
+pub fn tcp_connect(addr: SocketAddr) -> io::Result<net::TcpStream> {
```
kleimkuhler: This could probably still go in a `unix/tcp/...` directory. I think as we are able to just rely on `deref` for *most* of the `std::net` methods, this could still pick up a few specialized functions that Mio has to do.

src/sys/unix/tcp/stream.rs
```diff
@@ -1,159 +0,0 @@
-use crate::sys::unix::net::{new_ip_socket, socket_addr};
```
kleimkuhler: This is really good to see.

src/sys/windows/mod.rs
```diff
@@ -100,6 +191,42 @@ fn init() {
     });
 }
 
+pub fn tcp_connect(addr: SocketAddr) -> io::Result<net::TcpStream> {
```
kleimkuhler: Again probably could go in a `tcp` module just keep the size of this file down.

src/sys/windows/tcp.rs
```diff
@@ -1,308 +1,21 @@
...
-}
+use std::{fmt, io};
+use winapi::um::winsock2::{bind, closesocket, listen, SOCKET_ERROR, SOCK_STREAM};
 
 pub struct TcpListener {
     internal: Box<Mutex<Option<InternalState>>>,
     inner: net::TcpListener,
 }
 
-impl TcpStream {
```
kleimkuhler: 👍

Thomasdezeeuw:
@kleimkuhler thanks for the feedback. I don't think I have enough time to do this pr justice this week, so I'm going postpone it until next week. Meaning it will not land in v0.7-alpha1.

kleimkuhler:
@Thomasdezeeuw Yep no problem sounds good.

Thomasdezeeuw:
The reason for the CI failure I've solved in #1236.

Thomasdezeeuw:
@kleimkuhler I've rebased this on #1236, so that needs to be merged first. I did address all your comments.

Thomasdezeeuw:
I'm marking this ready to review as the CI is green, but only focus on the last commit as the other are already in #1236.

Thomasdezeeuw:
Rebased this on master.

PR #1201, opened 2019-12-09
Add more tests poll registry

dtacalau:


tests/poll.rs
```diff
@@ -201,10 +208,112 @@ fn registry_behind_arc() {
     handle2.join().unwrap();
 }
 
+// Testing on Windows only where this must fail by design, on other platforms
```
kleimkuhler: Nit: Comment style is pretty different than the rest of Mio. I'd prefer something like:

```
// Testing on Windows ..
//
// This test checks the following register/deregister ..
```

And the remove the `/* .. */` comment block inside.

tests/poll.rs
```diff
@@ -201,10 +208,112 @@ fn registry_behind_arc() {
...
+// behavior is undefined.
+#[test]
+#[cfg(target_os = "Windows")]
+fn register_deregister_multiple_poll_instances() {
+    /*
+    This test checks the following register/deregister contrain:
```
kleimkuhler: Typo: "constraint"

tests/poll.rs
```diff
@@ -201,10 +208,112 @@ fn registry_behind_arc() {
...
+    A handle can be passed back to `register` after it has been
```
kleimkuhler: Is this true for `reregister` as well? If so, should that be test? If not, could there be a comment explaining why?

tests/poll.rs
```diff
@@ -201,10 +208,112 @@ fn registry_behind_arc() {
...
+
+    poll1
+        .registry()
+        .deregister(&mut listener)
+        .expect("unable to deregister listener");
+}
+
+#[test]
+fn reregister_overrides_interest() {
+    /*
```
kleimkuhler: Same nit about comment style

tests/poll.rs
```diff
@@ -201,10 +208,112 @@ fn registry_behind_arc() {
...
+    This test checks the followin reregister constraints
```
kleimkuhler: typo: "following"

Missing the ":" after "constraints".

tests/poll.rs
```diff
@@ -201,10 +208,112 @@ fn registry_behind_arc() {
...
+
+    expect_no_events(&mut poll, &mut events);
+
+    barrier.wait();
+    thread_handle.join().expect("unable to join thread");
+}
+
+#[test]
+fn register_reregister_token_usage() {
+    /*
```
kleimkuhler: ^ Same as above

tests/poll.rs
```diff
@@ -220,6 +329,56 @@ pub fn double_register() {
...
+}
+
+// Testing on Windows only where this must fail by design, on other platforms
+// behavior is undefined, for instance on kqueue platforms deregistering w/o
+// registering works.
+#[test]
+#[cfg(target_os = "windows")]
+fn deregister_without_register() {
+    /*
+    This test checks the following reregister constraint applied to deregister:
```
kleimkuhler: Wording: "This test checks the following deregister constraint:"

I wouldn't rely on the previous test to describe this

tests/poll.rs
dtacalau: ok, will redo the comments style

tests/poll.rs
dtacalau: done, corrected

tests/poll.rs
dtacalau: done, corrected.

tests/poll.rs
dtacalau: ok, corrections done

tests/poll.rs
dtacalau: redone.

tests/poll.rs
dtacalau: I've added a test for the following sequence `poll1.register(listener); poll2.reregister(listener);`. Is this what you suggested? Thanks

tests/poll.rs
dtacalau: redone

dtacalau:
@kleimkuhler thanks for reviewing. The comments also needed updates after we've documented the undefined behavior for Poll.

 I've also added another undefined behavior note about double register.

Thomasdezeeuw:
I would like to postpone this pr until something like #1195 is merged. Then we can provide a more platform consistent behaviour. For example `IoSource` (with debug assertions enabled) could keep track of registration much like Windows does already, expanding upon `SelectorId` perhaps.

dtacalau:
> I would like to postpone this pr until something like #1195 is merged.

I've added a separate PR with tests which do not verify the registry undefined behavior. We can merge those and keep this opened for the other tests.

dtacalau:
Rebased on master. @kleimkuhler

tests/poll.rs
```diff
@@ -473,6 +473,58 @@ fn poll_ok_after_cancelling_pending_ops() {
...
+// of `Poll`, otherwise the behavior is undefined.
+//
+// This test is done on Windows and epoll platforms where reregistering a
+// source without a previous register is defined behavior that fail with an
+// error code.
+//
+// On kqueue platforms reregistering w/o registering works but that's not a
+// test goal, so it is not tested.
+#[test]
+#[cfg(any(target_os = "linux", target_os = "windows"))]
```
Thomasdezeeuw: As `TcpListener` uses `IoSource` this should work on all platforms, see https://github.com/tokio-rs/mio/blob/c5f52893ffd4b641d1a457f0197804e69ad666e4/src/io_source.rs#L246-L266

tests/poll.rs
```diff
@@ -473,6 +473,58 @@ fn poll_ok_after_cancelling_pending_ops() {
...
+fn reregister_without_register() {
+    let poll = Poll::new().expect("unable to create Poll instance");
+
+    let mut listener = TcpListener::bind(any_local_address()).unwrap();
+
+    assert_eq!(
```
Thomasdezeeuw: Can you use [`assert_error`](https://github.com/tokio-rs/mio/blob/c5f52893ffd4b641d1a457f0197804e69ad666e4/tests/util/mod.rs#L176-L188) to ensure the error string at least mentioned re-registering?

tests/poll.rs
```diff
@@ -473,6 +473,58 @@ fn poll_ok_after_cancelling_pending_ops() {
...
+// of `Poll`, otherwise the behavior is undefined.
+//
+// This test is done on Windows and epoll platforms where deregistering a
+// source without a previous register is defined behavior that fail with an
+// error code.
+//
+// On kqueue platforms deregistering w/o registering works but that's not a
+// test goal, so it is not tested.
+#[test]
+#[cfg(any(target_os = "linux", target_os = "windows"))]
```
Thomasdezeeuw: Same two comments as on the other tests.

Thomasdezeeuw:
Closing in favour of #1360. Thanks @dtacalau, this will get merged!

Issue #1205, opened 2019-12-09
Sometimes poll with None timeout returns with no events


dtacalau:
I noticed this behavior when writing tests, so I wrote a simple test to isolate it.

This is a bug because according to poll doc:
```
A `timeout` of `None` means that
    /// `poll` will block until a readiness event has been received.
```

dtacalau:
Repro test:

```
fn repro() {
    // Test to verify issue 1205 does not reproduce again
    init();

    let mut poll = Poll::new().unwrap();
    let mut events = Events::with_capacity(128);

    let mut listener = TcpListener::bind(any_local_address()).unwrap();

    poll.registry()
        .register(&mut listener, ID1, Interest::READABLE)
        .unwrap();

    poll.poll(&mut events, Some(std::time::Duration::from_millis(0)))
        .unwrap();
    assert!(events.iter().count() == 0);

    let _stream = TcpStream::connect(listener.local_addr().unwrap()).unwrap();

    poll.registry().deregister(&mut listener).unwrap();


    poll.poll(&mut events, None).unwrap();
    assert!(events.iter().count() > 1);
}
```

dtacalau:
Fixed by #1206

PR #1206, opened 2019-12-09
fix for issue #1205, increase event count only for valid events, that…

dtacalau:
… is events for sockets which have not been deleted

### DIFF
--- a/tests/poll.rs
+++ b/tests/poll.rs
@@ -1,5 +1,5 @@
 use mio::net::{TcpListener, TcpStream};
-use mio::{event, Events, Interest, Poll, Registry, Token};
+use mio::{event, Events, Interest, Poll, Registry, Token, Waker};
 
 use std::net;
 use std::sync::{Arc, Barrier};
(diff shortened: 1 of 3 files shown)


tests/poll.rs
```diff
@@ -201,6 +204,49 @@ fn registry_behind_arc() {
     handle2.join().unwrap();
 }
 
+#[test]
+fn poll_no_timeout_not_affected_by_deregister() {
```
Thomasdezeeuw: Can you move this to the `regressions.rs` file, then the comment below won't be needed.

tests/poll.rs
```diff
@@ -201,6 +204,49 @@ fn registry_behind_arc() {
...
+    // Test to verify issue 1205 does not reproduce again
+    init();
```
Thomasdezeeuw: You can use `init_with_poll`, then you don't need create the `poll` and `events` below.

tests/poll.rs
dtacalau: done

tests/poll.rs
dtacalau: done

Thomasdezeeuw:
Thanks.

PR #1207, opened 2019-12-09
added a test with registry ops on multiple threads

dtacalau:

### DIFF
--- a/tests/poll.rs
+++ b/tests/poll.rs
@@ -1,6 +1,7 @@
 #![cfg(all(feature = "os-poll", feature = "tcp"))]
 
-use mio::net::{TcpListener, TcpStream};
+use mio::event::Source;
+use mio::net::{TcpListener, TcpStream, UdpSocket};
 use mio::{event, Events, Interest, Poll, Registry, Token};
 use std::net;
 use std::sync::{Arc, Barrier};
@@ -9,7 +10,14 @@ use std::time::Duration;
 use std::{fmt, io};
 
 mod util;
-use util::{any_local_address, assert_send, assert_sync, init};
+
+use util::{
+    any_local_address, assert_send, assert_sync, expect_events, init, init_with_poll, ExpectEvent,
+};
+
+const ID1: Token = Token(1);
+const ID2: Token = Token(2);
+const ID3: Token = Token(3);
 
 #[test]
 fn is_send_and_sync() {
@@ -199,6 +207,107 @@ fn registry_behind_arc() {
 
     handle1.join().unwrap();
     handle2.join().unwrap();
+}
+
+/// Call all registration operations, ending with `source` being registered with `token` and `final_interests`.
+pub fn registry_ops_flow(
+    registry: &Registry,
+    source: &mut dyn Source,
+    token: Token,
+    init_interests: Interest,
+    final_interests: Interest,
+) -> io::Result<()> {
+    registry.register(source, token, init_interests).unwrap();
+    registry.deregister(source).unwrap();
+
+    registry.register(source, token, init_interests).unwrap();
+    registry.reregister(source, token, final_interests)
+}
+
+#[test]
+fn registry_operations_are_thread_safe() {
+    let (mut poll, mut events) = init_with_poll();
+
+    let registry = Arc::new(poll.registry().try_clone().unwrap());
+    let registry1 = Arc::clone(&registry);
+    let registry2 = Arc::clone(&registry);
+    let registry3 = Arc::clone(&registry);
+
+    let barrier = Arc::new(Barrier::new(4));
+    let barrier1 = Arc::clone(&barrier);
+    let barrier2 = Arc::clone(&barrier);
+    let barrier3 = Arc::clone(&barrier);
+
+    let mut listener = TcpListener::bind(any_local_address()).unwrap();
+    let addr = listener.local_addr().unwrap();
+
+    // Expect that multiple register/deregister/reregister work fine on multiple
+    // threads. Main thread will wait before the expect_events for all other 3
+    // threads to do their work. Otherwise expect_events timeout might be too short
+    // for all threads to complete, and call might fail.
+
+    let handle1 = thread::spawn(move || {
+        registry_ops_flow(
+            &registry1,
+            &mut listener,
+            ID1,
+            Interest::READABLE,
+            Interest::READABLE,
+        )
+        .unwrap();
+
+        barrier1.wait();
+        barrier1.wait();
+    });
+
+    let handle2 = thread::spawn(move || {
+        let mut udp_socket = UdpSocket::bind(any_local_address()).unwrap();
+        registry_ops_flow(
+            &registry2,
+            &mut udp_socket,
+            ID2,
+            Interest::WRITABLE,
+            Interest::WRITABLE.add(Interest::READABLE),
+        )
+        .unwrap();
+
+        barrier2.wait();
+        barrier2.wait();
+    });
+
+    let handle3 = thread::spawn(move || {
+        let mut stream = TcpStream::connect(addr).unwrap();
+        registry_ops_flow(
+            &registry3,
+            &mut stream,
+            ID3,
+            Interest::READABLE,
+            Interest::READABLE | Interest::WRITABLE,
+        )
+        .unwrap();
+
+        barrier3.wait();
+        barrier3.wait();
+    });
+
+    // wait for threads to finish before expect_events
+    barrier.wait();
+    expect_events(
+        &mut poll,
+        &mut events,
+        vec![
+            ExpectEvent::new(ID1, Interest::READABLE),
+            ExpectEvent::new(ID2, Interest::WRITABLE),
+            ExpectEvent::new(ID3, Interest::WRITABLE),
+        ],
+    );
+
+    // Let the threads return.
+    barrier.wait();
+
+    handle1.join().unwrap();
+    handle2.join().unwrap();
+    handle3.join().unwrap();
 }
 
 // On kqueue platforms registering twice (not *re*registering) works.


tests/poll.rs
```diff
@@ -201,6 +205,98 @@ fn registry_behind_arc() {
...
+            .register(&mut stream, ID3, Interest::READABLE)
+            .unwrap();
+        registry3
+            .reregister(&mut stream, ID3, Interest::READABLE | Interest::WRITABLE)
+            .unwrap();
+
+        barrier3.wait();
+    });
+
+    std::thread::sleep(std::time::Duration::from_millis(1000));
```
Thomasdezeeuw: Is this sleep needed? `expect_events` below already loops/polls three time.

tests/poll.rs
```diff
@@ -201,6 +205,98 @@ fn registry_behind_arc() {
     handle2.join().unwrap();
 }
 
+#[test]
+fn registry_operations_are_thread_safe() {
+    init();
```
Thomasdezeeuw: You can use `init_will_poll`.

tests/poll.rs
Thomasdezeeuw: If the timeout is too short you can use `barrier.wait()` before the call to `expect_events` (and call it twice at the of the threads above), to allow the threads to do there work.

tests/poll.rs
```diff
@@ -201,6 +205,98 @@ fn registry_behind_arc() {
...
+                &mut listener1,
+                ID1,
+                Interest::READABLE.add(Interest::WRITABLE),
+            )
+            .unwrap();
+
+        barrier1.wait();
+    });
+
+    let handle2 = thread::spawn(move || {
```
Thomasdezeeuw: Why do we need another `TcpListener`? Can't we use a `UdpSocket` instead, then we have all three main types (UDS not included).

tests/poll.rs
```diff
@@ -201,6 +205,98 @@ fn registry_behind_arc() {
...
+            .unwrap();
+        registry1.deregister(&mut listener1).unwrap();
+        registry1
+            .register(&mut listener1, ID1, Interest::READABLE)
+            .unwrap();
+        registry1
+            .reregister(
+                &mut listener1,
+                ID1,
+                Interest::READABLE.add(Interest::WRITABLE),
```
Thomasdezeeuw: `WRITABLE` makes no sense for a `TcpListener`.

tests/poll.rs
```diff
@@ -201,6 +205,98 @@ fn registry_behind_arc() {
...
+            .unwrap();
+        registry2.deregister(&mut listener2).unwrap();
+        registry2
+            .register(&mut listener2, ID2, Interest::READABLE)
+            .unwrap();
+        registry2
+            .reregister(
+                &mut listener2,
+                ID2,
+                Interest::READABLE.add(Interest::WRITABLE),
```
Thomasdezeeuw: `WRITABLE` makes no sense for a `TcpListener`.

tests/poll.rs
```diff
@@ -201,6 +205,98 @@ fn registry_behind_arc() {
...
+
+    let registry1 = Arc::clone(&registry);
+    let registry2 = Arc::clone(&registry);
+    let registry3 = Arc::clone(&registry);
+    let barrier1 = Arc::clone(&barrier);
+    let barrier2 = Arc::clone(&barrier);
+    let barrier3 = Arc::clone(&barrier);
+
+    let handle1 = thread::spawn(move || {
+        registry1
```
Thomasdezeeuw: Since these thread do mostly the same thing can you create another function for it to reduce the code duplication, some like:

```rust
/// Call all registration operations, ending with `source` being registered with `token` and `final_interests`.
fn registration_ops<S>(registry: &Registry, source: &mut io::Source, token: Token, init_interests: Interest, final_interests: Interest) -> io::Result<()> {
        registry.register(&mut source, token, init_interests).unwrap();
        registry.deregister(&mut source).unwrap();
        registry.register(&mut source, token, init_interests).unwrap();
        registry.reregister(&mut source, token, final_interests)
}
```

tests/poll.rs
dtacalau: ok, done

tests/poll.rs
dtacalau: ok, good idea, will use UdpSocket

tests/poll.rs
dtacalau: don't really need it, removed.

tests/poll.rs
dtacalau: don't really need it, removed.

tests/poll.rs
dtacalau: Thanks for the idea. Will use an extra barrier.wait call since it express better what I want to achieve here.

tests/poll.rs
dtacalau: great idea, will use it, thanks

tests/util/mod.rs
```diff
@@ -258,3 +258,18 @@ macro_rules! expect_read {
         assert_eq!(address, source);
     }};
 }
+
+/// Call all registration operations, ending with `source` being registered with `token` and `final_interests`.
+pub fn registry_ops_flow(
```
Thomasdezeeuw: Since its only used in `tests/poll.rs` it can live there.

tests/poll.rs
Thomasdezeeuw: Then you don't need this `init` line.

tests/poll.rs
```diff
@@ -201,6 +208,94 @@ fn registry_behind_arc() {
...
+
+    let barrier = Arc::new(Barrier::new(4));
+    let barrier1 = Arc::clone(&barrier);
+    let barrier2 = Arc::clone(&barrier);
+    let barrier3 = Arc::clone(&barrier);
+
+    let mut listener = TcpListener::bind(any_local_address()).unwrap();
+    let addr = listener.local_addr().unwrap();
+
+    let mut udp_socket = UdpSocket::bind(any_local_address()).unwrap();
```
Thomasdezeeuw: Small thing; this can be moved into the `thread::spawn`.

tests/poll.rs
```diff
@@ -201,6 +208,94 @@ fn registry_behind_arc() {
...
+        barrier3.wait();
+    });
+
+    // wait for threads to finish before expect_events
+    barrier.wait();
+    expect_events(
+        &mut poll,
+        &mut events,
+        vec![
+            ExpectEvent::new(ID1, Interest::READABLE),
```
Thomasdezeeuw: No writable for the `UdpSocket`?

tests/poll.rs
```diff
@@ -201,6 +208,94 @@ fn registry_behind_arc() {
...
+    let barrier1 = Arc::clone(&barrier);
+    let barrier2 = Arc::clone(&barrier);
+    let barrier3 = Arc::clone(&barrier);
+
+    let mut listener = TcpListener::bind(any_local_address()).unwrap();
+    let addr = listener.local_addr().unwrap();
+
+    let mut udp_socket = UdpSocket::bind(any_local_address()).unwrap();
+
+    /* Expect that multiple register/deregister/reregister work fine on multiple
```
Thomasdezeeuw: Small thing; can you can change this to use `// ` style comment. I don't see `/* ... */` to often any more.

tests/util/mod.rs
dtacalau: ok, moved.

tests/poll.rs
dtacalau: ok, moved.

tests/poll.rs
dtacalau: done, added writable instead of readable

tests/poll.rs
dtacalau: sure, old C habits here :)

tests/poll.rs
Thomasdezeeuw: I still don't see an `ExpectEvent` for `ID2`.

dtacalau: done, added, thanks

dtacalau:
Rebased on master after #1192 merge. @Thomasdezeeuw

tests/poll.rs
```diff
@@ -201,6 +209,107 @@ fn registry_behind_arc() {
...
+    // threads. Main thread will wait before the expect_events for all other 3
+    // threads to do their work. Otherwise expect_events timeout might be too short
+    // for all threads to complete, and call might fail.
+
+    let handle1 = thread::spawn(move || {
+        registry_ops_flow(
+            &registry1,
+            &mut listener,
+            ID1,
+            Interest::READABLE,
```
kleimkuhler: For extra testing maybe the initial interest can be `WRITABLE`, and then reregisters with `READABLE`?

dtacalau: good idea, changed.

dtacalau:
rebased on master.

tests/poll.rs
Thomasdezeeuw: `WRITABLE` doesn't make sense for `TcpListener` as there is no such event.

tests/poll.rs
```diff
@@ -201,6 +209,107 @@ fn registry_behind_arc() {
...
+    // threads to do their work. Otherwise expect_events timeout might be too short
+    // for all threads to complete, and call might fail.
+
+    let handle1 = thread::spawn(move || {
+        registry_ops_flow(
+            &registry1,
+            &mut listener,
+            ID1,
+            Interest::WRITABLE,
+            Interest::WRITABLE.add(Interest::READABLE),
```
Thomasdezeeuw: Can you remove the `WRITABLE` again, it shouldn't be use with `TcpListener`.

dtacalau: done

tests/poll.rs
dtacalau: removed

Thomasdezeeuw:
Thanks.

PR #1210, opened 2019-12-09
added specific tests for Events:: methods

dtacalau:


tests/events.rs
```diff
@@ -0,0 +1,35 @@
...
+
+#[test]
+fn events_all() {
+    let mut events = Events::with_capacity(16);
+    assert_eq!(events.capacity(), 16);
+    assert!(events.is_empty());
+
+    let mut poll = Poll::new().expect("unable to create new Poll instance");
+    let token = Token(10);
+    let waker = Waker::new(poll.registry(), token).expect("unable to create waker");
```
Thomasdezeeuw: Why not use `WAKE_TOKEN`?

tests/events.rs
```diff
@@ -0,0 +1,35 @@
...
+
+const WAKE_TOKEN: Token = Token(10);
+
+#[test]
+fn events_all() {
+    let mut events = Events::with_capacity(16);
+    assert_eq!(events.capacity(), 16);
+    assert!(events.is_empty());
+
+    let mut poll = Poll::new().expect("unable to create new Poll instance");
```
Thomasdezeeuw: Small nit; I don't think you need to use `expect`, `unwrap` will do.

tests/events.rs
```diff
@@ -0,0 +1,35 @@
...
+        .unwrap();
+
+    assert!(!events.is_empty());
+
+    for event in events.iter() {
+        assert_eq!(event.token(), token);
+        assert!(event.is_readable());
+    }
+
+    assert!(!events.is_empty());
```
Thomasdezeeuw: This block tests the same thing as the for loop above.

tests/events.rs
```diff
@@ -0,0 +1,35 @@
+use std::time::Duration;
+
+use mio::{Events, Poll, Token, Waker};
+
+const WAKE_TOKEN: Token = Token(10);
+
```
Thomasdezeeuw: Can you move the `event_source.rs` tests into this file? We're have too many tests file as it is and the file name is better.

tests/events.rs
dtacalau: ok, done

tests/events.rs
dtacalau: used init_with_poll.

tests/events.rs
dtacalau: removed

tests/events.rs
dtacalau: ok, done.

dtacalau:
> LGTM. Blocked on #1192.

Rebased on master after #1192 merge.

Issue #1189, opened 2019-12-05
Find source of possible polling delay


dtacalau:
This was initially reported here #1051 (comment).

The polling delay is still happening on latest Windows master:
![Capture](https://user-images.githubusercontent.com/2316254/70261776-d2aacc80-179b-11ea-8bde-0b37599e7ad9.JPG)

dtacalau:
@Thomasdezeeuw please assign this to me, thanks.

dtacalau:
The polling delay is exactly the time needed for GetQueuedCompletionStatusEx to return.

I've tested multiple scenarios in my attempt to diagnose the delay better. Here's what I've ruled out as causes of the polling delay:
1. Time needed to get the two threads started is close to 0 msec. 
2. Time for the listener to do the bind + register and the time for stream to connect + register are close to 0 msec.
3. Sync issues because of multiple active registers are not the cause of the delay.
The polling delay appears even if I do only one register: one of the listener  the stream. 

The polling delay does not appear in the following situations:
1. Doing the listener and the stream on the main thread.
2. Doing the listener and the stream on one thread, not the main thread.

So I'm inclined to think this delay has something to do with the internals of completion ports when handing events from multiple threads. Will talk to Mark Livs to see if I can get more insights.

dtacalau:
@MarkLivsc found the cause of the polling delay. These are his words describing the problem:
`what's happening is the connect thread attempts to connect to port 9999 before the listen thread has a chance to start listening. the packet is dropped (no one's listening). then ~500ms later it's retransmitted because there's been no response. by this point the listen thread has started bound + started listening, which causes the connect to go through right away
adding some synchronization between the connect and listen threads makes the test pass:
```
let handle1 = thread::spawn(move  {
            let mut listener = TcpListener::bind(addr).unwrap();
            listenbarrier2.wait();
[...]

        let handle2 = thread::spawn(move  {
            listenbarrier3.wait();
            let mut stream = TcpStream::connect(addr).unwrap();
```
`

dtacalau:
Closed, nothing to fix here, it was a test problem: connecting before listener.

PR #1212, opened 2019-12-10
Added a regression test for issue #1189 to verify the polling delay i…

dtacalau:
…s gone.


dtacalau:
@Thomasdezeeuw My goal was to have a test to check the poll timeout is consistent, no long delays out of nowhere. But, I think you're right, not needed because this test checks timeouts, it won't be reliable when run on CI anyway.

Issue #1191, opened 2019-12-06
mio should not panic when afd poll return error


dtacalau:
Currently when Afd.poll `NtDeviceIoControlFile` returns any error code other than `ERROR_IO_PENDING` and `ERROR_INVALID_HANDLE` mio just panics. I think it would be better to handle all error codes and pass the decision on how to handle the error to the mio user. 

For instance, let's say a user polls a TcpListener, when Afd.poll return any error code then the two codes above, do not panic, but save the error code in the listener SockState, also do not re-add the SockState to the update queue. TcpListener should provide the mio user a method to retrieve the last error. When getting the error user could choose to deregister the socket, or take other actions.

Thomasdezeeuw:
Could you point to the panic?

dtacalau:
> Could you point to the panic?

@Thomasdezeeuw https://github.com/tokio-rs/mio/blob/0a55a121f0a2646376e062a2ecbbdf3f03902598/src/sys/windows/selector.rs#L565

Thomasdezeeuw:
We should go through all the `unwrap` in the Windows implementation make sure the all make sense and see if we can handle them more gracefully.

dtacalau:
@Thomasdezeeuw good idea, I've opened #1198 to not forget about this, though I'd push this check after the custom polling implementation.

dtacalau:
@Thomasdezeeuw please assign this to me, thanks!

carllerche:
@piscisaureus Do you know what the error cases are?

carllerche:
According to @piscisaureus, there are no error cases that can be handled. We should include a panic w/ an error message and worry about it when we have a bug report.

Thomasdezeeuw:
@piscisaureus @carllerche  is it like `close`, where even though you can't handle the error, you shouldn't retry the operation because you don't know if the file descriptor is closed or not?

Thomasdezeeuw:
Punting this to v0.7 (not alpha 1).

Thomasdezeeuw:
Fixed by #1215.

PR #1215, opened 2019-12-11
A partial fix for issue #1191. Do not panic mio on sockstate error, i…

dtacalau:
…nstead keep the err sock states and just retry them later.

This is a partial fix for 0.7 because it just handles the errors, and makes the sockstates available next poll instead of panicking. Pushing the error to user is more complicated as it would require new api added. I'd suggest to leave this part until custom IOCP handlers are implemented and  #1195 is merged.

### DIFF
--- a/src/sys/windows/selector.rs
+++ b/src/sys/windows/selector.rs
@@ -104,12 +104,18 @@ pub struct SockState {
     poll_status: SockPollStatus,
     delete_pending: bool,
 
+    // last raw os error
+    error: Option<i32>,
+
     pinned: PhantomPinned,
 }
 
 impl SockState {
     fn update(&mut self, self_arc: &Pin<Arc<Mutex<SockState>>>) -> io::Result<()> {
         assert!(!self.delete_pending);
+
+        // make sure to reset previous error before a new update
+        self.error = None;
 
         if let SockPollStatus::Pending = self.poll_status {
             if (self.user_evts & afd::KNOWN_EVENTS & !self.pending_evts) == 0 {
@@ -122,7 +128,11 @@ impl SockState {
                  * events that the user is interested in. Therefore, cancel the pending
                  * poll operation; when we receive it's completion package, a new poll
                  * operation will be submitted with the correct event mask. */
-                self.cancel()?;
+                 if let Err(e) = self.cancel() {
+                    self.error = e.raw_os_error();
+                    return Err(e);
+                }
+                return Ok(());
             }
         } else if let SockPollStatus::Cancelled = self.poll_status {
             /* The poll operation has already been cancelled, we're still waiting for
@@ -156,6 +166,7 @@ impl SockState {
                         self.mark_delete();
                         return Ok(());
                     } else {
+                        self.error = e.raw_os_error();
                         return Err(e);
                     }
                 }
@@ -166,6 +177,7 @@ impl SockState {
         } else {
             unreachable!("Invalid poll status during update, {:#?}", self)
         }
+
         Ok(())
     }
 
@@ -246,6 +258,10 @@ impl SockState {
             self.delete_pending = true;
         }
     }
+
+    fn has_error(&self) -> bool {
+        self.error.is_some()
+    }
 }
 
 cfg_net! {
@@ -262,6 +278,7 @@ cfg_net! {
                 user_data: 0,
                 poll_status: SockPollStatus::Idle,
                 delete_pending: false,
+                error: None,
                 pinned: PhantomPinned,
             })
         }
@@ -455,16 +472,18 @@ impl SelectorInner {
 
     unsafe fn update_sockets_events(&self) -> io::Result<()> {
         let mut update_queue = self.update_queue.lock().unwrap();
-        loop {
-            let sock = match update_queue.pop_front() {
-                Some(sock) => sock,
-                None => break,
-            };
+        for sock in update_queue.iter_mut() {
             let mut sock_internal = sock.lock().unwrap();
             if !sock_internal.is_pending_deletion() {
-                sock_internal.update(&sock).unwrap();
+                let _ = sock_internal.update(&sock);
             }
         }
+
+        // remove all sock which do not have error, they have afd op pending
+        update_queue.retain(|sock| {
+            sock.lock().unwrap().has_error()
+        });
+
         self.afd_group.release_unused_afd();
         Ok(())
     }
@@ -561,6 +580,8 @@ cfg_net! {
                 state.lock().unwrap().set_event(event);
             }
 
+            // FIXME: a sock which has_error true should not be re-added to
+            // the update queue because it's already there.
             self.queue_state(state);
             unsafe { self.update_sockets_events_if_polling() }
         }


src/sys/windows/selector.rs
```diff
@@ -122,7 +127,12 @@ impl SockState {
                  * events that the user is interested in. Therefore, cancel the pending
                  * poll operation; when we receive it's completion package, a new poll
                  * operation will be submitted with the correct event mask. */
-                self.cancel()?;
+                let result = self.cancel();
```
Thomasdezeeuw: I haven't been able to find any documentation on `NtCancelIoFileEx` (the system call made here), do you have a documentation @dtacalau?

I want to know what kind of errors can be returned here because maybe we can ignore some of them.

dtacalau: @Thomasdezeeuw not much, seems it can return any NTSTATUS. The only I was able to found: 
- some of these unit tests check result of `NtCancelIoFileEx` https://git.reactos.org/?p=reactos.git;a=blob;hb=6d0c07c44fae0a93c57f8d3410545e0f3c201287;f=modules/rostests/winetests/ntdll/pipe.c#l384
- Win list of NTSTATUS codes https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55

Thomasdezeeuw: Always great when you're developing against an API for which you don't have docs, right?

dtacalau: rhetorical question I guess :)

kleimkuhler: Let's just handle the result directly with:

```
if let Err(e) = self.cancel() {
    self.error = e.raw_os_error();
    return Err(e);
}
```

If errors are able to be ignored (after we stumble across *some* documentation 😖) we can just make it a `match`.

src/sys/windows/selector.rs
```diff
@@ -464,16 +481,28 @@ impl SelectorInner {
 
     unsafe fn update_sockets_events(&self) -> io::Result<()> {
         let mut update_queue = self.update_queue.lock().unwrap();
-        loop {
-            let sock = match update_queue.pop_front() {
-                Some(sock) => sock,
-                None => break,
-            };
+        let mut update_err: bool = false;
+
```
kleimkuhler: Extra line snuck in

src/sys/windows/selector.rs
```diff
@@ -464,16 +481,28 @@ impl SelectorInner {
...
+        for sock in update_queue.iter_mut() {
```
kleimkuhler: Can just use an `unwrap_or_else` here since we don't use the `Ok` value:

```
for sock in update_queue.iter_mut() {
    let mut sock_internal = sock.lock().unwrap();
    if !sock_internal.is_pending_deletion() {
        sock_internal
            .update(&sock)
            .unwrap_or_else(|_| update_err = true);
    }
}
```

src/sys/windows/selector.rs
```diff
@@ -104,13 +104,18 @@ pub struct SockState {
     poll_status: SockPollStatus,
     delete_pending: bool,
 
+    error: Option<i32>, //last raw os error
```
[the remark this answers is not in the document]
dtacalau: ok, moved above.

src/sys/windows/selector.rs
dtacalau: ok

src/sys/windows/selector.rs
dtacalau: removed, btw I like many extra lines :).

src/sys/windows/selector.rs
dtacalau: ok, thanks

src/sys/windows/selector.rs
```diff
@@ -104,13 +104,19 @@ pub struct SockState {
     poll_status: SockPollStatus,
     delete_pending: bool,
 
+    //last raw os error
```
kleimkuhler: Missing space

src/sys/windows/selector.rs
```diff
@@ -246,6 +258,10 @@ impl SockState {
             self.delete_pending = true;
         }
     }
+
+    pub fn has_error(&self) -> bool {
```
kleimkuhler: Doesn't need to be pub

src/sys/windows/selector.rs
dtacalau: fixed

src/sys/windows/selector.rs
dtacalau: ok, removed pub

dtacalau:
I've done a change in the logic of `update_sockets_events`. The following block should be done no matter if a sock update failed or not, because if we do it only on error then we end up with duplicates in the update_queue:
```
        update_queue.retain(|sock| {
            sock.lock().unwrap().has_error()
        });
```

I've also found another issue we need to address: `reregister done on a sock which has_error will result in having the same socked added twice in the update queue`. Will not result in a bug, just that we'll call update twice, 2nd call being useless. To fix this `self.add_socket_to_update_queue(socket);` has to be called only if `sock.has_error is false`. 
@Thomasdezeeuw any idea how can I achieve this? I got into mutability issues, cannot move out of a lock, so I cannot check for has_error and add it to the queue at the while sock.lock is being held.

kleimkuhler:
@dtacalau I still need to familiarize more with the Windows impl, but could `set_event` do the check and return `false` if `SockState.error` is `true`? We hold the lock right before going into that block of code so it could just return early

kleimkuhler:
@dtacalau Also I would appreciate it for changes in general if you can not force push commits 🙂. It makes it difficult to review PRs. I like to be able to leave a review, and then when a review is re-requested I can see what *changed* since the last time.

When a force push happens, the "viewed already" cache resets and I don't have a way to tell what is new and what I've reviewed.

I wouldn't worry about the git log either because commits should be squashed in a PR.

dtacalau:
> @dtacalau I still need to familiarize more with the Windows impl, but could `set_event` do the check and return `false` if `SockState.error` is `true`? We hold the lock right before going into that block of code so it could just return early

@kleimkuhler Yeah that would work, By changing `set_event` to return `has_error` we avoid double locking. Another way would be to check `has_error` under lock, like this `sock.lock().unwrap().has_error()`:
```
if (sock.lock().unwrap().set_event(event)) {
}
self.update_sockets_events_if_polling()?;
```

dtacalau:
> @dtacalau Also I would appreciate it for changes in general if you can not force push commits . 

Ok, got it, sorry for that.

kleimkuhler:
@dtacalau Yea so we still want to call `socket.set_sock_state(Some(sock));` after `set_event` it looks like. In that case, `set_event` could return `false` if `error.is_some()`, and then we only update based on that result. So something like:

```
let should_update = sock.lock().unwrap().set_event(event);
if should_update {
    unsafe {
        self.update_sockets_events_if_polling()?;
    }
}
```

dtacalau:
rebased on master

@kleimkuhler I'm not quite happy with the solution from https://github.com/tokio-rs/mio/pull/1215#issuecomment-566166194. It works but it feels like a hack `conveniently modifying the return of set_event for this fix`. What if we need it to return something else for another issue we might find :)? 

I'm going to leave it as it is until I figure something cleaner. WDYT?

Thomasdezeeuw:
Thanks @dtacalau, sorry for the late review.

PR #1208, opened 2019-12-09
Multiple doc notes

dtacalau:
reregister and deregister are thread-safe
undefined behavior for reregister and deregister w/o prev register
undefined behavior to register a source with another poll instance after deregister

Fixes #1202 and #1203.

PR #1219, opened 2019-12-12
Add more tests poll registry (short from PR 1201)

dtacalau:
This is a spin from #1201 contains tests / changes which do not verify the registry undefined behavior. These changes have already been review in #1201 .

### DIFF
--- a/tests/poll.rs
+++ b/tests/poll.rs
@@ -12,7 +12,8 @@ use std::{fmt, io};
 mod util;
 
 use util::{
-    any_local_address, assert_send, assert_sync, expect_events, init, init_with_poll, ExpectEvent,
+    any_local_address, assert_send, assert_sync, expect_events, expect_no_events, init,
+    init_with_poll, start_listener, ExpectEvent,
 };
 
 const ID1: Token = Token(1);
@@ -310,10 +311,70 @@ fn registry_operations_are_thread_safe() {
     handle3.join().unwrap();
 }
 
-// On kqueue platforms registering twice (not *re*registering) works.
+// This test checks the following reregister constraints:
+// The `reregister` arguments fully override the previous values. In other
+// words, if a socket is registered with [`readable`] interest and the call
+// to `reregister` specifies [`writable`], then read interest is no longer
+// requested for the handle.
+#[test]
+fn reregister_overrides_interest() {
+    let (mut poll, mut events) = init_with_poll();
+
+    let barrier = Arc::new(Barrier::new(2));
+    let (thread_handle, address) = start_listener(1, Some(barrier.clone()), false);
+
+    let mut stream = TcpStream::connect(address).unwrap();
+
+    poll.registry()
+        .register(&mut stream, ID1, Interest::WRITABLE)
+        .expect("unable to register TCP stream");
+
+    poll.registry()
+        .reregister(&mut stream, ID2, Interest::READABLE)
+        .expect("unable to register TCP stream");
+
+    expect_no_events(&mut poll, &mut events);
+
+    barrier.wait();
+    thread_handle.join().expect("unable to join thread");
+}
+
+// This test checks the following constraints:
+// - reregister can use the same token as register
+// - reregister can use different token from register
+// - multiple reregister are ok
+#[test]
+fn register_reregister_token_usage() {
+    let poll = Poll::new().expect("unable to create Poll instance");
+
+    let mut listener1 = TcpListener::bind(any_local_address()).unwrap();
+
+    poll.registry()
+        .register(&mut listener1, ID1, Interest::READABLE)
+        .expect("unable to register listener");
+
+    poll.registry()
+        .reregister(&mut listener1, ID1, Interest::READABLE)
+        .expect("unable to register listener");
+
+    poll.registry()
+        .reregister(&mut listener1, ID2, Interest::READABLE)
+        .expect("unable to register listener");
+}
+
+// This test checks the following register constraint:
+// The event source must **not** have been previously registered with this
+// instance of `Poll`, otherwise the behavior is undefined.
+//
+// This test is done on Windows only because Windows is the only platform where
+// there is a defined behavior by design, that is: registering twice must
+// fail with an error code.
+//
+// On kqueue platforms registering twice (not *re*registering) works, but that
+// is not a test goal, so it is not tested.
 #[test]
 #[cfg(any(target_os = "linux", target_os = "windows"))]
-pub fn double_register() {
+pub fn double_register_different_token() {
     init();
     let poll = Poll::new().unwrap();
 
--- a/tests/poll.rs
+++ b/tests/poll.rs
@@ -12,8 +12,7 @@ use std::{fmt, io};
 mod util;
 
 use util::{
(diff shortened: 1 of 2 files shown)


src/poll.rs
```diff
@@ -411,6 +411,9 @@ impl Registry {
     /// The readiness interest for an `event::Source` can be changed at any time
     /// by calling [`reregister`].
     ///
+    /// The event source must **not** have been previously registered with this
```
kleimkuhler: I'm not sure this comment should be included in the PR. I welcome new tests, but I think this comment can add some confusion to `Poll` as it currently reads.

Below in `# Notes`, the caution of registering sources and what that means for the lifetime of the source is already conveyed.

dtacalau: Could you point me to those `#Notes`?

tests/poll.rs
```diff
@@ -201,10 +208,70 @@ fn registry_behind_arc() {
...
+    poll.registry()
+        .reregister(&mut listener1, ID2, Interest::READABLE)
+        .expect("unable to register listener");
+}
+
+// This test checks the following register constraint:
+// The event source must **not** have been previously registered with this
+// instance of `Poll`, otherwise the behavior is undefined.
+//
+// This test is done on Windows only because Windows is the only platform where
```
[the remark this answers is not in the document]
dtacalau: In which way you think `Linux` should be mentioned? If we consider mentioning `Linux` then we should mention it for all 3 cases: double register(this one), reregister w/o register and deregister w/o register. The behavior seem to be the same for all 3 of them. Thoughts?

tests/poll.rs
```diff
@@ -201,10 +208,70 @@ fn registry_behind_arc() {
     handle2.join().unwrap();
 }
 
-// On kqueue platforms registering twice (not *re*registering) works.
+// This test checks the following reregister constraints:
+// The `reregister` arguments fully override the previous values. In other
+// words, if a socket is registered with [`readable`] interest and the call
```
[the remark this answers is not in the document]
dtacalau: This is a copy paste from the `poll.rs`. You think I should do this change everywhere?

dtacalau:
> Thanks @dtacalau for splitting this out. I left a few comments, but I'm cautious about adding this > > additional comment to `Poll`.

@kleimkuhler My goal with adding this comment to `Poll` was to prevent double register. Could be reworded, suggestions :)?

tests/poll.rs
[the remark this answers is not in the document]
kleimkuhler: No I don't think we need to do that as part of this PR, but at least a comment being added can switch to that.

src/poll.rs
kleimkuhler: Yep just below a few lines [here](https://github.com/tokio-rs/mio/pull/1219/files#diff-d7937c2d56ffbacce8c1a6acba567e1cR417)!

tests/poll.rs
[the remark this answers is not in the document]
kleimkuhler: Similar to the above, I think for a comment being *added* we can just make the change here.

Something like: "This test is done only on Windows and epoll platforms because registering a source twice is defined behavior that fails with an error".

kleimkuhler:
> @kleimkuhler My goal with adding this comment to `Poll` was to prevent double register. Could be reworded, suggestions :)?

Thanks that's a little more clear. I think in this case since we are defining behavior that errors *if* conditions are met, we can add an `# Error` section that says that. Similar to how errors are defined in Tokio comments, something like:

```
/// ..
/// The readiness interest for an `event::Source` can be changed at any time
/// by calling [`reregister`].
///
/// # Errors
///
/// [`IOCP`] and [`epoll`] platforms will error if the I/O source being
/// registered was previously registered with this `Poll` instance without
/// first being deregistered. The caller must ensure this is properly
/// handled.
///
/// [`kqueue`] platforms will allow consecutive registers of the same I/O
/// source to `Poll` instance.
///
/// # Notes
///
/// ..
```

For the kqueue section, we can continue saying something like:

> On [`kqueue`] platforms this behavior is undefined.

However, [kqueue](https://www.freebsd.org/cgi/man.cgi?query=kqueue&sektion=2) `KV_ADD` does say this is okay. @Thomasdezeeuw may have some input on why we haven't defined it before, but I think we're okay to at least define it for IOCP and epoll.

tests/poll.rs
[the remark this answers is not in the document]
dtacalau: ok, done

tests/poll.rs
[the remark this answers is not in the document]
dtacalau: ok, got it

dtacalau:
@kleimkuhler I'd not expand on `undefined behavior`.  While reviewing this https://github.com/tokio-rs/mio/pull/1208 Carl said not to expand on what `undefined behavior` means because it might get complicated to explore all cases on all platforms. Best to put it simple, `this is undefined behavior, do not use it`. On another note if we expand it and explain it maybe `some users` would be tempted to use it because it works on platform x. I think it's ok to document it in tests, and test properly what can be tested, i.e Windows.

@Thomasdezeeuw what do you think about adding the doc error section as Kevin suggested here https://github.com/tokio-rs/mio/pull/1219#issuecomment-565199346 ?

dtacalau:
rebased on master

tests/poll.rs
```diff
@@ -310,10 +311,69 @@ fn registry_operations_are_thread_safe() {
...
+// to `reregister` specifies `WRITABLE`, then read interest is no longer
+// requested for the handle.
+#[test]
+fn reregister_overrides_interest() {
+    let (mut poll, mut events) = init_with_poll();
+
+    let barrier = Arc::new(Barrier::new(2));
+    let (thread_handle, address) = start_listener(1, Some(barrier.clone()), false);
+
+    let mut stream = TcpStream::connect(address).unwrap();
```
Thomasdezeeuw: Why not use an `UdpSocket`, then we don't have to start a `TcpListener`.

tests/poll.rs
```diff
@@ -310,10 +311,69 @@ fn registry_operations_are_thread_safe() {
...
 }
 
-// On kqueue platforms registering twice (not *re*registering) works.
+// This test checks the following reregister constraints:
+// The `reregister` arguments fully override the previous values. In other
+// words, if a socket is registered with `READABLE` interest and the call
+// to `reregister` specifies `WRITABLE`, then read interest is no longer
+// requested for the handle.
+#[test]
+fn reregister_overrides_interest() {
```
Thomasdezeeuw: I think the two tests can be merged; changing both the `Token` and `Interest` in a single call to reregister. If you can from `ID1`, `READABLE` to `ID2`, `WRITABLE` you can `expect_event(ID2, WRITABLE)` after using a `UdpSocket`.
Then you don't need to both with starting the listener and the barrier.

tests/poll.rs
dtacalau: yup, good idea, dropped the listener

tests/poll.rs
dtacalau: ok, I've merged the tests, thanks

dtacalau:
@kleimkuhler I've dropped the commit with the undefined behavior notes from this PR. Will open another PR for that.

PR #1220, opened 2019-12-12
added a test to check register ops done during a pending poll.

dtacalau:

### DIFF
--- a/tests/poll.rs
+++ b/tests/poll.rs
@@ -315,7 +315,6 @@ fn register_during_poll() {
     let (mut poll, mut events) = init_with_poll();
 
     let registry = Arc::new(poll.registry().try_clone().unwrap());
-    let registry1 = registry.clone();
 
     let barrier = Arc::new(Barrier::new(2));
     let barrier1 = Arc::clone(&barrier);
@@ -324,7 +323,7 @@ fn register_during_poll() {
         let mut stream = UdpSocket::bind(any_local_address()).unwrap();
 
         barrier1.wait();
-        registry1
+        registry
             .register(&mut stream, ID1, Interest::WRITABLE)
             .unwrap();
 
--- a/tests/poll.rs
+++ b/tests/poll.rs
@@ -320,52 +320,29 @@ fn register_during_poll() {
     let barrier = Arc::new(Barrier::new(2));
     let barrier1 = Arc::clone(&barrier);
 
-    let mut listener = TcpListener::bind(any_local_address()).unwrap();
-    let addr = listener.local_addr().unwrap();
-    registry
-        .register(&mut listener, ID1, Interest::READABLE)
-        .unwrap();
-
     let handle1 = thread::spawn(move || {
-        let mut listener2 = TcpListener::bind(any_local_address()).unwrap();
-        let addr2 = listener2.local_addr().unwrap();
+        let mut stream = UdpSocket::bind(any_local_address()).unwrap();
 
-        // Main thread should be blocked in poll before entering this register.
-        // This is done to check register done during poll works
-        registry1
-            .register(&mut listener2, ID3, Interest::READABLE)
-            .unwrap(); // this should go through update_if_polling
-
-        let mut stream = TcpStream::connect(addr).unwrap();
         barrier1.wait();
-
         registry1
-            .register(&mut stream, ID2, Interest::READABLE | Interest::WRITABLE)
+            .register(&mut stream, ID1, Interest::WRITABLE)
             .unwrap();
 
-        // this is done to check we get an event for a register done during poll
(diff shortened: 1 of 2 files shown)


kleimkuhler:
I like the idea of this test, but I'm not sure this is guaranteed to test what it's being added for. It is pretty race-y with how it is currently implemented, and it's not guaranteed that when `listener2` is registered, the main thread is blocking in a `Poll:poll`. `expect_events` does multiple iterations of `poll`, and the barriers in their current placement don't necessarily block things where they need to be.

To start, it could be helpful to directly `poll.poll` once, and still kind of *hope* that when the first thread is blocked in that call the second thread tries to register, but again that is not guaranteed.

The guaranteed fix would somehow pass in a `Barrier` to `Poll::poll` and have it block in there, but that is **way** more of a change that I don't think we need to do.

So this isn't really a helpful comment, but maybe if you look again at where the wait points are and try to sync it up a little more?

Thomasdezeeuw:
I agree with @kleimkuhler. To make it a little less race-y, you could call `barrier.wait()` just before a direct call to `register` and `Poll::poll` (with a long-ish timeout, e.g. 5 seconds). It doesn't guarantee that calls will be concurrent, but on local machines with enough cores available it will likely result in errors if the call is not made, hopefully failing the test.

dtacalau:
@Thomasdezeeuw @kleimkuhler I agree with both of you. Try to see this test as a best effort test which should work most of the time.  I could redo it with poll(None) w/o expect_events. Second opinion?

Long story:
Initially I've written it using Poll.poll with timeouts, but then I remembered what Carl said about timeouts and CI, and I've redone it w/o timeouts using Poll.poll(None), but got failures from time to time. Next I've reconsidered, again, and reimplemented it with expect_events. I thought expect_events is best because it is used in lots of tests. 

I've run this test with prints many times and most of the times register was done during the pending poll.  So, to wrap it up, I would not add timeouts because of CI limitations, if current expect_events timeouts, 500ms, are not enough, than 5sec would not cut it most of the times too. Having barrier inside poll would be best, but can't be done atm.

Thomasdezeeuw:
@dtacalau I created https://github.com/dtacalau/mio/pull/1 to simplify this pr.

dtacalau:
> @dtacalau I created [dtacalau#1](https://github.com/dtacalau/mio/pull/1) to simplify this pr.

@Thomasdezeeuw While your test is simpler it doesnt fix the race problems. Both thread will race after the first barrier.wait call. Let me think about it.

Thomasdezeeuw:
> @Thomasdezeeuw While your test is simpler it doesnt fix the race problems. Both thread will race after the first barrier.wait call. Let me think about it.

Isn't that what this test was testing? Or have I misunderstood the purpose of the test?

dtacalau:
> > @Thomasdezeeuw While your test is simpler it doesnt fix the race problems. Both thread will race after the first barrier.wait call. Let me think about it.
> 
> Isn't that what this test was testing? Or have I misunderstood the purpose of the test?

@Thomasdezeeuw What I meant was my test and your test suffer from the same `possible` race condition between poll and register, but, nevermind, I've accepted you PR. 

I've also rebased on master.

dtacalau:
I've opened #1226 to track the CI hanging test

tests/poll.rs
```diff
@@ -310,6 +310,43 @@ fn registry_operations_are_thread_safe() {
...
+
+    let registry = Arc::new(poll.registry().try_clone().unwrap());
+
+    let barrier = Arc::new(Barrier::new(2));
+    let barrier1 = Arc::clone(&barrier);
+
+    let handle1 = thread::spawn(move || {
+        let mut stream = UdpSocket::bind(any_local_address()).unwrap();
+
+        barrier1.wait();
```
kleimkuhler: Thoughts on adding some millisecond sleep here? That way both threads unblock at the barrier. The poll thread immediately goes into a 5 second poll, and this thread waits just a little bit so we can get closer to "trying" to register during a poll.

dtacalau:
> LGTM, @kleimkuhler suggestion about a small sleep before registering would make this even better.

I've adedd the timeout before register.

dtacalau:
@Thomasdezeeuw I've reverted your commit https://github.com/tokio-rs/mio/pull/1220/commits/c20132df7e65f0f95125a3786a78fd21ebc6fd6f which was causing all Linux CI to fail

dtacalau:
@kleimkuhler , @Thomasdezeeuw I think this can be merged.

kleimkuhler:
@dtacalau And thanks for taking care of all the reviews on this! I think this test does a good job trying to trigger the test purpose.

PR #1221, opened 2019-12-12
Add tcp feature flag to events test

kleimkuhler:
#1210 introduced a TCP dependency to the `event` tests. Without a module cfg, it
warns about the use of `mio::net`.


Thomasdezeeuw:
Thanks.

PR #1222, opened 2019-12-12
Added a test to verify cancelling pending poll requests works as expe…

dtacalau:
…cted.

Test will register an event source, submit a poll request, reregister the
event source to cancel it, expect to get the same event as if the initial
pending poll request was not cancelled.

### DIFF
--- a/tests/poll.rs
+++ b/tests/poll.rs
@@ -9,6 +9,7 @@ use std::thread::{self, sleep};
 use std::time::Duration;
 use std::{fmt, io};
 
+#[macro_use]
 mod util;
 
 use util::{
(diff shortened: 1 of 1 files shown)


tests/poll.rs
```diff
@@ -310,6 +311,67 @@ fn registry_operations_are_thread_safe() {
...
+        .register(&mut listener, ID1, Interest::READABLE)
+        .unwrap();
+
+    // Call a dummy poll just to submit an afd poll request
+    poll.poll(&mut events, Some(Duration::from_millis(0)))
+        .unwrap();
+
+    // this reregister will cancel the previous pending poll op
+    registry
+        .reregister(&mut listener, ID1, Interest::READABLE | Interest::WRITABLE)
```
Thomasdezeeuw: `WRITABLE` doesn't make sense.

Thomasdezeeuw: You might want to use `ID2`/`ID3` so you can ensure below that the proper event got returned in `expect_events` below.

dtacalau: I know, I just use it to be have something different in the `reregister` call. While not having a practical usage, it helps during debugging to have different flags set on reregister. I'd rather keep it than remove it.

dtacalau: Good idea, will change it.

dtacalau: I added a comment with its purpose, but if you have a strong opinion against it I'll remove it.

Thomasdezeeuw: I often look at tests of a crate as example usage. So if we have invalid code in our tests we set a bad precedent that's why I like to avoid it.

dtacalau: good point, will remove writable

tests/poll.rs
```diff
@@ -9,6 +9,7 @@ use std::thread::{self, sleep};
 use std::time::Duration;
 use std::{fmt, io};
 
+#[macro_use]
```
Thomasdezeeuw: Not needed, we don't use any macros from the `util` module.

tests/poll.rs
```diff
@@ -310,6 +311,69 @@ fn registry_operations_are_thread_safe() {
...
+        .register(&mut listener, ID1, Interest::READABLE)
+        .unwrap();
+
+    // Call a dummy poll just to submit an afd poll request
+    poll.poll(&mut events, Some(Duration::from_millis(0)))
+        .unwrap();
+
+    // This reregister will cancel the previous pending poll op.
+    // The token is different from the register done above, so it can ensure
+    // the proper event got returned exoect_events below.
```
Thomasdezeeuw: Type-o: exoect_events -> expect_events

tests/poll.rs
dtacalau: ok, will remove

tests/poll.rs
dtacalau: fixed

dtacalau:
Rebased on master.

dtacalau:
@kleimkuhler , @Thomasdezeeuw I think this can be merged.

PR #1224, opened 2019-12-13
Remove intra-doc links

kleimkuhler:
Any documentation links that were `crate` or path dependent have been updated
to use the relative paths of the generated Rustdoc output.


src/poll.rs
```diff
@@ -413,17 +408,22 @@ impl Registry {
     ///
     /// # Notes
     ///
+    /// Callers must ensure that if a source being registered with a `Poll`
```
Thomasdezeeuw: 👍

src/poll.rs
```diff
@@ -184,16 +184,11 @@ use std::{fmt, io};
 /// data to be copied into an intermediate buffer before it is passed to the
 /// kernel.
 ///
-/// Notifications generated by [`SetReadiness`] are handled by an internal
```
Thomasdezeeuw: 👍

hawkw:
Hmm, my understanding is that intra-doc links should be _preferred_ since relative links can sometimes break due to path differences between docs.rs and local docs — @carllerche may remember the specifics?

src/lib.rs
```diff
@@ -17,8 +17,8 @@
 //!
 //! For more detail, see [`Poll`].
 //!
-//! [`Poll`]: struct.Poll.html
-//! [`Events`]: struct.Events.html
+//! [`Poll`]: ../mio/struct.Poll.html
```
Thomasdezeeuw: `../mio` can be replaced with `./`.

src/lib.rs
```diff
@@ -30,7 +30,7 @@
 //! ## Guide
 //!
 //! A getting started guide is available in the
-#![cfg_attr(feature = "guide", doc = "[`guide`](crate::guide) module.")]
+#![cfg_attr(feature = "guide", doc = "[`guide`](../mio/guide/index.html) module.")]
```
Thomasdezeeuw: `../mio` can be replaced with `./`.

Thomasdezeeuw:
@kleimkuhler thanks for doing this, last issue for the alpha1 milestone.

kleimkuhler:
@hawkw Intra-doc links are a nightly feature. `cargo doc --open` on master just had broken links in all those places. Only if a user knew to use `cargo +nightly doc --open` did they work properly.

kleimkuhler:
The same issues exists on Tokio master as well right now. If docs are not compiled locally with nightly then most of the links do not work 😕

Thomasdezeeuw:
And to add in the fun the nightly compiler doesn't give any warnings that you're using a nightly only feature.

PR #1229, opened 2019-12-18
Format all CI files

kleimkuhler:
## Motivation

The last few times I've changed these files for testing CI in draft PRs or for
upstream changes, I've had to turn off formatting so that my change didn't
include these. This reformats all the `.yaml` files related to CI so that in the
future this is not the extremely mild incovenience that it currently is :P


Thomasdezeeuw:
Thanks.

Issue #1226, opened 2019-12-16
regression test issue_1205 hangs on all Linux CI


dtacalau:
https://dev.azure.com/tokio-rs/Tokio/_build/results?buildId=4073&view=logs&j=95c0de06-86ba-5b83-6acc-e8fdf9d57564&t=804b0c6d-d844-5f90-18fc-e53e483bd33c

![Screenshot_1](https://user-images.githubusercontent.com/2316254/70918515-d1915f00-2027-11ea-87af-bc8c64d68216.jpg)

Thomasdezeeuw:
So the only log is that its running for longer then 60 seconds?

dtacalau:
> So the only log is that its running for longer then 60 seconds?

@Thomasdezeeuw Yes, and the fact that it hanged on all Linux CIs during the same run.

Thomasdezeeuw:
See https://github.com/tokio-rs/mio/pull/1231#discussion_r359766938, my change was invalid.

PR #1230, opened 2019-12-18
Always use a timeout calling Poll::poll in tests

Thomasdezeeuw:
This is to prevent #1226 from happening too some extend, see the docs of `NO_TIMEOUT` in `tests/util/mod.rs`.

### DIFF
--- a/tests/close_on_drop.rs
+++ b/tests/close_on_drop.rs
@@ -1,12 +1,13 @@
 #![cfg(feature = "tcp")]
 
+use self::TestState::{AfterRead, Initial};
 use log::debug;
 use mio::net::{TcpListener, TcpStream};
 use mio::{Events, Interest, Poll, Token};
 use std::io::Read;
+
 mod util;
-use self::TestState::{AfterRead, Initial};
-use util::{any_local_address, init};
+use util::{any_local_address, init, NO_TIMEOUT};
 
 const SERVER: Token = Token(0);
 const CLIENT: Token = Token(1);
@@ -112,7 +113,8 @@ pub fn close_on_drop() {
 
     // == Run test
     while !handler.shutdown {
-        poll.poll(&mut events, None).unwrap();
+        poll.poll(&mut events, NO_TIMEOUT).unwrap();
+        assert!(!events.is_empty(), "expected at least one event");
 
         for event in &events {
             if event.is_readable() {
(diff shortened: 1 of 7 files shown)


dtacalau:
While I understand the problem this patch is trying to fix, I think the patch has its own problems:
1. poll(None) vs poll(timeout) exercise different code paths. Removing all poll(None) from tests means its code paths are no longer tested.
2. Having an assert(Poll(timeout)) failing doesn't seem more data than knowing a Poll(None) hangs.

Maybe the CI can be configured to terminate a test earlier than 1 hour?

kleimkuhler:
@dtacalau Brings up a good point about testing the `None` case for `Poll::poll`. I'd like to see this change go in, but maybe add a simple test that still uses `None` and expect an event event that should reliably happen? It wouldn't need need to be anything that tests additional behavior.

Thomasdezeeuw:
I agree with @dtacalau and been looking into other solutions. I found the `--ensure-time` flag which:

> Treat excess of the test execution time limit as
                        error.
                        Threshold values for this option can be configured via
                        `RUST_TEST_TIME_UNIT`, `RUST_TEST_TIME_INTEGRATION`
                        and
                        `RUST_TEST_TIME_DOCTEST` environment variables.
                        Expected format of environment variable is
                        `VARIABLE=WARN_TIME,CRITICAL_TIME`.
                        `CRITICAL_TIME` here means the limit that should not
                        be exceeded by test.

See `cargo test -- --help`. But its a nightly compiler only thing, see https://github.com/rust-lang/rust/pull/64873 and https://github.com/rust-lang/rust/issues/64888. I haven't been able to get it to actually run with the flag set.

Thomasdezeeuw:
Oh you need to pass `-Zunstable-options`. But still will only work on nightly only compilers.

Thomasdezeeuw:
`--ensure-time` won't work at all as the test isn't interrupted, so we do need a change like this.

Or we could just use the [`timeout`](http://man7.org/linux/man-pages/man1/timeout.1.html) command.

kleimkuhler:
@Thomasdezeeuw Unfortunately it doesn't look like there is a great alternative for that on Windows. It looks like it some kind of equivalent could be used if CI was setup to run through Powerhsell?

Thomasdezeeuw:
Not doing it this way and I haven't found an alternative to `timeout` on Windows/macOS.

Thomasdezeeuw:
Something the following could work in the future.

```
RUST_TEST_TIME_UNIT="5000,5000" RUST_TEST_TIME_INTEGRATION="5000,5000" RUST_TEST_TIME_DOCTEST="5000,5000" cargo test --all-features -- -Z unstable-options --ensure-time
```

PR #1231, opened 2019-12-18
Fix a Clippy warning

Thomasdezeeuw:


dtacalau:
This change was also causing #1220 CI to fail. I've reverted it from #1220 and it passed.

kleimkuhler:
The test no longer needs to wrap `waker` in an `Arc` with this change.

@dtacalau I'm starting to think this test may just be flaky and that adding back the `.clone()` just happened to hit trigger a successful run..

tests/regressions.rs
```diff
@@ -60,7 +60,6 @@ fn issue_1205() {
     let (mut poll, mut events) = init_with_poll();
 
     let waker = Arc::new(Waker::new(poll.registry(), WAKE_TOKEN).unwrap());
```
kleimkuhler: `waker` doesn't need to be wrapped in an `Arc` anymore.

Thomasdezeeuw: I know but that was part of the regression test so I just left it.

tests/regressions.rs
```diff
@@ -79,7 +78,7 @@ fn issue_1205() {
     // spawn a waker thread to wake the poll call below
     let handle = thread::spawn(move || {
         thread::sleep(Duration::from_millis(500));
-        waker1.wake().expect("unable to wake");
+        waker.wake().expect("unable to wake");
```
Thomasdezeeuw: I just realised this change drops the `waker` here and than `Poll` will never get a wakup event (see the `Waker` docs). So this change is invalid.

I'll send a change to ignore the Clippy warning.

PR #1232, opened 2019-12-19
Draft: Check against feature combinations

kleimkuhler:
## Draft

This may add a significant amount of time to CI, so this is starting out as a
draft to see what the change is. If it is too long, I'm going to just manually
add the combinations that should be tested.

The main difference between the feature powerset and what I would manually add
is the combinations will not include `guide`, reducing the amount of checks by
half.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,6 +21,13 @@ jobs:
       displayName: Test --release
       cmd: test --release
 
+  # Stable feature powerset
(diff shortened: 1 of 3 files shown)


kleimkuhler:
Closing in favor of #1244. Seems unnecessary to add a new cargo subcommand install. @Thomasdezeeuw recommended just sticking with a script which makes more sense for this.

PR #1234, opened 2019-12-22
Add `UdpSocket::peer_addr`

### LINKED ISSUE #977: Support for `peer_addr` for `UdpSocket`

messense:
Closes #977

### DIFF
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -166,6 +166,28 @@ impl UdpSocket {
         self.sys.local_addr()
     }
 
+    /// Returns the socket address of the remote peer this socket was connected to.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// # use std::error::Error;
+    /// #
+    /// # fn main() -> Result<(), Box<dyn Error>> {
+    /// use mio::net::UdpSocket;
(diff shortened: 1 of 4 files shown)


src/net/udp.rs
```diff
@@ -166,6 +166,28 @@ impl UdpSocket {
...
+    /// ```no_run
+    /// # use std::error::Error;
+    /// #
+    /// # fn main() -> Result<(), Box<dyn Error>> {
+    /// use mio::net::UdpSocket;
+    ///
+    /// let bind_addr = "127.0.0.1:0".parse()?;
+    /// let socket = UdpSocket::bind(bind_addr)?;
+    /// let addr = "192.168.0.1:41203".parse()?;
+    /// socket.connect(addr)?;
```
Thomasdezeeuw: You can create another `UdpSocket` that binds to an address and use that address to in this call to connect.

messense:
Closing because I don't have the time to work on it.

PR #1236, opened 2019-12-28
Change UdpSocket to use IoSource

Thomasdezeeuw:
Depends on #1195 (included), also see #1196.

Hopefully this will pass all tests, then I can work on #1196.


src/lib.rs
```diff
@@ -49,6 +49,8 @@ mod waker;
 pub mod event;
 
 cfg_net! {
```
kleimkuhler: `io_souce` is only used with the UDP feature for this change. Therefore, to prevent warnings on builds such as `cargo check --features tcp`, this should be changed to:

```
cfg_net! {
    pub mod net;
}

// `io_source` is only used by the UDP feature. This will be changed to
// `cfg_net!` once the other net resources use `io_source`.
cfg_udp! {
    mod io_source;
}
```

src/sys/mod.rs
```diff
@@ -61,13 +62,17 @@ cfg_os_poll! {
...
-        pub(crate) use self::unix::UdpSocket;
+        pub(crate) use self::unix::udp;
     }
 
     cfg_uds! {
         pub(crate) use self::unix::{UnixDatagram, UnixListener, UnixStream};
         pub use self::unix::SocketAddr;
     }
+
+    cfg_any_os_util! {
```
kleimkuhler: `IoSourceState` is also only used by the UDP features in this change, so this should be changed to `cfg_udp!`

src/sys/shell/mod.rs
```diff
@@ -16,12 +16,32 @@ cfg_tcp! {
...
+    pub mod udp;
 }
 
 #[cfg(unix)]
 cfg_uds! {
     mod uds;
     pub(crate) use self::uds::{UnixDatagram, UnixListener, UnixStream};
 }
+
+cfg_net! {
```
kleimkuhler: `IoSourceState` and it's methods are only used by UDP, so similarly this should be `cfg_udp!`

src/sys/unix/mod.rs
```diff
@@ -31,15 +31,36 @@ cfg_os_poll! {
...
+        pub mod udp;
     }
 
     cfg_uds! {
         mod uds;
         pub use self::uds::SocketAddr;
         pub(crate) use self::uds::{UnixDatagram, UnixListener, UnixStream};
     }
+
+    cfg_net! {
```
kleimkuhler: Similar to the above: `cfg_udp!`

src/lib.rs
Thomasdezeeuw: I know but I intend to switch all net types to use `IoSource`, e.g. see #1196, so I took a bit of a shortcut to prefer future merge conflicts.

src/sys/mod.rs
Thomasdezeeuw: I've changed this to `cfg_net!`.

src/sys/windows/selector.rs
```diff
@@ -357,10 +357,11 @@ impl Selector {
 }
 
 cfg_net! {
-    use super::SocketState;
+    use super::{SocketState, InternalState};
     use crate::Token;
     use std::os::windows::io::AsRawSocket;
 
+
```
kleimkuhler: Extra line!

Thomasdezeeuw: This is the downside of having a large amount of code inside a macro calls; no rustfmt :(

kleimkuhler: Yea pretty annoying 😔

kleimkuhler: Yea pretty annoying 😔

PR #1237, opened 2019-12-29
Use tempfile instead of tempdir

JohnTitor:
tempdir has been deprecated and using tempfile is recommended.


Thomasdezeeuw:
Can do without this dependency? I don't really have the time to review another dependency right now.

JohnTitor:
The implementation here is hard a bit. Actually, tempfile is using tempdir API (see: https://github.com/Stebalien/tempfile/pull/33) so the behavior shouldn't be changed. And it's just a dev-dependency so the effect is quite small.
But if you still have concerns, I'll close this or leave until you review it.

Thomasdezeeuw:
Removing the dependency wasn't too much work: #1239.

Thomasdezeeuw:
#1239 removed the dependency.

PR #1238, opened 2020-01-02
Use released bytes 0.5

JohnTitor:
bytes 0.5 is out now, we can use it without git rev.

### DIFF
--- a/tests/udp_socket.rs
+++ b/tests/udp_socket.rs
@@ -757,16 +757,18 @@ fn send_recv_udp(mut tx: UdpSocket, mut rx: UdpSocket, connected: bool) {
             if event.is_readable() {
                 if let LISTENER = event.token() {
                     debug!("We are receiving a datagram now...");
-                    let cnt = unsafe {
-                        if !handler.connected {
-                            handler.rx.recv_from(handler.rx_buf.bytes_mut()).unwrap().0
-                        } else {
-                            handler.rx.recv(handler.rx_buf.bytes_mut()).unwrap()
-                        }
+                    let data = &mut handler.rx_buf.bytes_mut();
+                    let data = unsafe {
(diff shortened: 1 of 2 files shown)


tests/udp_socket.rs
```diff
@@ -757,16 +757,18 @@ fn send_recv_udp(mut tx: UdpSocket, mut rx: UdpSocket, connected: bool) {
...
                 if let LISTENER = event.token() {
                     debug!("We are receiving a datagram now...");
-                    let cnt = unsafe {
-                        if !handler.connected {
-                            handler.rx.recv_from(handler.rx_buf.bytes_mut()).unwrap().0
-                        } else {
-                            handler.rx.recv(handler.rx_buf.bytes_mut()).unwrap()
-                        }
+                    let data = &mut handler.rx_buf.bytes_mut();
+                    let data = unsafe {
```
Thomasdezeeuw: Why not use the [`AsMut`](https://docs.rs/bytes/0.5.3/bytes/struct.BytesMut.html#impl-AsMut%3C%5Bu8%5D%3E) implementation?

JohnTitor: I tried with `let data = &mut rx_buf.as_mut()` but `rx_buf.as_ref()` returns empty so later assertion is failed.

Thomasdezeeuw: I tried the code locally and I see what you mean.

However this code isn't what the old `bytes_mut` did. The old `bytes_mut` returned a slice of the remaining capacity, this returns a slice to the already used bytes.

Thomasdezeeuw:
Sorry to do this again @JohnTitor, I've created pr #1240 to remove the dependency. We don't really need it any more. @JohnTitor thank you for putting the time in!

JohnTitor:
Okay, I'm closing this then. Nice work @Thomasdezeeuw!

PR #1239, opened 2020-01-02
Remove the tempdir dev-dependency

Thomasdezeeuw:
tempdir is being deprecated and merged with the tempfile crate. But
instead of reviewing another crate it took less time to replicate the
small amount of functionality we need in the test code ourselves.

Alternative to #1237.


Dushistov:
This is made in presumption that all tests run inside docker always? 
Because of in compare to `tempfile` if you run two instances of `cargo test` on the same machine that would cause random failures, because of different instaces  of `cargo test` would use the same files.

Thomasdezeeuw:
@Dushistov that is why I added https://github.com/tokio-rs/mio/pull/1239/files#diff-2e4938596e921e8ab04f6272f8dbbe3dR21-R25, to remove the temp dir in `init`. So only running the same test concurrently would be a problem.

At no point was/is/or will be running inside docker a requirement.

PR #1240, opened 2020-01-03
Remove the bytes dev-dependency

Thomasdezeeuw:
The types and traits from the bytes crate were only used in two tests
for the UdpSocket. But a Vec<u8> (now) has all the same functionality,
so we don't need the crate anymore.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,9 +52,6 @@ ntapi  = "0.3"
 lazy_static = "1.4.0"
 
 [dev-dependencies]
-# Bytes v0.4 still depends on winapi v0.2, but Bytes v0.5 is released yet. So
-# for testing we'll use the git version.
-bytes      = { version = "0.5.0", git = "https://github.com/tokio-rs/bytes", rev = "79e4b2847f27137faaf149d116a352cbeae47fd1" }
 env_logger = { version = "0.6.2", default-features = false }
 tempdir    = "0.3.7"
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
The failure seems unrelated to this change.

```log
error: truncating cast: the value 18446744073709551615 requires 64 bits but the target type is only 32 bits
   --> src\sys\windows\selector.rs:650:57
    |
650 |                 return Err(io::Error::from_raw_os_error(INVALID_SOCKET as i32));
    |                                                         ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[deny(const_err)]` on by default
```

I guess a new lint was added to the nightly compiler? @dtacalau do you know why we don't just use `io::Error::last_os_error` here? Looking at [the docs](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsaioctl#return-value) that seems like the right thing to do.

Thomasdezeeuw:
Created #1241 for the CI failure.

dtacalau:
> @dtacalau do you know why we don't just use `io::Error::last_os_error` here? Looking at [the docs](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsaioctl#return-value) that seems like the right thing to do.

@Thomasdezeeuw don't know why, `io::Error::last_os_error` seems better indeed.

PR #1242, opened 2020-01-03
Change TcpListener to use IoSource

Thomasdezeeuw:
Depends on #1196.


Thomasdezeeuw:
Rebased on master.

Thomasdezeeuw:
Well this is interesting. I completely forgot to set the accepted `TcpStream` to non-blocking and on macOS it just works, but on Linux it doesn't.

Thomasdezeeuw:
I just tested FreeBSD and it works with setting `TcpStream` to non-blocking as well.

PR #1243, opened 2020-01-03
Change UnixDatagram to use IoSource

Thomasdezeeuw:


src/sys/unix/uds/datagram.rs
```diff
@@ -1,166 +1,90 @@
...
-
-    pub(crate) fn unbound() -> io::Result<UnixDatagram> {
-        let socket = new_socket(libc::AF_UNIX, libc::SOCK_DGRAM)?;
-        Ok(unsafe { UnixDatagram::from_raw_fd(socket) })
-    }
-
-    pub(crate) fn local_addr(&self) -> io::Result<SocketAddr> {
-        SocketAddr::new(|sockaddr, socklen| {
-            syscall!(getsockname(self.inner.as_raw_fd(), sockaddr, socklen))
+pub(crate) fn bind(path: &Path) -> io::Result<net::UnixDatagram> {
```
kleimkuhler: Not a blocker, but I think the readability of combinators becomes more difficult in a chain two or three operations long. I prefer `?` style for this function.

Thomasdezeeuw: I'm starting to like the functional more and more so it starts to creep into changes like this. But I see where you're coming from, so I changed it back.

PR #1244, opened 2020-01-03
Add cargo-hack for feature powerset checks

### LINKED ISSUE #1214: Test feature flag combinations in CI
## Motivation

Once #1192 merges, Mio will have several feature flags that can be used in different combinations of each other. CI should test that the combinations are correct. Currently, it would just test `--all-features`.

## Solutions

1. I've been using [cargo-hack](https://github.com/taiki-e/cargo-hack)'s `--feature-powerset`. **Cons**: It's a little expensive to run right now because it mixes in `guide` doubling the number of combinations that needs to be tested. CI could download cargo hack and run it in place of `cargo check` steps.

2. Explicitly test the combinations we care about. **Cons**: A combination could be missed, and it would need to be updated as features are added (probably not that often..)

kleimkuhler:
Check combinations will only be performed on the three main stable targets
(Linux, MacOS, and Windows).

Closes #1214


ci/azure-test-stable.yml
```diff
@@ -25,6 +25,10 @@ jobs:
       - template: azure-install-rust.yml
         parameters:
           rust_version: ${{ parameters.rust_version }}
+      
+      - ${{ if eq(parameters.cmd, 'test') }}:
+          - script: ./bin/check-powerset
```
Thomasdezeeuw: I don't know if Windows has bash but many try `bash ./bin/check-powerset`?

Otherwise we might need to port it to PowerShell.

kleimkuhler: Yea unforutnately `bash ..` does not work. I could add a second `.ps1` file that does a similar thing?

Or it could go back to using `cargo-hack`. Do you have a preference?

Thomasdezeeuw: I'm fine either option.

kleimkuhler:
It's running properly. In the "Test Linux", "Test MacOS",  and "Test Windows" jobs there is a new "Check feature powerset" step that has run.

Thomasdezeeuw:
Ah I see. A couple of import warnings are generated, e.g. https://dev.azure.com/tokio-rs/Tokio/_build/results?buildId=4438&view=logs&j=429a383f-e486-57ac-3982-9ac65641a32a&t=1f926ffb-0ec7-5f9b-ab05-76fbb04597a8, do we want to fail on that?

kleimkuhler:
Yep working on cleaning those up now. Hadn't realized Windows also had warnings going on when fixing `to_sockaddr_storage` in #1249.

I would be okay with those failing by default now. I think Tokio has that behavior already?

Thomasdezeeuw:
I've also got a branch going fixing them. I'm ok with failing.

Thomasdezeeuw:
I think the second commit in #1251 fixes all of them. If you change this to fail on warnings and rebase on #1251 (after its merged) we can see if everything is fixed.

kleimkuhler:
Sounds good I'll wait for #1251 to merge!

Thomasdezeeuw:
Thanks @kleimkuhler, great work.

PR #1247, opened 2020-01-04
Use accept4(2) on platforms that support it

### LINKED ISSUE #1213: Use accept4 where possible
In `sys::TcpListener::accept`.

Thomasdezeeuw:
Fixes #1213.

~~Depends on #1242.~~

### DIFF
--- a/src/sys/unix/tcp.rs
+++ b/src/sys/unix/tcp.rs
@@ -1,9 +1,9 @@
 use std::io;
-use std::mem::size_of;
+use std::mem::{size_of, MaybeUninit};
 use std::net::{self, SocketAddr};
-use std::os::unix::io::FromRawFd;
+use std::os::unix::io::{AsRawFd, FromRawFd};
 
-use crate::sys::unix::net::{new_ip_socket, socket_addr};
+use crate::sys::unix::net::{new_ip_socket, socket_addr, to_socket_addr};
 
 pub fn connect(addr: SocketAddr) -> io::Result<net::TcpStream> {
     new_ip_socket(addr, libc::SOCK_STREAM)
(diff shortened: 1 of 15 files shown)


src/sys/windows/tcp.rs
```diff
@@ -64,3 +64,8 @@ pub fn bind(addr: SocketAddr) -> io::Result<net::TcpListener> {
         .map(|_| unsafe { net::TcpListener::from_raw_socket(socket as StdSocket) })
     })
 }
+
+pub fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream, SocketAddr)> {
+    // The non-blocking state of `listener` is inherited.
```
kleimkuhler: Is the source for this assumption the [Remark](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-accept#remarks)? I don't think the source needs to be an explicit comment, but just want to make sure this comment isn't assuming from something else.

src/sys/unix/tcp.rs
```diff
@@ -50,3 +50,52 @@ pub fn bind(addr: SocketAddr) -> io::Result<net::TcpListener> {
...
+            addr.as_mut_ptr() as *mut _,
+            &mut length,
+            libc::SOCK_CLOEXEC | libc::SOCK_NONBLOCK,
+        ))
+        .map(|socket| unsafe { net::TcpStream::from_raw_fd(socket) })
+    }?;
+
+    // But not all platforms have the `accept4(2)` call. Luckily BSD (derived)
+    // OSes inherit the non-blocking flag from the listener, so we just have to
+    // set `CLOEXEC`.
```
kleimkuhler: > When using  accept(), portable programs should not rely on the O_NONBLOCK
> and O_ASYNC properties and  the signal destination being inherited, but
> should set them explicitly using fcntl(2); accept4() sets these properties
> consistently, but may not be fully portable across UNIX platforms.

This makes sense. The "portable" programs are eiter going to have `accept4`, or we can assume inherit `NONBLOCK`.

src/sys/windows/tcp.rs
Thomasdezeeuw: I was indeed going from those docs, I've added a link in the last commit anyway. Always useful for people (and future me when I forget).

src/sys/unix/tcp.rs
Thomasdezeeuw: I've manually checked all the OSes and the list and they all inherit the `NONBLOCK` property, it seems only Linux really diverges from this.

PR #1252, opened 2020-01-07
Expand usage of SelectorId

Thomasdezeeuw:
This expands the checks in SelectorId, adding checks for use in
event::Source::{reregister, deregister}.

In addition this moves the SelectorId to the io_source module, as its
the only user.

@dtacalau after this additional tests like in #1201 will give more consistent results :)

### DIFF
--- a/src/io_source.rs
+++ b/src/io_source.rs
@@ -3,12 +3,12 @@ use std::ops::{Deref, DerefMut};
 use std::os::unix::io::AsRawFd;
 #[cfg(windows)]
 use std::os::windows::io::AsRawSocket;
+#[cfg(debug_assertions)]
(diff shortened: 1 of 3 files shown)


src/io_source.rs
```diff
@@ -198,3 +206,87 @@ where
         fmt::Debug::fmt(&self.inner, f)
     }
 }
+
+/// Used to associate an `IoSource` with a `sys::Selector`.
+#[derive(Debug)]
+#[cfg(debug_assertions)]
```
kleimkuhler: nit: It's not order dependent, but could the `cfg` come the `derive`? I think it's more more important to see that a type is conditionally compiled than the attributes it derives *when* it's included.

Thomasdezeeuw: Done.

Issue #1122, opened 2019-11-05
Reconsider `transparent` on `Event`


carllerche:
I'm mostly opening this issue as I'm not sure what the benefit of having `transparent` on `Event` is. It may be that I am blind, but I thought I would open the issue instead of forgetting to ask :)

Thomasdezeeuw:
I think I added this because at some point we did `&Event as *const Event as *const libc::kevent`, which needed the transparent attribute. But I don't think we need it any more.

PR #1253, opened 2020-01-07
Add back repr(transparent) to Event

Thomasdezeeuw:
I removed the attribute in pull request #1155, however we still need
this attribute for Event::from_sys_event_ref.

Original issue: #1122.

### DIFF
--- a/src/event/event.rs
+++ b/src/event/event.rs
@@ -12,6 +12,7 @@ use std::fmt;
 /// [`Poll::poll`]: ../struct.Poll.html#method.poll
 /// [`Poll`]: ../struct.Poll.html
 /// [`Token`]: ../struct.Token.html
+#[repr(transparent)]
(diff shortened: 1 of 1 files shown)


kleimkuhler:
Has `Event::from_sys_event_ref` not been working correctly since getting rid of the `repr`?

Thomasdezeeuw:
It has. But I believe its not guaranteed that `Event` and `sys::Event` have the same layout without `repr(transparent)`, so this: https://github.com/tokio-rs/mio/blob/77991fbd705a8e30ec798ba4be2e008df84f591f/src/event/event.rs#L182-L188 is not actually guaranteed to be valid without it. The compiler could decide that `Event` needs some padding (for what ever reason), which is completely valid, but now `Event` and `sys::Event` have different layouts, thus references to the two can't be converted to/from each other.

Issue #1211, opened 2019-12-10
Introduce non-libstd socket builder methods


kleimkuhler:
## Motivation

#1029 removed the `net` resources APIs to match those of `libstd`. This change also allowed Mio to remove it's dependency on `net2` which was used to provide these methods.

As part of the Mio update in Tokio--and mentioned briefly in the other issues--these methods would be helpful to provide again. For example, they are currently [removed](https://github.com/tokio-rs/tokio/pull/1767/files#diff-91404b27ce0ae7138a049039d5674df7L365-L510) as part of the Mio update PR in Tokio.

`socket2-rs`/`net2` provide a lot of additional methods that Mio does not really need and reintroduces a dependency when the syscalls would be fairly straightforward to implement directly. That being said, it's a dependency that can be relied on for the things we would need it for.

Alternatively, a new API could be introduced that provides builder methods for sockets (such as `set_keepalive`/`keepalive`/`set_ttl`/etc.) This approach would not introduce a new dependency. Something along the lines of providing a `TcpSocket` that has these methods and is consumed with a `connect -> TcpStream`.

Speaking with @carllerche offline about this I think the second approach of introducing something along the lines of `TcpSocket` builder is what Mio should move forward with, but there is more design conversation that should go into it.

Thomasdezeeuw:
I've outed thing opinion before, but I'll repeat it here. I don't think it wise to duplicate the effort of a `net2`/`socket` crate. We already have the `FromRawFd`/`FromRawSocket` and `from_std` implementations that allow for advanced usage of all our `net` type (not just `TcpStream`, as the other types also have useful options that the std lib versions don't provide).

In the past I've also tried adding a way to set options on `TcpListener`, see https://github.com/tokio-rs/mio/pull/1068. I think it would be best if we would add the required methods to std lib instead, then everybody can benefit.

Thomasdezeeuw:
We now have `TcpSocket` and #1381 track switching to socket2.

PR #1258, opened 2020-01-10
Consolidate existing socket system calls into private `Socket` type

kleimkuhler:
## Motivation

This PR introduces `sys::Socket` that is internal to Mio and consolidates all the existing socket system calls into a single type.

There has been a lot of discussion about how Mio should provide additional methods for setting certain socket options. Currently, it is left up to the user to pull in an additional dependency (`socket2-rs`/`net2`) and use the provided `As`/`From`/`Into` socket impls.

There are however various system calls for working with sockets in Mio that are specific to setting non-blocking behavior and using as few system calls as possible. They include functionality that creat sockets to be non-blocking by default, using `accept4` on targets that provide it, and relying on inheriting certain socket statuses where possible.

## Solution

`Socket` is a `RawFd`/`RawSocket` wrapper type that provides the necessary methods for consolidating all the `sys::target::net` socket sytem calls into a single type.

Since `Socket` works at the target's socket layer, any future changes related to additional functionality or reported bugs can take place in these methods and apply to all the types that use them.

**Additionally**, `Socket` provides the necessary building block for providing a public `net` socket types. This will allow users of Mio to set additional socket options without worrying about separate system calls for setting non-blocking mode, and avoid pulling in additional dependencies. An example of the public socket type can be found #1259.

### DIFF
--- a/src/sys/unix/socket.rs
+++ b/src/sys/unix/socket.rs
@@ -98,12 +98,16 @@ impl Socket {
     }
 
     #[cfg(any(feature = "tcp", feature = "udp"))]
-    pub(crate) fn from_addr(addr: SocketAddr, socket_type: libc::c_int) -> Result<Self> {
+    pub(crate) fn from_addr(
+        addr: SocketAddr,
+        socket_type: libc::c_int,
+        protocol: libc::c_int,
+    ) -> Result<Self> {
         let domain = match addr {
             SocketAddr::V4(..) => libc::AF_INET,
             SocketAddr::V6(..) => libc::AF_INET6,
         };
-        Self::new(domain, socket_type, 0)
+        Self::new(domain, socket_type, protocol)
     }
 
     #[cfg(feature = "tcp")]
--- a/src/sys/unix/socket.rs
+++ b/src/sys/unix/socket.rs
@@ -55,7 +55,49 @@ impl Socket {
                 })
         });
 
-        socket.map(|socket| Socket { fd: socket })
+        socket.map(|socket| unsafe { Socket::from_raw_fd(socket) })
+    }
+
+    #[cfg(feature = "uds")]
+    pub(crate) fn pair(
+        domain: libc::c_int,
+        socket_type: libc::c_int,
+        protocol: libc::c_int,
+    ) -> Result<(Self, Self)> {
+        #[cfg(any(
+            target_os = "android",
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "linux",
+            target_os = "netbsd",
+            target_os = "openbsd"
+        ))]
+        let socket_type = socket_type | libc::SOCK_NONBLOCK | libc::SOCK_CLOEXEC;
+
+        let mut fds = [-1; 2];
+        syscall!(socketpair(domain, socket_type, protocol, fds.as_mut_ptr()))?;
+
+        // Darwin and Solaris do not have SOCK_NONBLOCK or SOCK_CLOEXEC.
+        //
+        // In order to set those flags, additional `fcntl` sys calls must be
+        // performed. If a `fnctl` fails after the sockets have been created,
+        // the file descriptors will leak. Creating `pair` above ensures that if
+        // there is an error, the file descriptors are closed.
+        #[cfg(any(target_os = "ios", target_os = "macos", target_os = "solaris"))]
+        syscall!(fcntl(fds[0], libc::F_SETFD, libc::FD_CLOEXEC))
+            .and_then(|_| syscall!(fcntl(fds[0], libc::F_SETFL, libc::O_NONBLOCK)))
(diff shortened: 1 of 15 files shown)


src/sys/unix/tcp.rs
```diff
@@ -1,101 +1,30 @@
...
-            // from closing since we can't pass back two errors.
-            let _ = unsafe { libc::close(socket) };
-            err
-        })
-        .map(|_| unsafe { net::TcpListener::from_raw_fd(socket) })
-    })
+    let socket = Socket::from_addr(addr, libc::SOCK_STREAM)?;
+    socket.bind(addr)?;
+    socket.listen(1024)?;
+    unsafe { Ok(net::TcpListener::from_raw_fd(socket.into_raw_fd())) }
```
oskardrums: This pattern is quite risky: if any method fails between creating the Socket and casting it into another type, an FD will leak.
A good alternative may be implementing Drop for Socket so the fd gets closed no matter what. That will also reduce some code duplication :)

src/sys/unix/socket.rs
```diff
@@ -0,0 +1,274 @@
...
+
+        let socket = syscall!(socket(domain, socket_type, protocol))?;
+
+        // Darwin and Solaris do not have SOCK_NONBLOCK or SOCK_CLOEXEC.
+        //
+        // In order to set those flags, additional `fcntl` sys calls must be
+        // performed. If a `fnctl` fails after the socket has been created,
+        // `close` ensures the socket does not leak.
+        #[cfg(any(target_os = "ios", target_os = "macos", target_os = "solaris"))]
+        syscall!(fcntl(socket, libc::F_SETFL, libc::O_NONBLOCK))
```
oskardrums: This closure maybe better suited as a method that sets an arbitrary fd to be nonblocking and cloexec etc., so the same logic will be easily applicable after using from_raw_fd on an arbitrary FD (maybe also F_GETFL the original flags before setting).

oskardrums:
I would love to see this merged, really useful for allowing support for other kinds of sockets (SCTP, raw...)

src/sys/unix/tcp.rs
kleimkuhler: The `Socket` methods this dispatches to will close the FD on errors. Here is `bind2` for [example](https://github.com/kleimkuhler/mio/blob/0c22d3d57d16ebefa0e75606c42b3190d8e7a6a7/src/sys/unix/socket.rs#L156-L161). This currently doesn't happen for `Socket::listen` so that should be added!

If the `bind` or `listen` then fail, the FD would be closed and it wouldn't try to create a `TcpListener`.

src/sys/unix/socket.rs
kleimkuhler: > so the same logic will be easily applicable after using from_raw_fd on an arbitrary FD

Yes that could be useful! For the targets where this currently runs, we know these flags are not set by default and the FD was just created, so an extra `F_GETFL` would probably be unnecessary.

For this PR however I was trying to keep it as similar as possible without changing too much, since the intent was to just introduce the `Socket` type. I'm not sure if this PR will land due to discussion above, but if it makes it in in the future for supporting other kinds of sockets it would be something to discuss!

kleimkuhler: Edit: Ah sorry discussion briefly took place in the Tokio discord and not on this PR. Maybe @Thomasdezeeuw can echo some of their thoughts on the PR as well. Essentially, we still aren't sure if a `Socket` type (private or not) still makes sense in Mio.

src/sys/unix/tcp.rs
oskardrums: Well, as I said IMO it's too risky, because whatever new code that will use this interface in the future won't be aware of this implicit "critical section" in which FDs may leak.
So `Socket::from_addr` should either return a fully safe-to-use socket, or a clearly WIP-dont-drop-me-now socket.

kleimkuhler: Oh okay yes I see what you mean. Yep that is good point. I mentioned in my comment below that there may be some further discussion if this is even a change Mio wants to consider, but I can add a `Drop` impl for `Socket` to reduce some code duplication and ensure a leak doesn't occur for methods where there isn't an explicit `libc::close`.

Thomasdezeeuw:
I've mentioned this on the Discord chat, but I forgot to add it here. If we don't expose a `TcpSocket` type (see #1211) I don't see this pr as adding benefit at the moment. It add 100 lines of codes, without a change in functionality.

I intend to look into a new crate (likely [`socket2`](https://crates.io/crates/socket2)) as a new place for this kind of code. It would be a low-level library that provides all options/function that the OS' exposes in a nicer Rust interface, but it will not do anything cross-platform, i.e. if a function is not available in the OS/libc its not available in the crate.

kleimkuhler:
Closing due to the conversation above.

PR #1259, opened 2020-01-10
draft: Introduce `TcpSocket` (Unix right now)

kleimkuhler:
Depends on #1258 

Draft to test targets other than MacOS and Linux.


kleimkuhler:
Closing due to conversation in #1258

PR #1260, opened 2020-01-10
Update README to work with latest API

sezna:
I was setting up a new project with mio and noticed the README example is just a touch out of date. This fixes that.


Thomasdezeeuw:
Thanks @sezna.

PR #1261, opened 2020-01-14
Update winapi dependency to 0.3.8.

arlosi:
This change updates the winapi dependency to 0.3.8, and the miow dependency to 0.3.3.

winapi 0.2 does not support compilation on Windows arm64, which prevents this
crate from being used on that platform.

Most of this change is simple, except for the tcp `schedule_accept` function. This change was required because the signature of `accept_overlapped` in winapi changed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ name          = "mio"
 # - Update CHANGELOG.md.
 # - Update doc URL.
 # - Create git tag
-version       = "0.6.21"
+version       = "0.6.22"
 license       = "MIT"
 authors       = ["Carl Lerche <me@carllerche.com>"]
 description   = "Lightweight non-blocking IO"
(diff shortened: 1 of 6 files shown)


Thomasdezeeuw:
This would be a breaking change for the v0.6 branch and based on that it would not be accepted. Luckily the 0.7 branch (on master) already uses winapi 0.3.

arlosi:
Do you have an estimate for when a stable (non-alpha) 0.7 release will be published to crates.io?

Thomasdezeeuw:
Honestly the alpha is actually quite stable, haven't had any bug reports and I don't expect the API to change in 0.7 stable. I'm ok with doing a stable release this month, but we don't have a release date yet. It will go into Tokio 0.3, which releases in two months or so I believe (100% sure).

arlosi:
Thanks. My end goal is to build rustc without winapi 0.2.

The dependency chain I'm looking at is: rustc -> tokio 0.2 -> mio 0.6 -> winapi 0.2

Another option to break the chain would be updating tokio 0.2 to build with mio 0.7. Do you think that would be acceptable? Or do you do you expect I would need to wait until tokio 0.3 is released?

Thomasdezeeuw:
I'm afraid you'll have to wait until Tokio 0.3 is released.

PR #1262, opened 2020-01-20
Use const fn for new

yjhmelody:


src/interest.rs
```diff
@@ -96,7 +96,7 @@ impl ops::BitOr for Interest {
 
     #[inline]
     fn bitor(self, other: Self) -> Self {
-        Interest(unsafe { NonZeroU8::new_unchecked(self.0.get() | other.0.get()) })
+        self.add(other)
```
udoprog: Was this accidentally included?

yjhmelody: No, just reduce the duplicated code

Thomasdezeeuw:
Thanks @yjhmelody.

Issue #1152, opened 2019-11-15
Port to Solaris


psumbera:
Currently mio is using for Solaris epoll interface. But Solaris doesn't have this interface. Only Illumos has it.

There are two things to be done:
- To distinguish between Solaris and Illumos.
- Implement selector on Solaris (probably via even ports API)

Thomasdezeeuw:
So you're saying that the Solaris port never worked?

@psumbera do you use Solaris and are you willing to contribute such a port? Otherwise we maybe should just drop support for Solaris if it never worked in the first place.

/cc @carllerche

carllerche:
Well, we need some way to test it in CI... none of the mio devs have a solaris setup.

psumbera:
Mio is used by Rust. And Solaris really needs Rust to build Firefox. From Rust version 1.38 something changed and it now fails to link on Solaris because of this issue (though I know Mio was used also with previous Rust versions without the issue).

I'm Solaris engineer so I have access to Solaris. Unfortunately I'm not very good with Rust. But anyway I will try to come with something or will seek for help.

Thomasdezeeuw:
Can you try testing Mio (by running `cargo test`) and give us an output of the error, then we can try and fix it.

Also could you help with setting up the CI for Solaris? Going forward we would like to a bit stricter about what platforms we support and having a CI for it is one of them.

psumbera:
> Can you try testing Mio (by running `cargo test`) and give us an output of the error, then we can try and fix it.

```
error: linking with `gcc` failed: exit code: 1
  |
[...]

PR #1263, opened 2020-01-22
Port MIO library to Solaris OS

batrla:
Solaris OS (unlike its fork Illumos) has no epoll(7) interface. The port of MIO
library to Solaris OS adds support of standard POSIX poll(2) interface and
implements edge-triggered semantic in a user-space.

Please review the code change, I'm willing to make updates if needed.
The PR fixes some minor defects I found.

Cargo test output is clean, no failure:

$ grep test\ result ~/tmp/x
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 16 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 14 passed; 0 failed; 5 ignored; 0 measured; 0 filtered out
test result: ok. 31 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 10 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 10 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 35 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out


asomers:
AFAIK there's no hosted CI solution for Oracle Solaris.  So mio can't ever maintain this port.  Nor should we, I think.  Oracle has all but abandoned Solaris.  How much unpaid effort should we put into supporting an abandoned, non-free OS?

Thomasdezeeuw:
@psumbera was working on a Solaris port in #1152, could provide us with a CI solution?

batrla:
> AFAIK there's no hosted CI solution for Oracle Solaris.  So mio can't
> ever maintain this port.

The implementation uses poll(2) interface which is defined by POSIX, so 
this port can be theoretically tested on any Unix based system. It should 
work also on Linux (I haven't verified it yet). It's a matter of adding
configuration option / switch / feature in port code to fallback
from epoll/kqueue to poll(2) and ensuring it's OK.

> Nor should we, I think.  Oracle has all but abandoned Solaris.  How much
> unpaid effort should we put into supporting an abandoned, non-free OS?

As far as I know Oracle declares Solaris 11 OS is supported at least till 
2031/2034 per support plan. Yes, it's non-free OS and I'm not 
aware of free CI solution for it. On the other hand I'm willing
to add way to unit test the suggested fix on Linux.

batrla:
I modified the code to automatically test MIO port to POSIX poll(2) on Linux pipeline in Azure, please see an update in CI tasks.

Thomasdezeeuw:
I do think that running on the actual OS is a requirement for Mio to support it. I also don't want to support `poll(2)` on Linux so that route is a no-go for me.

psumbera:
What can we do to have Solaris in CI? Can we help with the setup?

Thomasdezeeuw:
What we need is something/somewhere to run the Solaris OS. Can Solaris be virtualised somehow/somewhere (without cost to the maintainers of Mio)?

psumbera:
Oracle Solaris is available in Oracle cloud, or we can run eventually tests somewhere, plus it can be run in VirtualBox.  But how to connect it to Github CI?

Thomasdezeeuw:
@psumbera sorry for the late response. Maybe something like Jenkins (https://jenkins.io/solutions/github/) on a virtual machine on Oracle cloud would work?

psumbera:
Jenkins might be the solution. I wonder whether you need to have access to the machine or it's enough if we set it up for you. In any case we need to explore possibilities to get the machine first.

Thomasdezeeuw:
@psumbera I don't think we need access to the machine. All we need to know is that the tests passed, so the output log and some kind of feedback to GitHub would suffice.

That would mean that if we find a Solaris specific problem we need to relay that to someone with access to a Solaris machine to debug it.

batrla:
If there is a native Solaris CI, then I'll update the fix to use port_associate(3C) instead and take out poll(2) unless there's an interest to keep it. Event ports semantic is closer to what MIO does today and it simplifies the code change.

Thomasdezeeuw:
@batrla changing to use `port_associate(3c)` sounds good to me, you can remove `poll(2)` as no other platforms will make use of it.

Thomasdezeeuw:
This port still doesn't seem to have commitment to any maintainers, so I'm going to close this.

psumbera:
Following would hopefully resolve the issue too: https://github.com/tokio-rs/mio/pull/1602

PR #1264, opened 2020-01-23
Expand on documentation

### LINKED ISSUE #1223: Expand documentation of Mio features
After #1192 Mio makes much more use of features, but these are only documented shortly in [`Cargo.toml`](https://github.com/tokio-rs/mio/blob/3ebce3e276f58126c25bdf9b606fe4676c0013c1/Cargo.toml#L26-L38). When documentation is generated, such as for doc.rs, all features are enabled so it might be surprising for new users that the need to enable a bunch of features.

I propose two solutions two this:
* Add documentation about the available features in the main crate docs, i.e. in `lib.rs`. This ways its clear for users when the open the documentation the first type.
* Add the required feature(s) per type.
  Preferably we add a box like the boxes added for unstable/nightly features (e.g. the yellow nightly-only box https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#associatedconstant.UNINIT). But I don't think that is possible to do for user code atm.
  The alternative is just adding a notes section to each type to let the user know what feature(s) are required.

Thomasdezeeuw:
Collection of four small commits that revise some of the documentation.

Closed #1223.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -80,8 +80,20 @@ cfg_any_os_util! {
 pub mod guide {
     //! # Getting started guide.
     //!
-    //! Using Mio starts by creating a [`Poll`] instance, which reads events from
-    //! the OS and puts them into [`Events`].
+    //! In this guide we'll do the following:
+    //!
+    //! 1. Create a [`Poll`] instance (and learn what it is).
+    //! 2. Register an [event source].
+    //! 3. Create an event loop.
+    //!
+    //! At the end you'll have a very small (but quick) TCP server that accepts
+    //! connections and then drops (disconnects) them.
+    //!
+    //! ## 1. Creating a `Poll` instance
+    //!
+    //! Using Mio starts by creating a [`Poll`] instance, which monitors events
+    //! from the OS and puts them into [`Events`]. This allows us to execute I/O
+    //! operations based on what operations are ready.
     //!
     //! [`Poll`]: ../struct.Poll.html
     //! [`Events`]: ../event/struct.Events.html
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
Maybe this should be moved to a new `docs` directory. I want to add the following:

* Platforms supported
* Upgrade guide
* Getting started guide (currently already in)
* Description of features (this pr)

src/lib.rs
```diff
@@ -67,21 +82,77 @@ pub use poll::{Poll, Registry};
...
+#[cfg(feature = "extra-docs")]
+pub mod features {
+    //! # Mio's optional features.
+    //!
+    //! This document describes the available features in Mio.
+    //!
+    #![cfg_attr(feature = "os-poll", doc = "## `os-poll` (enabled)")]
+    #![cfg_attr(not(feature = "os-poll"), doc = "## `os-poll` (disabled)")]
+    //!
+    //! Mio by default provides only a shell implementation, that `panic!`s the
```
kleimkuhler: nit: I don't think the comma after `implementation` is needed

src/lib.rs
```diff
@@ -67,21 +82,77 @@ pub use poll::{Poll, Registry};
...
+    //! moment it is actually run. To run it requires OS support, this is
```
kleimkuhler: how about: `.. it is actually run. OS support is required in order to run and is enabled by activating the `os-poll` feature`.

src/lib.rs
```diff
@@ -67,21 +82,77 @@ pub use poll::{Poll, Registry};
...
+    //!
+    #![cfg_attr(feature = "tcp", doc = "* `tcp` (enabled)")]
+    #![cfg_attr(not(feature = "tcp"), doc = "* `tcp` (disabled)")]
+    //! : includes `TcpStream` and `TcpListener`,
+    #![cfg_attr(feature = "udp", doc = "* `udp` (enabled)")]
+    #![cfg_attr(not(feature = "udp"), doc = "* `udp` (disabled)")]
+    //! : includes `UdpSocket`, and
+    #![cfg_attr(feature = "uds", doc = "* `uds` (enabled)")]
+    #![cfg_attr(not(feature = "uds"), doc = "* `uds` (disabled)")]
+    //! : includes `UnixDatagram`, `UnixListener`, `UnixStream` and `SocketAddr`.
```
kleimkuhler: Could the net types here (`TcpStream`, `TcpListener`, `UdpSocket`, etc.) have links provided at the end of this doc section so that they are click throughs?

For example line 119 `//! : includes 'TcpStream' and 'TcpListener'` could be changed to `//! : includes ['TcpStream'] and ['TcpListener']`

kleimkuhler:
I'm in support of a a new `docs` directory. I like this change, but I think if maybe the feature was changed to `docs` and the `guide` module was just included with that, it would be a good way to have the extra documentation and examples included.

Thomasdezeeuw:
@kleimkuhler I've move the documentation to a new `docs` directory and addressed all your comments.

kleimkuhler:
So I think I misunderstood your comment above. I did like the fact that `guide` was a feature flagged module and included its documentation and examples that were checked against.

For example, we also saw with #1260 that code in `.md` files can easily fall out of sync with API changes. Could we keep the pattern we currently have and conditionally include the `docs` module where the examples include code as doc tests?

Thomasdezeeuw:
@kleimkuhler I've been trying to convert the the docs into doc tests, but haven't got the best results. The changes are in https://github.com/Thomasdezeeuw/mio/commit/fcbb9ead2a9d019d6165cb3e9f8231cc202a475f.

It has two problems:
* It currently doesn't work on stable as it requires `external_doc`, but maybe we can work around that.
* It requires that all examples are *full* examples, i.e. include all `use` statements, use all variables etc. Especially for the getting started guide this add a lot of noise to the code examples and silly code like `# drop(poll)` to silence the unused variable warning. I've tried adding `#![allow(unused_variables)]` but that didn't work.

kleimkuhler:
@Thomasdezeeuw Sorry about not responding to your comment above.

Not sure what your plans on with this branch, but I'm still not really sure changing an entire example into markdown is the right move here.. I'm tempted to say for this PR maybe keep that as part of the doc comment? I know the API doesn't (or hopefully with 0.7 alpha being released) that much, but seems like it will still lead to incorrect examples at some point.

We already have examples in `examples/` that I think do a pretty good job at offering basic uses of Mio types. Maybe include the couple few-liners you have that show making a `Poll` instance and such, but keep the larger example as a doc comment or point to examples?

Thomasdezeeuw:
@kleimkuhler I'm ok with moving it back. However I don't much time for this branch at the moment.

kleimkuhler:
@Thomasdezeeuw No problem

Thomasdezeeuw:
@kleimkuhler I've reverted this to rustdoc comments, can you review & merge?

PR #1271, opened 2020-02-10
Implement ExactSizeIterator and FusedIterator for events::Iter

Thomasdezeeuw:
I know @carllerche was against this previously because maybe not all platforms could support it. But this pr only enables it for release build to enforce it is not part of the stable API.

### DIFF
--- a/src/event/events.rs
+++ b/src/event/events.rs
@@ -208,7 +208,38 @@ impl<'a> Iterator for Iter<'a> {
         self.pos += 1;
         ret
     }
+
+    #[doc(hidden)]
+    #[cfg(not(debug_assertions))]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let len = self.len();
+        (len, Some(len))
+    }
+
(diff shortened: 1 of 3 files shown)


carllerche:
Why does conditionally adding it not include it as part of the stable API? Also, what is the drive to add this?

Thomasdezeeuw:
This won't make the API available normally, it's only available when the `--release` flag is passed to cargo. So if code that depends on this API is run without the release flag it won't compile. The reason for the change is performance: the `ExactSizeIterator` and `TrustedLen` (unstable) traits allow the compiler to further optimise the code. So even though the methods on it aren't available for the developer, the compiler can make use of them in release build when desugaring `for` loops. Same goes for `FusedIterator`, which should really only be used in combination with `Iterator::fuse` anyway, where `FusedIterator` simply provide an optimised version that does the same thing.

Thomasdezeeuw:
It seems this isn't going to happen.

PR #1283, opened 2020-03-18
waker: add a test using multiple wakers

Keruspe:
When we register multiple wakers and different wakers are activated, poll should receive events for all of the activated ones.

This used to work fine with Registrations in mio 0.6.

It seems to work fine with mio 0.7 at least on linux and windows but I suspect there's a problem on macos, unless I missed some other problem on my side. Hopefully CI will confirm or infirm this statement and this test covers a scenario which wasn't covered yet.

### DIFF
--- a/tests/waker.rs
+++ b/tests/waker.rs
@@ -106,6 +106,34 @@ fn waker_multiple_wakeups_different_thread() {
     handle2.join().unwrap();
 }
 
+#[test]
+fn multiple_wakers_same_thread() {
+    init();
+
+    let mut poll = Poll::new().expect("unable to create new Poll instance");
+    let mut events = Events::with_capacity(10);
+
+    let token1 = Token(10);
+    let token2 = Token(11);
+    let waker1 = Waker::new(poll.registry(), token1).expect("unable to create waker");
+    let waker2 = Waker::new(poll.registry(), token2).expect("unable to create waker");
+
+    waker1.wake().expect("unable to wake");
+    waker2.wake().expect("unable to wake");
+
+    let mut missing = vec![token1, token2];
+
+    poll.poll(&mut events, Some(Duration::from_millis(100)))
+        .unwrap();
+    assert!(!events.is_empty());
+    for event in events.iter() {
+        assert!(event.is_readable());
+        assert!(missing.contains(&event.token()));
+        missing.retain(|tk| *tk != event.token());
+    }
+    assert!(missing.is_empty());
+}
+
 fn expect_waker_event(poll: &mut Poll, events: &mut Events, token: Token) {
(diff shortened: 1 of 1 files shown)


Keruspe:
Fixed rustfmt

Thomasdezeeuw:
Although this currently works we do have the following in the docs:

> Only a single `Waker` should active per [`Poll`], if multiple threads need access to the `Waker` it can be shared via for example an `Arc`.

Source: https://github.com/tokio-rs/mio/blob/fefbd790be734ce2313c3dc51878973956bc325d/src/waker.rs#L19-L22

I added it mostly as a precaution to ensure we could switch `Waker` strategies later as needed, but I think (at the moment) this test would send the wrong signal. Of course we can discuss whether or not we want to support multiple `Waker`s per `Poll`.

Thomasdezeeuw:
And the macOS/iOS tests also seem to fail.

Keruspe:
Yeah, so CI confirms my suspicion.

I missed this part in the doc, that's a bummer. There's no way to have two custom events on a same poll with mio 0.7 then?

Thomasdezeeuw:
I'm afraid not. `Waker` is really designed to just wake a thread from polling, with the idea being that there is a user space queue (outside of Mio) to which the events are added.

I don't know what your use case is, but I generally recommend using a [channel](https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html) of some kind to create a user space queue, followed by calling `Waker::wake` to ensure the polling thread gets the events. When the polling thread then receives the `Token` registered with the `Waker` poll the user space queue and handle the events.

Keruspe:
Yeah, I dealt with that differently in the end.

Sorry for the noise because of missing this point in the doc. I wonder if it could be made more explicit in the API.

I guess we can close this if you don't pan on supporting having multiple wakers per poll

Thomasdezeeuw:
I wouldn't know how to make it more explicit in the API without complicating it needlessly. We could add a "status" generic parameters, something like:

```rust
struct Poll<S> {
    // ..
}

enum NoWaker { }
enum HasWaker { }

impl Poll<NoWaker> {
    fn new( /* ... */ ) -> Poll<NoWaker> {
        // ..
    }

    // Can only call this we we don't have a `Waker` yet.
    fn new_waker(self, /* ... */) -> io::Result<(Poll<HasWaker>, Waker)> {
        // ...
    }
}
```

But that would make it more complex, don't know if that is worth it.

Keruspe:
Well, could be simpler: `Poll::new_waker(&self, token: Token) -> Option<Waker>` returning `Some` the first time and `None` the next ones. Or maybe a result with a specific error

Thomasdezeeuw:
True, but then we would have to keep track of that somewhere which would cost a byte.

PR #1285, opened 2020-03-22
Fix TCP example

### LINKED ISSUE #1284: v0.7 is broken on Windows?
It seems the tcp_server example can only accept one client.

Reproduce steps:

1 compile and run the tcp_server example:
https://github.com/tokio-rs/mio/blob/master/examples/tcp_server.rs

2 compile the connect example in tokio:
https://github.com/tokio-rs/tokio/blob/master/examples/connect.rs

3 run connect.exe twice

result: only the connecttion from first connect.exe is accecpted.

If the tcp_server is ported back to v0.6.0, there is no such problem.

Thomasdezeeuw:
This now responds to edge-trigger events properly.

Fixes #1284

/cc @cg31

### DIFF
--- a/examples/tcp_server.rs
+++ b/examples/tcp_server.rs
@@ -43,10 +43,25 @@ fn main() -> io::Result<()> {
 
         for event in events.iter() {
             match event.token() {
-                SERVER => {
-                    // Received an event for the TCP server socket.
-                    // Accept an connection.
-                    let (mut connection, address) = server.accept()?;
+                SERVER => loop {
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
CI is failing because `armv7s-apple-ios` (iOS on arm v7) and `i386-apple-ios` (iOS 32bit) are no longer supported by rustc/rustup.

kleimkuhler:
Once this gets master merged in, should be good to go

PR #1287, opened 2020-03-23
Remove 32bit iOS and replace ARM with ARM 64 bit iOS in CI

Thomasdezeeuw:
Both i386 and armv7s iOS targets are no longer supported by Rust (or
Apple for the matter), so remove them both from the CI.

And use the aarch64 (ARM 64 bit) iOS target instead.

### DIFF
--- a/ci/azure-cross-compile.yml
+++ b/ci/azure-cross-compile.yml
@@ -10,15 +10,11 @@ jobs:
           vmImage: macOS-10.13
           target: x86_64-apple-ios
 
-        iOS_32:
+        iOS_ARM64:
           vmImage: macOS-10.13
-          target: i386-apple-ios
+          target: aarch64-apple-ios
 
-        iOS_ARM:
-          vmImage: macOS-10.13
-          target: armv7s-apple-ios
-
-        Android:
+        Android_ARM:
           vmImage: ubuntu-16.04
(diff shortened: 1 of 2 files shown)


kleimkuhler:
Tokio had to bump MacOS versions recently to 10.14 as well. Looks like same needs to happen here.

ci/azure-cross-compile.yml
```diff
@@ -7,18 +7,14 @@ jobs:
     strategy:
       matrix:
         iOS_64:
-          vmImage: macOS-10.13
+          vmImage: macOS-10.15
```
kleimkuhler: Any reason this is `10.15` and not `10.14`? I think we should keep this in line with [Tokio's versions](https://github.com/djc/tokio/blob/cadbedc3d633f24c069e4a8ef36108148d145ec7/ci/azure-check-features.yml#L8-L9)

Thomasdezeeuw: No reason, 10.15 is just latest so we don't have to update next year. If Tokio tests on 10.14 and we test on 10.15 then we got both covered. But I don't really have a strong preference either way, so if you want me to change it I'm ok with it.

kleimkuhler: I'm okay with that.

PR #1290, opened 2020-03-27
Enable all mio features in Rust Playground

peterjoel:
Some common mio examples do not currently work in Rust playground (https://play.rust-lang.org/), because modules like `net` are not available with default features.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,9 +52,7 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [package.metadata.playground]
-default-features = true
 features = ["os-poll", "os-util", "tcp", "udp", "uds"]
-all-features = false
 
 [[example]]
 name = "tcp_server"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,11 @@ net2       = "0.2.33"
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
+[package.metadata.playground]
+default-features = true
+features = ["os-poll", "os-util", "tcp", "udp", "uds"]
+all-features = false
+
 [[example]]
 name = "tcp_server"
 required-features = ["os-poll", "tcp"]


Keruspe:
Shouldn't this be in the features section rather than in the examples one?

peterjoel:
> Shouldn't this be in the features section rather than in the examples one?

Yes, indeed. I will change that.

Thomasdezeeuw:
I have no idea if this would work or not, is there any documentation for the playground? Couldn't it just use `--all-features`? That would make it easier to maintain for us, compared to having to keep this feature up-to-date (enabling all other features).

peterjoel:
@Thomasdezeeuw https://github.com/integer32llc/rust-playground/issues/192#issuecomment-318836449

Thomasdezeeuw:
Looking at https://github.com/integer32llc/rust-playground/issues/192#issuecomment-388589736, could we do the following?

```toml
# Enable all features on the playground.
[package.metadata.playground]
all-features = true
```

Do we have any way to to test this, besides publishing a new version and trying?

peterjoel:
@Thomasdezeeuw I will have a go at building and running the playground locally.

shepmaster:
> Couldn't it just use --all-features

In general, no, as it’s possible to have conflicting sets of features or [features that require nightly](https://github.com/chronotope/chrono/issues/373).  Of course, a crate can opt in to all features if it knows that’s feasible for itself.

Cargo.toml
```diff
@@ -51,6 +51,11 @@ net2       = "0.2.33"
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
+[package.metadata.playground]
+default-features = true
+features = ["os-poll", "os-util", "tcp", "udp", "uds"]
+all-features = false
```
kleimkuhler: How about just specifying the features needed?

```suggestion
[package.metadata.playground]
features = ["os-poll", "os-util", "tcp", "udp", "uds"]
```

`all-features` is `false` by default, so I don't think we need to explicitly set that.

Additionally, Mio has no default features, so `default-features = true` is not adding any features.

Thomasdezeeuw:
Thanks!

PR #1292, opened 2020-03-28
Add test for sending empty datagram

kleimkuhler:
macos 10.15.4, fixed a bug in kqueue which would cause this test to hang. Now
that is fixed, this test ensures the behavior remains correct for all targets.

### DIFF
--- a/tests/udp_socket.rs
+++ b/tests/udp_socket.rs
@@ -29,6 +29,39 @@ const ID2: Token = Token(3);
 const ID3: Token = Token(4);
 
 #[test]
+fn empty_datagram() {
+    const EMPTY: &[u8] = b"";
+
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
Since 10.15.4 is very new (I don't even have it installed yet) let's wait a bit before merging. It also seems the CI doesn't have the update yet.

Thomasdezeeuw (approved):

Thomasdezeeuw:
macOS 10.15.4 is old enough (10.15.7 being the current version), I think this can be safely merged. I tried to restart the CI, but I don't think it actually did restart.

Thomasdezeeuw:
This got merged in #1392.

PR #1294, opened 2020-04-14
Add support for illumos target (0.6.x)

pfmooney:
With support for an illumos target [on its way](https://github.com/rust-lang/rust/pull/71145) into rust, the cfg guards for `solaris` should be updated to include `illumos` as well.

PR #1295, opened 2020-04-15
Add support for illumos target

pfmooney:
As the master branch counterpart to the 0.6.x PR (#1294), this adds support for the Solaris-like (but distinct) `illumos` target.  When support is [merged](https://github.com/rust-lang/rust/pull/71145) into rust, it would be nice for mio to be immediately usable.

### DIFF
--- a/src/sys/unix/net.rs
+++ b/src/sys/unix/net.rs
@@ -27,6 +27,7 @@ pub(crate) fn new_socket(
         target_os = "android",
         target_os = "dragonfly",
         target_os = "freebsd",
+        target_os = "illumos",
         target_os = "linux",
         target_os = "netbsd",
         target_os = "openbsd"
--- a/src/sys/unix/selector/epoll.rs
+++ b/src/sys/unix/selector/epoll.rs
@@ -21,10 +21,15 @@ pub struct Selector {
 
 impl Selector {
     pub fn new() -> io::Result<Selector> {
-        // According to libuv `EPOLL_CLOEXEC` is not defined on Android API <
-        // 21. But `EPOLL_CLOEXEC` is an alias for `O_CLOEXEC` on all platforms,
-        // so we use that instead.
-        syscall!(epoll_create1(libc::O_CLOEXEC)).map(|ep| Selector {
+        // According to libuv, `EPOLL_CLOEXEC` is not defined on Android API <
+        // 21. But `EPOLL_CLOEXEC` is an alias for `O_CLOEXEC` on that platform,
+        // so we use it instead.
+        #[cfg(target_os = "android")]
+        let flag = libc::O_CLOEXEC;
+        #[cfg(not(target_os = "android"))]
+        let flag = libc::EPOLL_CLOEXEC;
+
+        syscall!(epoll_create1(flag)).map(|ep| Selector {
             #[cfg(debug_assertions)]
             id: NEXT_ID.fetch_add(1, Ordering::Relaxed),
             ep,
@@ -212,6 +217,7 @@ pub mod event {
     }
 }
 
+#[cfg(target_os = "android")]
 #[test]
 fn assert_close_on_exec_flag() {
     // This assertion need to be true for Selector::new.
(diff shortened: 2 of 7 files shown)


src/sys/unix/selector/epoll.rs
```diff
@@ -212,6 +217,7 @@ pub mod event {
     }
 }
 
+#[cfg(target_os = "linux")]
```
Thomasdezeeuw, line 220: This needs to be `android`.

src/sys/unix/net.rs
```diff
@@ -29,7 +29,8 @@ pub(crate) fn new_socket(
         target_os = "freebsd",
         target_os = "linux",
         target_os = "netbsd",
-        target_os = "openbsd"
+        target_os = "openbsd",
+        target_os = "illumos"
```
Thomasdezeeuw, line 33: Nit: can you sort these alphabetically, same for all places you've added `illumos` target.

src/sys/unix/selector/epoll.rs
pfmooney: Since it was on Android (for older versions) where `EPOLL_CLOEXEC` is missing, and only on Linux where this constant equality is assured, I thought it made more sense to perform the test on the platform where it was guaranteed to have the proper context.

src/sys/unix/net.rs
pfmooney: Yep.  I'll fix this one and the others.

src/sys/unix/selector/epoll.rs
Thomasdezeeuw: The point of the test was to ensure that `O_CLOEXEC` equals `EPOLL_CLOEXEC` on all platforms, as I couldn't find any documentation that those constants were guaranteed to be equal. But with this change we don't use `O_CLOEXEC` any more on all platforms except for Android, so we still need to check it on later versions of Android that this assertion is true.

pfmooney: Fixed

pfmooney:
> Does this need a bump in `libc` version? I can image there will be changes required to that crate as well.

Yep, fixed.

> Also we would need CI support, running at least `cargo check`, for Illumos to be supported by Mio.

I'm waiting on the [PR](https://github.com/rust-lang/rust/pull/71145) to add support for the target into rustc right now.  I have tested locally with a toolchain built from that to ensure that the results compile (and pass at least as many tests as the `x86_64-sun-solaris` target does).  Considering that the `illumos` triple is going to effectively subsume many (most?) instances of the `solaris` one, we wanted to get ahead of it so popular crates like mio will be supported ASAP once support shows up in nightly builds.

src/sys/unix/selector/epoll.rs
```diff
@@ -212,6 +217,7 @@ pub mod event {
     }
 }
 
+#[cfg(target_os = "android")]
```
kleimkuhler, line 220: Any reason this was added? I think this test should still be okay for other targets.

pfmooney: Since it's only on android where that define was missing?  Using `O_CLOEXEC` in the place of `EPOLL_CLOEXEC` works on Linux, but asserting that they be equal elsewhere doesn't seem valuable unless that substitution is necessary.  In the case of illumos, the two are not equal.

kleimkuhler: Okay makes sense. This test seems a little odd in that if we can compare it with `EPOLL_CLOEXEC`, we might as well use it in `Selector::new`, but I don't think it needs to be removed or anything as part of this.

Thomasdezeeuw: @kleimkuhler I asked for this, see https://github.com/tokio-rs/mio/pull/1295#discussion_r408866646.

pfmooney:
> This looks good once it is rebased.

Rebased, with the conflicts addressed.

Thomasdezeeuw:
@pfmooney thanks!

PR #1300, opened 2020-04-20
Update FreeBSD version in CI

kleimkuhler:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,8 +1,8 @@
 freebsd_instance:
-  image: freebsd-11-2-release-amd64
+  image: freebsd-12-1-release-amd64
 
 task:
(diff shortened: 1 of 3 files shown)


asomers:
This should be ok.  But there's a risk that we accidentally begin to depend on a feature from FreeBSD 12.0 or 12.1, and mio breaks on 11.3.  To guarantee backwards-compatibility, it would be better to do
```yaml
  image_family: freebsd-11-3-snap
```

kleimkuhler:
This bumps the minimum MacOS version from 10.13 to 10.14 as well so that CI starts passing again. Those images are no longer available on Azure. Similar to the FreeBSD version bump, these are minor bumps so that should be fine in order to get `v0.6.x` branch passing again.

PR #1301, opened 2020-04-20
Change FreeBSD CI name

kleimkuhler:
These versions don't get bumped that often, but it's been a bit of a pain changing them and having to update the name of the required FreeBSD check. Now that the required check is just `FreeBSD`, this updates the name to reflect that.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,7 +5,7 @@ env:
   RUST_BACKTRACE: full
 
 task:
-  name: FreeBSD 12.1 amd64
+  name: FreeBSD
   setup_script:
     - pkg install -y curl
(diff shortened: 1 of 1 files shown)


asomers:
I just realized that the 0.6.x branch is testing on 11.3-stable, whereas master is testing on 12.1-release.  Shouldn't we be testing the same OS version of both branches?

kleimkuhler:
The reason I bumped #1300 back to 11.3 is because it is the `v0.6.x` branch. I think we should keep version bumps as minimal as possible in order for CI to pass.

`master` being on 12.1 lines up with [tokio](https://github.com/tokio-rs/tokio/blob/master/.cirrus.yml#L2) so I think this is fine as is.

PR #1304, opened 2020-04-24
Windows: Fix afd::POLL_ABORT and POLL_CONNECT_FAIL handling

### LINKED ISSUE #1303: afd::POLL_ABORT mapping
I noticed that this is mapped as `write_closed` in mio. Should this also trigger `read_closed`?

I observe that [wepoll maps it as EPOLLHUP](https://github.com/piscisaureus/wepoll/blob/8128da416cbfb67602dcc0d352b0da8bca3c93d0/wepoll.c#L1766) and that [libuv maps it as "readable"](https://github.com/libuv/libuv/blob/16910d23a34b8b8c313d24824fd85383aa8a8d40/src/win/poll.c#L192).

goffrie:
This does two things:
1. POLL_ABORT is considered both read_closed and write_closed, much like EPOLLHUP.
2. POLL_ABORT and POLL_CONNECT_FAIL events are properly handled by the edge-triggered-event emulation.

Fixes #1303.

### DIFF
--- a/src/sys/windows/selector.rs
+++ b/src/sys/windows/selector.rs
@@ -1,6 +1,9 @@
 use super::afd::{self, Afd, AfdPollInfo};
 use super::io_status_block::IoStatusBlock;
 use super::Event;
+use crate::sys::event::{
+    ERROR_FLAGS, READABLE_FLAGS, READ_CLOSED_FLAGS, WRITABLE_FLAGS, WRITE_CLOSED_FLAGS,
+};
 use crate::sys::Events;
 use crate::Interest;
 
@@ -719,12 +722,11 @@ fn interests_to_afd_flags(interests: Interest) -> u32 {
     let mut flags = 0;
 
     if interests.is_readable() {
-        // afd::POLL_DISCONNECT for is_read_hup()
-        flags |= afd::POLL_RECEIVE | afd::POLL_ACCEPT | afd::POLL_DISCONNECT | afd::POLL_ABORT;
+        flags |= READABLE_FLAGS | READ_CLOSED_FLAGS | ERROR_FLAGS;
     }
 
     if interests.is_writable() {
-        flags |= afd::POLL_SEND | afd::POLL_ABORT;
+        flags |= WRITABLE_FLAGS | WRITE_CLOSED_FLAGS | ERROR_FLAGS;
     }
 
     flags
--- a/src/sys/windows/selector.rs
+++ b/src/sys/windows/selector.rs
@@ -720,11 +720,11 @@ fn interests_to_afd_flags(interests: Interest) -> u32 {
 
     if interests.is_readable() {
         // afd::POLL_DISCONNECT for is_read_hup()
-        flags |= afd::POLL_RECEIVE | afd::POLL_ACCEPT | afd::POLL_DISCONNECT;
+        flags |= afd::POLL_RECEIVE | afd::POLL_ACCEPT | afd::POLL_DISCONNECT | afd::POLL_ABORT;
     }
 
     if interests.is_writable() {
-        flags |= afd::POLL_SEND;
+        flags |= afd::POLL_SEND | afd::POLL_ABORT;
     }
 
     flags
--- a/tests/tcp.rs
+++ b/tests/tcp.rs
@@ -574,6 +574,8 @@ fn connect_error() {
     }
 
     assert!(l.take_error().unwrap().is_some());
+
+    expect_no_events(&mut poll, &mut events);
 }
 
 #[test]
--- a/tests/tcp_stream.rs
+++ b/tests/tcp_stream.rs
@@ -8,6 +8,7 @@ use std::net::{self, Shutdown, SocketAddr};
 use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd};
 use std::sync::{mpsc::channel, Arc, Barrier};
 use std::thread;
+use std::time::Duration;
 
 #[macro_use]
 mod util;
@@ -641,7 +642,14 @@ fn tcp_reset_close_event() {
         vec![ExpectEvent::new(ID1, Readiness::READ_CLOSED)],
     );
 
-    expect_no_events(&mut poll, &mut events);
+    // Make sure we quiesce. `expect_no_events` seems to flake sometimes on mac/freebsd.
+    loop {
+        poll.poll(&mut events, Some(Duration::from_millis(100)))
+            .expect("poll failed");
+        if events.iter().count() == 0 {
+            break;
+        }
+    }
 }
 
 #[test]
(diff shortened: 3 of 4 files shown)


tests/tcp.rs
```diff
@@ -574,6 +574,8 @@ fn connect_error() {
     }
 
     assert!(l.take_error().unwrap().is_some());
+
+    expect_no_events(&mut poll, &mut events);
```
goffrie, line 578: This was previously failing because we did not consume POLL_CONNECT_FAIL.

tests/tcp_stream.rs
```diff
@@ -603,6 +603,42 @@ fn tcp_shutdown_server_write_close_event() {
...
+    // Server goes away.
+    drop(server_stream);
+
+    expect_events(
+        &mut poll,
+        &mut events,
+        vec![ExpectEvent::new(ID1, Readiness::READ_CLOSED)],
+    );
+
+    expect_no_events(&mut poll, &mut events);
```
goffrie, line 639: This was previously failing because we did not consume POLL_ABORT.

Thomasdezeeuw:
I'm not sure about this fix.

@piscisaureus what do you think?

goffrie:
> I'm not sure about this fix.

What do you find worrying about it?

Thomasdezeeuw:
I don't know if `POLL_ABORT` should trigger `is_read_closed`, that why I'm asking @piscisaureus he knows way more about AFD than I do.

piscisaureus:
This patch looks correct to me.

> I don't know if POLL_ABORT should trigger is_read_closed, that why I'm asking

I think ABORT and CONNECT_FAIL should more or less trigger every event. The point is to get the user to make an attempt to read/write the socket. When they do they'll observe the actual error.

goffrie:
> perhaps not RDHUP (read_close) as this indicates a graceful half-close which allows the user, as an optimization, to _skip_ a read() call.

I see; should it be `is_error` instead then? Or `is_readable`?

piscisaureus:
> I see; should it be is_error instead then? Or is_readable?

I think that would be better.

PS: I'm applying what I know about the semantics behind epoll flags. Mio's semantics might be subtly different.

That said, ABORT and CONNECT_FAIL are both error conditions and should be treated as such.

Thomasdezeeuw:
> PS: I'm applying what I know about the semantics behind epoll flags. Mio's semantics might be subtly different.
> 
> That said, ABORT and CONNECT_FAIL are both error conditions and should be treated as such.

Would the error be returned using `take_error` (`getsockopt(SOL_SOCKET, SO_ERROR)`)? If that is the case I would say it should trigger `is_error`.

piscisaureus:
> Would the error be returned using `take_error` (`getsockopt(SOL_SOCKET, SO_ERROR)`)? If that is the case I would say it should trigger `is_error`.

I think SO_ERROR is only used for non-blocking connect() failures, but I'm not 100% sure to be honest. AFD_POLL_ABORT is semantically closest to EPOLLHUP so let's just do what epoll.rs does in practice.

That means:
- AFD_POLL_CONNECT_FAIL -> `readable`, `writable`, `read_closed`, `write_closed`, `error`. These mappings are not all explicitly stated in epoll.rs, but linux reports [practically all events](https://github.com/piscisaureus/wepoll/blob/8128da416cbfb67602dcc0d352b0da8bca3c93d0/wepoll.c#L1769-L1771) when connect() fails.
- AFD_POLL_ABORT -> `readable`, `writable`, `read_closed`, `write_closed`. 

I personally do not agree with reporting `read_closed` when you detect a non-graceful disconnect, but epoll.rs does it anyway so let's not have that discussion now.

goffrie:
Updated per discussion.

piscisaureus:
LGTM

src/sys/windows/selector.rs
```diff
@@ -719,12 +722,11 @@ fn interests_to_afd_flags(interests: Interest) -> u32 {
     let mut flags = 0;
 
     if interests.is_readable() {
-        // afd::POLL_DISCONNECT for is_read_hup()
-        flags |= afd::POLL_RECEIVE | afd::POLL_ACCEPT | afd::POLL_DISCONNECT;
+        flags |= READABLE_FLAGS | READ_CLOSED_FLAGS | ERROR_FLAGS;
```
Thomasdezeeuw, line 725: Since both `epoll` and `kqueue` always register error flags (e.g. `EPOLLERR`) should we just do the same for Windows? /cc @piscisaureus

piscisaureus: It looks like that's getting done right?

piscisaureus:
> Does the `tcp_shutdown_client_read_close_event` (tests/tcp_stream.rs:506) test pass now as well?

It bothers me a bit that PRs in this repo move at an absolute glacial pace and there's always another sticking point or another nitpick.

Don't take it too personal please, I understand that y'all (plural) are doing your best, but this is one such occasion where you can do that 1 minute of work and enable that test yourself, or letting this PR drag on for another week.

piscisaureus:
^-- there's the answer

goffrie:
I undid the above commit since it failed, but indeed tcp_shutdown_client_read_close_event still does not pass as-is on Windows.

Thomasdezeeuw:
> > Does the `tcp_shutdown_client_read_close_event` (tests/tcp_stream.rs:506) test pass now as well?
> 
> It bothers me a bit that PRs in this repo move at an absolute glacial pace and there's always another sticking point or another nitpick.
> 
> Don't take it too personal please, I understand that y'all (plural) are doing your best, but this is one such occasion where you can do that 1 minute of work and enable that test yourself, or letting this PR drag on for another week.

I get your point, but I generally don't have access to a Windows machine making that "1 minute of work" not quite a minute long. And as a side note you can approve and merge PRs yourself, you requested my review and that is something I noted.

Thomasdezeeuw:
FreeBSD CI seems to be failing on a unrelated test, but I don't have time to look into right now.

```log
---- tcp_reset_close_event stdout ----
[ERROR tcp_stream::util] unexpected event: Event { token: Token(0), readable: true, writable: false, error: true, read_closed: true, write_closed: false, priority: false, aio: false, lio: false }
[ERROR tcp_stream::util] unexpected event: Event { token: Token(0), readable: false, writable: true, error: true, read_closed: false, write_closed: true, priority: false, aio: false, lio: false }
thread 'tcp_reset_close_event' panicked at 'received events, but didn't expect any, see above', tests/util/mod.rs:172:9
```

goffrie:
That's a test I added in this PR. It seems to be a bit flaky, so I adjusted the test to work around it.

Thomasdezeeuw:
@goffrie thanks for the pr. @piscisaureus thanks for the review.

goffrie:
Thanks!

PR #1305, opened 2020-04-24
implement event::Source on &IoSource and &TcpStream

Keruspe:
As of now, it's implemented for IoSource and TcpStream but not for their references. The only operation on the underlying TcpStream from the stdlib is as_raw_fd/as_raw_socket, iiuc, which does not require a mutable state.
Having event::Source implemented on &TcpStream would allow registerig a TcpStream without it being mut. Given Read and Write are also implemented on &TcpStream, I think this would make sense?

### DIFF
--- a/src/io_source.rs
+++ b/src/io_source.rs
@@ -232,41 +232,6 @@ where
     }
 }
 
-#[cfg(windows)]
-impl<T> event::Source for &IoSource<T>
-where
-    T: AsRawSocket,
-{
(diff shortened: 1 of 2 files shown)


Keruspe:
Since it cannot currenlty work on windows, this makes it way less useful.

Thomasdezeeuw:
@Keruspe you've found the reason why it was removed in v0.7 :)

PR #1309, opened 2020-04-30
draft: Add additional write without reregistering

kleimkuhler:
I do not have a Windows env easily accessible right now and am using CI for testing Windows

### DIFF
--- a/tests/udp_socket.rs
+++ b/tests/udp_socket.rs
@@ -334,6 +334,8 @@ fn smoke_test_connected_udp_socket(mut socket1: UdpSocket, mut socket2: UdpSocke
     expect_read!(socket1.recv(&mut buf), DATA2);
     expect_read!(socket2.recv(&mut buf), DATA1);
 
+    assert_would_block(socket2.recv(&mut buf));
+
     checked_write!(socket1.send(DATA1));
     expect_events(
         &mut poll,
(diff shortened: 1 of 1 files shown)


tests/udp_socket.rs
```diff
@@ -334,6 +334,14 @@ fn smoke_test_connected_udp_socket(mut socket1: UdpSocket, mut socket2: UdpSocke
     expect_read!(socket1.recv(&mut buf), DATA2);
     expect_read!(socket2.recv(&mut buf), DATA1);
 
+    checked_write!(socket1.send(DATA1));
+    expect_events(
```
Thomasdezeeuw, line 338: I think this is because we didn't hit `WouldBlock` before reading from socket2, which is a requirement for re-registering: https://github.com/tokio-rs/mio/blob/976f2354d0e8fbbb64fba3bf017d7131f9c369a0/src/sys/windows/mod.rs#L81-L86

Could you try adding a `socket2.recv` call that expects `WouldBlock` to test this guess?

kleimkuhler: Hm yea good point I can try that.

Thomasdezeeuw:
Seems like not hitting `WouldBlock` is the problem.

If I remember correctly @dtacalau and I had this conversion for Windows before. Basically with `epoll(2)`/`kqueue(2)` reading all data will re-arm the edge trigger, however Mio requires (per the documentation) the user to hit `WouldBlock` first. So on Windows we decided to only re-arm when `WouldBlock` was hit to not double re-arm it (i.e. when reading didn't fill the entire buffer -> re-arm and then the user would read again, hit `WouldBlock` and re-arm it again).

kleimkuhler:
@Thomasdezeeuw Yep I remember that conversation now. It looks like this is handled properly in the `udp_server` example, but it is where this was originally reported from.. can double check with this user.

PR #1311, opened 2020-04-30
Cargo.toml: exclude more CI files

elmarco:


Thomasdezeeuw:
Maybe we should use include instead, something like `include = ["/Cargo.toml", "/src/**/*.rs", "/README.md", "/LICENSE"]`?

elmarco:
I leave that decision up to the maintainers

Thomasdezeeuw:
> I leave that decision up to the maintainers

Well that would be me :)

I've tested it locally and we need the examples as well:

```toml
include       = ["/Cargo.toml", "/src/**/*.rs", "/examples", "/README.md", "/LICENSE"]
```

@carllerche @kleimkuhler any opinions about excluding tests and additional files from the package uploaded to crates.io?

Thomasdezeeuw:
@asomers `Cargo.toml` has two [`[[example]]` sections](https://github.com/tokio-rs/mio/blob/976f2354d0e8fbbb64fba3bf017d7131f9c369a0/Cargo.toml#L57-L63) and cargo commands will fail if the files are missing, i.e. `cargo build` fails with:

```
error: failed to parse manifest at `mio/Cargo.toml`

Caused by:
  can't find `tcp_server` example, specify example.path
```

kleimkuhler:
I don't have a major preference on `exclude` vs `include`. If we go with `exclude` and start excluding tests, is that common or necessary? I would assume the `#[test]` attribute already excludes in publishing, so doing it on `tests/` may not be necessary?

Thomasdezeeuw:
The `exclude`/`include` attribute determines what will be packaged and shipped to `crates.io` and end-users. Its not really critical but in most cases you really only need the source code and license. Removing the tests and other files makes the package smaller.

elmarco:
It also makes packaging for distros a bit easier. Fedora packaging is what prompted me to make this patch.

kleimkuhler:
Sounds good. Again I don't have a major preference; I lean towards the `include` that @Thomasdezeeuw suggested above, but would also be good with the current change!

Thomasdezeeuw:
@elmarco can you this to match https://github.com/tokio-rs/mio/commit/6df2fb1abb91488dd6d527c43d58216b8f6e2347?

PR #1315, opened 2020-05-05
Windows: Implement a workaround for non-AFD sockets.

### LINKED ISSUE #1314: Non-AFD fallback?
Would it be in scope for mio to have a [fallback like libuv does](https://github.com/libuv/libuv/blob/16910d23a34b8b8c313d24824fd85383aa8a8d40/src/win/poll.c#L561) for non-MSAFD sockets on Windows?

(I ask because we are trying to deploy some software that uses mio 0.7 and a very significant number of users are hitting errors in registration, presumably due to this, so we're going to need a fallback at some layer no matter what. Mio 0.6 actually does work for these users, but it has other issues.)

goffrie:
In some system configurations the AFD apis don't work. Fall back to
spawning a new thread for each registration and using
WSAWaitForMultipleEvents.

I tested this (i.e. ran the test suite) in a Windows 7 VM with some
network intercepting drivers installed. Without this change, registrations
fail with the error

```
The system detected an invalid pointer address in attempting to use a pointer argument in a call. (os error 10014)
```

(The specific software I installed was Qustodio, but I can't vouch for it being safe or useful for anything else.)

Closes #1314.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ libc = "0.2.62"
 
 [target.'cfg(windows)'.dependencies]
 miow   = "0.3.3"
-winapi = { version = "0.3", features = ["winsock2", "mswsock"] }
+winapi = { version = "0.3", features = ["winsock2", "mswsock", "synchapi"] }
 ntapi  = "0.3"
 lazy_static = "1.4.0"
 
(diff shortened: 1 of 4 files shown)


goffrie:
I'm not sure what to do about testing this codepath in CI, maybe a `--cfg`?

goffrie:
Closing this per discussion in #1314

PR #1316, opened 2020-05-09
On windows, try some alternative IOCTLs if SIO_BASE_HANDLE fails

piscisaureus:
Fixes: #1314

### DIFF
--- a/src/sys/windows/selector.rs
+++ b/src/sys/windows/selector.rs
@@ -527,7 +527,8 @@ impl SelectorInner {
 cfg_net! {
     use std::mem::size_of;
     use std::ptr::null_mut;
-    use winapi::um::mswsock::SIO_BASE_HANDLE;
+    use winapi::um::mswsock;
+    use winapi::um::winsock2::WSAGetLastError;
     use winapi::um::winsock2::{WSAIoctl, SOCKET_ERROR};
 
     impl SelectorInner {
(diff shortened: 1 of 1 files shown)


piscisaureus:
Closing & reopening in an attempt to get CI status reporting out of its rut..

src/sys/windows/selector.rs
```diff
@@ -623,29 +624,60 @@ cfg_net! {
...
         }
     }
+
+    fn get_base_socket(raw_socket: RawSocket) -> io::Result<RawSocket> {
+        let res = try_get_base_socket(raw_socket, mswsock::SIO_BASE_HANDLE);
+        if let Ok(base_socket) = res {
+            return Ok(base_socket);
+        }
+
+        // The `SIO_BASE_HANDLE` should not be intercepted by LSPs, therefore
```
Thomasdezeeuw, line 656: Nit: I don't know what "LSP" is in this context, but if its a well known term you can ignore this comment.

piscisaureus:
> LGTM, is there any way we can add a test for this?

The answer is ... it's complicated. 

In theory, yes; you'd have to add a test job where you install the qustodio LSP (without user interaction) before running the tests. I do it for wepoll with two different LSPs: https://github.com/piscisaureus/wepoll/blob/4fe5f50f22681ba680ff44df4152ed3296a904f3/.appveyor.yml#L85-L95. Those LSPs don't need this fix BTW so no sense in adding a CI job with one of them.

But the short answer is that it's gonna be really finicky to get it to work. Let's see if we can first get some feedback from @goffrie.

src/sys/windows/selector.rs
piscisaureus: Layered Service Provider
https://en.wikipedia.org/wiki/Layered_Service_Provider
Good luck, you still don't know what it is 😉

PR #1321, opened 2020-05-14
Set `SO_NOSIGPIPE` on all sockets on Apple systems.

twittner:
Previously this option was only set for UDP sockets but is equally important for TCP. [Like `libstd`](https://github.com/rust-lang/rust/blob/7c34d8d6629506a596215886e5fc4bb2b04b00ae/src/libstd/sys/unix/net.rs#L79) this PR enables the option on all sockets if compiled for Apple systems.


Thomasdezeeuw:
Thanks @twittner.

PR #1325, opened 2020-05-25
RUSTSEC-2020-0016: Replace net2 with socket2

smklein:
`cargo audit` now passes without error.


Thomasdezeeuw:
Socket2 isn't ready for production, even if its only used in a single test, e.g. see https://github.com/alexcrichton/socket2-rs/pull/81 (merged 4 days as of today). I'm personally working on it, but it will take some time as it isn't very high priority for me at the moment.

PR #1329, opened 2020-06-12
Add a check that a single Waker is active per Poll instance

Thomasdezeeuw:
Ensuring the API is used properly (at least during tests).

See #1283

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -613,6 +613,15 @@ impl Registry {
             .try_clone()
             .map(|selector| Registry { selector })
     }
+
+    /// Internal check to ensure only a single `Waker` is active per [`Poll`]
+    /// instance.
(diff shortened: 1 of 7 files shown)


Thomasdezeeuw:
It seems that `panic!`king on FreeBSD causes `SIGSEGV: invalid memory reference` (see https://cirrus-ci.com/task/6501661407444992?command=amd64_test#L265, I've also tried it locally on a FreeBSD machine). @asomers what do you think the best way to proceed is? Replace the call to `panic!` with `log::error!` on FreeBSD, ignoring the test?

Thomasdezeeuw:
Rebased this on latest master.

Darksonn (approved):
Seems correct to me.

Issue #1324, opened 2020-05-25
Replace net2 with socket2


petreeftime:
net2 is marked as deprecated and the suggestion is to use the socket2 crate instead. `cargo audit` in particular will complain about net2 dependency.

petreeftime:
It seems this is duplicated.

Thomasdezeeuw:
See #1319, #1322.

smklein:
If I submit a PR that replaces net2 with socket2 (it's compiling; passing all tests), will that be looked at?

(I understand the "if it ain't broke, don't fix it" mentality, but I'm just checking if there is opposition to this change)

smklein:
https://github.com/tokio-rs/mio/pull/1325

Issue #1319, opened 2020-05-10
switch from net2 to socket2


kallisti5:
net2 is now deprecated.   mio should switch to socket2 which is supported.

Thomasdezeeuw:
Mio doesn't use net2 anymore.

kallisti5:
https://github.com/tokio-rs/mio/blob/40df934a11b05233a7796c4de19a4ee06bc4e03e/Cargo.toml#L48 ??

Thomasdezeeuw:
I know its only used in a single test: https://github.com/tokio-rs/mio/blob/0ba2ed8bb1e3af999024653d587b9c25f8d4b57c/tests/tcp.rs#L472

So its not really a high priority.

kallisti5:
Ok. So all of net2 which is broken on some platforms gets pulled in and compiled by every crate which depends on tokio and mio for a single test.

Thomasdezeeuw:
No it’s a development dependency which means it’s only used when run ‘cargo test’ in the Mio crate

kallisti5:
weird... the Cargo.lock tells another story.

```
[[package]]
name = "mio"
version = "0.6.22"
source = "registry+https://github.com/rust-lang/crates.io-index"
dependencies = [
 "cfg-if 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
 "fuchsia-zircon 0.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
 "fuchsia-zircon-sys 0.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
 "iovec 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
 "kernel32-sys 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
 "libc 0.2.69 (registry+https://github.com/rust-lang/crates.io-index)",
 "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
 "miow 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
 "net2 0.2.34 (registry+https://github.com/rust-lang/crates.io-index)",
 "slab 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
 "winapi 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
]
```

Keruspe:
@kallisti5 That's mio 0.6, latest (and fixed) version is 0.7

[...]

PR #1330, opened 2020-06-20
Replace deprecated dev-dependency net2 by socket2

shenek:
net2 crate was officially deprecated in favor of socket2

see https://rustsec.org/advisories/RUSTSEC-2020-0016.html

This patch replaces the dependency and fixes the test.


Thomasdezeeuw:
See #1325, #1324,  #1322 and #1319.

PR #1342, opened 2020-07-28
net2 -> socket2

meehow:
https://rustsec.org/advisories/RUSTSEC-2020-0016


Thomasdezeeuw:
Please use the search function: #1330 and many more.

PR #1345, opened 2020-08-11
Add API to hook into Windows IOCP loop

darkwisebear:
An API similar to the one of mio 0.6 is now available so that custom objects can be added to the IOCP that is used to implement Poll. Relates to https://github.com/tokio-rs/mio/issues/1047.

This change isn't complete yet but in a draft state as I wanted to gather some feedback on the suggested API.

Since sockets are polled by using a single connection to afd.sys, there is no 1:1 relationship between a user-provided token and the completion key of the IOCP. The implementation therefore uses a slab to keep track of the registered handlers and the mapping of the user-provided token to the reported events.

Other notable things about the change:
* The Overlapped type now uses a boxed trait object instead of a function pointer. While this is resource-heavier, it also enables writing more clean completion handlers as they can now directy carry state instead of having to do some pointer magic or accessing some sort of global state.
* There is anew type "Readiness" to convert the completion events into polling events. There is also a similar (equally named) type inside the tests and maybe this type is misplaced inside the Windows module.
* I still need to write more tests and add them to this PR before it can be accepted. Working on it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -89,6 +89,68 @@ pub mod unix {
     pub use crate::sys::SourceFd;
 }
 
+cfg_any_os_util! {
+/// Windows-only extensions to the mio crate.
+///
+/// Mio on windows is currently implemented with IOCP for a high-performance
+/// implementation of asynchronous I/O. On top of the IOCP, a connection to the "Ancillary Function
+/// Driver" provides socket services. However, this approach doesn't work for other handle types
+/// like files and named pipes. They need to be connected directly to the IO completion port like
+/// this is done on unix-like systems  with the `SourceFd` type. The purpose of this module is to
+/// similarly provide a mechanism for foreign I/O types to get hooked up into the IOCP event loop.
+///
+/// This module provides types for interfacing with a custom IOCP handle:
+///
+/// * `Binding` - this type is intended to govern binding with mio's `Registry`
+///   type. Each I/O object should contain an instance of `Binding` that's
+///   interfaced with for the implementation of the `Source` trait. The
+///   `register`, `reregister`, and `deregister` methods for the `Source` trait
+///   may use methods provided by `Binding` during their operations.
+///
+///   Also note that for types which represent streams of bytes the mio
+///   interface of *readiness* doesn't map directly to the Windows model of
+///   *completion*. This means that types will have to perform internal
+///   buffering to ensure that a readiness interface can be provided.
+///
+/// * `Overlapped` - this type is intended to be used as the concrete instances
+///   of the `OVERLAPPED` type that most win32 methods expect. It's crucial, for
+///   safety, that all asynchronous operations are initiated with an instance of
+///   `Overlapped` and not another instantiation of `OVERLAPPED`.
+///
+///   Mio's `Overlapped` type is created with an object implementing `CompletionHandler` that
+///   is notified on completion of the associated asynchronous operation. The provided
+///   `OVERLAPPED_ENTRY` type is defined in the `winapi` crate. Whenever a completion is posted to
+///   an IOCP object the `OVERLAPPED` that was signaled will be interpreted as
+///   `Overlapped` in the mio crate and the `CompletionHandler` will be invoked.
+///   Through this object, and through the `OVERLAPPED` pointer, implementations can handle
+///   management of I/O events.
+///
+/// * `Readiness` - this type is used to indicate whether an IOCP client is ready to start the next
+///   IO operation. This is done either by returning a `Readiness` value from the
+///   `CompletionHandler` or by using `Binding::inject_event`.
+///
+/// * `CompletionHandler` - this trait is implemented by the IOCP client and is used as a callback
+///   when an asynchronous operation has finished. Currently it only implements one callback but may
+///   be extended in the future.
+///
+/// When put together these types enable custom Windows handles to be
+/// registered with mio's event loops. The `Binding` type is used to associate
+/// handles and the `Overlapped` type is used to execute I/O operations. When
+/// the I/O operations are completed an object implementing `CompletionHandler` is called which
+/// typically emits a `Readiness` that is then reported to the user who then may start the next
+/// asynchronous operation.
+#[cfg(windows)]
+    pub mod windows {
+        //! Windows only extensions.
+        pub use crate::sys::{
+            Binding,
+            CompletionCallback,
+            Overlapped,
+            Readiness,
+        };
+    }
+}
+
 // Enable with `cargo doc --features extra-docs`.
 #[cfg(feature = "extra-docs")]
 pub mod features {
--- a/src/sys/windows/selector.rs
+++ b/src/sys/windows/selector.rs
@@ -5,48 +5,57 @@ use crate::sys::event::{
     ERROR_FLAGS, READABLE_FLAGS, READ_CLOSED_FLAGS, WRITABLE_FLAGS, WRITE_CLOSED_FLAGS,
 };
 use crate::sys::Events;
+use crate::sys::windows::iocp_handler::{
+    IocpHandler, IocpHandlerRegistry
+};
 use crate::Interest;
 
-use miow::iocp::{CompletionPort, CompletionStatus};
-use miow::Overlapped;
+use miow::iocp::CompletionStatus;
 use std::collections::VecDeque;
 use std::marker::PhantomPinned;
 use std::os::windows::io::RawSocket;
 use std::pin::Pin;
 #[cfg(debug_assertions)]
-use std::sync::atomic::AtomicUsize;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 use std::{io, ptr};
 use winapi::shared::ntdef::NT_SUCCESS;
 use winapi::shared::ntdef::{HANDLE, PVOID};
 use winapi::shared::ntstatus::STATUS_CANCELLED;
-use winapi::shared::winerror::{ERROR_INVALID_HANDLE, ERROR_IO_PENDING, WAIT_TIMEOUT};
+use winapi::shared::winerror::{ERROR_INVALID_HANDLE, ERROR_IO_PENDING};
 use winapi::um::minwinbase::OVERLAPPED;
 
 /// Overlapped value to indicate a `Waker` event.
 //
 // Note: this must be null, `SelectorInner::feed_events` depends on it.
-pub const WAKER_OVERLAPPED: *mut Overlapped = ptr::null_mut();
+pub const WAKER_OVERLAPPED: *mut miow::Overlapped = ptr::null_mut();
 
 #[derive(Debug)]
 struct AfdGroup {
-    cp: Arc<CompletionPort>,
+    cp: Arc<IocpHandlerRegistry>,
     afd_group: Mutex<Vec<Arc<Afd>>>,
+    completion_handler: AfdCompletionPortEventHandler,
 }
 
 impl AfdGroup {
-    pub fn new(cp: Arc<CompletionPort>) -> AfdGroup {
+    pub fn new(cp: Arc<IocpHandlerRegistry>, completion_handler: AfdCompletionPortEventHandler) -> AfdGroup {
         AfdGroup {
             afd_group: Mutex::new(Vec::new()),
             cp,
+            completion_handler
         }
     }
 
     pub fn release_unused_afd(&self) {
         let mut afd_group = self.afd_group.lock().unwrap();
         afd_group.retain(|g| Arc::strong_count(&g) > 1);
+    }
+}
+
+impl Drop for AfdGroup {
+    fn drop(&mut self) {
+        self.release_unused_afd();
     }
 }
 
@@ -75,7 +84,7 @@ cfg_net! {
         }
 
         fn _alloc_afd_group(&self, afd_group: &mut Vec<Arc<Afd>>) -> io::Result<()> {
-            let afd = Afd::new(&self.cp)?;
+            let afd = Afd::new(&self.cp, self.completion_handler.clone())?;
             let arc = Arc::new(afd);
             afd_group.push(arc);
             Ok(())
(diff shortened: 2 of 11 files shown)


src/lib.rs
```diff
@@ -89,6 +89,68 @@ pub mod unix {
...
+/// like files and named pipes. They need to be connected directly to the IO completion port like
+/// this is done on unix-like systems  with the `SourceFd` type. The purpose of this module is to
+/// similarly provide a mechanism for foreign I/O types to get hooked up into the IOCP event loop.
+///
+/// This module provides types for interfacing with a custom IOCP handle:
+///
+/// * `Binding` - this type is intended to govern binding with mio's `Registry`
+///   type. Each I/O object should contain an instance of `Binding` that's
+///   interfaced with for the implementation of the `Source` trait. The
+///   `register`, `reregister`, and `deregister` methods for the `Source` trait
```
Thomasdezeeuw, line 107: I see that `Binding` only has `register_handle`, how would I implement de/re-register?

Or this is that part simply not implemented yet?

src/sys/windows/selector.rs
```diff
@@ -371,14 +411,233 @@ impl Selector {
...
+
+    impl Overlapped {
+        /// Creates a new `Overlapped` which will invoke the provided `cb` callback
+        /// whenever it's triggered.
+        ///
+        /// The returned `Overlapped` must be used as the `OVERLAPPED` passed to all
+        /// I/O operations that are registered with mio's event loop. When the I/O
+        /// operation associated with an `OVERLAPPED` pointer completes the event
+        /// loop will invoke the object provided as `callback`.
+        pub fn new<C: CompletionCallback+'static>(callback: C) -> Self {
```
Thomasdezeeuw, line 613: Why do we need a trait with a single method? Can't we use`C: FnMut(&OVERLAPPED_ENTRY) + static`?

src/sys/windows/selector.rs
```diff
@@ -371,14 +411,233 @@ impl Selector {
...
+    /// that is then converted into `mio::Event` and delivered to the user.
+    ///
+    /// This trait is implemented for closures as well.
+    pub trait CompletionCallback {
+        /// Called on completion of the associated operation.
+        ///
+        /// The callback will be called once the operation that it is bound to via a call to
+        /// `Overlapped::new` has been completed and the IOCP is signalled. By returning a `Readiness`
+        /// value, the callback may deliver a `mio::Event` to the user.
+        fn complete_operation(&mut self, entry: &OVERLAPPED_ENTRY) -> Option<Readiness>;
```
Thomasdezeeuw, line 492: Personally not the biggest fan out exposing types from different crates, especially when those crates don't have a stable API.

src/sys/windows/selector.rs
```diff
@@ -371,14 +411,233 @@ impl Selector {
...
+    /// This type is required to be used for all IOCP operations on handles that are
+    /// registered with an event loop. The event loop will receive notifications
+    /// over `OVERLAPPED` pointers that have completed, and it will cast that
+    /// pointer to a pointer to this structure and invoke the associated callback.
+    #[repr(C)]
+    pub struct Overlapped {
+        inner: UnsafeCell<miow::Overlapped>,
+        #[cfg(debug_assertions)]
+        canary: u32,
+        callback: Box<dyn CompletionCallback>,
```
Thomasdezeeuw, line 596: Maybe `fn(&OVERLAPPED_ENTRY) -> Option<Readiness>` would be better, than we don't have to allocate.

src/sys/windows/selector.rs
```diff
@@ -371,14 +411,233 @@ impl Selector {
...
+        token: Token
+    }
+
+    impl Binding {
+        /// Creates a new blank binding ready to be inserted into an I/O
+        /// object.
+        ///
+        /// Won't actually do anything until `register_handle` has been called during a call to
+        /// `Registry::register`.
+        pub fn new(registry: &Registry, token: Token) -> Self {
```
Thomasdezeeuw, line 541: Is there a reason why you don't require an `event::Source` on creation of `Binding`?

src/sys/windows/selector.rs
```diff
@@ -371,14 +411,233 @@ impl Selector {
...
+        /// scheduled.
+        pub fn inject_event(&self, readiness: Readiness) {
+            let event = readiness.into_event(self.token);
+            self.selector.injected_events.lock().unwrap().push(event)
+        }
+
+        /// Check whether the provided registry is the same as the one used during creation of the
+        /// binding.
+        ///
+        /// IOCP driven sources cannot unregister from their completion port before they get closed.
```
Thomasdezeeuw, line 575: No `event::Source` in Mio can be registered with another `Poll`, so this is true for all platforms. See `IoSource`'s `selector_id` field.

src/sys/windows/selector.rs
```diff
@@ -371,14 +411,233 @@ impl Selector {
...
+            self.selector.cp.register_handle(handle, handler.into())
+        }
+
+        /// Inject an event to the event loop.
+        ///
+        /// In case it is required to generate an event even though there was no callback due to a
+        /// completed IO operation, this method can be used to mark the associated token as ready, e.g.
+        /// in case it is necessary to signal an initial readiness before any IO operation has been
+        /// scheduled.
+        pub fn inject_event(&self, readiness: Readiness) {
```
Thomasdezeeuw, line 567: I'm not sure about exposing such a function. Is the case you describe in the docs real? I personally never needed that.

src/sys/windows/selector.rs
```diff
@@ -371,14 +411,233 @@ impl Selector {
...
+                canary: OVERLAPPED_CANARY,
+                callback: Box::new(callback) as Box<dyn CompletionCallback>,
+            }
+        }
+
+        /// Get the underlying `Overlapped` instance as a raw pointer.
+        ///
+        /// This can be useful when only a shared borrow is held and the overlapped
+        /// pointer needs to be passed down to winapi.
+        pub fn as_mut_ptr(&self) -> *mut OVERLAPPED {
```
Thomasdezeeuw, line 626: We don't really expose non-Mio types.

src/sys/windows/selector.rs
Thomasdezeeuw: Also can't Mio provide that implementation? Or is there special handling per type of handle?

darkwisebear: I was torn on this too, but I've chosen to introduce a trait because it can be extended in a source-compatible way by adding more trait methods with default impls if we find more cases of hooks that could be costomized by mio drivers (e.g. a "pre-poll hook").

There is a blanket impl for FnMuts as well, so they can be used, see the provided sample test case, which is doing exactly that.

src/sys/windows/selector.rs
darkwisebear: I don't think that this happens here: This method resembles the old fn(&OVERLAPPED_ENTRY) from mio 0.6, so except from the OVERLAPPED_ENTRY (which is a stable Windows API struct) we do not expose other types. But maybe I misunderstand your comment.

src/sys/windows/selector.rs
darkwisebear: This is correct. I've chosen to use a boxed trait here because I was looking at mio_named_pipes which does some pointer magic to get its state back inside the callback functions fore read, write and connect. By using dynamic dispatch, we can avoid that and have more safe code, but if resource usage beats that argument (plus the "future hooks" argument from above) I'd be willing to change this to a pure function pointer.

src/sys/windows/selector.rs
darkwisebear: Ah, I see, so this is redundand and can be removed. I'll do that.

src/sys/windows/selector.rs
darkwisebear: Ack, will change that one.

src/sys/windows/selector.rs
darkwisebear: While working on that PR, I used mio_named_pipes as a sample driver and forward ported it. There were cases (example: Establishing the first read after a successful connection) where in case of an error an immediate readiness flag was raised to be able to deliver the error. I also would love to get rid of that call as it requires a "side channel" for events but I couldn't think of a better solution yet.

src/sys/windows/selector.rs
darkwisebear: Actually I just need the handle, not the whole source. I split this into two parts as I wanted to give the user a chance to store the `Binding` instance before it starts firing events, so the usual sequence in a driver is:
- Initialize the `Option<Binding>` member of your custom struct with `None`
- Inside `Source::register`, you then set it to `Some(Binding::new(..))`
- *then** give it the handle to work with using `register_handle`

src/lib.rs
darkwisebear: Its not implemented yet because I'm unsure how that should/could be done as in Windows, there is no way of removing a handle from an IOCP except by closing that handle. So the only thing we should do here is to remove the handler from the list of IO handlers, which should be done by a Drop implementation of that type. reregistration would only support changing the token; I will add both functionalities.

Thomasdezeeuw:
@darkwisebear having though about this a bit more I think trying to `SourceFd` like API first would be the best way forward to make externally created `event::Source`s (e.g. unix/anonymous pipes). I don't know if that is possible however. If not lets try to minimise the API to begin with, e.g. only expose functions like `inject_event` once requested.

darkwisebear:
> @darkwisebear having though about this a bit more I think trying to `SourceFd` like API first would be the best way forward to make externally created `event::Source`s (e.g. unix/anonymous pipes). I don't know if that is possible however.

The major problem I see when trying to mimic `SourceFd` is that IOCP doesn't tell you what operation it was that finished when IOCPs `get_many` returns an event for a certain token (remember: IOCP is completion-based as opposed to readiness-based). Therefore we cannot (in the general case) deduce from a returned IOCP event what operation might now be available on a certain handle/source. I don't know what happens is if you simply signal read AND write readiness on event arrival, but that just doesn't sound right tbh :)

The only idea I have here is to provide premade `OVERLAPPED` factories that the user needs to use for the operations that require `OVERLAPPED`s so that mio knows what kind of `Event` has to be generated. However, `OVERLAPPED`s will also transport information like bytes transferred and the file offset back to the caller, information that is not available through the current mio event API - which I think is deliberate as other systems like async IO on Unix will return read/written bytes immediately, so this is a non-issue there. However, on Windows you usually at least need to know the number of bytes read/written to update internal buffers, maybe also file positions, depending on the driver.

All in all I'd say I agree with starting with a leaner API but I'm convinced that IOCP and `SourceFd` won't be a good match and we need a broader, more IOCP specific API like the one from mio 0.6.x that this PR is inspired by. In addition I'm not sure whether mio-named-pipes can be refactored such that this API will be sufficient. But if it turns out to be infeasible, we can still add the required APIs.

darkwisebear:
@carllerche @Thomasdezeeuw Is this approach still relevant / desired in any form?

Thomasdezeeuw:
@carllerche can you look at this? Seeing as an API akin to `SourceFd` doesn't seem possible it would require a bit more Windows knowledge.

Thomasdezeeuw:
@darkwisebear do you plan on continuing this work?

darkwisebear:
@Thomasdezeeuw Since there was no additional feedback I didn't pursue this PR any further but used an internal patched version of mio for my projects. Now there is https://github.com/tokio-rs/tokio/pull/3760 which provides `NamedPipe` which was the reason I proposed this change. Are there other use cases for such an internal API and would it suffice/have a chance to be integrated like this?

Thomasdezeeuw:
> @Thomasdezeeuw Since there was no additional feedback I didn't pursue this PR any further but used an internal patched version of mio for my projects. Now there is [tokio-rs/tokio#3760](https://github.com/tokio-rs/tokio/pull/3760) which provides `NamedPipe` which was the reason I proposed this change. Are there other use cases for such an internal API and would it suffice/have a chance to be integrated like this?

I'm afraid I'm not familiar enough with Windows to answer that.

Also, sorry about the lack of feedback, I've been a bit busy over the last year or so, but I have a bit more time for Mio now.

darkwisebear:
> I'm afraid I'm not familiar enough with Windows to answer that.

I need to know two things:

1. Are there more use cases for the proposed API?
2. Is the API as well as the implementation (at least the aproach, the code itself code needs some love anyway since a lot of code has changed since then) close to a state that would lead to a final merge to master?

The core question now is: How/who can we answer those two questions? I would feel a lot more comfortable continuing to work if those were answered.

> Also, sorry about the lack of feedback, I've been a bit busy over the last year or so, but I have a bit more time for Mio now.

No worries, it was meant as an offer to the community in case this was needed, didn't have strong feelings about merging that. If there is still need, we can continue working on it.

roblabla:
> Are there more use cases for the proposed API?

I can answer this one at least. I subscribed to this issue ages ago, because I was looking to integrate windows Filter Ports (`FltGetMessage` & co, used to communicate with drivers) into tokio. There are a few more APIs like this (most notably, ones based on `DeviceIoControl`) that currently cannot be integrated into mio cleanly.

Thomasdezeeuw:
> > Are there more use cases for the proposed API?
> 
> I can answer this one at least. I subscribed to this issue ages ago, because I was looking to integrate windows Filter Ports (`FltGetMessage` & co, used to communicate with drivers) into tokio. There are a few more APIs like this (most notably, ones based on `DeviceIoControl`) that currently cannot be integrated into mio cleanly.

@roblabla could you maybe describe some (rough) requirement that you would need for Mio to work cleanly, and if this pr would (partially) implement those requirements.

Noah-Kennedy:
I'm in a similar situation to @roblabla here.

Thomasdezeeuw:
I'm closing this due to inactivity. @darkwisebear if you, or anyone else, want to continue this please don't feel discouraged to do so.

Thank you @darkwisebear for your efforts here.

Noah-Kennedy:
@Thomasdezeeuw I am not roblabla, but I have a use case where I need to support some custom sockets on windows. I would be willing to contribute to getting an API like this in place for mio.

Thomasdezeeuw:
> @Thomasdezeeuw I have a use case where I need to support some custom sockets on windows. I would be willing to contribute to getting an API like this in place for mio.

Sounds good. I can't offer much help with the implementation as I'm no Windows expert, but some do sometimes visit our Mio discord channel so you could ask for help there.

PR #1351, opened 2020-09-29
Add windows NamedPipe

carllerche:
Tokio needs windows named pipe support in order to upgrade to Mio 0.7.  I thought I would take a quick initial stab at this.

It is based on https://github.com/bbqsrc/mio-named-pipes, but simplified some. Instead of shoehorning this into the `Source` trait, I dod registration in `new`. Also, since the `Registration` internal queue is gone in Mio 0.6, I built the API on top of `std::task::Waker` instead. As this is essentially adding an IOCP based API on top of Mio in order to power Tokio, it seemed silly to add a shim to make the API "seem" like epoll and then layer wakers back on.

### DIFF
--- a/src/sys/windows/afd.rs
+++ b/src/sys/windows/afd.rs
@@ -194,7 +194,10 @@ cfg_net! {
                     ));
                 }
                 let fd = File::from_raw_handle(afd_helper_handle as RawHandle);
-                // Increment by 2 to reserve space for other types of handles
+                // Increment by 2 to reserve space for other types of handles.
+                // Non-AFD types (currently only NamedPipe), use odd numbered
+                // tokens. This allows the selector to differentate between them
+                // and dispatch events accordingly.
                 let token = NEXT_TOKEN.fetch_add(2, Ordering::Relaxed) + 2;
                 let afd = Afd { fd };
                 cp.add_handle(token, &afd.fd)?;
--- a/src/sys/windows/afd.rs
+++ b/src/sys/windows/afd.rs
@@ -194,7 +194,8 @@ cfg_net! {
                     ));
                 }
                 let fd = File::from_raw_handle(afd_helper_handle as RawHandle);
-                let token = NEXT_TOKEN.fetch_add(1, Ordering::Relaxed) + 1;
+                // Increment by 2 to reserve space for other types of handles
+                let token = NEXT_TOKEN.fetch_add(2, Ordering::Relaxed) + 2;
                 let afd = Afd { fd };
                 cp.add_handle(token, &afd.fd)?;
                 match SetFileCompletionNotificationModes(
--- a/src/sys/windows/afd.rs
+++ b/src/sys/windows/afd.rs
@@ -7,9 +7,7 @@ use std::io;
 use std::mem::size_of;
 use std::os::windows::io::AsRawHandle;
 use std::ptr::null_mut;
-use winapi::shared::ntdef::{
-    HANDLE, LARGE_INTEGER, NTSTATUS, PVOID, ULONG,
-};
+use winapi::shared::ntdef::{HANDLE, LARGE_INTEGER, NTSTATUS, PVOID, ULONG};
 use winapi::shared::ntstatus::{STATUS_NOT_FOUND, STATUS_PENDING, STATUS_SUCCESS};
 
 const IOCTL_AFD_POLL: ULONG = 0x00012024;
--- a/src/sys/windows/named_pipe.rs
+++ b/src/sys/windows/named_pipe.rs
@@ -1,0 +1,654 @@
+//! Windows named pipes bindings for mio.
+//!
+//! This crate implements bindings for named pipes for the mio crate. This
+//! crate compiles on all platforms but only contains anything on Windows.
+//! Currently this crate requires mio 0.6.2.
+//!
+//! On Windows, mio is implemented with an IOCP object at the heart of its
+//! `Poll` implementation. For named pipes, this means that all I/O is done in
+//! an overlapped fashion and the named pipes themselves are registered with
+//! mio's internal IOCP object. Essentially, this crate is using IOCP for
+//! bindings with named pipes.
+//!
+//! Note, though, that IOCP is a *completion* based model whereas mio expects a
+//! *readiness* based model. As a result this crate, like with TCP objects in
+//! mio, has internal buffering to translate the completion model to a readiness
+//! model. This means that this crate is not a zero-cost binding over named
+//! pipes on Windows, but rather approximates the performance of mio's TCP
+//! implementation on Windows.
+//!
+//! # Trait implementations
+//!
+//! The `Read` and `Write` traits are implemented for `NamedPipe` and for
+//! `&NamedPipe`. This represents that a named pipe can be concurrently read and
+//! written to and also can be read and written to at all. Typically a named
+//! pipe needs to be connected to a client before it can be read or written,
+//! however.
+//!
+//! Note that for I/O operations on a named pipe to succeed then the named pipe
+//! needs to be associated with an event loop. Until this happens all I/O
+//! operations will return a "would block" error.
+//!
+//! # Managing connections
+//!
+//! The `NamedPipe` type supports a `connect` method to connect to a client and
+//! a `disconnect` method to disconnect from that client. These two methods only
+//! work once a named pipe is associated with an event loop.
+//!
+//! The `connect` method will succeed asynchronously and a completion can be
+//! detected once the object receives a writable notification.
+//!
+//! # Named pipe clients
+//!
+//! Currently to create a client of a named pipe server then you can use the
+//! `OpenOptions` type in the standard library to create a `File` that connects
+//! to a named pipe. Afterwards you can use the `into_raw_handle` method coupled
+//! with the `NamedPipe::from_raw_handle` method to convert that to a named pipe
+//! that can operate asynchronously. Don't forget to pass the
+//! `FILE_FLAG_OVERLAPPED` flag when opening the `File`.
+
+use crate::{poll, Registry, Token};
+
+use std::cell::UnsafeCell;
+use std::ffi::OsStr;
+use std::fmt;
+use std::io::{self, Read};
+use std::mem;
+use std::os::windows::io::*;
+use std::slice;
+use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering::SeqCst;
+use std::sync::{Arc, Mutex};
+use std::task::{Context, Poll, Waker};
+
+// use mio::windows;
+// use mio::{Evented, Poll, PollOpt, Ready, Registration, SetReadiness, Token};
+use miow::iocp::CompletionStatus;
+use miow::pipe;
+use winapi::shared::winerror::*;
+use winapi::um::ioapiset::*;
+use winapi::um::minwinbase::*;
+
+macro_rules! offset_of {
+    ($t:ty, $($field:ident).+) => (
+        &(*(0 as *const $t)).$($field).+ as *const _ as usize
+    )
+}
+
+macro_rules! overlapped2arc {
+    ($e:expr, $t:ty, $($field:ident).+) => ({
+        let offset = offset_of!($t, $($field).+);
+        debug_assert!(offset < mem::size_of::<$t>());
+        Arc::from_raw(($e as usize - offset) as *mut $t)
+    })
+}
+
+fn would_block() -> io::Error {
+    io::ErrorKind::WouldBlock.into()
+}
+
+/// Representation of a named pipe on Windows.
+///
+/// This structure internally contains a `HANDLE` which represents the named
+/// pipe, and also maintains state associated with the mio event loop and active
+/// I/O operations that have been scheduled to translate IOCP to a readiness
+/// model.
+pub struct NamedPipe {
+    inner: Arc<Inner>,
+}
+
+struct Inner {
+    handle: pipe::NamedPipe,
+
+    connect: Overlapped,
+    connecting: AtomicBool,
+
+    read: Overlapped,
+    write: Overlapped,
+
+    io: Mutex<Io>,
+
+    pool: Mutex<BufferPool>,
+}
+
+struct Io {
+    read: State,
+    read_waker: Option<Waker>,
+    write: State,
+    write_waker: Option<Waker>,
+    connect_error: Option<io::Error>,
+}
+
+enum State {
+    None,
+    Pending(Vec<u8>, usize),
+    Ok(Vec<u8>, usize),
+    Err(io::Error),
+}
+
+fn _assert_kinds() {
+    fn _assert_send<T: Send>() {}
+    fn _assert_sync<T: Sync>() {}
+    _assert_send::<NamedPipe>();
+    _assert_sync::<NamedPipe>();
+}
+
+impl NamedPipe {
+    /// Creates a new named pipe at the specified `addr` given a "reasonable
+    /// set" of initial configuration options.
+    pub fn new<A: AsRef<OsStr>>(addr: A, registry: &Registry, token: Token) -> io::Result<NamedPipe> {
+        let pipe = pipe::NamedPipe::new(addr)?;
+        NamedPipe::from_raw_handle(pipe.into_raw_handle(), registry, token)
+    }
+
+    /// TODO: Dox
+    pub fn from_raw_handle(handle: RawHandle, registry: &Registry, token: Token) -> io::Result<NamedPipe> {
+        // Create the pipe
+        let pipe = NamedPipe {
+            inner: Arc::new(Inner {
+                // Safety: not really unsafe
+                handle: unsafe { pipe::NamedPipe::from_raw_handle(handle) },
+                // transmutes to straddle winapi versions (mio 0.6 is on an
+                // older winapi)
+                connect: Overlapped::new(connect_done),
+                connecting: AtomicBool::new(false),
+                read: Overlapped::new(read_done),
+                write: Overlapped::new(write_done),
+                io: Mutex::new(Io {
+                    read: State::None,
+                    read_waker: None,
+                    write: State::None,
+                    write_waker: None,
+                    connect_error: None,
+                }),
+                pool: Mutex::new(BufferPool::with_capacity(2)),
+            }),
+        };
+
+        // Register the handle w/ the IOCP handle
+        poll::selector(registry).inner.cp.add_handle(usize::from(token), &pipe.inner.handle)?;
+
+        // Queue the initial read
+        pipe.inner.post_register();
+
+        Ok(pipe)
+    }
+
+    /// Attempts to call `ConnectNamedPipe`, if possible.
+    ///
+    /// This function will attempt to connect this pipe to a client in an
+    /// asynchronous fashion. If the function immediately establishes a
+    /// connection to a client then `Ok(())` is returned. Otherwise if a
+    /// connection attempt was issued and is now in progress then a "would
+    /// block" error is returned.
+    ///
+    /// When the connection is finished then this object will be flagged as
+    /// being ready for a write, or otherwise in the writable state.
+    ///
+    /// # Errors
+    ///
+    /// This function will return a "would block" error if the pipe has not yet
+    /// been registered with an event loop, if the connection operation has
+    /// previously been issued but has not yet completed, or if the connect
+    /// itself was issued and didn't finish immediately.
+    ///
+    /// Normal I/O errors from the call to `ConnectNamedPipe` are returned
+    /// immediately.
+    pub fn connect(&self) -> io::Result<()> {
+        // "Acquire the connecting lock" or otherwise just make sure we're the
+        // only operation that's using the `connect` overlapped instance.
+        if self.inner.connecting.swap(true, SeqCst) {
+            return Err(would_block());
+        }
+
+        // Now that we've flagged ourselves in the connecting state, issue the
+        // connection attempt. Afterwards interpret the return value and set
+        // internal state accordingly.
+        let res = unsafe {
+            let overlapped = self.inner.connect.as_mut_ptr() as *mut _;
+            self.inner.handle.connect_overlapped(overlapped)
+        };
+
+        match res {
+            // The connection operation finished immediately, so let's schedule
+            // reads/writes and such.
+            Ok(true) => {
+                self.inner.connecting.store(false, SeqCst);
+                Inner::post_register(&self.inner);
+                Ok(())
+            }
+
+            // If the overlapped operation was successful and didn't finish
+            // immediately then we forget a copy of the arc we hold
+            // internally. This ensures that when the completion status comes
+            // in for the I/O operation finishing it'll have a reference
+            // associated with it and our data will still be valid. The
+            // `connect_done` function will "reify" this forgotten pointer to
+            // drop the refcount on the other side.
+            Ok(false) => {
+                mem::forget(self.inner.clone());
+                Err(would_block())
+            }
+
+            // TODO: are we sure no IOCP notification comes in here?
+            Err(e) => {
+                self.inner.connecting.store(false, SeqCst);
+                Err(e)
+            }
+        }
+    }
+
+    /// Takes any internal error that has happened after the last I/O operation
+    /// which hasn't been retrieved yet.
+    ///
+    /// This is particularly useful when detecting failed attempts to `connect`.
+    /// After a completed `connect` flags this pipe as writable then callers
+    /// must invoke this method to determine whether the connection actually
+    /// succeeded. If this function returns `None` then a client is connected,
+    /// otherwise it returns an error of what happened and a client shouldn't be
+    /// connected.
+    pub fn take_error(&self) -> io::Result<Option<io::Error>> {
+        Ok(self.inner.io.lock().unwrap().connect_error.take())
+    }
+
+    /// Disconnects this named pipe from a connected client.
+    ///
+    /// This function will disconnect the pipe from a connected client, if any,
+    /// transitively calling the `DisconnectNamedPipe` function. If the
+    /// disconnection is successful then this object will no longer be readable
+    /// or writable.
+    ///
+    /// After a `disconnect` is issued, then a `connect` may be called again to
+    /// connect to another client.
+    pub fn disconnect(&self) -> io::Result<()> {
+        self.inner.handle.disconnect()
+    }
+
+    /// TODO: dox
+    pub fn read(&mut self, cx: &mut Context<'_>, buf: &mut [u8]) -> Poll<io::Result<usize>> {
+        let mut state = self.inner.io.lock().unwrap();
+        match mem::replace(&mut state.read, State::None) {
+            // In theory not possible with `ready_registration` checked above,
+            // but return would block for now.
+            State::None => {
+                state.read_waker = Some(cx.waker().clone());
+                Poll::Pending
+            }
+
+            // A read is in flight, still waiting for it to finish
+            State::Pending(buf, amt) => {
+                state.read = State::Pending(buf, amt);
+                state.read_waker = Some(cx.waker().clone());
+                Poll::Pending
+            }
+
+            // We previously read something into `data`, try to copy out some
+            // data. If we copy out all the data schedule a new read and
+            // otherwise store the buffer to get read later.
+            State::Ok(data, cur) => {
+                let n = {
+                    let mut remaining = &data[cur..];
+                    remaining.read(buf)?
+                };
+                let next = cur + n;
+                if next != data.len() {
+                    state.read = State::Ok(data, next);
+                } else {
+                    self.inner.put_buffer(data);
+                    Inner::schedule_read(&self.inner, &mut state);
+                }
+                Poll::Ready(Ok(n))
+            }
+
+            // Looks like an in-flight read hit an error, return that here while
+            // we schedule a new one.
+            State::Err(e) => {
+                Inner::schedule_read(&self.inner, &mut state);
+                if e.raw_os_error() == Some(ERROR_BROKEN_PIPE as i32) {
+                    Poll::Ready(Ok(0))
+                } else {
+                    Poll::Ready(Err(e))
+                }
+            }
+        }
+    }
+
+    /// TODO: dox
+    pub fn write(&mut self, cx: &mut Context<'_>, buf: &[u8]) -> Poll<io::Result<usize>> {
+        // Make sure there's no writes pending
+        let mut io = self.inner.io.lock().unwrap();
+        match io.write {
+            State::None => {}
+            _ => {
+                io.write_waker = Some(cx.waker().clone());
+                return Poll::Pending;
+            }
+        }
+
+        // Move `buf` onto the heap and fire off the write
+        let mut owned_buf = self.inner.get_buffer();
+        owned_buf.extend(buf);
+        Inner::schedule_write(&self.inner, owned_buf, 0, &mut io);
+        Poll::Ready(Ok(buf.len()))
+    }
+}
+
+impl AsRawHandle for NamedPipe {
+    fn as_raw_handle(&self) -> RawHandle {
+        self.inner.handle.as_raw_handle()
+    }
+}
+
+impl fmt::Debug for NamedPipe {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.inner.handle.fmt(f)
+    }
+}
+
+impl Drop for NamedPipe {
+    fn drop(&mut self) {
+        // Cancel pending reads/connects, but don't cancel writes to ensure that
+        // everything is flushed out.
+        unsafe {
+            if self.inner.connecting.load(SeqCst) {
+                drop(cancel(&self.inner.handle, &self.inner.connect));
+            }
+
+            let io = self.inner.io.lock().unwrap();
+
+            match io.read {
+                State::Pending(..) => {
+                    drop(cancel(&self.inner.handle, &self.inner.read));
+                }
+                _ => {}
+            }
+        }
+    }
+}
+
+impl Inner {
+    /// Schedules a read to happen in the background, executing an overlapped
+    /// operation.
+    ///
+    /// This function returns `true` if a normal error happens or if the read
+    /// is scheduled in the background. If the pipe is no longer connected
+    /// (ERROR_PIPE_LISTENING) then `false` is returned and no read is
+    /// scheduled.
+    fn schedule_read(me: &Arc<Inner>, io: &mut Io) -> bool {
+        // Check to see if a read is already scheduled/completed
+        match io.read {
+            State::None => {}
+            _ => return true,
+        }
+
+        // Allocate a buffer and schedule the read.
+        let mut buf = me.get_buffer();
+        let e = unsafe {
+            let overlapped = me.read.as_mut_ptr() as *mut _;
+            let slice = slice::from_raw_parts_mut(buf.as_mut_ptr(), buf.capacity());
+            me.handle.read_overlapped(slice, overlapped)
+        };
+
+        match e {
+            // See `connect` above for the rationale behind `forget`
+            Ok(_) => {
+                io.read = State::Pending(buf, 0); // 0 is ignored on read side
+                mem::forget(me.clone());
+                true
+            }
+
+            // If ERROR_PIPE_LISTENING happens then it's not a real read error,
+            // we just need to wait for a connect.
+            Err(ref e) if e.raw_os_error() == Some(ERROR_PIPE_LISTENING as i32) => false,
+
+            // If some other error happened, though, we're now readable to give
+            // out the error.
+            Err(e) => {
+                io.read = State::Err(e);
+                if let Some(waker) = io.read_waker.take() {
+                    waker.wake();
+                }
+                true
+            }
+        }
+    }
+
+    fn schedule_write(me: &Arc<Inner>, buf: Vec<u8>, pos: usize, io: &mut Io) {
+        // Very similar to `schedule_read` above, just done for the write half.
+        let e = unsafe {
+            let overlapped = me.write.as_mut_ptr() as *mut _;
+            me.handle.write_overlapped(&buf[pos..], overlapped)
+        };
+
+        match e {
+            // See `connect` above for the rationale behind `forget`
+            Ok(_) => {
+                io.write = State::Pending(buf, pos);
+                mem::forget(me.clone())
+            }
+            Err(e) => {
+                io.write = State::Err(e);
+                if let Some(waker) = io.write_waker.take() {
+                    waker.wake();
+                }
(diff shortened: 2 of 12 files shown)


carllerche:
The reason I am opting to add this to mio proper is because #1047 is not done, so `NamedPipe` cannot be added outside of mio proper.

Thomasdezeeuw:
Wants me to move https://github.com/Thomasdezeeuw/mio-pipe into Mio as well? I was planning on proposing that for a while.

carllerche:
Sure! In general, I think it is fair to have mio constructs for common OS primitives.

Thomasdezeeuw:
@carllerche created an issue for the Mio-pipe move: https://github.com/tokio-rs/mio/issues/1354.

carllerche:
Ok, I got this building and tests passing. I would appreciate a review. This is an upgrade of the `mio-named-pipes` crate. I could use a review now, but let me get https://github.com/tokio-rs/tokio/pull/2893 building w/ this before merging.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
+//! Windows named pipes bindings for mio.
+//!
+//! This crate implements bindings for named pipes for the mio crate. This
```
Thomasdezeeuw, line 3: Outdated docs and you might want to move them to the `NamePipe` type as that is the only thing that is exported.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+use std::mem;
+use std::os::windows::io::*;
+use std::slice;
+use std::sync::atomic::{AtomicUsize, AtomicBool};
+use std::sync::atomic::Ordering::{Relaxed, SeqCst};
+use std::sync::{Arc, Mutex};
+use std::task::{Context, Poll, Waker};
+
+// use mio::windows;
+// use mio::{Evented, Poll, PollOpt, Ready, Registration, SetReadiness, Token};
```
Thomasdezeeuw, line 65: Nit: Remove.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+use miow::iocp::CompletionStatus;
+use miow::pipe;
+use winapi::shared::winerror::*;
+use winapi::um::ioapiset::*;
+use winapi::um::minwinbase::*;
+
+macro_rules! offset_of {
```
Thomasdezeeuw, line 72: This is only valid if the type has `#[repr(C)]`, if possible we might put some kind of bound/limit on it. Or at least a comment. Also see `Inner` struct.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+    /// or writable.
+    ///
+    /// After a `disconnect` is issued, then a `connect` may be called again to
+    /// connect to another client.
+    pub fn disconnect(&self) -> io::Result<()> {
+        self.inner.handle.disconnect()
+    }
+
+    /// TODO: dox
+    pub fn read(&mut self, cx: &mut Context<'_>, buf: &mut [u8]) -> Poll<io::Result<usize>> {
```
Thomasdezeeuw, line 284: Non of the other types of Mio use the task/`Future` system. So this seems out of place. This should implement the `io::Read` and `io::Write` traits instead?

src/sys/windows/afd.rs
```diff
@@ -196,7 +194,8 @@ cfg_net! {
                     ));
                 }
                 let fd = File::from_raw_handle(afd_helper_handle as RawHandle);
-                let token = NEXT_TOKEN.fetch_add(1, Ordering::Relaxed) + 1;
+                // Increment by 2 to reserve space for other types of handles
+                let token = NEXT_TOKEN.fetch_add(2, Ordering::Relaxed) + 2;
```
Thomasdezeeuw, line 198: Maybe add a comment referring to `NamePipe`'s `NEXT_TOKEN`? Saying that the odd tokens are used by it.

tests/regressions.rs
```diff
@@ -61,6 +61,7 @@ fn issue_776() {
 }
 
 #[test]
+#[ignore]
```
Thomasdezeeuw, line 64: Does this fail?

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+    None,
+    Pending(Vec<u8>, usize),
+    Ok(Vec<u8>, usize),
+    Err(io::Error),
+}
+
+// Odd tokens are for named pipes
+static NEXT_TOKEN: AtomicUsize = AtomicUsize::new(1);
+
+fn _assert_kinds() {
```
Thomasdezeeuw, line 133: Nit: can you move this to the `tests/win_named_pipe.rs` file? The assertion function are in `tests/util`.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+use std::sync::atomic::{AtomicUsize, AtomicBool};
+use std::sync::atomic::Ordering::{Relaxed, SeqCst};
+use std::sync::{Arc, Mutex};
+use std::task::{Context, Poll, Waker};
+
+// use mio::windows;
+// use mio::{Evented, Poll, PollOpt, Ready, Registration, SetReadiness, Token};
+use miow::iocp::CompletionStatus;
+use miow::pipe;
+use winapi::shared::winerror::*;
```
Thomasdezeeuw, line 68: Nit: I'm not the biggest fan of `*` imports.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+    /// set" of initial configuration options.
+    pub fn new<A: AsRef<OsStr>>(
+        addr: A,
+        registry: &Registry,
+    ) -> io::Result<NamedPipe> {
+        let pipe = pipe::NamedPipe::new(addr)?;
+        NamedPipe::from_raw_handle(pipe.into_raw_handle(), registry)
+    }
+
+    /// TODO: Dox
```
Thomasdezeeuw, line 151: Docs.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+    pub fn from_raw_handle(
```
Thomasdezeeuw, line 152: This must be `unsafe` because it calls [`FromRawHandle::from_raw_handle`](https://doc.rust-lang.org/std/os/windows/io/trait.FromRawHandle.html) on `pipe::NamedPipe`, the contract of which the caller must uphold.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+        handle: RawHandle,
+        registry: &Registry,
+    ) -> io::Result<NamedPipe> {
+        // Generate a token
+        let token = NEXT_TOKEN.fetch_add(2, Relaxed) + 2;
+
+        // Create the pipe
+        let pipe = NamedPipe {
+            inner: Arc::new(Inner {
+                // Safety: not really unsafe
```
Thomasdezeeuw, line 162: Suggestion:
```
// Safety: the call must ensure the handle is valid.
```

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+                handle: unsafe { pipe::NamedPipe::from_raw_handle(handle) },
+                // transmutes to straddle winapi versions (mio 0.6 is on an
```
Thomasdezeeuw, line 164: This comment seems outdated, referring to Mio 0.6?

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+        registry: &Registry,
+    ) -> io::Result<NamedPipe> {
+        let pipe = pipe::NamedPipe::new(addr)?;
+        NamedPipe::from_raw_handle(pipe.into_raw_handle(), registry)
+    }
+
+    /// TODO: Dox
+    pub fn from_raw_handle(
+        handle: RawHandle,
+        registry: &Registry,
```
Thomasdezeeuw, line 154: Most type are registered after creation, not in the same creation function. Is that possible for this type?

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+    /// been registered with an event loop, if the connection operation has
+    /// previously been issued but has not yet completed, or if the connect
+    /// itself was issued and didn't finish immediately.
+    ///
+    /// Normal I/O errors from the call to `ConnectNamedPipe` are returned
+    /// immediately.
+    pub fn connect(&self) -> io::Result<()> {
+        // "Acquire the connecting lock" or otherwise just make sure we're the
+        // only operation that's using the `connect` overlapped instance.
+        if self.inner.connecting.swap(true, SeqCst) {
```
Thomasdezeeuw, line 216: Nit: this can be lowered to `AcqRel`.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+            // in for the I/O operation finishing it'll have a reference
+            // associated with it and our data will still be valid. The
+            // `connect_done` function will "reify" this forgotten pointer to
+            // drop the refcount on the other side.
+            Ok(false) => {
+                mem::forget(self.inner.clone());
+                Err(would_block())
+            }
+
+            // TODO: are we sure no IOCP notification comes in here?
```
Thomasdezeeuw, line 249: TODO.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+    /// connected.
+    pub fn take_error(&self) -> io::Result<Option<io::Error>> {
+        Ok(self.inner.io.lock().unwrap().connect_error.take())
+    }
+
+    /// Disconnects this named pipe from a connected client.
+    ///
+    /// This function will disconnect the pipe from a connected client, if any,
+    /// transitively calling the `DisconnectNamedPipe` function. If the
+    /// disconnection is successful then this object will no longer be readable
```
Thomasdezeeuw, line 274: I'm not sure if this is true or not, but wouldn't you be able to read the last bytes written before calling `disconnect`? Or are all bytes in the pipe the moment you call `disconnect` lost?

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+        self.inner.handle.fmt(f)
+    }
+}
+
+impl Drop for NamedPipe {
+    fn drop(&mut self) {
+        // Cancel pending reads/connects, but don't cancel writes to ensure that
+        // everything is flushed out.
+        unsafe {
+            if self.inner.connecting.load(SeqCst) {
```
Thomasdezeeuw, line 369: Nit: this can be lowered to `AcqRel`.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+        // Allocate a buffer and schedule the read.
+        let mut buf = me.get_buffer();
+        let e = unsafe {
+            let overlapped = me.read.as_mut_ptr() as *mut _;
+            let slice = slice::from_raw_parts_mut(buf.as_mut_ptr(), buf.capacity());
+            me.handle.read_overlapped(slice, overlapped)
+        };
+
+        match e {
+            // See `connect` above for the rationale behind `forget`
```
Thomasdezeeuw, line 409: Nit: `NamedPipe::connect`.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+            if let State::None = io.write {
+                if let Some(waker) = io.write_waker.take() {
+                    waker.wake();
+                }
+            }
+        }
+    }
+
+    fn get_buffer(&self) -> Vec<u8> {
+        self.pool.lock().unwrap().get(8 * 1024)
```
Thomasdezeeuw, line 466: Nit: 8kb seems like a lot, most Unix pipe buffer 4kb.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+    }
+
+    fn put_buffer(&self, buf: Vec<u8>) {
+        self.pool.lock().unwrap().put(buf)
+    }
+}
+
+unsafe fn cancel<T: AsRawHandle>(handle: &T, overlapped: &Overlapped) -> io::Result<()> {
+    let ret = CancelIoEx(handle.as_raw_handle(), overlapped.as_mut_ptr() as *mut _);
+    if ret == 0 {
```
Thomasdezeeuw, line 476: Nit: can you add a comment 0 is correct? It surprised me and I had to look it up, might safe other readers some time.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+///
+/// This structure internally contains a `HANDLE` which represents the named
+/// pipe, and also maintains state associated with the mio event loop and active
+/// I/O operations that have been scheduled to translate IOCP to a readiness
+/// model.
+pub struct NamedPipe {
+    inner: Arc<Inner>,
+}
+
+struct Inner {
```
Thomasdezeeuw, line 100: I believe for this to work with `overlapped2arc`, or more specifically `offset_of`, this should be `#[repr(C)]`.

See https://github.com/Gilnaa/memoffset/pull/43, https://github.com/rust-lang/rust/issues/48956#issuecomment-544506419 and https://github.com/rust-lang/rust/issues/64490#issuecomment-657089537.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+fn connect_done(status: &OVERLAPPED_ENTRY) {
+    let status = CompletionStatus::from_entry(status);
+
+    // Acquire the `Arc<Inner>`. Note that we should be guaranteed that
+    // the refcount is available to us due to the `mem::forget` in
+    // `connect` above.
+    let me = unsafe { overlapped2arc!(status.overlapped(), Inner, connect) };
+
+    // Flag ourselves as no longer using the `connect` overlapped instances.
+    let prev = me.connecting.swap(false, SeqCst);
```
Thomasdezeeuw, line 492: Nit: this can be lowered to `AcqRel`.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+        let res = unsafe {
+            let overlapped = self.inner.connect.as_mut_ptr() as *mut _;
+            self.inner.handle.connect_overlapped(overlapped)
+        };
+
+        match res {
+            // The connection operation finished immediately, so let's schedule
+            // reads/writes and such.
+            Ok(true) => {
+                self.inner.connecting.store(false, SeqCst);
```
Thomasdezeeuw, line 232: Nit: this can be lowered to `AcqRel`.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+            // `connect_done` function will "reify" this forgotten pointer to
+            // drop the refcount on the other side.
+            Ok(false) => {
+                mem::forget(self.inner.clone());
+                Err(would_block())
+            }
+
+            // TODO: are we sure no IOCP notification comes in here?
+            Err(e) => {
+                self.inner.connecting.store(false, SeqCst);
```
Thomasdezeeuw, line 251: Nit: this can be lowered to `AcqRel`.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+    let status = CompletionStatus::from_entry(status);
+
+    // Acquire the `Arc<Inner>`. Note that we should be guaranteed that
+    // the refcount is available to us due to the `mem::forget` in
+    // `connect` above.
+    let me = unsafe { overlapped2arc!(status.overlapped(), Inner, connect) };
+
+    // Flag ourselves as no longer using the `connect` overlapped instances.
+    let prev = me.connecting.swap(false, SeqCst);
+    assert!(prev, "wasn't previously connecting");
```
Thomasdezeeuw, line 493: Can we include `mio::windows::NamedPipe` in the message if it does ever come up?

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+            }
+            Err(e) => {
+                debug_assert_eq!(status.bytes_transferred(), 0);
+                io.read = State::Err(e);
+            }
+        }
+    }
+
+    // Flag our readiness that we've got data.
+    if let Some(waker) = io.read_waker.take() {
```
Thomasdezeeuw, line 538: This should not call a waker, but add an event to the selector.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+                }
+            }
+        }
+    }
+
+    fn post_register(me: &Arc<Inner>) {
+        let mut io = me.io.lock().unwrap();
+        if Inner::schedule_read(&me, &mut io) {
+            if let State::None = io.write {
+                if let Some(waker) = io.write_waker.take() {
```
Thomasdezeeuw, line 458: This should add an event to the selector, not call a waker.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+                        waker.wake();
+                    }
+                } else {
+                    Inner::schedule_write(&me, buf, new_pos, &mut io);
+                }
+            }
+            Err(e) => {
+                debug_assert_eq!(status.bytes_transferred(), 0);
+                io.write = State::Err(e);
+                if let Some(waker) = io.write_waker.take() {
```
Thomasdezeeuw, line 576: Should add an event to the selector.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+    };
+
+    unsafe {
+        match me.handle.result(status.overlapped()) {
+            Ok(n) => {
+                debug_assert_eq!(status.bytes_transferred() as usize, n);
+                let new_pos = pos + (status.bytes_transferred() as usize);
+                if new_pos == buf.len() {
+                    me.put_buffer(buf);
+                    if let Some(waker) = io.write_waker.take() {
```
Thomasdezeeuw, line 566: Should add an event to the selector.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+    }
+
+    fn get(&mut self, default_cap: usize) -> Vec<u8> {
+        self.pool
+            .pop()
+            .unwrap_or_else(|| Vec::with_capacity(default_cap))
+    }
+
+    fn put(&mut self, mut buf: Vec<u8>) {
+        if self.pool.len() < self.pool.capacity() {
```
Thomasdezeeuw, line 603: Do we want to limit ourselves to just 2 buffers? If the client is using more we might as well keep them around. Or did you find a case where the pool is ever growing?

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+                io.write = State::Err(e);
+                if let Some(waker) = io.write_waker.take() {
+                    waker.wake();
+                }
+            }
+        }
+    }
+}
+
+// Based on https://github.com/tokio-rs/mio/blob/13d5fc9/src/sys/windows/buffer_pool.rs
```
Thomasdezeeuw, line 584: Do we need to refer to our own code?

src/sys/windows/overlapped.rs
```diff
@@ -0,0 +1,64 @@
...
+    }
+}
+
+impl fmt::Debug for Overlapped {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Overlapped").finish()
+    }
+}
+
+// Overlapped's APIs are marked as unsafe Overlapped's APIs are marked as
```
Thomasdezeeuw, line 60: `Overlapped's APIs are marked as unsafe` twice. And the functions aren't marked as unsafe.

src/sys/windows/overlapped.rs
```diff
@@ -0,0 +1,64 @@
...
+            inner: UnsafeCell::new(miow::Overlapped::zero()),
+            callback: cb,
+        }
+    }
+
+    /// Get the underlying `Overlapped` instance as a raw pointer.
+    ///
+    /// This can be useful when only a shared borrow is held and the overlapped
+    /// pointer needs to be passed down to winapi.
+    pub fn as_mut_ptr(&self) -> *mut OVERLAPPED {
```
Thomasdezeeuw, line 49: This should be `&mut self` as we're returning a `*mut` pointer.

src/sys/windows/overlapped.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[cfg(feature = "os-util")]
+impl Overlapped {
+    /// Creates a new `Overlapped` which will invoke the provided `cb` callback
+    /// whenever it's triggered.
+    ///
+    /// The returned `Overlapped` must be used as the `OVERLAPPED` passed to all
+    /// I/O operations that are registered with mio's event loop. When the I/O
+    /// operation associated with an `OVERLAPPED` pointer completes the event
+    /// loop will invoke the function pointer provided by `cb`.
```
Thomasdezeeuw, line 37: Comment seems outdated, e.g. Mio doesn't have an event loop.

src/sys/windows/selector.rs
```diff
@@ -438,18 +438,8 @@ impl SelectorInner {
...
-                if len == 0 {
-                    continue;
-                }
-                return Ok(());
-            }
-        } else {
-            self.select2(&mut events.statuses, &mut events.events, timeout)?;
-            return Ok(());
-        }
+        self.select2(&mut events.statuses, &mut events.events, timeout)?;
```
Thomasdezeeuw, line 441: There is a reason why the code was a little complicated: https://github.com/tokio-rs/mio/issues/1106, https://github.com/tokio-rs/mio/commit/bbd82672ab5adab7cd68e765810d1d1809b90c46. Are you sure this still works?

src/sys/windows/selector.rs
```diff
@@ -438,18 +438,8 @@ impl SelectorInner {
...
+        return Ok(());
```
Thomasdezeeuw, line 442: Nit: if we keep this you can remove this line and `?;` from the previous one.

tests/win_named_pipe.rs
```diff
@@ -0,0 +1,224 @@
...
+        match $e {
+            Ok(e) => e,
+            Err(e) => panic!("{} failed with {}", stringify!($e), e),
+        }
+    };
+}
+
+fn server(registry: &mio::Registry) -> (NamedPipe, String) {
+    let num: u64 = rand::thread_rng().gen();
+    let name = format!(r"\\.\pipe\my-pipe-{}", num);
```
Thomasdezeeuw, line 28: Nit: we could just use the test name here, then we don't need to add `rand` as dev-dependency.

tests/win_named_pipe.rs
```diff
@@ -0,0 +1,224 @@
+#![cfg(windows)]
+
+use std::fs::OpenOptions;
+use std::io;
+use std::os::windows::fs::*;
+use std::os::windows::io::*;
+use std::task::{Context, Poll};
+use std::time::Duration;
+
+// use mio::{Events, Poll, PollOpt, Ready, Token};
```
Thomasdezeeuw, line 10: Remove.

tests/win_named_pipe.rs
```diff
@@ -0,0 +1,224 @@
...
+use mio::windows::NamedPipe;
+use rand::Rng;
+use winapi::um::winbase::*;
```
Thomasdezeeuw, line 13: Nit: not the biggest fan of `*` imports.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,610 @@
...
+//! `FILE_FLAG_OVERLAPPED` flag when opening the `File`.
+
+use crate::{poll, Registry};
+use crate::sys::windows::Overlapped;
+
+use std::ffi::OsStr;
+use std::fmt;
+use std::io::{self, Read};
+use std::mem;
+use std::os::windows::io::*;
```
Thomasdezeeuw, line 57: Nit: not the biggest fan of `*` imports.

tests/win_named_pipe.rs
```diff
@@ -0,0 +1,224 @@
+#![cfg(windows)]
+
+use std::fs::OpenOptions;
+use std::io;
+use std::os::windows::fs::*;
```
Thomasdezeeuw, line 5: Nit: not the biggest fan of `*` imports.

tests/win_named_pipe.rs
```diff
@@ -0,0 +1,224 @@
...
+
+fn pipe(registry: &mio::Registry) -> (NamedPipe, NamedPipe) {
+    let (pipe, name) = server(registry);
+    (pipe, client(&name, registry))
+}
+
+static data: &[u8] = &[100; 4096];
+
+#[test]
+fn writable_after_register() {
```
Thomasdezeeuw, line 50: These tests should be updated to not use the task system.

Thomasdezeeuw (changes requested):
I'm mostly OK with the code, however this uses the task/`Future` system while other Mio types do not. I don't think we should change to it now. I'm added a number of inline comments, but a large number of them are small nits.

src/sys/windows/named_pipe.rs
carllerche: I couldn't think of a different way to implement this w/o any overhaul of mio, adding back the custom readiness queue. That seemed much more overkill than this. Since IOCP is a callback system, we need some way to signal the callback fired. Using `Waker` in std is the easiest way I can think of doing this (it's essentially just an argument-less callback).

Since this is mostly to support Tokio upgrading to Mio 0.7, I would  be OK flagging the type as `doc(hidden)` if you want. Thoughts?

carllerche: This type is primarily here to support the Tokio upgrade, so I could also flag this as `#[doc(hidden)]` for now.

Thomasdezeeuw: We could do something similar to what we do with the `Waker` here: https://github.com/tokio-rs/mio/blob/086251c49dc3ec5dba1483980bc0023a9ea4ee94/src/sys/windows/selector.rs#L491-L505

We would call the callback, but instead of the callback calling `task::Waker.wake`, the code linked above in `SelectorInner::feed_events` would create a new `Event` and add it to the events.

carllerche: I investigated this strategy, there are a number of scenarios in which we need to be able to fire an event w/o having access to the the events list. For example, if `NamedPipe::connect()` is called, a write event needs to be able to fire, but no events list is present. Simply saving off the ready state is not sufficient because something needs to wake up the selector.

carllerche: Ok, I worked around this by using the IOCP handle to queue events.

tests/regressions.rs
carllerche: Ah, I need to remove this in a separate PR probably... the test is not correct. Spurious wakeups should be permitted.

src/sys/windows/named_pipe.rs
carllerche: This isn't even true. The caller can pass in a random value and there is no risk of memory unsafety. The `unsafe` annotation is from early rust days when `unsafe` was used more liberally to indicate the caller needed to try not to write bugs.

src/sys/windows/named_pipe.rs
carllerche: This is not obviously true to me. How would this interact w/ the stores?

src/sys/windows/named_pipe.rs
carllerche: This was polled from the original impl :shrug: will delete.

src/sys/windows/named_pipe.rs
carllerche: This comment is no longer accurate (based on an older impl), i will remove.

src/sys/windows/named_pipe.rs
carllerche: I don't believe this is true. Switching to `AcqRel` permits it to load stale values, i.e. we can see values that are not the "newest".

src/sys/windows/named_pipe.rs
carllerche: The pool is scoped per NamedPipe and there is at most 1 in flight buffer per direction (max of 2).

src/sys/windows/selector.rs
carllerche: Spurious wakeups are legal. Either way, I am reverting.

tests/win_named_pipe.rs
carllerche: Unless the tests are run in multiple threads :P

carllerche: (multiple processes)

carllerche:
@Thomasdezeeuw Ok, the feedback should be addressed. I opted to stay w/ `SeqCst` as the lowering was not obviously correct to me and I didn't want to do a deeper analysis. Because `load` and `store` is being used, w/o `SeqCst` there is no guarantee that the atomic ops see the "latest value".

carllerche:
Ok, I tested that this works for Tokio. :+1:

Thomasdezeeuw:
> @Thomasdezeeuw Ok, the feedback should be addressed. I opted to stay w/ `SeqCst` as the lowering was not obviously correct to me and I didn't want to do a deeper analysis. Because `load` and `store` is being used, w/o `SeqCst` there is no guarantee that the atomic ops see the "latest value".

If I understood correctly, the only benefit `SeqCst` provides over `AcqRel` is that it guarantees the ordering in relation to atomic operations on on data location. In other words if we had two `AtomicBool`s and we did a store on 1 and then on 2, that order has to be preserved with `SeqCst`, while with `AcqRel` the compiler/hardware is allowed to switch them around. But as we're only using a single atomic location that shouldn't matter.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,671 @@
...
+                }),
+                pool: Mutex::new(BufferPool::with_capacity(2)),
+            }),
+        }
+    }
+}
+
+impl Read for NamedPipe {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        <&NamedPipe as Read>::read(&mut &*self, buf)
```
Thomasdezeeuw, line 259: Nit: this doesn't create a `&dyn Read`  pointer does it? Why not use `(&*self).read(buf)`?

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,671 @@
...
+
+        // Move `buf` onto the heap and fire off the write
+        let mut owned_buf = self.inner.get_buffer();
+        owned_buf.extend(buf);
+        Inner::schedule_write(&self.inner, owned_buf, 0, &mut io, None);
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> io::Result<()> {      
+        Ok(())
```
Thomasdezeeuw, line 350: This combined with the  scheduled write above might give issues were a user thinks the bytes have been written to the pipe and closes the application, but it's not actually the case.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,671 @@
...
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> io::Result<()> {      
+        Ok(())  
+    }
+}
+
+impl Source for NamedPipe {
+    fn register(&mut self, registry: &Registry, token: Token, interest: Interest) -> io::Result<()> {
```
Thomasdezeeuw, line 355: Other types failed to register twice with ```Err(io::Error::new(io::ErrorKind::AlreadyExists, "I/O source already registered with a `Registry`"))```, at least in debug mode. We'll want to do the same here.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,671 @@
...
+        io.read_interest = interest.is_readable();
+        io.write_interest = interest.is_writable();
+        drop(io);
+
+        Inner::post_register(&self.inner, None);
+
+        Ok(())
+    }
+
+    fn reregister(&mut self, _registry: &Registry, token: Token, interest: Interest) -> io::Result<()> {
```
Thomasdezeeuw, line 375: Same as for `register`, we fail with ```Err(io::Error::new(io::ErrorKind::NotFound, "I/O source not registered with `Registry`"))```if not already registered.

src/sys/windows/named_pipe.rs
```diff
@@ -0,0 +1,671 @@
...
+        io.read_interest = interest.is_readable();
+        io.write_interest = interest.is_writable();
+        drop(io);
+
+        Inner::post_register(&self.inner, None);
+
+        Ok(())
+    }
+
+    fn deregister(&mut self, _registry: &Registry) -> io::Result<()> {
```
Thomasdezeeuw, line 388: Same as for `reregister`, same error.

tests/win_named_pipe.rs
Thomasdezeeuw: True, but we already use something like this in `temp_file` (in `tests/util/mod.rs::240`).

tests/win_named_pipe.rs
```diff
@@ -0,0 +1,267 @@
+#![cfg(windows)]
```
Thomasdezeeuw, line 1: Nit: These tests can use some of the test utilities in `tests/util`, such as `expect_events`, but that can be done later.

tests/win_named_pipe.rs
```diff
@@ -0,0 +1,267 @@
...
+        if let Some(event) = e.iter().find(|e| e.token() == Token(0)) {
+            if event.is_writable() {
+                break;
+            }
+        }
+    }
+}
+
+#[test]
+fn write_then_drop() {
```
Thomasdezeeuw, line 181: Re: not implementing `flush` (or a no-op implementation) above. I think running this test with the client in a different process that stops right after the write (line 194) would fail.

Thomasdezeeuw (approved):
Couple more small things, but LGTM.

src/sys/windows/named_pipe.rs
Thomasdezeeuw: True, it isn't about memory unsafety. However a lot of types that implement `FromRawHandle` expect the handle to be valid once created. This is a part of the contract the user of `FromRawHandle` must uphold, but the compiler can't check, and for that reason it uses `unsafe` to notify the user that not every value passed is valid (even though the compiler allows it).

I've always seen `unsafe` having a double function: 1) saying to the compiler to leave me alone, I know what I'm doing (at the call side), and 2) informing the caller needs to ensure something the compiler can't check (e.g. functions contracts such as using valid handles). We should really have two keywords for it, but here we are.

carllerche:
> > @Thomasdezeeuw Ok, the feedback should be addressed. I opted to stay w/ `SeqCst` as the lowering was not obviously correct to me and I didn't want to do a deeper analysis. Because `load` and `store` is being used, w/o `SeqCst` there is no guarantee that the atomic ops see the "latest value".
> 
> If I understood correctly, the only benefit `SeqCst` provides over `AcqRel` is that it guarantees the ordering in relation to atomic operations on on data location. In other words if we had two `AtomicBool`s and we did a store on 1 and then on 2, that order has to be preserved with `SeqCst`, while with `AcqRel` the compiler/hardware is allowed to switch them around. But as we're only using a single atomic location that shouldn't matter.

`SeqCst` gets you total order on a single atomic cell as well. `load(Acquire)` is permitted to return stale values as long as the coherence rules are followed. `store(Release)` has no requirement to immediately make the store visible to other threads as long as coherence rules are followed. It is a common mistake to make and I have made it a few times myself. If you want, you can play around w/ this w/ loom to see how it would behave.

carllerche:
@Thomasdezeeuw `AcqRel` does **not** guarantee that for stores and loads. If, given a single atomic cell, only read-modify-write operation are performed, then AcqRel is sufficient to establish a linear modification order. This is because of the guarantees that come w/ read-modify-write operations. However, once `load` and `store` is thrown into the mix, all bets are off.

I have seen this in practice and the [documentation](https://en.cppreference.com/w/cpp/atomic/memory_order) indicates as such.

carllerche:
Look at the ASM generated by this:

```rust
use std::sync::atomic::AtomicUsize;
use std::sync::atomic::Ordering::{Acquire, Release};

static CELL: AtomicUsize = AtomicUsize::new(0);


pub fn main() -> usize {
CELL.store(1, Release);
CELL.load(Acquire)
}
```

```asm
example::main:
        mov     qword ptr [rip + example::CELL], 1
        mov     rax, qword ptr [rip + example::CELL]
        ret

example::CELL:
        .zero   8
```

Plain `mov` instructions. No synchronization. Stale values can be obtained when a CPU loads the value from its cache w/o synchronizing across threads.

src/sys/windows/named_pipe.rs
carllerche: It does not create a dyn reader. I just copied the existing code. My guess is the intent is to make it clear it is calling the ref impl and not end up in an infinite loop.

carllerche:
Alright! CI passes :tada: let me investigate Tokio now.

Thomasdezeeuw:
> Look at the ASM generated by this:
> 
> ```rust
> use std::sync::atomic::AtomicUsize;
> use std::sync::atomic::Ordering::{Acquire, Release};
> 
> static CELL: AtomicUsize = AtomicUsize::new(0);
> 
> 
> pub fn main() -> usize {
> CELL.store(1, Release);
> CELL.load(Acquire)
> }
> ```
> 
> ```assembly
> example::main:
>         mov     qword ptr [rip + example::CELL], 1
>         mov     rax, qword ptr [rip + example::CELL]
>         ret
> 
> example::CELL:
>         .zero   8
> ```
> 
> Plain `mov` instructions. No synchronization. Stale values can be obtained when a CPU loads the value from its cache w/o synchronizing across threads.

The `mov` for the load is expected, as an <= 8 bytes load is always atomic on x86, but I would expect `xchg` for the store. I learned something new, thanks!

For cas operations it doesn't matter and I incorrectly assumed the same to be true for load/store operations.

```rust
use std::sync::atomic::{AtomicUsize, Ordering};

static CELL: AtomicUsize = AtomicUsize::new(0);

pub fn main() {
    CELL.fetch_add(1, Ordering::AcqRel);
    CELL.fetch_add(1, Ordering::SeqCst);
    CELL.load(Ordering::SeqCst);
}
```

```asm
example::main:
        lock            add     qword ptr [rip + example::CELL], 1
        lock            add     qword ptr [rip + example::CELL], 1
        mov     rax, qword ptr [rip + example::CELL]
        ret

example::CELL:
        .zero   8
```

https://rust.godbolt.org/z/1r5M4Y

bbqsrc:
I've deprecated and directed users of the old mio-named-pipes to mio 0.7. Thanks!

PR #1356, opened 2020-09-30
Add unix::pipe

### LINKED ISSUE #1354: Proposal: move Mio-pipe into Mio
The Mio-pipe crate provides a simple wrapper around a Unix pipe. It's currently ~300 lines of code, but I think that can be reduced if it's lives inside the Mio repo, mainly by using [`IoSource`](https://github.com/tokio-rs/mio/blob/784d6ffeab9df9ee2f47e3cec787577c740088c5/src/io_source.rs).

One current disadvantage is that it doesn't have Windows support. But @carllerche has been working on that in https://github.com/tokio-rs/mio/pull/1351.

Source: https://github.com/Thomasdezeeuw/mio-pipe
Docs: https://docs.rs/mio-pipe/0.1.1/mio_pipe

Thomasdezeeuw:
Adds one new function: unix::pipe, which is a wrapper around the pipe(2)
system call, and two new types: Sender and Receiver, wrappers around the
file descriptors.

This is a port of https://github.com/Thomasdezeeuw/mio-pipe, commit
8c3025edf128e90733e95327d88493887b93fcdd.

Closes #1354.


Thomasdezeeuw:
This adds 189 lines of code (381 total) in `sys::unix`, and ~200 lines of test code.

Thomasdezeeuw:
These feature matrices are getting out hand.. we're up to 65 (!) combinations now.

src/lib.rs
```diff
@@ -82,11 +84,20 @@ pub use poll::{Poll, Registry};
...
+)]
 pub mod unix {
     //! Unix only extensions.
+
+    #[cfg(feature = "os-util")]
+    #[cfg_attr(docsrs, doc(cfg(all(unix, feature = "os-util"))))]
     pub use crate::sys::SourceFd;
+    #[cfg(all(feature = "pipe", feature = "os-poll"))]
+    #[cfg_attr(docsrs, doc(cfg(all(unix, feature = "pipe", feature = "os-poll"))))]
+    pub use crate::sys::{pipe, Receiver, Sender};
```
carllerche, line 100: Having pipe handles named `Receiver` and `Sender` in `unix` is a bit unclear. I would either have it be `unix::pipe::Receiver` or `unix::PipeReceiver` (same w/ sender).

carllerche (approved):
Looks good to me. I did have a thought re: the handle names.

src/lib.rs
Thomasdezeeuw: I've created a new module `unix::pipe`, with `Sender`, `Receiver` and `new`. So it can be created with `pipe::new`, and receiver accessed with `pipe::Sender`.

PR #1362, opened 2020-10-07
remove publish = false

carllerche:
We are actively publishing 0.7.x


Thomasdezeeuw (approved):
Hit this a couple times to.

PR #1368, opened 2020-10-14
Drop net2 dev-dependency

Keruspe:
Now that we have out own TcpSocket builder inside of mio, just add a set_linger method to it and switch away from net2.

This doesn't bring socket2 in and hopefully will stop all the recurring "please switch from net2 to socket2" reports

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,6 @@ ntapi  = "0.3"
 
 [dev-dependencies]
 env_logger = { version = "0.6.2", default-features = false }
-net2       = "0.2.33"
 rand = "0.4"
 
 [package.metadata.docs.rs]
--- a/src/sys/unix/tcp.rs
+++ b/src/sys/unix/tcp.rs
@@ -55,8 +55,7 @@ pub(crate) fn set_reuseaddr(socket: TcpSocket, reuseaddr: bool) -> io::Result<()
         libc::SO_REUSEADDR,
         &val as *const libc::c_int as *const libc::c_void,
         size_of::<libc::c_int>() as libc::socklen_t,
-    ))?;
-    Ok(())
+    )).map(|_| ())
 }
 
 pub(crate) fn set_linger(socket: TcpSocket, dur: Option<Duration>) -> io::Result<()> {
@@ -70,8 +69,7 @@ pub(crate) fn set_linger(socket: TcpSocket, dur: Option<Duration>) -> io::Result
         libc::SO_LINGER,
         &val as *const libc::linger as *const libc::c_void,
         size_of::<libc::linger>() as libc::socklen_t,
-    ))?;
-    Ok(())
+    )).map(|_| ())
 }
 
 pub fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream, SocketAddr)> {
--- a/src/sys/unix/tcp.rs
+++ b/src/sys/unix/tcp.rs
@@ -1,6 +1,7 @@
 use std::io;
 use std::mem::{size_of, MaybeUninit};
 use std::net::{self, SocketAddr};
+use std::time::Duration;
 use std::os::unix::io::{AsRawFd, FromRawFd};
 
 use crate::sys::unix::net::{new_socket, socket_addr, to_socket_addr};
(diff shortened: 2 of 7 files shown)


Keruspe:
Oops, there were some stale commits in my local clone, dropped those.

Keruspe:
Clippy failure is unrelated (and probably requires updating the MSRV to "fix")

Thomasdezeeuw:
Mio v0.7 doesn't have net2 as dependency, only as dev-dependency for https://github.com/tokio-rs/mio/blob/a07d38bf61ee7ce54c36cbbab357881e073eae6f/tests/tcp.rs#L464-L545. See https://github.com/tokio-rs/mio/blob/a5381b3e477ece2ef0db57adafad145746bd33cb/Cargo.toml#L50.

The problem with all the issues for net2 is Tokio 0.2, which depends on Mio v0.6. But since Mio v0.6 is in maintenance mode I'm not going to replace a foundational crate for it at this stage. Hopefully Tokio 0.3, which depends on Mio v0.7, should be released soon.

Thomasdezeeuw:
So this is really a pr to add `TcpStream::set_linger`?

src/sys/unix/tcp.rs
```diff
@@ -58,6 +59,21 @@ pub(crate) fn set_reuseaddr(socket: TcpSocket, reuseaddr: bool) -> io::Result<()
...
+        l_onoff: if dur.is_some() { 1 } else { 0 },
+        l_linger: dur.map(|dur| dur.as_secs() as libc::c_int).unwrap_or_default(),
+    };
+    syscall!(setsockopt(
+        socket,
+        libc::SOL_SOCKET,
+        libc::SO_REUSEADDR,
+        &val as *const libc::linger as *const libc::c_void,
+        size_of::<libc::linger>() as libc::socklen_t,
+    ))?;
```
Thomasdezeeuw, line 73: Nit: don't need `?`, can use `map(|_| ())`.

Keruspe:
@Thomasdezeeuw It just adds TcpSocket::set_linger and replaces the use of net2 in tests by TcpSocket now that we have it, yep.

I know net2 is only a dev-dependency in mio 0.7, but if we even remove the dev-dependency, requests to add socket2 instead might stop, since it won't be listed at all in master?

src/sys/windows/tcp.rs
```diff
@@ -89,6 +90,29 @@ pub(crate) fn set_reuseaddr(socket: TcpSocket, reuseaddr: bool) -> io::Result<()
     }
 }
 
+pub(crate) fn set_linger(socket: TcpSocket, dur: Option<Duration>) -> io::Result<()> {
+    use winapi::ctypes::{c_char, c_int, c_ushort};
+    use winapi::um::winsock2::{linger, setsockopt, SOCKET_ERROR, SO_LINGER, SOL_SOCKET};
+    use std::mem::size_of;
```
Thomasdezeeuw, line 96: Nit: can you put these with the other use statements?

Keruspe:
And once tokio 0.3 is out, it will definitely stop, yep :+1:

src/sys/windows/tcp.rs
```diff
@@ -89,6 +90,29 @@ pub(crate) fn set_reuseaddr(socket: TcpSocket, reuseaddr: bool) -> io::Result<()
...
+    match unsafe { setsockopt(
+        socket,
+        SOL_SOCKET,
+        SO_LINGER,
+        &val as *const _ as *const c_char,
+        size_of::<linger>() as c_int,
+    ) } {
+        0 => Ok(()),
+        SOCKET_ERROR => Err(io::Error::last_os_error()),
+        _ => panic!("unexpected return value"),
```
Thomasdezeeuw, line 112: Let's not panic. Let's only return an error on `SOCKET_ERROR` and otherwise `Ok(())`.

Cargo.toml
```diff
@@ -47,7 +47,6 @@ ntapi  = "0.3"
 
 [dev-dependencies]
 env_logger = { version = "0.6.2", default-features = false }
-net2       = "0.2.33"
```
Thomasdezeeuw, line 50: Nice.

Thomasdezeeuw (commented):
Couple small things, but LGTM otherwise.

Thomasdezeeuw:
> @Thomasdezeeuw It just adds TcpSocket::set_linger and replaces the use of net2 in tests by TcpSocket now that we have it, yep.
> 
> I know net2 is only a dev-dependency in mio 0.7, but if we even remove the dev-dependency, requests to add socket2 instead might stop, since it won't be listed at all in master?

Let's hope that is true :)

src/sys/windows/tcp.rs
Keruspe: This is what set_reuseaddr does too just above

Keruspe:
Fixed the nits, and fixed them in set_reuseaddr too while at it

src/sys/windows/tcp.rs
Thomasdezeeuw: Oh than I missed that in reviewing at some point. Can you remove both? Ideally we don't create a panic branch that is never reached (assuming Windows behaves as documented).

Keruspe: Done

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @Keruspe.

PR #1370, opened 2020-10-19
Fix HUP notifications on windows

carllerche:
Currently, on windows, when receiving a "writeable" notification, HUP interest is disabled. This results in not receiving HUP notifications.

Fixes: https://github.com/tokio-rs/tokio/issues/2982

PR #1371, opened 2020-10-19
prepare v0.7.4 release

carllerche:
This fixes a lost event bug on windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+# 0.7.4
+
+## Fixes
+
+* lost "socket closed" events on windows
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw (changes requested):
I don't think the changes in #1370 should have been committed, so I'm -1 on this.

carllerche:
@Thomasdezeeuw i don’t see any disagreement with the change to the lib, just the test. The patch fixes a significant bug that is currently live. Can we push the patch and figure out then figure out the test?

Thomasdezeeuw (approved):

PR #1372, opened 2020-10-20
Follow up to  #1370

Thomasdezeeuw:


carllerche:
This test is not equivalent. The test will pass without the bug fix. This is because linger set to zero prevents the OS from performing an implicit graceful shutdown.

Thomasdezeeuw:
> This test is not equivalent. The test will pass without the bug fix. This is because linger set to zero prevents the OS from performing an implicit graceful shutdown.

Ah, then I miss understood the purpose of the test. I though that setting linger to zero would the test for run quicker.

Thomasdezeeuw:
Closing in favour of #1374.

PR #1373, opened 2020-10-20
Enable some more TcpStream tests on Windows

Thomasdezeeuw:
After commit 50c299aca56c4a26e5ed20c283007239fbe6a7a7 was merged some of
these tests might pass, the CI will show which.


Thomasdezeeuw:
These still all fail.

Issue #1332, opened 2020-06-24
TCP connect_error test is flaky on macOS


Thomasdezeeuw:
The last call to `expect_no_events` might be too strict: https://github.com/tokio-rs/mio/blob/68805cbeb7797e0d2653e861e009cc10c05bd4f9/tests/tcp.rs#L578

Error log and backtrace:

```log
---- connect_error stdout ----
[ERROR tcp::util] unexpected event: Event { token: Token(0), readable: false, writable: true, error: false, read_closed: false, write_closed: true, priority: false, aio: false, lio: false }
thread 'connect_error' panicked at 'received events, but didn't expect any, see above', tests/util/mod.rs:172:9
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1076
   5: std::io::Write::write_fmt
             at /rustc/feb3536eba10c2e4585d066629598f03d5ddc7c6/src/libstd/io/mod.rs:1537
   6: std::io::impls::<impl std::io::Write for alloc::boxed::Box<W>>::write_fmt
             at src/libstd/io/impls.rs:176
   7: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   8: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   9: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:198
  10: std::panicking::default_hook
             at src/libstd/panicking.rs:215
  11: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:477
  12: std::panicking::begin_panic
             at src/libcore/alloc/layout.rs:363
  13: tcp::util::expect_no_events
             at tests/util/mod.rs:172
  14: tcp::connect_error
[...]

PR #1374, opened 2020-10-20
tests: use our own TcpSocket instead of socket2

Keruspe:
Alternative for part of #1372

### DIFF
--- a/tests/tcp_stream.rs
+++ b/tests/tcp_stream.rs
@@ -751,7 +751,7 @@ fn start_listener(
 }
 
 #[test]
-fn hup() {
+fn hup_event_on_disconnect() {
     use mio::net::TcpListener;
 
     let (mut poll, mut events) = init_with_poll();
@@ -782,6 +782,7 @@ fn hup() {
     );
 
     let (sock, _) = listener.accept().unwrap();
+    // Prevent the OS from performing a graceful shutdown
     set_linger_zero(&sock);
     drop(sock);
 
--- a/tests/tcp_stream.rs
+++ b/tests/tcp_stream.rs
@@ -794,17 +794,17 @@ fn hup() {
 
 #[cfg(windows)]
 fn set_linger_zero(socket: &TcpStream) {
-    use socket2::Socket;
+    use mio::net::TcpSocket;
     use std::os::windows::io::{AsRawSocket, FromRawSocket};
 
-    let s = unsafe { Socket::from_raw_socket(socket.as_raw_socket()) };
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw:
@Keruspe I prefer this over my on changes in #1372. Can do three things for me:
* Change the test name to `hup_event_on_disconnect`.
* Add a comment to the `set_linger` call in the test that it is done to prevent a graceful shutdown, per @carllerche comment here: https://github.com/tokio-rs/mio/pull/1372#issuecomment-712899060.
* Pull d77ec615b2544dc061902b2b914896e417abeaaa (or make your own commit).

Then this LGTM.

You can ignore the CI, that is tracked in #1332.

Thomasdezeeuw (approved):
Thanks.

tests/tcp_stream.rs
```diff
@@ -794,17 +795,17 @@ fn hup() {
 
 #[cfg(windows)]
 fn set_linger_zero(socket: &TcpStream) {
-    use socket2::Socket;
+    use mio::net::TcpSocket;
     use std::os::windows::io::{AsRawSocket, FromRawSocket};
 
-    let s = unsafe { Socket::from_raw_socket(socket.as_raw_socket()) };
+    let s = unsafe { TcpSocket::from_raw_socket(socket.as_raw_socket()) };
```
carllerche, line 801: Our TCP socket closes on drop, we probably need to `mem::forget` it after setting linger.

carllerche (changes requested):
Thanks, minor request.

carllerche:
fwiw, I added the test threads = 1 after seeing some failure on linux as well.

Thomasdezeeuw:
> fwiw, I added the test threads = 1 after seeing some failure on linux as well.

But then we should just fix the failure(s). The number of concurrent test threads shouldn't influence the tests, and if it would I would consider it a bug.

carllerche:
I don't believe it had to do w/ the test being buggy. I believe it had to do w/ tests overlapping with a defunct TCP connection, i.e. what lingering protects against. I did not do a deep investigation into it, but assuming everything was correct before the PR, this would be the most likely explanation.

tests/tcp_stream.rs
Keruspe: Done

Thomasdezeeuw:
Ci failure seems unrelated, opened #1376, @carllerche you need to approve it to merge.

carllerche (approved):
Thanks :+1:

Thomasdezeeuw:
Thanks @Keruspe

PR #1375, opened 2020-10-20
Add SO_REUSEPORT support for TCP socket

zekisherif:
Tokio's `TcpSocket` doesn't currently support setting SO_REUSEPORT. 

This PR is to add this functionality (talked with @carllerche ).

Only enabling it for unix systems (except Solaris).

### DIFF
--- a/src/net/tcp/socket.rs
+++ b/src/net/tcp/socket.rs
@@ -84,12 +84,16 @@ impl TcpSocket {
 
     /// Sets the value of `SO_REUSEPORT` on this socket.
     /// Only supported available in unix
-    #[cfg(all(
-        unix,
-        not(target_os = "solaris")
-    ))]
+    #[cfg(all(unix, not(any(target_os = "solaris", target_os = "illumos"))))]
     pub fn set_reuseport(&self, reuseport: bool) -> io::Result<()> {
         sys::tcp::set_reuseport(self.sys, reuseport)
+    }
+
+    /// Get the value of `SO_REUSEPORT` on set for this socket.
+    /// Only supported available in unix
+    #[cfg(all(unix, not(any(target_os = "solaris", target_os = "illumos"))))]
+    pub fn get_reuseport(&self) -> io::Result<bool> {
+        sys::tcp::get_reuseport(self.sys)
     }
 
     /// Sets the value of `SO_LINGER` on this socket.
--- a/src/net/tcp/socket.rs
+++ b/src/net/tcp/socket.rs
@@ -82,6 +82,11 @@ impl TcpSocket {
         sys::tcp::set_reuseaddr(self.sys, reuseaddr)
     }
 
+    /// something
+    pub fn get_reuseaddr(&self) -> io::Result<bool> {
+        sys::tcp::get_reuseaddr(self.sys)
+    }
+
     /// Sets the value of `SO_REUSEPORT` on this socket.
     /// Only supported available in unix
     #[cfg(all(unix, not(any(target_os = "solaris", target_os = "illumos"))))]
--- a/src/net/tcp/socket.rs
+++ b/src/net/tcp/socket.rs
@@ -84,7 +84,10 @@ impl TcpSocket {
 
     /// Sets the value of `SO_REUSEPORT` on this socket.
     /// Only supported available in unix
-    #[cfg(unix)]
+    #[cfg(all(
+        unix,
+        not(target_os = "solaris")
+    ))]
     pub fn set_reuseport(&self, reuseport: bool) -> io::Result<()> {
         sys::tcp::set_reuseport(self.sys, reuseport)
     }
--- a/src/net/tcp/socket.rs
+++ b/src/net/tcp/socket.rs
@@ -82,6 +82,13 @@ impl TcpSocket {
         sys::tcp::set_reuseaddr(self.sys, reuseaddr)
     }
 
+    /// Sets the value of `SO_REUSEPORT` on this socket.
+    /// Only supported available in unix
+    #[cfg(unix)]
+    pub fn set_reuseport(&self, reuseport: bool) -> io::Result<()> {
+        sys::tcp::set_reuseport(self.sys, reuseport)
+    }
+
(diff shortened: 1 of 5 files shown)


src/net/tcp/socket.rs
```diff
@@ -82,6 +82,16 @@ impl TcpSocket {
         sys::tcp::set_reuseaddr(self.sys, reuseaddr)
     }
 
+    /// Sets the value of `SO_REUSEPORT` on this socket.
```
Thomasdezeeuw, line 85: Can you add `pub fn reuse_port(&self) -> io::Result<bool>` to retrieve the value?

src/net/tcp/socket.rs
```diff
@@ -82,6 +82,16 @@ impl TcpSocket {
...
+    /// Only supported available in unix
+    #[cfg(all(
+        unix,
+        not(target_os = "solaris")
```
Thomasdezeeuw, line 89: I don't know if illumos has this.

/cc @tokio-rs/illumos

tests/tcp_socket.rs
```diff
@@ -21,3 +21,15 @@ fn set_reuseaddr() {
...
     let _ = socket.listen(128).unwrap();
 }
+
+#[cfg(all(unix, not(target_os = "solaris")))]
+#[test]
+fn set_reuseport() {
+    let addr = "127.0.0.1:0".parse().unwrap();
+
+    let socket = TcpSocket::new_v4().unwrap();
+    socket.set_reuseport(true).unwrap();
```
Thomasdezeeuw, line 31: Get you add getting reading the setting back once you add that?

Thomasdezeeuw (changes requested):
Overall I'm good with this, some small changes.

src/net/tcp/socket.rs
pfmooney: Correct. It does not expose `SO_REUSEPORT`.
Including it in the negation with solaris would be wise for now.

Thanks for pointing that out.

zekisherif: added illumos

zekisherif:
Messed up the last commit. will fix

pfmooney:
One potential approach would be to emit the equivalent of `ENOTSUP` on those platforms which do not support the socket option.  It would allow building for those platforms, provided the consumer chose to handle the error rather than simply `unwrap()`ing it.  Otherwise it means compile-time errors and more ifdefs for that downstream software to be usable. *shrug*

zekisherif:
The test is failing because set_reuseport doesn't seem to actually apply the value (always setting it to zero).

Out of curiousity I tried to get the value set for `SO_REUSEADDR` and it also doesn't seem to actually apply the value.

zekisherif:
Going to test on a windows machine instead of relying on passing checks here

zekisherif:
@Thomasdezeeuw This is ready for review again. The 1 test failure I believe is flaky (failed once on my local machine, succeeded on successive attempts) and unrelated to my changes.

src/net/tcp/socket.rs
```diff
@@ -82,6 +82,25 @@ impl TcpSocket {
         sys::tcp::set_reuseaddr(self.sys, reuseaddr)
     }
 
+    /// something
```
Thomasdezeeuw, line 85: Docs.

src/sys/unix/tcp.rs
```diff
@@ -58,6 +59,50 @@ pub(crate) fn set_reuseaddr(socket: TcpSocket, reuseaddr: bool) -> io::Result<()
     )).map(|_| ())
 }
 
+pub(crate) fn get_reuseaddr(socket: TcpSocket) -> io::Result<bool> {
+    let mut optval: libc::c_int = unsafe { mem::zeroed() };
```
Thomasdezeeuw, line 63: Nit: this can just be `0`.

src/sys/unix/tcp.rs
```diff
@@ -58,6 +59,50 @@ pub(crate) fn set_reuseaddr(socket: TcpSocket, reuseaddr: bool) -> io::Result<()
...
+    let mut optlen= mem::size_of::<libc::c_int>() as libc::socklen_t;
```
Thomasdezeeuw, line 64: Nit: ss this rustfmted? `optlen=` is incorrect.

src/sys/unix/tcp.rs
```diff
@@ -58,6 +59,50 @@ pub(crate) fn set_reuseaddr(socket: TcpSocket, reuseaddr: bool) -> io::Result<()
...
+        libc::SOL_SOCKET,
+        libc::SO_REUSEPORT,
+        &val as *const libc::c_int as *const libc::c_void,
+        size_of::<libc::c_int>() as libc::socklen_t,
+    )).map(|_| ())
+}
+
+#[cfg(all(unix, not(any(target_os = "solaris", target_os = "illumos"))))]
+pub(crate) fn get_reuseport(socket: TcpSocket) -> io::Result<bool> {
+    let mut optval: libc::c_int = unsafe { mem::zeroed() };
```
Thomasdezeeuw, line 92: Nit: this can be `0`.

src/sys/windows/tcp.rs
```diff
@@ -87,6 +87,22 @@ pub(crate) fn set_reuseaddr(socket: TcpSocket, reuseaddr: bool) -> io::Result<()
     }
 }
 
+pub(crate) fn get_reuseaddr(socket: TcpSocket) -> io::Result<bool> {
+    let mut optval: c_char = unsafe { std::mem::zeroed() };
```
Thomasdezeeuw, line 91: Nit: this can be `0`.

Thomasdezeeuw (commented):
LGTM, apart from some small changes.

src/sys/unix/tcp.rs
zekisherif: I've run cargo fmt and it doesn't catch it. Fixed it manually.

Thomasdezeeuw (approved):

Issue #1381, opened 2020-10-29
Replace TcpSocket with socket2::Socket


Thomasdezeeuw:
Recently the socket2 crate moved to a new home: https://github.com/rust-lang/socket2-rs, and it is to become the socket implementation for the std lib. For v0.8 I would like to remove the `TcpSocket` type and instead let users use the `Socket` type from the `socket2` crate.

I personally not looking forward for the stream of prs adding each and every method `socket2::Socket` already has, such as #1379 and #1375. Since I also maintain socket2, I don't want to maintain the same code twice.

Thomasdezeeuw:
Not adding socket2 as a dependency, instead we're dropping the `TcpSocket`.

PR #1384, opened 2020-10-31
net: add `TcpSocket::set_{send, recv}_buffer_size`

hawkw:
This commit adds methods for setting `SO_RECVBUF` and `SO_SNDBUF` to
Mio's `TcpSocket` type. It would be nice to eventually expose these in
Tokio, so adding them to Mio is the first step. 

See tokio-rs/tokio#3082 for details.

### DIFF
--- a/src/net/tcp/socket.rs
+++ b/src/net/tcp/socket.rs
@@ -106,12 +106,12 @@ impl TcpSocket {
         sys::tcp::set_linger(self.sys, dur)
     }
 
-    /// Sets the value of `SO_RECVBUF` on this socket.
+    /// Sets the value of `SO_RCVBUF` on this socket.
     pub fn set_recv_buffer_size(&self, size: u32) -> io::Result<()> {
         sys::tcp::set_recv_buffer_size(self.sys, size)
     }
 
-    /// Get the value of `SO_RECVBUF` set on this socket.
+    /// Get the value of `SO_RCVBUF` set on this socket.
     pub fn get_recv_buffer_size(&self) -> io::Result<u32> {
         sys::tcp::get_recv_buffer_size(self.sys)
     }
--- a/src/net/tcp/socket.rs
+++ b/src/net/tcp/socket.rs
@@ -106,6 +106,16 @@ impl TcpSocket {
         sys::tcp::set_linger(self.sys, dur)
     }
 
+    /// Sets the value of `SO_RECVBUF` on this socket.
+    pub fn set_recv_buffer_size(&self, size: u32) -> io::Result<()> {
+        sys::tcp::set_recv_buffer_size(self.sys, size)
+    }
+
+    /// Sets the value of `SO_SNDBUF` on this socket.
+    pub fn set_send_buffer_size(&self, size: u32) -> io::Result<()> {
+        sys::tcp::set_send_buffer_size(self.sys, size)
+    }
+
     /// Returns the local address of this socket
     ///
     /// Will return `Err` result in windows if called before calling `bind`
--- a/src/net/tcp/socket.rs
+++ b/src/net/tcp/socket.rs
@@ -112,6 +112,25 @@ impl TcpSocket {
     }
 
     /// Get the value of `SO_RCVBUF` set on this socket.
+    ///
+    /// Note that if [`set_recv_buffer_size`] has been called on this socket
+    /// previously, the value returned by this function may not be the same as
+    /// the argument provided to `set_recv_buffer_size`. This is for the
+    /// following reasons:
+    ///
+    /// * Most operating systems have minimum and maximum allowed sizes for the
+    ///   receive buffer, and will clamp the provided value if it is below the
+    ///   minimum or above the maximum. The minimum and maximum buffer sizes are
+    ///   OS-dependent.
+    /// * Linux will double the buffer size to account for internal bookkeeping
+    ///   data, and returns the doubled value from `getsockopt(2)`. As per `man
+    ///   7 socket`:
+    ///   > Sets or gets the maximum socket receive buffer in bytes. The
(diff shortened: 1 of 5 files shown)


Thomasdezeeuw:
Can you also add the getters and a simple round-trip test?

hawkw:
> Can you also add the getters and a simple round-trip test?

Sure, will do!

tests/tcp_socket.rs
```diff
@@ -56,3 +56,39 @@ fn get_localaddr() {
...
+fn recv_buffer_size_roundtrips() {
+    fn test(size: u32) {
+        println!("testing recv buffer size: {}", size);
+        let socket = TcpSocket::new_v4().unwrap();
+        socket.set_recv_buffer_size(size).unwrap();
+        // As per `man socket(7)`:
+        // > Sets or gets the maximum socket receive buffer in bytes.  The
+        // > kernel doubles this value (to allow space for bookkeeping
+        // > overhead) when it is set using setsockopt(2), and this doubled
+        // > value is returned by getsockopt(2).
```
hawkw, line 88: I'm not sure if this is also the case on Windows/BSD or not...guess we'll find out when CI fails.

src/net/tcp/socket.rs
```diff
@@ -106,6 +106,26 @@ impl TcpSocket {
         sys::tcp::set_linger(self.sys, dur)
     }
 
+    /// Sets the value of `SO_RCVBUF` on this socket.
+    pub fn set_recv_buffer_size(&self, size: u32) -> io::Result<()> {
+        sys::tcp::set_recv_buffer_size(self.sys, size)
+    }
+
+    /// Get the value of `SO_RCVBUF` set on this socket.
+    pub fn get_recv_buffer_size(&self) -> io::Result<u32> {
```
hawkw, line 115: In general, I think the convention in Rust is to call getters like this `recv_buffer_size` rather than `get_recv_buffer_size`, but the existing methods had names beginning with `get_`, so I thought it was better to be consistent, especially since it looks like `get_localaddr` and `get_reuseport` [are now stable APIs and can't easily be renamed](https://github.com/tokio-rs/mio/blob/master/CHANGELOG.md#added).

tests/tcp_socket.rs
hawkw: ...cool, it looks like this [is, in fact, a Linux-specific behavior](https://dev.azure.com/tokio-rs/Tokio/_build/results?buildId=6430&view=logs&j=9ee66ee0-c6b6-50f2-32c8-2bebb82e1137&t=3b0adf12-f909-59d8-0849-13de19d8dc87&l=250)! Yay!

@Thomasdezeeuw, that brings up a question about what the right API to provide is. We could easily add a `#[cfg(target_os = "linux")]` in the tests and only expect a doubled value on Linux. However, since Mio's goal is to provide a cross-platform abstraction, would it make more sense to have the getters divide the value by 2 on Linux, so that the API behaves the same across all platforms? FWIW, I noticed that [`socket2` does _not_ have this behavior](https://github.com/rust-lang/socket2-rs/blob/b0f77842b3357dd571bb86b30b354bc55215f693/src/sys/unix.rs#L896-L901), but I'm not sure if that's relevant or not.

src/net/tcp/socket.rs
Thomasdezeeuw: You're right the `get_` shouldn't have been a thing, I missed that in reviewing. You can leave it as so it's at least consistent with the `TcpSocket` type, I intend on removing it for v0.8 anyway (#1381).

tests/tcp_socket.rs
Thomasdezeeuw: You can change this to `socket.get_recv_buffer_size().unwrap() >= size` with a comment saying that Linux behaviour is different, as long as we have a check that the option is correctly applied I'm fine with it.

Thomasdezeeuw (approved):
LGTM, see my comment in the test to fix the CI.

hawkw:
It looks like one of the test is [failing on FreeBSD CI](https://cirrus-ci.com/task/5730017084702720):
```
---- no_events_after_deregister stdout ----
thread 'no_events_after_deregister' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 57, kind: NotConnected, message: "Socket is not connected" }', tests/tcp_stream.rs:492:35
stack backtrace:
   0:  0x110dfb3 - <unknown>
   1:  0x113043e - <unknown>
   2:  0x109c89d - <unknown>
   3:  0x11066b8 - <unknown>
   4:  0x1110020 - <unknown>
   5:  0x110fcb5 - <unknown>
   6:  0x11106c0 - <unknown>
   7:  0x1110271 - <unknown>
   8:  0x110e3fa - <unknown>
   9:  0x1110221 - <unknown>
  10:  0x112ecda - <unknown>
  11:  0x112eb0a - <unknown>
  12:  0x105a0c4 - <unknown>
  13:  0x1055210 - <unknown>
  14:  0x1069573 - <unknown>
  15:  0x106f3f4 - <unknown>
  16:  0x10c1cd3 - <unknown>
  17:  0x10c02ab - <unknown>
  18:  0x109bc43 - <unknown>
  19:  0x10a10b9 - <unknown>
  20:  0x1117090 - <unknown>
  21: 0x211848ab - <unknown>


failures:
    no_events_after_deregister
```

It seems quite unlikely that the changes in this PR could have actually broken this, since the test that's failing doesn't do anything related to send/recv buffer size:
https://github.com/tokio-rs/mio/blob/7358a31e99630e9a3e78f482eed4e332341194ef/tests/tcp_stream.rs#L473-L506

I'm guessing this is flaky... @Thomasdezeeuw have you seen this test failing spuriously before?

PR #1385, opened 2020-10-31
net: add `TcpSocket::{set, get}_keepalive`

hawkw:
This commit adds methods to `TcpSocket` to enable/disable TCP keepalive,
and set the keepalive idle time, probe interval, and retries, when the
OS supports configuring those values.

Because of how TCP keepalive is configured in Winsock, it's necessary to
use a builder-type API: the interval and idle time are set together as
separate fields on a struct, and if either is not set, it clobbers the
previous value. This means we can't just call that API twice, once with
each value set. Unlike Unix OSes, Windows also doesn't provide any way
to access the current value of these parameters, so we can't just
populate the other fields on the struct with the previous values, either
(incidentally, this also means we can't provide getters on Windows).


src/net/tcp/socket.rs
```diff
@@ -106,6 +106,33 @@ impl TcpSocket {
         sys::tcp::set_linger(self.sys, dur)
     }
 
+    /// Sets whether keepalive messages are enabled to be sent on this socket.
+    ///
+    /// On Unix, this option will set the `SO_KEEPALIVE` as well as the
+    /// `TCP_KEEPALIVE` or `TCP_KEEPIDLE` option (depending on your platform).
```
Thomasdezeeuw, line 112: I've never been a fan of setting both options. `SO_KEEPALIVE` and `TCP_KEEPALIVE` are different, although related, options and should be able to be set separately. But I'm also ok with keeping this as is.

src/sys/windows/tcp.rs
```diff
@@ -149,6 +151,64 @@ pub(crate) fn set_linger(socket: TcpSocket, dur: Option<Duration>) -> io::Result
     }
 }
 
+pub(crate) fn set_keepalive(socket: TcpSocket, dur: Option<Duration>) -> io::Result<()> {
+    // Windows takes the keepalive timeout as a u32 of milliseconds.
+    let dur_ms = dur.map(|dur| {
+        let ms = dur.as_millis();
+        if ms > DWORD::max_value() as u128 {
+            winapi::um::winbase::INFINITE
```
Thomasdezeeuw, line 159: On Unix we use `i32::max_value()`, which I think we should here also (assuming `INFINITE` is actually infinite and not just the max value).

src/net/tcp/socket.rs
hawkw: Hmm, we could separate these and have a `set_keepalive(bool)` and a `set_keepalive_timeoutl(Duration)`... However, what behavior would we expect from `set_keepalive_timeout` if `SO_KEEPALIVE` is unset? Should it enable `SO_KEEPALIVE` first, or just return whatever error the OS sets in that case? 

I'm not immediately aware of what behavior we would expect from setting `TCP_KEEPIDLE`/`TCP_KEEPALIVE` when `SO_KEEPALIVE` is unset, but my guess is that it's OS-specific...

hawkw: I suppose we might also want a separate function to set the interval between keepalive probes...

src/sys/windows/tcp.rs
hawkw: `INFINITE` _is_ just the max value, but I thought it was better to use the winapi constants explicitly, just in case they are ever different?

src/net/tcp/socket.rs
carllerche: If we are going w/ socket options at this level, we *probably* should just expose them as they are provided by the OS?

hawkw: @carllerche just to make sure we're on the same page, you mean having separate methods for
- enabling/disabling keepalive (`SOL_SOCKET` + `SO_KEEPALIVE`)
- setting the keepalive time (`IPPROTO_TCP` + `TCP_KEEPIDLE`/`SO_KEEPALIVE`/`TCP_KEEPALIVE` depending on OS)
- setting the keepalive interval (`IPPROTO_TCP` + `TCP_KEEPINTVL` on Linux, i have no idea what this is on BSD...)
- setting the number of keepalive retries (`IPPROTO_TCP` + `TCP_KEEPCNT` on linux, not sure about BSD, winsock [doesn't support setting this](https://docs.microsoft.com/en-us/windows/win32/winsock/sio-keepalive-vals))

carllerche: I am not familiar enough w/ the portability details of the various options to comment. At the mio layer, we want to try to avoid being too opinionated. It's ok to have platform specific fns, but any given fn *should* behave the same across platforms.

hawkw: Okay, there are now separate API methods for setting each configurable value that the OS exposes. Let me know what you think!

hawkw:
Okay, I've updated this branch to use a builder-style API to workaround the issues with winsock, which I've described in the PR description. Let me know what you think!

src/net/tcp/socket.rs
```diff
@@ -168,7 +191,126 @@ impl TcpSocket {
...
+
+    /// Sets parameters configuring TCP keepalive probes for this socket.
+    ///
+    /// The supported parameters depend on the operating system, and are
+    /// configured using the [`TcpKeepalive`] struct. At a minimum, all systems
+    /// support configuring the [keepalive time]: the time after which the OS
+    /// will start sending keepalive messages on an idle connection.
+    ///
+    /// Additionally, this will enable TCP keepalive on this socket, if it is
+    /// not already enabled.
```
Thomasdezeeuw, line 215: Can you add a `# Notes` section saying that all parameters related to keepalive will be overwritten.

src/sys/unix/tcp.rs
```diff
@@ -202,6 +215,226 @@ pub(crate) fn get_send_buffer_size(socket: TcpSocket) -> io::Result<u32> {
...
+    }
+
+    impl Error for TcpKeepaliveError {
+        fn source(&self) -> Option<&(dyn Error + 'static)> {
+            Some(&self.inner)
+        }
+    }
+
+    fn error(msg: &'static str) -> impl Fn(io::Error) -> io::Error {
+        move |inner| io::Error::new(inner.kind(), TcpKeepaliveError { msg, inner })
```
Thomasdezeeuw, line 266: How much value does this additional message provide? Because `io::Error::new` allocates twice.

Thomasdezeeuw (approved):
Two small things, but LGTM.

This might be the nicest, full-featured TCP keepalive API I've seen, well done.

src/sys/unix/tcp.rs
hawkw: Hmm, I thought it was potentially useful to know which keepalive parameter couldn't be set, but it may not be worth the extra allocation. I can remove this and just forward the OS error unmodified?

Alternatively, we could just make the `TcpKeepaliveError` type part of the public API and return it directly, avoiding the additional allocation from returning it behind an `io::Error`. However, that seemed inconsistent with the rest of the `TcpSocket` API.

hawkw:
> This might be the nicest, full-featured TCP keepalive API I've seen, well done.

Thanks! :) I noticed that `socket2`'s implementation set the same value for the keepalive timeout and keepalive interval on Windows, and thought "hmm, that doesn't seem ideal"...and then I went down a bit of a rabbit-hole when it turned out that doing something reasonable on winsock was a bit harder than it looked at first...

src/sys/unix/tcp.rs
Thomasdezeeuw: Let's just forward the error unmodified, seems like the easiest solution.

src/net/tcp/socket.rs
hawkw: Hmm, that's actually only the case on Windows; on Unix, we can avoid overwriting previously set parameters. Do you think we should have the same behavior on Unix OSes even though it's less than ideal?

Thomasdezeeuw: Let's change it to that it *might* overwrite other keepalive parameters and that either only a single call should be made, or that all desired parameters should be set in all calls. I don't want to make unnecessary system calls on Unix.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @hawkw.

PR #1387, opened 2020-11-05
Support setting SO_LINGER on TcpSocket

zekisherif:
Because tokio `TcpStream` and mio `TcpStream` now use `std::net::TcpStream` as the underlying stream, there isn't a way to set linger options.

I'm aware of #1381 and the dislike of duplicating features. This is the last functionality I need.

The plan is to convert the stream by calling `as_raw_fd()` and set the linger option.

### DIFF
--- a/tests/tcp_socket.rs
+++ b/tests/tcp_socket.rs
@@ -2,6 +2,7 @@
 
 use mio::net::TcpSocket;
 use std::io;
+use std::time::Duration;
 
 #[test]
 fn is_send_and_sync() {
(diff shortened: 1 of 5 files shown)


tests/tcp_socket.rs
```diff
@@ -58,6 +59,19 @@ fn get_localaddr() {
...
 }
 
+#[test]
+fn set_linger() {
+    let addr = "127.0.0.1:0".parse().unwrap();
+
+    let socket = TcpSocket::new_v4().unwrap();
+    socket.set_linger(Some(Duration::from_secs(1))).unwrap();
+    assert_eq!(socket.get_linger().unwrap().unwrap().as_secs(), 1);
+
```
Thomasdezeeuw, line 69: Can you also test `set_ling(None)`?

Thomasdezeeuw (approved):
LGTM if you can fix the CI and the small change to the test.

Thomasdezeeuw:
Thanks @zekisherif.

zekisherif:
Just realized that the commit message should really be "Support getting SO_LINGER value".

The setting of the linger value is meant for tokio TcpStream not mio.

Thomasdezeeuw:
I missed it too, oh well.. it's committed now.

PR #1388, opened 2020-11-06
Don't assume memory layout of std::net::SocketAddr

### LINKED ISSUE #1386: mio::sys::unix::net::socket_addr assumes the layout of std::net::SocketAddrV{4,6} matches libc::sockaddr
@faern noted this in https://internals.rust-lang.org/t/why-are-socketaddrv4-socketaddrv6-based-on-low-level-sockaddr-in-6/13321

https://github.com/tokio-rs/mio/blob/27fbd5f04bb5f52a4d1c358cf0c04c6074a3d46b/src/sys/unix/net.rs#L78-L85

As far as I can tell there are no guarantees from `std` about the layout of `SocketAddrV{4,6}`, and this code could silently compile and cause UB elsewhere if the representation changes (e.g. if padding is added early on in the struct, resulting in C code reading uninitialized bytes).

faern:
Fixes #1386

This is so far only implemented in the `unix` implementation. A very similar thing has to be done for the Windows module. But I felt I could get this posted before that to get some feedback.

### DIFF
--- a/src/sys/unix/net.rs
+++ b/src/sys/unix/net.rs
@@ -84,14 +84,14 @@ impl SocketAddrCRepr {
 /// Converts a Rust `SocketAddr` into the system representation.
 pub(crate) fn socket_addr(addr: &SocketAddr) -> (SocketAddrCRepr, libc::socklen_t) {
     match addr {
-        SocketAddr::V4(ref a) => {
+        SocketAddr::V4(ref addr) => {
             // `s_addr` is stored as BE on all machine and the array is in BE order.
             // So the native endian conversion method is used so that it's never swapped.
-            let sin_addr = libc::in_addr { s_addr: u32::from_ne_bytes(a.ip().octets()) };
+            let sin_addr = libc::in_addr { s_addr: u32::from_ne_bytes(addr.ip().octets()) };
 
             let sockaddr_in = libc::sockaddr_in {
                 sin_family: libc::AF_INET as libc::sa_family_t,
-                sin_port: a.port().to_be(),
+                sin_port: addr.port().to_be(),
                 sin_addr,
                 sin_zero: [0; 8],
                 #[cfg(any(
@@ -108,13 +108,13 @@ pub(crate) fn socket_addr(addr: &SocketAddr) -> (SocketAddrCRepr, libc::socklen_
             let sockaddr = SocketAddrCRepr { v4: sockaddr_in };
             (sockaddr, mem::size_of::<libc::sockaddr_in>() as libc::socklen_t)
         }
-        SocketAddr::V6(ref a) => {
+        SocketAddr::V6(ref addr) => {
             let sockaddr_in6 = libc::sockaddr_in6 {
                 sin6_family: libc::AF_INET6 as libc::sa_family_t,
-                sin6_port: a.port().to_be(),
-                sin6_addr: libc::in6_addr { s6_addr: a.ip().octets() },
-                sin6_flowinfo: a.flowinfo(),
-                sin6_scope_id: a.scope_id(),
+                sin6_port: addr.port().to_be(),
+                sin6_addr: libc::in6_addr { s6_addr: addr.ip().octets() },
+                sin6_flowinfo: addr.flowinfo(),
+                sin6_scope_id: addr.scope_id(),
                 #[cfg(any(
                     target_os = "dragonfly",
                     target_os = "freebsd",
--- a/src/sys/unix/net.rs
+++ b/src/sys/unix/net.rs
@@ -124,6 +124,8 @@ pub(crate) fn socket_addr(addr: &SocketAddr) -> (SocketAddrCRepr, libc::socklen_
                     target_os = "openbsd"
                 ))]
                 sin6_len: 0,
+                #[cfg(any(target_os = "solaris", target_os = "illumos"))]
+                __sin6_src_id: 0,
             };
 
             let sockaddr = SocketAddrCRepr { v6: sockaddr_in6 };
--- a/src/sys/unix/net.rs
+++ b/src/sys/unix/net.rs
@@ -94,7 +94,7 @@ pub(crate) fn socket_addr(addr: &SocketAddr) -> (SocketAddrCRepr, libc::socklen_
                 sin_family: libc::AF_INET as libc::sa_family_t,
                 sin_port: a.port().to_be(),
                 sin_addr,
-                ..unsafe { mem::zeroed() }
+                sin_zero: [0; 8],
             };
 
             let sockaddr = SocketAddrCRepr { v4: sockaddr_in };
@@ -107,7 +107,15 @@ pub(crate) fn socket_addr(addr: &SocketAddr) -> (SocketAddrCRepr, libc::socklen_
                 sin6_addr: libc::in6_addr { s6_addr: a.ip().octets() },
                 sin6_flowinfo: a.flowinfo(),
                 sin6_scope_id: a.scope_id(),
-                ..unsafe { mem::zeroed() }
+                #[cfg(any(
+                    target_os = "dragonfly",
+                    target_os = "freebsd",
+                    target_os = "ios",
+                    target_os = "macos",
+                    target_os = "netbsd",
+                    target_os = "openbsd"
+                ))]
+                sin6_len: 0,
             };
 
             let sockaddr = SocketAddrCRepr { v6: sockaddr_in6 };
--- a/src/sys/unix/net.rs
+++ b/src/sys/unix/net.rs
@@ -1,4 +1,4 @@
-use std::net::SocketAddr;
+use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6};
 
 pub(crate) fn new_ip_socket(
     addr: SocketAddr,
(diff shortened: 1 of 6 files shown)


Thomasdezeeuw:
I was thinking about simply recreating `SocketAddr`/`libc::sockaddr_storage`, dropping the use of references to it. This would be especially useful if https://github.com/rust-lang/rust/pull/78802 is merged. I'll take a look at this this weekend.

faern:
You mean `mio` would not use the standard library `SocketAddr` type? Or you mean this PR should use `libc::sockaddr_storage` instead of the custom `SocketAddrCRepr`? The former does not sound good IMO. The intention with https://github.com/rust-lang/rust/pull/78802 was not to split the ecosystem, but rather unify it by allowing it into `core` etc.

Thomasdezeeuw:
> You mean `mio` would not use the standard library `SocketAddr` type? Or you mean this PR should use `libc::sockaddr_storage` instead of the custom `SocketAddrCRepr`? The former does not sound good IMO. The intention with [rust-lang/rust#78802](https://github.com/rust-lang/rust/pull/78802) was not to split the ecosystem, but rather unify it by allowing it into `core` etc.

No, Mio will continue to use `SocketAddr` from std lib. Instead I would create `libc::sockaddr_in(6)` manually from `SocketAddr` and for accept create `SocketAddr` from `libc::sockaddr_storage` (something we already for UDS).

faern:
I see. Makes sense :+1:. I will also look more into this in the weekend. I want to submit a PR to `socket2` as well since that's a bit blocking for my main PR on the standard library.

Thomasdezeeuw:
> I see. Makes sense 👍. I will also look more into this in the weekend. I want to submit a PR to `socket2` as well since that's a bit blocking for my main PR on the standard library.

For socket2 make a pr against the v0.3.x branch, master is for v0.4 which is quite a bit away from a release.

src/sys/unix/net.rs
```diff
@@ -64,32 +64,76 @@ pub(crate) fn new_socket(
...
+/// type over using `libc::sockaddr_storage` is that this type is exactly as large as it
+/// needs to be and not a lot larger. And it can be initialized cleaner from Rust.
+#[repr(C)]
+pub(crate) union SocketAddrCRepr {
+    v4: libc::sockaddr_in,
+    v6: libc::sockaddr_in6,
+}
+
+impl SocketAddrCRepr {
+    pub fn as_ptr(&self) -> *const libc::sockaddr {
```
Thomasdezeeuw, line 78: Nit: `pub(crate)`.

src/sys/unix/net.rs
```diff
@@ -64,32 +64,76 @@ pub(crate) fn new_socket(
...
+        self as *const _ as *const libc::sockaddr
```
Thomasdezeeuw, line 79: I'm not 100% on the rules when it comes to unions, I don't use them that often, so I have a question: does this require a `#[repr(transparent)]` attribute? Or is their a guarantee it contains no padding (at the start at least).

src/sys/unix/net.rs
```diff
@@ -64,32 +64,76 @@ pub(crate) fn new_socket(
...
+    }
+}
+
+/// Converts a Rust `SocketAddr` into the system representation.
+pub(crate) fn socket_addr(addr: &SocketAddr) -> (SocketAddrCRepr, libc::socklen_t) {
+    use std::mem;
```
Thomasdezeeuw, line 85: Nit: you move this to the start of the file. Before this file had bunch of `cfg`s on each function and thus inline the `use` statements, but that has been simplified recently.

src/sys/unix/net.rs
```diff
@@ -64,32 +64,76 @@ pub(crate) fn new_socket(
...
+        SocketAddr::V4(ref a) => {
+            // `s_addr` is stored as BE on all machine and the array is in BE order.
+            // So the native endian conversion method is used so that it's never swapped.
+            let sin_addr = libc::in_addr { s_addr: u32::from_ne_bytes(a.ip().octets()) };
+
+            let sockaddr_in = libc::sockaddr_in {
+                sin_family: libc::AF_INET as libc::sa_family_t,
+                sin_port: a.port().to_be(),
+                sin_addr,
+                ..unsafe { mem::zeroed() }
```
Thomasdezeeuw, line 97: Since only `sin_zero` is left can you remove the unsafe block and change it to `[0; 8]`? One less unsafe block.

src/sys/unix/net.rs
```diff
@@ -64,32 +64,76 @@ pub(crate) fn new_socket(
...
+            (sockaddr, mem::size_of::<libc::sockaddr_in>() as libc::socklen_t)
+        }
+        SocketAddr::V6(ref a) => {
+            let sockaddr_in6 = libc::sockaddr_in6 {
+                sin6_family: libc::AF_INET6 as libc::sa_family_t,
+                sin6_port: a.port().to_be(),
+                sin6_addr: libc::in6_addr { s6_addr: a.ip().octets() },
+                sin6_flowinfo: a.flowinfo(),
+                sin6_scope_id: a.scope_id(),
+                ..unsafe { mem::zeroed() }
```
Thomasdezeeuw, line 110: I think on the BSDs only `sin6_len` is left you can set it to `0` with the following `cfg`, again removing an unsafe block.

```
#[cfg(any(
    target_os = "dragonfly",
    target_os = "freebsd",
    target_os = "ios",
    target_os = "macos",
    target_os = "netbsd",
    target_os = "openbsd"
))]
sin6_len: 0,
```

src/sys/unix/net.rs
```diff
@@ -64,32 +64,76 @@ pub(crate) fn new_socket(
...
+
+            let sockaddr = SocketAddrCRepr { v6: sockaddr_in6 };
+            (sockaddr, mem::size_of::<libc::sockaddr_in6>() as libc::socklen_t)
+        }
     }
 }
 
-/// `storage` must be initialised to `sockaddr_in` or `sockaddr_in6`.
+/// Converts a `libc::sockaddr` compatible struct into a native Rust `SocketAddr`.
+/// SAFETY: `storage` must be initialised to `sockaddr_in` or `sockaddr_in6`.
```
Thomasdezeeuw, line 120: Nit: Can you make this safety a heading `# Safety` and say that the `ss_family` must be initialised correctly.

src/sys/unix/net.rs
```diff
@@ -64,32 +64,76 @@ pub(crate) fn new_socket(
...
 ) -> std::io::Result<SocketAddr> {
     match (*storage).ss_family as libc::c_int {
-        libc::AF_INET => Ok(SocketAddr::V4(
-            *(storage as *const libc::sockaddr_in as *const _),
-        )),
-        libc::AF_INET6 => Ok(SocketAddr::V6(
-            *(storage as *const libc::sockaddr_in6 as *const _),
-        )),
+        libc::AF_INET => {
+            let addr: &libc::sockaddr_in = &*(storage as *const libc::sockaddr_in);
```
Thomasdezeeuw, line 126: Can you add the following comment (as this would require an unsafe block):
// Safety: if the `ss_family` field is `AF_INET` than `storage` must be `sockaddr_in`.

src/sys/unix/net.rs
```diff
@@ -64,32 +64,76 @@ pub(crate) fn new_socket(
...
+            let ip = Ipv4Addr::from(addr.sin_addr.s_addr.to_ne_bytes());
+            let port = u16::from_be(addr.sin_port);
+            Ok(SocketAddr::V4(SocketAddrV4::new(ip, port)))
+        },
+        libc::AF_INET6 => {
+            let addr: &libc::sockaddr_in6 = &*(storage as *const libc::sockaddr_in6);
```
Thomasdezeeuw, line 132: Same safety comment as above.

Thomasdezeeuw (commented):
After looking into a bit I think this is indeed the best way forward. I've added some small inline comments but overall this looks good.

src/sys/unix/net.rs
faern: Does that make any difference? The entire type is only crate-internal anyway, so any method on it can't possibly become public outside of the crate. I would not expect `pub(crate)` to make any difference at all here. Does it? I can change it.

src/sys/unix/net.rs
faern: True! Good point. Yeah this code came from the more generic version supporting multiple different `sockaddr_in` implementations. I'll get it fixed.

src/sys/unix/net.rs
faern: Great. Done.

src/sys/unix/net.rs
faern: `#[repr(transparent)]` does not work on unions. It also only works on structs or enums with a single field.

I think this is currently correctly implemented. A Rust `#[repr(C)]` union is supposed to have the same memory layout as a C enum. And a pointer to a C enum should always be a valid pointer to each field in it.

src/sys/unix/net.rs
faern: `sin_len` is also a field on a number of platforms... I can conditionally add that as well, just need to figure out where exactly.

src/sys/unix/net.rs
faern: There is also a `pub __sin6_src_id: u32` field on solaris... This becomes a bit messy. Don't you think it's cleaner to just have the single `mem::zeroed()` instead? I think so at least.

src/sys/unix/net.rs
faern: Done. I made the change. Still not sure if it has any effect, but it does not hurt, and can maybe be more readable(?)

src/sys/windows/tcp.rs
```diff
@@ -1,14 +1,14 @@
...
+use std::net::{self, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6};
 use std::time::Duration;
 use std::convert::TryInto;
 use std::os::windows::io::FromRawSocket;
 use std::os::windows::raw::SOCKET as StdSocket; // winapi uses usize, stdlib uses u32/u64.
 
 use winapi::ctypes::{c_char, c_int, c_ushort};
 use winapi::shared::ws2def::{SOCKADDR_STORAGE, AF_INET, SOCKADDR_IN};
-use winapi::shared::ws2ipdef::SOCKADDR_IN6_LH;
+use winapi::shared::ws2ipdef::SOCKADDR_IN6;
```
faern, line 11: After I had bumped `winapi` to `0.3.9` I realized this code already used `SOCKADDR_IN6_LH`. Would you prefer I use that type instead of this alias that was introduced in `0.3.9`?

Thomasdezeeuw: According to https://docs.rs/winapi/0.3.9/winapi/shared/ws2ipdef/type.SOCKADDR_IN6.html, `SOCKADDR_IN6` is a type alias for `SOCKADDR_IN6_LH`, so it's the same. I don't really care to be honest.

Thomasdezeeuw: Although I coudn't find any docs of `SOCKADDR_IN6`, Windows's docs about `SOCKADDR_IN6_LH` pops up as first results

faern:
Now I have implemented the same for Windows. I might have missed some spot since any place that still just does the casting will just work with the current versions of Rust... That's the danger of that casting, the type system won't catch it.

I will need to get my custom toolchain built on top of https://github.com/rust-lang/rust/pull/78802 on Windows and try out this branch there...

src/sys/windows/tcp.rs
faern: Ok. Let's revert to `SOCKADDR_IN6_LH`. Will make the PR cause less fuzz and risk being harder to backport or adopt for people or whatever.

faern:
@Thomasdezeeuw Oh... `socket2` is a dependency to `mio` on Windows... So it sounds smoother if we fix it the other way around really. I currently can't build and run `mio` under my toolchain on Windows. Well I can hack around it with the Cargo `[patch]` section of course, but anyway, seems a bit backwards.

src/sys/windows/net.rs
```diff
@@ -41,15 +44,50 @@ pub(crate) fn new_socket(domain: c_int, socket_type: c_int) -> io::Result<SOCKET
...
-            size_of_val(addr) as c_int,
-        ),
-        SocketAddr::V6(ref addr) => (
-            addr as *const _ as *const SOCKADDR,
-            size_of_val(addr) as c_int,
-        ),
+        SocketAddr::V4(ref addr) => {
+            // `s_addr` is stored as BE on all machine and the array is in BE order.
+            // So the native endian conversion method is used so that it's never swapped.
+            let sin_addr = IN_ADDR { S_un: unsafe { mem::transmute(u32::from_ne_bytes(addr.ip().octets())) } };
```
faern, line 68: All these transmute are a bit of a hack. That's because unions in `winapi 0.3` are terrible. Optimally I should create them with `mem::zeroed()` or something first and then assign the value into it. I can do that. Mostly did this to get something to compile.

faern: Fixed in my PR to `socket2`. That PR does not use these ugly transmutes... I'll fix it here later.

src/sys/unix/net.rs
Thomasdezeeuw: I still prefer the `cfg` way because that forces someone to look at the field when porting to a new OS, rather than unknowingly setting the fields to zero, which may be incorrect.

src/sys/windows/tcp.rs
```diff
@@ -108,28 +108,33 @@ pub(crate) fn get_reuseaddr(socket: TcpSocket) -> io::Result<bool> {
...
+            } else if storage.ss_family as c_int == AF_INET6 {
+                let addr: &SOCKADDR_IN6_LH = unsafe { &*(&storage as *const _ as *const SOCKADDR_IN6_LH) };
+                let ip = Ipv6Addr::from(*unsafe { addr.sin6_addr.u.Byte() });
+                let port = u16::from_be(addr.sin6_port);
+                let scope_id = unsafe { *addr.u.sin6_scope_id() };
+                Ok(SocketAddr::V6(SocketAddrV6::new(ip, port, addr.sin6_flowinfo, scope_id)))
             } else {
-                let sock_addr : SocketAddrV6 = unsafe { *(storage as *const SOCKADDR_IN6_LH as *const _) };
-                Ok(sock_addr.into())
+                panic!("Invalid ss_family");
```
Thomasdezeeuw, line 134: Please return `Err(std::io::ErrorKind::InvalidInput.into())`, same as what we do on Unix.

Thomasdezeeuw (commented):
If you can fix the transmute and not panic in `get_localaddr` than this LGTM.

src/sys/windows/tcp.rs
faern: Ok. I thought it would be a bug rather than a state/user error to reach this branch. Is there any valid way to reach this branch? Hopefully not since the branch did not even exist before this PR. I usually prefer to panic on bugs. Can be confusing to users if they get errors when the problem is actually in this library.

Thomasdezeeuw: It indeed shouldn't happen (but then again which bugs should?). I think the only way this branch can be hit if if the socket isn't actually a TCP socket but a Unix socket, in which case it can be a Unix address. I think the code was written when WSL wasn't a thing and hence just `else` was used, but with WSL maybe Unix socket are/could be supported. In general I prefer to avoid panicking.

src/sys/windows/net.rs
faern: Fixed...

src/sys/windows/tcp.rs
faern: Ok. I don't fully agree to that bug handling strategy. But it's fine. It's changed. I'm now happy with the content of this PR.

faern:
@Thomasdezeeuw Would you like me to squash the git history a bit, or do you prefer it this way? Either way is fine with me.

Thomasdezeeuw:
I'll squash on merging.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @faern!

faern:
Thank you! I hope we can get this fix published in all relevant versions of `mio` and `socket2` soon :)

I realized I did not add it to the changelog. Will you do that or should I submit a follow up PR?

Thomasdezeeuw:
> Thank you! I hope we can get this fix published in all relevant versions of `mio` and `socket2` soon :)

I'll likely release new versions for both next week or so.

> I realized I did not add it to the changelog. Will you do that or should I submit a follow up PR?

No I usually update the changelog while preparing for a release.

PR #1400, opened 2020-11-16
Remove the libc prefix of EPOLL constants

uudiin:
These constants have been introduced in the use, no need to add the
libc crate prefix.


Thomasdezeeuw:
@uudiin you're importing the constants into the wrong context as the functions you use them in are in a scoped `event` module. But in general I don't think improves the code quality really, I rather not fill the global namespace with more constants, using the `libc::` prefix is fine.

Issue #1376, opened 2020-10-21
Test no_events_after_deregister failed on CI


Thomasdezeeuw:
While testing pr #1374. Log from https://dev.azure.com/tokio-rs/Tokio/_build/results?buildId=6371&view=logs&jobId=9ee66ee0-c6b6-50f2-32c8-2bebb82e1137&j=9ee66ee0-c6b6-50f2-32c8-2bebb82e1137&t=3b0adf12-f909-59d8-0849-13de19d8dc87, https://dev.azure.com/tokio-rs/e4c287c8-f114-46bc-a72c-9560911d83d6/_apis/build/builds/6371/logs/124.

```log
2020-10-20T19:34:07.9648420Z      Running target/debug/deps/tcp_stream-f2065a7326483e90
2020-10-20T19:34:08.0146140Z 
2020-10-20T19:34:08.0147600Z running 21 tests
2020-10-20T19:34:08.0148000Z test is_send_and_sync ... ok
2020-10-20T19:34:08.0159220Z test get_ttl_without_previous_set ... ok
2020-10-20T19:34:08.0176430Z thread '<unnamed>' panicked at 'error reading: Os { code: 9, kind: Other, message: "Bad file descriptor" }', tests/tcp_stream.rs:714:22
2020-10-20T19:34:08.0181250Z test get_nodelay_without_previous_set ... ok
2020-10-20T19:34:08.0181930Z test hup_event_on_disconnect ... ok
2020-10-20T19:34:08.0182300Z stack backtrace:
2020-10-20T19:34:08.0182660Z test reregistering ... ok
2020-10-20T19:34:08.0182970Z test raw_fd ... ok
2020-10-20T19:34:08.0183260Z test set_get_nodelay ... ok
2020-10-20T19:34:08.0183880Z test set_get_ttl ... ok
2020-10-20T19:34:08.0184200Z test shutdown_both ... ok
2020-10-20T19:34:08.0184650Z test shutdown_write ... ignored
2020-10-20T19:34:08.0184990Z test shutdown_read ... ok
2020-10-20T19:34:08.0189910Z test tcp_shutdown_client_both_close_event ... ok
2020-10-20T19:34:08.0193920Z test tcp_shutdown_client_read_close_event ... ok
2020-10-20T19:34:08.0197380Z test tcp_shutdown_client_write_close_event ... ok
2020-10-20T19:34:08.0201530Z test tcp_shutdown_server_write_close_event ... ok
[...]

PR #1401, opened 2020-11-17
Add `into_std` methods to structs with `from_std` methods

stevenroose:
It seems useful to have those.

### DIFF
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -78,6 +78,14 @@ impl TcpListener {
         }
     }
 
+    /// Convert into a `net::TcpListener` from the standard library.
+    ///
+    /// No assumptions should be made as to the state of the TCP listener.
+    /// It's up to the user to configure it for the intended usage.
+    pub fn into_std(self) -> net::TcpListener {
+        self.inner.into_inner()
+    }
+
     /// Accepts a new `TcpStream`.
     ///
     /// This may return an `Err(e)` where `e.kind()` is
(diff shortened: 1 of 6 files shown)


Thomasdezeeuw:
I don't think this is a good idea since the type in standard lib are generally consider blocking, while Mio's types are generally considered non-blocking.

stevenroose:
About the CI failure, it seems like it's unrelated. I don't get it locally, but I occasionally get other failures.

Thomasdezeeuw:
Yeah it's a known problem: #1376, I've restarted the CI.

Thomasdezeeuw:
> Well the same is true for `from_std`, right? 

That's true. 

> And there's a disclaimer about that on `from_std`, so I also added a disclaimer about that on `into_std`. I can make the disclaimer a bit more explicit about the non-blocking state?

The problem with going Mio -> std, is that Mio keeps some additional state (on Windows) that will get lost. For everything to guarantee to work correctly you would need to `deregister` before calling `into_std`. And `register` when doing the inverse again, i.e. `from_std`.

Can you add two tests for each type:

1) A tests that creates a Mio TcpStream/UdpSocket/etc. register it with `Poll`. Create an event for it (accept the stream/send a packet etc.). Next deregister it and call `into_std`. Then check that we get no events for the type (you can use `expect_no_events` for this).
2) Basically the same as above, create Mio type, register, create event, deregister and call `into_std`. But then call `from_std` and register it again. After that poll and we would expect to get an event (but I'm not 100% sure this will be the case).

Thomasdezeeuw:
@stevenroose do you plan on continuing this?

Thomasdezeeuw:
Closing this due to inactivity.

PR #1402, opened 2020-11-17
net: add SO_LINGER option for TcpStream

zekisherif:
It was requested in a [tokio PR](https://github.com/tokio-rs/tokio/pull/3143) that Mio TcpStream support setting linger option. This is so that the logic for converting a mio TcpStream into a TcpSocket stays within mio.

### DIFF
--- a/src/net/tcp/socket.rs
+++ b/src/net/tcp/socket.rs
@@ -334,6 +334,20 @@ impl Drop for TcpSocket {
     }
 }
 
+impl From<&TcpStream> for TcpSocket {
+    fn from(stream: &TcpStream) -> TcpSocket {
+        #[cfg(unix)]
+            {
+                unsafe { TcpSocket::from_raw_fd(stream.as_raw_fd()) }
+            }
+
+        #[cfg(windows)]
+            {
+                unsafe { TcpSocket::from_raw_socket(stream.as_raw_socket())}
+            }
+    }
+}
+
 #[cfg(unix)]
 impl IntoRawFd for TcpSocket {
     fn into_raw_fd(self) -> RawFd {
--- a/src/net/tcp/stream.rs
+++ b/src/net/tcp/stream.rs
@@ -5,6 +5,7 @@ use std::net::{self, Shutdown, SocketAddr};
 use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
 #[cfg(windows)]
 use std::os::windows::io::{AsRawSocket, FromRawSocket, IntoRawSocket, RawSocket};
+use std::time::Duration;
 
 use crate::io_source::IoSource;
 use crate::net::TcpSocket;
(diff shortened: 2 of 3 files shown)


src/net/tcp/stream.rs
```diff
@@ -123,6 +124,40 @@ impl TcpStream {
         self.inner.nodelay()
     }
 
+    /// Sets the value for the `SO_LINGER` option on this socket.
+    ///
+    /// When set, the closing/shutting down of the socket will not return
+    /// until all queued messages for the socket have been successfully
+    /// sent or the linger timeout has been reached.
+    pub fn set_linger(&self, dur: Option<Duration>) -> io::Result<()> {
```
Thomasdezeeuw, line 132: I think we should stick to the same API as std lib.

src/net/tcp/stream.rs
```diff
@@ -123,6 +124,40 @@ impl TcpStream {
...
+    /// For more information about this option, see [`set_linger`][link].
+    ///
+    /// [link]: #method.set_linger
+    pub fn linger(&self) -> io::Result<Option<Duration>> {
+        let sock = std::mem::ManuallyDrop::new(self.to_sock());
+
+        sock.get_linger()
+    }
+
+    fn to_sock(&self) -> TcpSocket {
```
Thomasdezeeuw, line 149: Why not `impl From<TcpStream> for TcpSocket`? And `impl From<TcpSocket> for TcpStream`.

Thomasdezeeuw (changes requested):

src/net/tcp/stream.rs
zekisherif: sorry about that. I was just on autopilot as I basically had this change already in `tokio::net::TcpStream`.

I'll update this.

src/net/tcp/stream.rs
zekisherif: std lib doesn't have `set_linger` option anymore. How else can we expose this feature?

I think the plan for tokio's TcpStream was to also stick to the same api as std lib but @carllerche says they've abandoned that plan.

zekisherif: So for tokio::net `TcpStream` @carllerche mentioned to me the plan was to follow std lib's api but it was abandoned.
I assumed we want mio's `TcpStream` to be similar to tokio's.

If we only add `From` traits here then we'd still have to have deal with `ManuallyDrop` at tokio `TcpStream` layer.

carllerche: @Thomasdezeeuw what is the std lib API? I don't see one? Could you clarify what API you mean?

Thomasdezeeuw: @carllerche that is my point. Not having `set_linger` on `TcpStream`.

src/net/tcp/socket.rs
```diff
@@ -334,6 +334,20 @@ impl Drop for TcpSocket {
     }
 }
 
+impl From<&TcpStream> for TcpSocket {
+    fn from(stream: &TcpStream) -> TcpSocket {
```
Thomasdezeeuw, line 338: If drop the `TcpSocket` now it will close the `TcpStream`'s file descriptor. Can't use `From<&TcpStream>`.

Thomasdezeeuw: This could work with something like socket2's `SockRef`: https://github.com/rust-lang/socket2/blob/9c4a404d5e44df1ecc56d62d35ee22109a886b5f/src/sockref.rs#L59. Maybe it's best to wait for #1381, which I'll work on (likely) next year.

Thomasdezeeuw:
Closing this as we're going to follow the standard library API. We can revisited this once `TcpStream` in the standard library supports linger: https://github.com/rust-lang/rust/issues/88494.

In the mean time the socket2 crate can be used for this.

PR #1404, opened 2020-11-25
Fix underflow in socketAddr::address

### LINKED ISSUE #1403: SocketAddr  attempt to subtract with overflow
I think there might be an issue [here](https://github.com/tokio-rs/mio/blob/master/src/sys/unix/uds/socketaddr.rs#L29-L45):
```
    fn address(&self) -> AddressKind<'_> {
        let offset = path_offset(&self.sockaddr);
        let len = self.socklen as usize - offset;
        let path = unsafe { &*(&self.sockaddr.sun_path as *const [libc::c_char] as *const [u8]) };

        // macOS seems to return a len of 16 and a zeroed sun_path for unnamed addresses
        if len == 0
            || (cfg!(not(any(target_os = "linux", target_os = "android")))
                && self.sockaddr.sun_path[0] == 0)
        {
            AddressKind::Unnamed
        } else if self.sockaddr.sun_path[0] == 0 {
            AddressKind::Abstract(&path[1..len])
        } else {
            AddressKind::Pathname(OsStr::from_bytes(&path[..len - 1]).as_ref())
        }
    }
```
From the [linux man pages for "recv"](https://man7.org/linux/man-pages/man2/recv.2.html):
```
   recvfrom()
       recvfrom() places the received message into the buffer buf.  The
       caller must specify the size of the buffer in len.

       If src_addr is not NULL, and the underlying protocol provides the
       source address of the message, that source address is placed in the
       buffer pointed to by src_addr.  In this case, addrlen is a value-re‐
       sult argument.  Before the call, it should be initialized to the size
       of the buffer associated with src_addr.  Upon return, addrlen is up‐
       dated to contain the actual size of the source address.  The returned
       address is truncated if the buffer provided is too small; in this
       case, addrlen will return a value greater than was supplied to the
       call.
```
Specifically this part `Upon return, addrlen is updated to contain the actual size of the source address`. I believe if the related socket is not bound to an address, i.e. the value for `self.socklen` would be set to `0`, you would get an error, because in the following line you'd be subtracting some positive value (in my case `2`) from `0`:
```
        let len = self.socklen as usize - offset;
```
Resulting in the following runtime error:
```
thread 'main' panicked at 'attempt to subtract with overflow', /home/ANT.AMAZON.COM/wssmts/.cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.7.6/src/sys/unix/uds/socketaddr.rs:31:19
```

Thomasdezeeuw:
With a regression test for issue #1403.

Fixes  #1403

### DIFF
--- a/src/sys/unix/uds/socketaddr.rs
+++ b/src/sys/unix/uds/socketaddr.rs
@@ -28,6 +28,10 @@ enum AddressKind<'a> {
 impl SocketAddr {
     fn address(&self) -> AddressKind<'_> {
         let offset = path_offset(&self.sockaddr);
+        // Don't underflow in `len` below.
+        if (self.socklen as usize) < offset {
+            return AddressKind::Unnamed;
+        }
         let len = self.socklen as usize - offset;
         let path = unsafe { &*(&self.sockaddr.sun_path as *const [libc::c_char] as *const [u8]) };
 
--- a/tests/regressions.rs
+++ b/tests/regressions.rs
@@ -1,14 +1,15 @@
 #![cfg(all(feature = "os-poll", feature = "net"))]
 
-use mio::net::{TcpListener, TcpStream};
-use mio::{Events, Interest, Poll, Token, Waker};
 use std::io::{self, Read};
 use std::sync::Arc;
 use std::time::Duration;
 use std::{net, thread};
 
+use mio::net::{TcpListener, TcpStream};
+use mio::{Events, Interest, Poll, Token, Waker};
+
 mod util;
-use util::{any_local_address, init, init_with_poll};
+use util::{any_local_address, init, init_with_poll, temp_file};
 
 const ID1: Token = Token(1);
 const WAKE_TOKEN: Token = Token(10);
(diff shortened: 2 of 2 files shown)


kleimkuhler (approved):
I think this solution is good. Another solution I'll suggest is a little more similar to std since this is ultimately where a lot of this implementation is from (reasoning described in comments).

In `SocketAddr::from_parts`, we check the length there and if `0`, we set it to the value of `path_offset`. So something like:

```
pub(crate) fn from_parts(sockaddr: libc::sockaddr_un, mut socklen: libc::socklen_t) -> SocketAddr {
    if socklen == 0 {
        socklen = path_offset(&sockaddr) as libc::socklen_t
    }
    SocketAddr { sockaddr, socklen }
}
```

Then in in `SocketAddr::address`, `self.socklen` is always at least the same value as `offset`. Ultimately this is exactly the same behavior as what we have in this change so it's the difference of where we actually check `socklen`.

[Here](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/ext/net.rs#L160-L163) is the check in std.

So I'm good with either, but changing `from_parts` keeps it closer to std.

Thomasdezeeuw:
I don't think the solutions differ all that much. Upside to this change is that it keeps the underlying data untouched. Downside is we need to ensure check this run every access, but the same is true for std's check in `from_raw`.

PR #1405, opened 2020-11-27
Bump net2 and miow to get rid of invalid memory layout assumption

### LINKED ISSUE #1390: Backport pr #1388 to v0.6
Pr: https://github.com/tokio-rs/mio/pull/1388
Issue: https://github.com/tokio-rs/mio/issues/1386

faern:
Fixes #1390

After getting [`net2 0.2.36`](https://github.com/deprecrated/net2-rs/pull/106) and [`miow 0.2.2`](https://github.com/yoshuawuyts/miow/pull/40) fixed and published we can bump `mio` so it forces the usage of dependencies not making invalid memory layout assumptions on `std::net::SocketAddr`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ default = ["with-deprecated"]
 [dependencies]
 log      = "0.4"
 slab     = "0.4.0"
-net2     = "0.2.29"
+net2     = "0.2.36"
 iovec    = "0.1.1"
 cfg-if   = "0.1.9"
 
@@ -44,7 +44,7 @@ libc   = "0.2.54"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.2.6"
-miow   = "0.2.1"
+miow   = "0.2.2"
 kernel32-sys = "0.2"
 
 [dev-dependencies]
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -32,7 +32,7 @@ jobs:
   parameters:
     name: minrust
     displayName: Min Rust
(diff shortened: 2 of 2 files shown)


faern:
So, does `mio 0.6` support Rust back to 1.18 and this is not negotiable? Then we have a problem. Because `net2` won't build on so old Rust. And it's nothing I did with the `0.2.36` changes. `net2` has not built worked with so old Rust for many versions from what I can find.

faern:
Since these other crates in the dependency chain does not regard an MSRV bump as a breaking change, it's somewhat hard for `mio` to provide that guarantee. I'd argue that the best would be to merge this, bump the MSRV to 1.31.0 (the oldest version I can make it work with) and simply let the (almost guaranteed) very few people using something older simply not bump their `mio` dependency. They currently can't bump their `net2` or `miow` dependency anyway :shrug:.

Every yearly Rust survey have showed that very very few use anything older than the latest stable Rust version.

Thomasdezeeuw (approved):

Thomasdezeeuw:
@carllerche I don't use Mio v0.6 any more, do you know of any use cases that needs to support < Rust 1.31?

carllerche:
It should be fine. We just need to support 6 month old Rust.

Thomasdezeeuw:
Thanks for all the work @faern.

faern:
@Thomasdezeeuw Will you release `0.6.23` now or does that release wait for something more?

Thomasdezeeuw:
Waiting on #1353.

PR #1410, opened 2020-12-01
Release v0.6.23

Thomasdezeeuw:


CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+# 0.6.23 (Dec 01, 2020)
+
+### Changed
+- **BREAKING**: Increased the MSRV from 1.18.0 (Jun 8, 2017) to 1.31.0 (Dec 6,
+  2018)
+  (https://github.com/tokio-rs/mio/commit/4879e0d32ddfd98e762fc87240e594a3ad8fca30).
```
faern, line 6: Not my call to make. But a bit inconsistent to call it a breaking change and not have a breaking version number? Unless those are two different definitions of breaking, but that becomes messy. It still supports two year old Rust versions, so will affect (hopefully) no one. Maybe include this information, but not in such an alarming fashion?

faern (approved):
:+1:

CHANGELOG.md
Thomasdezeeuw: I agree, it's inconsistent. However we're not going to create a new 0.x.0 based on the v0.6.x series. I put breaking in front of it because technically it's a breaking change, but I doubt many people are running a three year old compiler at this point and can't update to a two year old compiler.

I'm open to suggestions to reduce alarming nature of the statement. My main goal is to catch peoples attention who glance over the changelog and than hopefully go: *that doesn't affect me*.

faern: I'm fine either way. Personally I would maybe prefix it with `** NOTE **` or `** MSRV **` or something to not use the same word that semver uses since that is what creates a possible confusion/inconsistency. Whether or not an MSRV bump is a breaking change is debatable, but this in not the time nor place for that discussion at all :)

Thomasdezeeuw: Changed it to `**MSRV**`.

carllerche (approved):

Thomasdezeeuw:
Released v0.6.23.

PR #1413, opened 2020-12-01
Fix a typo in example code

zeenix:
Since the comment above the `reregister` call says "Even though the token is the same", I'm guessing the author intended to pass the same Token as the call to `register` above.


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @zeenix.

PR #1415, opened 2020-12-07
Report EPOLLERR as is_read_closed

c98:
The description of EPOLLOUT in the epoll_ctl(2) manual only covers
part of it. In fact, if the peer is interrupted abnormally
(such as disconnection, shutdown), the local read and write will
receive this event.

### DIFF
--- a/src/event/event.rs
+++ b/src/event/event.rs
@@ -79,6 +79,7 @@ impl Event {
     /// | [OS selector] | Flag(s) checked |
     /// |---------------|-----------------|
     /// | [epoll]       | `EPOLLHUP`, or  |
+    /// |               | only `EPOLLERR`, or |
     /// |               | `EPOLLIN` and `EPOLLRDHUP` |
     /// | [kqueue]      | `EV_EOF`        |
     ///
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
This is already reported as `Event::is_error`, so I'm not sure it should be included in `is_read_closed`. Can you add a test case which triggers this case (and not trigger it without this patch)?

Also when you say "shutdown'' do you mean `shutdown(2)` or something else? Because `EPOLLERR` shouldn't trigger in case of `shutdown(2)`.

c98:
Thanks for your reply.

This scenario describes the events that can only be sensed by the local end after it initiates a read/write call when an abnormal situation occurs on the opposite end (such as unplugging the network cable or shutting down the power).

Personally think that the logic of `is_read_close and` and `is_write_close` in this case (`EPOLLERR`) needs to be consistent ;-)

kleimkuhler:
@c98 I agree with Thomas on this that it should stay as `Event::is_error` for now unless you can provide a test case where it really makes sense.

Checking for it on `is_writed_closed` makes sense because that is mentioned in [epoll_ctl](https://man7.org/linux/man-pages/man2/epoll_ctl.2.html) docs in the `EPOLLERR` section.

Also take a look at #1110 if you'd like for a description of why we check which events right now.

Thomasdezeeuw:
Since there hasn't been any follow up I'm going to close this. @c98 if you have a case in which this change is needed we'll reconsider this pr.

PR #1416, opened 2020-12-07
impl Clone for Event

LuoZijun:


Thomasdezeeuw:
I think `Event` is too big for `Copy`, e.g. `kevent` is 32 bytes. Do you have a use case for `Clone`?

LuoZijun:
@Thomasdezeeuw 

Yes, for example, when I receive a readable event, but I don't want to start processing the event immediately, so I want to put the event in a delay queue and then process it at a suitable time. At this time, I need Event to be Clone.

LuoZijun:
@Thomasdezeeuw `Copy` has removed.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @LuoZijun.

PR #1425, opened 2020-12-28
Bring back UdpSocket::only_v6 (fixes #1424)

### LINKED ISSUE #1424: Can we have UdpSocket::only_v6() back?
This used to be in 0.6, but was removed in 0.7. We're using it in Quinn to detect dual-stack sockets on darwin:

https://github.com/quinn-rs/quinn/blob/main/quinn/src/platform/unix.rs#L44

djc:
Here's a minimal attempt, let me know what else is needed.

### DIFF
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -511,6 +511,12 @@ impl UdpSocket {
         self.inner.leave_multicast_v6(multiaddr, interface)
     }
 
+    /// Get the value of the `IPV6_V6ONLY` option on this socket.
+    #[allow(clippy::trivially_copy_pass_by_ref)]
+    pub fn only_v6(&self) -> io::Result<bool> {
+        sys::udp::only_v6(&self.inner)
+    }
+
     /// Get the value of the `SO_ERROR` option on this socket.
     ///
     /// This will retrieve the stored error in the underlying socket, clearing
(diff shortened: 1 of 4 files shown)


src/net/udp.rs
```diff
@@ -511,6 +511,13 @@ impl UdpSocket {
         self.inner.leave_multicast_v6(multiaddr, interface)
     }
 
+    /// Get the value of the `IPV6_V6ONLY` option on this socket.
+    #[cfg(all(unix, feature = "os-poll"))]
```
Thomasdezeeuw, line 515: Windows has this as well, see https://github.com/rust-lang/socket2/blob/35c26913ec3c0205fd1d9c3946ec0abcb50e3a4c/src/socket.rs#L1170-L1175.

And I'm afraid this needs a fake implementation in the shell, e.g. see: https://github.com/tokio-rs/mio/blob/ec28a8dcdd09bf2d7ad4030b7330c86834827124/src/sys/shell/tcp.rs#L16-L18

djc:
Pff, that was a bit painful. Please have a look!

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @djc.

djc:
Could you publish this on crates.io?

Thomasdezeeuw:
I want to merge https://github.com/tokio-rs/mio/pull/1422 first, it has a couple of fixes.

PR #1430, opened 2021-01-05
Add Event::as_raw_event

### LINKED ISSUE #990: Consider exposing OS specific event
In #983 I initially gave the user read only access to the OS event, that is `kevent`/`epoll_event`, etc. This is an issue to gather feedback whether or not this should be exposed or not.

Thomasdezeeuw:
Returns the OS specific event.

Closes #990.

### DIFF
--- a/src/event/event.rs
+++ b/src/event/event.rs
@@ -174,6 +174,17 @@ impl Event {
         sys::event::is_lio(&self.inner)
     }
 
+    /// Returns the OS specific event.
+    ///
+    /// The returned type depends on the OS and is thus not guaranteed. Portable
+    /// programs should not depends on this function and should only use the
+    /// addtional information store in the OS specific event as an optimisation.
+    #[cfg(all(unix, feature = "os-poll", feature = "os-ext"))]
+    #[cfg_attr(docsrs, doc(cfg(all(unix, feature = "os-poll", feature = "os-ext"))))]
+    pub fn as_raw_event(&self) -> &sys::Event {
+        &self.inner
+    }
+
     /// Create a reference to an `Event` from a platform specific event.
     pub(crate) fn from_sys_event_ref(sys_event: &sys::Event) -> &Event {
         unsafe {


Darksonn (approved):

Thomasdezeeuw:
I'm wondering if this doesn't need some scarier docs? It's *very* OS specific and currently only available on Unix.

src/event/event.rs
```diff
@@ -174,6 +174,17 @@ impl Event {
         sys::event::is_lio(&self.inner)
     }
 
+    /// Returns the OS specific event.
+    ///
+    /// The returned type depends on the OS and is thus not guaranteed. Portable
+    /// programs should not depends on this function and should only use the
```
djarek, line 180:
```suggestion
    /// programs should not depend on this function and should only use the
```
Typo

carllerche:
If we do this, we should make *very* sure `sys::Event` is a type defined in `mio` itself and not re-exporting a `libc` or `winapi` type.

Thomasdezeeuw:
> If we do this, we should make _very_ sure `sys::Event` is a type defined in `mio` itself and not re-exporting a `libc` or `winapi` type.

But this literally returns `kevent`/`epoll_event` that is the whole point of this method. I mean to use to e.g. to determine the amount of bytes available for reading. For `kqueue(2)` this is stored in `kevent.data` with the `EVFILT_READ` flag.

carllerche:
We can return a struct defined by `mio` w/ `#[repr(C)]` and a layout that matches the event struct. By exporting a struct defined in a separate crate, we would bind the public mio API to that crate. Should `libc` release v0.3, we would not be able to update. We made this mistake w/ Mio 0.6 & winapi.

carllerche:
Personally, if the goal is to expose platform specific data, I would define platform specific accessor fns on `Event` to get that data.

src/event/event.rs
```diff
@@ -174,6 +174,17 @@ impl Event {
...
+    /// addtional information store in the OS specific event as an optimisation.
+    #[cfg(all(unix, feature = "os-poll", feature = "os-ext"))]
+    #[cfg_attr(docsrs, doc(cfg(all(unix, feature = "os-poll", feature = "os-ext"))))]
+    pub fn as_raw_event(&self) -> &sys::Event {
```
carllerche, line 184: This binds the mio public API to a libc version. If the goal is to provide access to platform specific data, we can define accessors on the current `Event` struct.

carllerche (changes requested):
IMO we should not expose the platform struct and instead provide platform specific accessors.

Thomasdezeeuw:
@carllerche you're right about binding to a specific libc version, didn't really think about that.

Maybe it's better to provide additional methods instead, e.g. `bytes_readable(&self) -> Option<usize>` or something. A possible problem with that is however that we'll have a bunch of API that might only work on a single platform. For example `kqueue` supports signals, however `epoll` and `IOCP` don't, so than we get API that only works on kqueue. I'll think about it a bit more, but I'll close this for now.

PR #1434, opened 2021-01-05
Update libc dependency on unix to v0.2.71

### LINKED ISSUE #1433: libc version causes MacOS compilation failure
When trying to build on MacOS, cargo fails with the following error seen [here](https://github.com/alantech/alan/runs/1652201136#step:4:417) (failed to find `TCP_KEEPINTVL` and `TCP_KEEPCNT` in crate `libc`). Fails both the CI checks and on a local machine. After doing some investigation, it seems that [this PR](https://github.com/rust-lang/libc/pull/1776) introduced the variables to the `libc` crate starting with `0.2.71`, whereas `mio` currently requires exactly `0.2.69` on all Unix builds (including MacOS).

Patching our full build with `mio:0.7.7` but with `libc:0.2.71` (not `0.2.72`) allows the local build to pass, so it seems that the only requirement for builds to pass is to upgrade the `libc` version

Edit:
- tested on GitHub Actions using `macos-latest` (10.15)
- tested locally on Catalina (10.15.7 19H114)

cdmistman:
fixes #1433

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ net = []
 log = "0.4.8"
 
 [target.'cfg(unix)'.dependencies]
-libc = "0.2.71"
+libc = "0.2.82"
 
 [target.'cfg(windows)'.dependencies]
 miow   = "0.3.6"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ net = []
 log = "0.4.8"
 
 [target.'cfg(unix)'.dependencies]
-libc = "0.2.69"
+libc = "0.2.71"
 
 [target.'cfg(windows)'.dependencies]
 miow   = "0.3.6"


Thomasdezeeuw:
Could you wait until https://github.com/rust-lang/libc/pull/2014 is merged and update to 0.2.82? It fixes macOS aarch64 and socket2 (a dependency in v0.8) needs it as well.

Also this would go into v0.8. If you want it in v0.7.x could you open another pr against the v0.7.x branch?

Thomasdezeeuw:
@cdmistman libc v04.82 is out, can you update to that?

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @cdmistman.

cdmistman:
thank you!

PR #1436, opened 2021-01-06
Added peer addr function

carlosb1:


src/net/udp.rs
```diff
@@ -161,6 +161,33 @@ impl UdpSocket {
         self.inner.local_addr()
     }
 
+    /// Returns the socket address of the remote peer this socket was connected to.
+    ///
+    /// # Examples
+    ///
+    ///
+    #[cfg_attr(all(feature = "os-poll", not(target_os = "freebsd")), doc = "```")]
```
Thomasdezeeuw, line 169: Why are you excluding FreeBSD?

Thomasdezeeuw (approved):
LGTM with a single comment.

src/net/udp.rs
carlosb1: I copied the description for the `addr` function, I can remove this exception if it not necessary.

Thomasdezeeuw: I guess you copied from `local_addr`? The comment above it explains why, I don't think it's needed for `peer_addr`, so you can remove it.

carlosb1: Yes.. I have the same doubts for other `cfg_attr` , I don-t have clear the objective for the `os-poll` feature

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @carlosb1.

PR #1442, opened 2021-01-20
Use socket2::Socket for TcpStream

Thomasdezeeuw:
This includes a single breaking change TcpStream::peek now now accepts
an uninitialised buffer ([MaybeUninit<u8>]), rather than an initialised
one.

This also add TcpStream::recv which is the same as the io::Read::read
implementation, but accept an unitialised buffer.

Depends on #1431.


Thomasdezeeuw:
We're not adding socket2 as Mio dependency.

Issue #1445, opened 2021-01-24
Android SECCOMP forbids accept4 on x86


jshearer:
It appears that when running native code on Android x86 (such as in the emulator), the seccomp profile forbids syscall 364 which maps to `accept4`. Would a PR to include android x86 in the list of targets that get `accept` instead of `accept4` be approved?

```
01-24 14:43:55.326 18621 18621 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
01-24 14:43:55.326 18621 18621 F DEBUG   : Build fingerprint: 'google/sdk_gphone_x86/generic_x86:10/QSR1.200715.002/6695061:userdebug/dev-keys'
01-24 14:43:55.326 18621 18621 F DEBUG   : Revision: '0'
01-24 14:43:55.326 18621 18621 F DEBUG   : ABI: 'x86'
01-24 14:43:55.328 18621 18621 F DEBUG   : Timestamp: 2021-01-24 14:43:55-0500
01-24 14:43:55.328 18621 18621 F DEBUG   : pid: 18525, tid: 18607, name: tokio-runtime-w  >>> com.xxx <<<
01-24 14:43:55.328 18621 18621 F DEBUG   : uid: 10134
01-24 14:43:55.328 18621 18621 F DEBUG   : signal 31 (SIGSYS), code 1 (SYS_SECCOMP), fault addr --------
01-24 14:43:55.328 18621 18621 F DEBUG   : Cause: seccomp prevented call to disallowed x86 system call 364
01-24 14:43:55.328 18621 18621 F DEBUG   :     eax 0000016c  ebx 00000066  ecx b4c5b3d0  edx b4c5b3cc
01-24 14:43:55.328 18621 18621 F DEBUG   :     edi d233b034  esi 00080800
01-24 14:43:55.328 18621 18621 F DEBUG   :     ebp bcbd2e48  esp b4c5b378  eip ed8b4ef9
01-24 14:43:55.331 18621 18621 F DEBUG   :
01-24 14:43:55.331 18621 18621 F DEBUG   : backtrace:
01-24 14:43:55.331 18621 18621 F DEBUG   :       #00 pc 00000ef9  [vdso] (__kernel_vsyscall+9)
[...]

PR #1446, opened 2021-01-24
Use accept() instead of accept4() on x86 Android

jshearer:
See #1445 for details. I can confirm that this change _does_ work on x86 android emulator, where it did not before.

### DIFF
--- a/src/sys/unix/tcp.rs
+++ b/src/sys/unix/tcp.rs
@@ -423,7 +423,12 @@ pub fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream, Socket
     // On platforms that support it we can use `accept4(2)` to set `NONBLOCK`
     // and `CLOEXEC` in the call to accept the connection.
     #[cfg(any(
-        target_os = "android",
+        // Android x86's seccomp profile forbids calls to `accept4(2)`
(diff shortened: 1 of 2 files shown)


src/sys/unix/tcp.rs
```diff
@@ -444,7 +447,15 @@ pub fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream, Socket
     // But not all platforms have the `accept4(2)` call. Luckily BSD (derived)
     // OSes inherit the non-blocking flag from the listener, so we just have to
     // set `CLOEXEC`.
-    #[cfg(any(target_os = "ios", target_os = "macos", target_os = "solaris"))]
+    #[cfg(any(
+        all(
+            target_arch = "x86",
+            target_os = "android"
```
Thomasdezeeuw, line 453: Can you add a comment saying why we can't use `accept4(2)` on Android x86, also linking to https://github.com/tokio-rs/mio/issues/1445.

jshearer: Done! :)

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @jshearer.

PR #1451, opened 2021-01-30
`accept()`d socket needs to be set to `O_NONBLOCK`

### LINKED ISSUE #1450: `accept()` code-path doesn't set socket to `O_NONBLOCK`, resulting in blocking reads
I spent a good couple of days debugging this one, and I learned so much! It appears that a while ago, [a commit](https://github.com/tokio-rs/mio/commit/b775034106ac402596ad72cedf376f613edc6268) was introduced that assumes that the file descriptors provided by `accept()` will inherit their parent socket's blocking attribute. This... seems to not be the case on Android, as after collecting a few `strace` outputs, I discovered that `recvfrom()` calls were blocking where they very much shouldn't have been, resulting in all sorts of trouble. 

```
##### SOCKET 93 CREATED HERE
[pid  6747] socket(AF_UNIX, SOCK_STREAM, 0 <unfinished ...>
[pid  6747] <... socket resumed> )      = 93

#### SET TO NONBLOCK HERE
[pid  6747] fcntl64(93, F_SETFL, O_RDONLY|O_NONBLOCK <unfinished ...>
[pid  6747] <... fcntl64 resumed> )     = 0
[pid  6747] fcntl64(93, F_SETFD, FD_CLOEXEC <unfinished ...>

###### BINDS HERE 
[pid  6747] bind(93, {sa_family=AF_UNIX, sun_path="xxx"}, 108 <unfinished ...>
[pid  6747] <... bind resumed> )        = 0

###### LISTENS HERE 
[pid  6747] listen(93, 1024 <unfinished ...>
[pid  6747] <... listen resumed> )      = 0

###### CONNECTION ACCEPTED HERE, SWITCH OVER TO FD 101
[pid  6746] accept4(93,  <unfinished ...>
[pid  6746] <... accept4 resumed> {sa_family=AF_UNIX}, [110->2], 0) = 101
[pid  6746] fcntl64(101, F_SETFD, FD_CLOEXEC <unfinished ...>
[pid  6746] <... fcntl64 resumed> )     = 0

##### DIES HERE
[pid  6746] recvfrom(101,  <unfinished ...>

#### NEVER RESUMES
```

After testing a patch locally that sets the resulting fd's `O_NONBLOCK` flag, this blocking turned nonblocking as expected. All is right in the world!

Reading the [docs](https://man7.org/linux/man-pages/man2/accept.2.html), it appears that 
> On Linux, the new socket returned by accept() does _not_ inherit
       file status flags such as `O_NONBLOCK` and `O_ASYNC` from the
       listening socket.  This behavior differs from the canonical BSD
       sockets implementation.  Portable programs should not rely on
       inheritance or noninheritance of file status flags and always
       explicitly set all required flags on the socket returned from
       `accept()`.

Even the NetBSD [docs](https://man.netbsd.org/accept.2#COMPATIBILITY) say
>  The `accept()` implementation makes the new file descriptor inherit file
     flags (like O_NONBLOCK) from the listening socket.  It's a traditional
     behaviour for BSD derivative systems.  On the other hand, there are
     implementations which don't do so.  Linux is an example of such implemen-
     tations.  **Portable programs should not rely on either of the behaviours.**

Emphasis mine

jshearer:
I also copied the fail-handling logic from `sys/unix/net.rs` to actually close the socket on error, as described by the comment.

Closes #1450

### DIFF
--- a/src/sys/unix/tcp.rs
+++ b/src/sys/unix/tcp.rs
@@ -465,7 +465,15 @@ pub fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream, Socket
             &mut length
         ))
         .map(|socket| unsafe { net::TcpStream::from_raw_fd(socket) })
-        .and_then(|s| syscall!(fcntl(s.as_raw_fd(), libc::F_SETFD, libc::FD_CLOEXEC)).map(|_| s))
+        .and_then(|s| {
+            syscall!(fcntl(s.as_raw_fd(), libc::F_SETFD, libc::FD_CLOEXEC))?;
+    
+            // See https://github.com/tokio-rs/mio/issues/1450
+            #[cfg(all(target_arch = "x86",target_os = "android"))]
+            syscall!(fcntl(s.as_raw_fd(), libc::F_SETFL, libc::O_NONBLOCK))?;
+            
+            Ok(s)
+        })
     }?;
 
     // This is safe because `accept` calls above ensures the address
--- a/src/sys/unix/uds/listener.rs
+++ b/src/sys/unix/uds/listener.rs
@@ -80,7 +80,13 @@ pub(crate) fn accept(listener: &net::UnixListener) -> io::Result<(UnixStream, So
         // Ensure the socket is closed if either of the `fcntl` calls
         // error below.
         let s = unsafe { net::UnixStream::from_raw_fd(socket) };
-        syscall!(fcntl(socket, libc::F_SETFD, libc::FD_CLOEXEC)).map(|_| s)
+        syscall!(fcntl(socket, libc::F_SETFD, libc::FD_CLOEXEC))?;
+
+        // See https://github.com/tokio-rs/mio/issues/1450
+        #[cfg(all(target_arch = "x86",target_os = "android"))]
+        syscall!(fcntl(socket, libc::F_SETFL, libc::O_NONBLOCK))?;
+        
+        Ok(s)
     });
 
     socket


Thomasdezeeuw:
I've commented on #1450, but I don't think this is the right solution as all OSs that do use the path should have the behaviour that inherits the non-blocking property of the listening socket. Android doesn't event follow the `accept(2)` path, but uses the `accept4(2)` with `SOCK_NONBLOCK` path.

src/sys/unix/uds/listener.rs
```diff
@@ -80,7 +80,13 @@ pub(crate) fn accept(listener: &net::UnixListener) -> io::Result<(UnixStream, So
         // Ensure the socket is closed if either of the `fcntl` calls
         // error below.
         let s = unsafe { net::UnixStream::from_raw_fd(socket) };
-        syscall!(fcntl(socket, libc::F_SETFD, libc::FD_CLOEXEC)).map(|_| s)
+        syscall!(fcntl(socket, libc::F_SETFL, libc::O_NONBLOCK))
+            .and_then(|_| syscall!(fcntl(socket, libc::F_SETFD, libc::FD_CLOEXEC)))
+            .map_err(|e| {
```
Thomasdezeeuw, line 85: This will close the socket twice as `s` of type `net::UnixStream` will also close the socket when dropped, so this can be removed.

src/sys/unix/uds/listener.rs
```diff
@@ -80,7 +80,13 @@ pub(crate) fn accept(listener: &net::UnixListener) -> io::Result<(UnixStream, So
         // Ensure the socket is closed if either of the `fcntl` calls
         // error below.
         let s = unsafe { net::UnixStream::from_raw_fd(socket) };
-        syscall!(fcntl(socket, libc::F_SETFD, libc::FD_CLOEXEC)).map(|_| s)
+        syscall!(fcntl(socket, libc::F_SETFL, libc::O_NONBLOCK))
```
Thomasdezeeuw, line 83: Can you limit this to x86 Android.

Thomasdezeeuw, line 83: Can you also do this in `src/sys/unix/tcp.rs`, after `stream` is created on line 469. Also limited to x86 Android.

https://github.com/tokio-rs/mio/blob/527e843b0e11d585c0b20a8a3e4df966766f8bfa/src/sys/unix/tcp.rs#L461-L469

Thomasdezeeuw, line 83: And finally can you add a comment referring to #1450.

Thomasdezeeuw (changes requested):

jshearer:
I think I addressed your comments. My only concern is that my implementation might not be very idiomatic (`let mut` then mutating conditionally based on `cfg!`). Thoughts?

src/sys/unix/tcp.rs
```diff
@@ -465,7 +465,19 @@ pub fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream, Socket
             &mut length
         ))
         .map(|socket| unsafe { net::TcpStream::from_raw_fd(socket) })
-        .and_then(|s| syscall!(fcntl(s.as_raw_fd(), libc::F_SETFD, libc::FD_CLOEXEC)).map(|_| s))
+        .and_then(|s| {
+            let mut f = syscall!(fcntl(s.as_raw_fd(), libc::F_SETFD, libc::FD_CLOEXEC));
```
Thomasdezeeuw, line 469: You can use `syscall!(...)?` here and ignore the result (now stored in `f`).

src/sys/unix/tcp.rs
```diff
@@ -465,7 +465,19 @@ pub fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream, Socket
...
+    
+            // See https://github.com/tokio-rs/mio/issues/1450
+            if cfg!(all(
```
Thomasdezeeuw, line 472: No need to use `if`, the following should work.

```
#[cfg(all(arget_arch = "x86", target_os = "android"))
syscall!(fcntl(s.as_raw_fd(), libc::F_SETFL, libc::O_NONBLOCK))?;
```

src/sys/unix/tcp.rs
```diff
@@ -465,7 +465,19 @@ pub fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream, Socket
...
+                target_arch = "x86",
+                target_os = "android"
+            )) {
+                f = f.and_then(|_| syscall!(fcntl(s.as_raw_fd(), libc::F_SETFL, libc::O_NONBLOCK)));
+            }
+            
+            f.map(|_| s)
```
Thomasdezeeuw, line 479: This can be `Ok(s)`.

src/sys/unix/tcp.rs
jshearer: Do you mean like this? Because this actually doesn't seem to work 🤔 
```rust
let stream = {
        syscall!(accept(
            listener.as_raw_fd(),
            addr.as_mut_ptr() as *mut _,
            &mut length
        ))
        .map(|socket| unsafe { net::TcpStream::from_raw_fd(socket) })
        .and_then(|s| {
            syscall!(fcntl(s.as_raw_fd(), libc::F_SETFD, libc::FD_CLOEXEC))?;
    
            // See https://github.com/tokio-rs/mio/issues/1450
            #[cfg(all(arget_arch = "x86", target_os = "android"))]
            syscall!(fcntl(s.as_raw_fd(), libc::F_SETFL, libc::O_NONBLOCK))?;
            
            Ok(s)
        })
    }?;
```

jshearer: `arget_arch` d'oh

It works fine 🙄

jshearer:
After fixing my typo, tests pass on Android with your suggestions :)

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @jshearer.

PR #1455, opened 2021-02-08
Use SO_LINGER_SEC on macOS for TcpSocket::set/get_linger

### LINKED ISSUE #1454: Darwin's default SO_LINGER is not in seconds
In `set_linger` [here](https://github.com/tokio-rs/mio/blob/103d2c751efefa2a0a49d2e2b3785065351038fc/src/sys/unix/tcp.rs#L133) and `get_linger` [here](https://github.com/tokio-rs/mio/blob/103d2c751efefa2a0a49d2e2b3785065351038fc/src/sys/unix/tcp.rs#L161), the code assumes that the value of `l_linger` is in seconds. This is true on Linux, but not on Darwin. On Darwin, the default `SO_LINGER` is measured in ticks, not seconds (see [this XNU header](https://github.com/apple/darwin-xnu/blob/8f02f2a044b9bb1ad951987ef5bab20ec9486310/bsd/sys/socket.h#L136)).

There is another possible value for `SO_LINGER` that *does* cause the value to be measured in seconds (in that same XNU header, [here](https://github.com/apple/darwin-xnu/blob/8f02f2a044b9bb1ad951987ef5bab20ec9486310/bsd/sys/socket.h#L138) or [here](https://github.com/apple/darwin-xnu/blob/8f02f2a044b9bb1ad951987ef5bab20ec9486310/bsd/sys/socket.h#L187)), but it's not being used and doesn't appear in the libc crate at all.

Thomasdezeeuw:
Fixes #1454.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ net = []
 log = "0.4.8"
 
 [target.'cfg(unix)'.dependencies]
-libc = "0.2.82"
+libc = "0.2.86"
 
 [target.'cfg(windows)'.dependencies]
 miow   = "0.3.6"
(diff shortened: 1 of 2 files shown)


kleimkuhler (approved):
👍

PR #1458, opened 2021-02-17
DragonFlyBSD: refine kqueue-related types and values

cmusser:
Backport from master to 0.7.x

- Make sure that Filter and Flag types are defined. The sizes
  for these kevent fields are the same as on FreeBSD and OpenBSD

- Don't define the NOTE_BACKGROUND EVFILT_TIMER flag. This flag
  value is exclusive to Apple platforms.

### DIFF
--- a/src/sys/unix/selector/kqueue.rs
+++ b/src/sys/unix/selector/kqueue.rs
@@ -19,7 +19,7 @@ type Count = libc::c_int;
 type Count = libc::size_t;
 
 // Type of the `filter` field in the `kevent` structure.
-#[cfg(any(target_os = "freebsd", target_os = "openbsd"))]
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks again @cmusser.

cmusser:
great, thanks for merging those so quickly. Do we need to do anything additional to get it into Tokio (PR?) or do they regularly keep up to date on mio?

Thomasdezeeuw:
I justed released Mio v0.7.8, so after a `cargo update` everything should work.

cmusser:
yup, ok. just built a couple projects here (both Tokio based) and they build successfully now.

PR #1461, opened 2021-02-21
Fix error handling in NamedPipe::write

blackbeam:
Hi. This is a part of tokio-rs/tokio#3388.

I found an issue with `NamedPipe::write`, namely it hangs in `WouldBlock` if other side is disconnected. The problem is in error handling:

1. Errors are stashed into the state ([here][1] and [here][2]);
2. And then ignored with `WouldBlock` without changing the state ([here][3]).

Another improvement introduced in this PR is better handling of immediately finished operations. For immediately failed operations this means, that the caller won't falsely receive `Ok(usize)`. For immediately succeeded operations this means, that the caller could track the progress with better precision.

[1]: https://github.com/tokio-rs/mio/blob/master/src/sys/windows/named_pipe.rs#L503
[2]: https://github.com/tokio-rs/mio/blob/master/src/sys/windows/named_pipe.rs#L626
[3]: https://github.com/tokio-rs/mio/blob/master/src/sys/windows/named_pipe.rs#L325

### DIFF
--- a/src/sys/windows/named_pipe.rs
+++ b/src/sys/windows/named_pipe.rs
@@ -324,6 +324,12 @@ impl<'a> Write for &'a NamedPipe {
 
         match io.write {
             State::None => {}
+            State::Err(_) => match mem::replace(&mut io.write, State::None) {
+                State::Err(e) => return Err(e),
+                // `io` is locked, so this branch is unreachable
+                _ => unreachable!(),
+            },
+            // any other state should be handled in `write_done`
             _ => {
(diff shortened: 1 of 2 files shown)


carllerche (approved):
At a skim, the behavior change looks correct to me. I will let @Thomasdezeeuw comment on any style feedback.

Thomasdezeeuw:
The couple of `unreachable!` statement aren't great, but I don't see a better alternative. I'll merge this as it fixes a bug.

PR #1466, opened 2021-02-25
Fallback to accept if os doesn't support accept4

### LINKED ISSUE #1465: Backward compatibility for systems that doesn't support accept4
When running on some old systems, like an old Android, they doesn't have the `accept4` syscall. When running programs on these platforms, we will get an error: `Function not implemented (os error 38)`.

It would be better to automatically fallback to `accept` when `accept4` returns this errno `ENOSYS`.

zonyitoo:
fixes #1465

### DIFF
--- a/src/sys/unix/tcp.rs
+++ b/src/sys/unix/tcp.rs
@@ -421,6 +421,27 @@ pub fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream, Socket
     let mut addr: MaybeUninit<libc::sockaddr_storage> = MaybeUninit::uninit();
     let mut length = size_of::<libc::sockaddr_storage>() as libc::socklen_t;
 
+    fn sys_accept(
+        listener: &net::TcpListener,
+        addr: &mut MaybeUninit<libc::sockaddr_storage>,
+        length: &mut libc::socklen_t,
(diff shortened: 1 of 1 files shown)


src/sys/unix/tcp.rs
```diff
@@ -438,12 +459,36 @@ pub fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream, Socket
...
+                addr.as_mut_ptr() as *mut _,
+                &mut length,
+                libc::SOCK_CLOEXEC | libc::SOCK_NONBLOCK,
+            ));
+
+            match result {
+                Ok(s) => Ok(s),
+                Err(err) => {
+                    match err.raw_os_error() {
+                        Some(libc::ENOSYS) => {
```
nicokoch, line 478: `EPERM` should probably be added to as a fallback condition here, because if `accept4` is blocked by seccomp, that is going to be returned.

Rust std usually does it like this when they use newer syscalls.

zonyitoo: Ok.

zonyitoo:
@nicokoch CI passed.

Thomasdezeeuw:
Per the discussion in #1465, this will not be accepted in its current form.

PR #1469, opened 2021-03-05
accept4 is not defined in libc for android

elwerene:
In libc 0.2.87 there's no `accept4` in [unix/linux_like/android/mod.rs](https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/android/mod.rs).
It is defined in [unix/linux_like/linux/mod.rs](https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/linux/mod.rs#L3129), but this module is [not loaded](https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/mod.rs#L1711) on android

### DIFF
--- a/src/sys/unix/tcp.rs
+++ b/src/sys/unix/tcp.rs
@@ -424,12 +424,6 @@ pub fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream, Socket
     // On platforms that support it we can use `accept4(2)` to set `NONBLOCK`
     // and `CLOEXEC` in the call to accept the connection.
     #[cfg(any(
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
This is a regression in libc, not Mio. See https://github.com/rust-lang/libc/issues/2101.

Thomasdezeeuw:
For a quick fix add the following to your `Cargo.toml`:

```toml
[dependencies]
libc = "=0.2.86" # Note the `=` here.
```

elwerene:
@Thomasdezeeuw I had to add libc to fix my build as the dependency to libc without `=` is defined in https://github.com/tokio-rs/mio/blob/master/Cargo.toml#L49

PR #1478, opened 2021-03-22
Add missing `mstcpip` feature flag to `winapi` dependency.

emgre:
When `miow` released 0.3.7, it moved its requirements of `socket2` as a dev dependencies. This caused the `mstcpip` feature flag of `winapi` to not be set anymore, although it is used by winapi:

https://github.com/tokio-rs/mio/blob/14c78d865b220167a3a22e5f1e2a1058cc81260c/src/sys/windows/tcp.rs#L11

It produces the following symptoms: https://github.com/tokio-rs/tokio/pull/3639/checks?check_run_id=2169888185#step:5:124%5C%5C%5C

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ libc = "0.2.86"
 
 [target.'cfg(windows)'.dependencies]
 miow   = "0.3.6"
-winapi = { version = "0.3", features = ["winsock2", "mswsock"] }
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @emgre.

PR #1480, opened 2021-03-30
test(sys) Doctest imports `unix::SourceFd` which is behind the `os-ext` feature

Hywan:
This patch fixes this doctest. It imports `mio::unix::SourceFd`, but
the `mio::unix` module exists if and only if the `os-ext` feature is
enabled.

To reproduce:

```sh
$ # without this patch
$ cargo test --features os-poll,net --doc -- sourcefd
<fails>
$ cargo test --features os-poll,os-ext,net --doc -- sourcefd
<fails>
$
$ # with this patch
$ cargo test --features os-poll,net --doc -- sourcefd
<success>
$ cargo test --features os-poll,os-ext,net --doc -- sourcefd
<success>
```

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -27,7 +27,7 @@ jobs:
       name: nightly
       displayName: Nightly
       # Pin nightly to avoid being impacted by breakage
-      rust_version: nightly-2020-12-31
+      rust_version: nightly-2020-03-24
       benches: true
 
   # This represents the minimum Rust version supported by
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw (approved):
Good catch.

Hywan:
Clippy error has been fixed with https://github.com/tokio-rs/mio/pull/1480/commits/10b2a667cc196e478dbf13eebf47f9469ef63d23. It's not related to this patch though.

There is an error with `anyhow` on Rust nightly, it's not related to this patch neither. Not sure if I must fix that myself (it probably imply fixing the CI somehow).

Thomasdezeeuw:
The nightly version used on the CI was outdated. The nightly version required by anyhow (dependency of cargo hack, not Mio) was to new. I've update the nightly version to fix it.

Thomasdezeeuw:
Thanks @Hywan

PR #1487, opened 2021-05-10
Replace x86_64-sun-solaris with x86_64-pc-solaris

Thomasdezeeuw:
https://github.com/rust-lang/rust/pull/82216 removed the
x86_64-sun-solaris target from rustup, changing it to use
x86_64-pc-solaris instead.

Related issues:
 * https://github.com/rust-lang/rust/issues/85098


Thomasdezeeuw:
Merged into #1485, see https://github.com/tokio-rs/mio/pull/1485#issuecomment-840617270.

PR #1490, opened 2021-05-23
Set RUSTUP_IO_THREADS=1 to avoid FreeBSD CI failure

taiki-e:
Workaround for https://github.com/rust-lang/rustup/issues/2774 mentioned in https://github.com/rust-lang/rustup/issues/2774#issuecomment-843835682.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -3,6 +3,8 @@ freebsd_instance:
 
 env:
   RUST_BACKTRACE: full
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
There is already a work-around/fix for Rustup in https://github.com/rust-lang/rustup/pull/2779.

taiki-e:
Oh, I didn't notice about there is a PR that fixes the problem... In that case, it might make more sense to wait for it.

PR #1502, opened 2021-06-21
feat: Make `IoSource` public

Hywan:
Hello,

This patch exposes `IoSource` to the user. It's the simplest path to “expose the `Selector`” to the user. I understand the `sys::Selector` has not the same API depending on the platform (e.g. Windows), so we need an abstraction layer. I believe `IoSource` is the best abstraction layer to register a raw FD (without going through the `net` module) onto the `sys::Selector`.

Thoughts?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -54,6 +54,8 @@ pub mod event;
 
 cfg_io_source! {
     mod io_source;
+
+    pub use io_source::IoSource;
 }
 
 cfg_net! {


Thomasdezeeuw:
The problem with this, as discussed in https://github.com/tokio-rs/mio/pull/1195, is that on Windows this is somewhat pointless as IOCP (underlying poll implementation) only supports sockets and we already expose those. On Unix we already have `SourceFd`, so it doesn't really provide much benefit.

Hywan:
I understand that `mio::windows::NamedPipe` and `mio::unix::SourceFd` can be used to reach the selector with `register` etc. However, it's two different API; thus it's up to the user to handle `target_family = "unix"` and `… = "windows"` to juggle between the two API.

Exposing `IoSource` would provide a single uniform API.

What do you think?

We could also consider to make `mio::IoSource` public if the `os-ext` feature is turned on (in addition to `os-poll` and `net`). Thoughts?

Thomasdezeeuw:
> I understand that `mio::windows::NamedPipe` and `mio::unix::SourceFd` can be used to reach the selector with `register` etc. However, it's two different API; thus it's up to the user to handle `target_family = "unix"` and `… = "windows"` to juggle between the two API.

`NamedPipe`and `SourceFd` are not equal. `NamedPipe` is a wrapper around Window's named pipe. `SourceFd` is a convenience wrapper for Unix, but it's up to the user to ensure the file descriptor type is supported.

So again, this doesn't work on Windows. That is reason why we didn't made it public in the first place and that limitation hasn't been lifted.

Hywan:
Thank you for the clarifications.

PR #1501, opened 2021-06-21
feat: Move `poll::selector` to `Registry::selector`

Hywan:
Hello,

This patch moves the `poll::selector` function to `Registry::selector`. It then updates the rest of the code accordingly. I believe it makes the code simpler to read and to understand.

PR #1503, opened 2021-06-22
feat: Implement `ExactSizeIterator` for `events::Iter`

Hywan:
The size of `events::Iter` is known. The method `Iter::size_hint`
returns the asme lower and upper size. It fulfills the invariant of
`ExatSizeIterator` and thus it allows us to implement
`ExactSizeIterator` with the default implementation for `len` and
`is_empty` safely.

In most usecases, using `ExactSizeIterator.len` is more appropriate
than `Iterator.count` because it doesn't consume the iterator.

### DIFF
--- a/src/event/events.rs
+++ b/src/event/events.rs
@@ -223,6 +223,8 @@ impl<'a> Iterator for Iter<'a> {
     }
 }
 
+impl<'a> ExactSizeIterator for Iter<'a> {}
+
 impl fmt::Debug for Events {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_list().entries(self).finish()


Hywan:
Tests look good. Clippy is raising errors, which are fixed in https://github.com/tokio-rs/mio/pull/1501, I can cherry-pick the commits if needed.

Thomasdezeeuw:
Also see https://github.com/tokio-rs/mio/pull/1271.

Hywan:
Thank you for the cross-link. However, I don't see how it impacts this PR exactly. The current `Iterator::size_hint` implementation returns the same value for both the lower and upper size, so `ExactSizeIterator` can be implemented for free, with no more modification of the code.

Moreover, it would allow to know the size of `Events` without having to consume the iterator, which is nice in more usecase.

Is there anything I've missed here?

Thomasdezeeuw:
Sorry @Hywan I should have given a bit more context. The pr #1271 has more or less the same changes as this pr; letting users determine the number of events in `Events`. The problem is that we might want to supports OSes which can't determine how many events are collected. That has been the main factor in deciding against it. Furthermore the usefulness of such implementations is somewhat limited; few people actually put the events as-is in a another collection, this is where Rust can optimise using traits such as `ExactSizeIterator`.

Thomasdezeeuw:
Closing this due to age.

Issue #1444, opened 2021-01-24
mio (and tokio) sockets are completely broken under wine


whitequark:
Testcase:
```rust
#[tokio::main]
async fn main() {
    let addr = std::net::SocketAddr::from(([0,0,0,0], 1234));
    tokio::net::TcpListener::bind(addr).await.unwrap();
}
```

On Windows, it succeeds:
```
> mio-bug.exe

```

On wine (any version), it crashes:
```
$ wine mio-bug.exe
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "File not found." }', src\main.rs:3:88
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Note the very unusual and confusing `ENOENT` error code. To understand how it gets produced, we can use the following command:
```
$ strace -e '!write' env WINEDEBUG=relay,ws2_32 wine mio-bug.exe
```
<details>
<summary>unedited part of the log</summary>

```
0140:Call ws2_32.socket(00000002,00000001,00000000) ret=14007a67d
0140:Call ntdll.RtlAllocateHeap(00010000,00000008,00000048) ret=7f42dbcaae68
0140:Ret  ntdll.RtlAllocateHeap() retval=00058a00 ret=7f42dbcaae68
0140:Call ntdll.RtlInitUnicodeString(0021ddd0,7f42dbcb7b40 L"\\Device\\Afd") ret=7f42dbcaa8cc
0140:Ret  ntdll.RtlInitUnicodeString() retval=00000018 ret=7f42dbcaa8cc
0140:Call ntdll.NtOpenFile(0021ddb8,c0100000,0021ddf0,0021dde0,00000000,00000000) ret=7f42dbcaa947
rt_sigprocmask(SIG_BLOCK, [HUP INT USR1 USR2 ALRM CHLD IO], [], 8) = 0
writev(3, [{iov_base=",\0\0\0\26\0\0\0\0\0\0\0\0\0\20\300\2\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., iov_len=64}, {iov_base="\\\0D\0e\0v\0i\0c\0e\0\\\0A\0f\0d\0", iov_len=22}], 2) = 86
read(4, "\0\0\0\0\0\0\0\0\200\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., 64) = 64
[...]

PR #1505, opened 2021-07-02
Move Wine (6.11+) to supported

Jake-Shadle:
As noted in https://github.com/tokio-rs/mio/issues/1444#issuecomment-867685922, Wine's latest version, 6.11, has implemented the functionality necessary for mio to work correctly under wine.

While it will indeed take a while for 6.11 or higher versions to be available most distros' default package manager, Wine can still be installed in many cases directly from packages made by Wine themselves, or built from source, so I feel that listing it in the supported section with the minimum necessary version (just as with Android) is a more optimistic portrayal of support rather than a user needing to look at #1444 and read the comments to find that it is indeed supported once they acquire the necessary minimum version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@ Currently supported platforms:
 * Windows
 * iOS
 * macOS
+* Wine (version 6.11+, see [issue #1444])
 
 There are potentially others. If you find that Mio works on another
 platform, submit a PR to update the list!
@@ -156,7 +157,6 @@ This uses the Windows AFD system to access socket readiness events.
 ### Unsupported
 
 * Haiku, see [issue #1472]
-* Wine, see [issue #1444]
 
 [issue #1472]: https://github.com/tokio-rs/mio/issues/1472
 


Thomasdezeeuw (approved):
I'm good with this, although I don't use Wine myself so I'm going to have trust you that Wine 6.11 indeed works.

I'll look at the CI this weekend, since it's unrelated to the changes you can ignore it.

Jake-Shadle:
If you are game, I can make a PR to add testing under wine to CI, as I recently redid this for the main project I work on, which was the reason I was interested in #1444 to begin with.

Thomasdezeeuw:
@Jake-Shadle yes would appreciate that very much.

Thomasdezeeuw:
So, about the CI. We choose a "random" port we expect not be in use and guess what... another test using an OS assigned random port used that exact port :). So the `accept` test accepted an connection from the `connect_error` test, both of which didn't expect a connection at time. Restarting the CI.

Thomasdezeeuw:
Thanks @Jake-Shadle.

PR #1507, opened 2021-07-03
Update dependency

remilauzier:


Darksonn (commented):
What's the point of this? These updates are unnecessary because the new versions are semver compatible with the one already listed, and if mio is compatible with `log` version `0.4.x` with `x ≥ 8`, then there's no reason to force the user to use `x ≥ 14`.

(the dev-dependencies is another story and worth updating)

Thomasdezeeuw (approved):
Ok with updating just the dev-dependencies.

PR #1510, opened 2021-07-22
feature: added `IP_TOS` methods to `TcpSocket`

saiintbrisson:
This pull request adds two methods to `TcpSocket`: `set_tos` and `get_tos`. `IP_TOS` is useful in some applications where selecting the type-of-service is necessary.
It is important to note that the Microsoft docs state that `IP_TOS` should not be used without the Quality of Service API, although it does not have any impact on functionality.

### DIFF
--- a/src/sys/windows/tcp.rs
+++ b/src/sys/windows/tcp.rs
@@ -9,8 +9,8 @@ use std::time::Duration; // winapi uses usize, stdlib uses u32/u64.
 
 use winapi::ctypes::{c_char, c_int, c_ulong, c_ushort};
 use winapi::shared::mstcpip;
-use winapi::shared::ws2def::{AF_INET, AF_INET6, SOCKADDR_IN, SOCKADDR_STORAGE};
-use winapi::shared::ws2ipdef::SOCKADDR_IN6_LH;
+use winapi::shared::ws2def::{AF_INET, AF_INET6, IPPROTO_IP, SOCKADDR_IN, SOCKADDR_STORAGE};
+use winapi::shared::ws2ipdef::{IP_TOS, SOCKADDR_IN6_LH};
 
 use winapi::shared::minwindef::{BOOL, DWORD, FALSE, LPDWORD, LPVOID, TRUE};
 use winapi::um::winsock2::{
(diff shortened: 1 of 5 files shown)


Thomasdezeeuw:
@SaiintBrisson any chance you can use the socket2 library for this, e.g. via [`SockRef`](https://docs.rs/socket2/0.4.0/socket2/struct.SockRef.html)? My plan for v0.8 is still to remove the `TcpSocket` type.

tests/tcp_socket.rs
```diff
@@ -56,6 +56,26 @@ fn set_keepalive() {
...
+    target_os = "solaris",
+    target_os = "illumos",
+)))]
+#[test]
+fn set_tos() {
+    let addr = "127.0.0.1:0".parse().unwrap();
+
+    let tos = 0x10;
+    let socket = TcpSocket::new_v4().unwrap();
+    socket.set_tos(tos).unwrap();
```
Thomasdezeeuw, line 71: Can you add a `assert!(socket.get_tos() != tos);`, so we know the test actually tests something.

tests/tcp_socket.rs
```diff
@@ -56,6 +56,26 @@ fn set_keepalive() {
...
+    assert!(socket.get_tos().unwrap() == tos);
+
+    socket.bind(addr).unwrap();
```
Thomasdezeeuw, line 74: No need to bind our listener below, you can let the socket be dropped.

src/sys/windows/tcp.rs
```diff
@@ -290,6 +290,39 @@ pub(crate) fn get_keepalive(socket: TcpSocket) -> io::Result<bool> {
...
 }
 
+pub(crate) fn set_tos(socket: TcpSocket, tos: u32) -> io::Result<()> {
+    match unsafe {
+        setsockopt(
+            socket,
+            IPPROTO_IP,
+            IP_TOS,
+            &tos as *const _ as *const c_char,
+            size_of::<c_ulong>() as c_int,
```
Thomasdezeeuw, line 300: This must be `size_of::<c_int>`. Also see https://github.com/rust-lang/socket2/blob/1c67209c2eda5708b3093900a59befbc4a981b13/src/socket.rs#L1171-L1205.

saiintbrisson:
Hi! I'm really really sorry for the delay, I haven't been able to keep up with open-source contributions lately. I managed to use `socket2` so I'll close this PR. Thank you for the feedback!

PR #1514, opened 2021-09-28
Remove TcpSocket type

Thomasdezeeuw:
The socket2 crate provide all the functionality and more. Furthermore
supporting all socket options is beyond the scope of Mio.

The easier migration is to the socket2 crate, using the Socket or
SockRef types.

The migration for Tokio is tracked in
https://github.com/tokio-rs/tokio/issues/4135.


Thomasdezeeuw:
Related https://github.com/nix-rust/nix/issues/1548 and https://github.com/bitflags/bitflags/issues/255.

PR #1525, opened 2021-11-06
Deregister connection before dropping it in TCP example

### LINKED ISSUE #1269: deregister question in example
In example tcp_server.rs, when the connection socket is done or failed, it seems that no deregister from the registry.. Is it a leak?

Thomasdezeeuw:
Closes #1269


Noah-Kennedy (approved):
LGTM!

PR #1529, opened 2021-11-07
Add section about raw fd to portability guidelines

### LINKED ISSUE #1186: Document interaction with RawSocket (do not read / write directly)
We need to tell users what they are allowed to do when getting the handle of a TcpListener/TcpStream/UdpSocket through as_raw_handle/as_raw_socket on Windows.

Target: v1.0-release

Thomasdezeeuw:
Closes #1186


Noah-Kennedy (approved):
Looks good!

PR #1530, opened 2021-11-07
Add note about short receive on datagram sockets

### LINKED ISSUE #1174: Investigate and document UDP recv/peek(_from) behavior on *nix when using a smaller buffer.
On Windows, if the data is larger than the buffer specified, the buffer is filled with the first part of the data, and recv/peek(_from) returns the error WSAEMSGSIZE(10040). The excess data is lost. 

What does *nix do?

Ref PR https://github.com/tokio-rs/mio/pull/1167/files#diff-c022871452a328dc8c9cd10130eb1572R233

Thomasdezeeuw:
Talking about the differences between OSs.

Closes #1174

### DIFF
--- a/src/net/mod.rs
+++ b/src/net/mod.rs
@@ -6,6 +6,24 @@
 //! matter the target platform.
 //!
 //! [portability guidelines]: ../struct.Poll.html#portability
+//!
+//! # Notes
+//!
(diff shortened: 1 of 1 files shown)


Noah-Kennedy (approved):
Looks great!

PR #1549, opened 2022-02-15
Add TCP networking for target_os = "wasi"

haraldh:
With

* https://github.com/bytecodealliance/wasmtime/pull/3711
* https://github.com/rust-lang/rust/pull/93158

merged, mio can have limited support for TCP networking for the `wasm32-wasi` target.

I also modified the `tcp_server` example.

```
$ cargo +nightly build --target wasm32-wasi --release --example tcp_server --features="os-poll net"
   Compiling cfg-if v1.0.0
   Compiling wasi v0.10.2+wasi-snapshot-preview1
   Compiling log v0.4.14
   Compiling libc v0.2.112
   Compiling ppv-lite86 v0.2.15
   Compiling wasi v0.11.0+wasi-snapshot-preview1
   Compiling getrandom v0.2.3
   Compiling rand_core v0.6.3
   Compiling env_logger v0.8.4
   Compiling rand_chacha v0.3.1
   Compiling mio v0.8.0 (/home/harald/git/mio)
   Compiling rand v0.8.4
    Finished release [optimized] target(s) in 2.92s

$ wasmtime run --tcplisten 127.0.0.1:9000 --env 'LISTEN_FDS=1' target/wasm32-wasi/release/examples/tcp_server.wasm
Using preopened socket FD 3
You can connect to the server using `nc`:
 $ nc <IP> <PORT>
You'll see our welcome message and anything you type will be printed here.
```

### DIFF
--- a/examples/tcp_server.rs
+++ b/examples/tcp_server.rs
@@ -13,6 +13,7 @@ const SERVER: Token = Token(0);
 // Some data we'll send over the connection.
 const DATA: &[u8] = b"Hello world!\n";
 
+#[cfg(not(target_os = "wasi"))]
 fn main() -> io::Result<()> {
     env_logger::init();
 
@@ -180,4 +181,9 @@ fn would_block(err: &io::Error) -> bool {
 
 fn interrupted(err: &io::Error) -> bool {
     err.kind() == io::ErrorKind::Interrupted
+}
+
+#[cfg(target_os = "wasi")]
+fn main() {
+    panic!("can't bind to an address with wasi")
 }
--- a/examples/udp_server.rs
+++ b/examples/udp_server.rs
@@ -1,14 +1,16 @@
 // You can run this example from the root of the mio repo:
 // cargo run --example udp_server --features="os-poll net"
 use log::warn;
-use mio::net::UdpSocket;
 use mio::{Events, Interest, Poll, Token};
 use std::io;
 
 // A token to allow us to identify which event is for the `UdpSocket`.
 const UDP_SOCKET: Token = Token(0);
 
+#[cfg(not(target_os = "wasi"))]
 fn main() -> io::Result<()> {
+    use mio::net::UdpSocket;
+
     env_logger::init();
 
     // Create a poll instance.
@@ -19,6 +21,7 @@ fn main() -> io::Result<()> {
 
     // Setup the UDP socket.
     let addr = "127.0.0.1:9000".parse().unwrap();
+
     let mut socket = UdpSocket::bind(addr)?;
 
     // Register our socket with the token defined above and an interest in being
@@ -74,4 +77,9 @@ fn main() -> io::Result<()> {
             }
         }
     }
+}
+
+#[cfg(target_os = "wasi")]
+fn main() {
+    panic!("can't bind to an address with wasi")
 }
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -1,6 +1,8 @@
 use std::net::{self, SocketAddr};
 #[cfg(unix)]
 use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
+#[cfg(target_os = "wasi")]
+use std::os::wasi::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
 #[cfg(windows)]
 use std::os::windows::io::{AsRawSocket, FromRawSocket, IntoRawSocket, RawSocket};
 use std::{fmt, io};
@@ -9,6 +11,7 @@ use crate::io_source::IoSource;
 use crate::net::TcpStream;
 #[cfg(unix)]
 use crate::sys::tcp::set_reuseaddr;
+#[cfg(not(target_os = "wasi"))]
 use crate::sys::tcp::{bind, listen, new_for_addr};
 use crate::{event, sys, Interest, Registry, Token};
 
@@ -52,6 +55,7 @@ impl TcpListener {
     /// 2. Set the `SO_REUSEADDR` option on the socket on Unix.
     /// 3. Bind the socket to the specified address.
     /// 4. Calls `listen` on the socket to prepare it to receive new connections.
+    #[cfg(not(target_os = "wasi"))]
     pub fn bind(addr: SocketAddr) -> io::Result<TcpListener> {
         let socket = new_for_addr(addr)?;
         #[cfg(unix)]
@@ -213,5 +217,32 @@ impl FromRawSocket for TcpListener {
     /// non-blocking mode.
     unsafe fn from_raw_socket(socket: RawSocket) -> TcpListener {
         TcpListener::from_std(FromRawSocket::from_raw_socket(socket))
+    }
+}
+
+#[cfg(target_os = "wasi")]
+impl IntoRawFd for TcpListener {
+    fn into_raw_fd(self) -> RawFd {
+        self.inner.into_inner().into_raw_fd()
+    }
+}
+
+#[cfg(target_os = "wasi")]
+impl AsRawFd for TcpListener {
+    fn as_raw_fd(&self) -> RawFd {
+        self.inner.as_raw_fd()
+    }
+}
+
+#[cfg(target_os = "wasi")]
+impl FromRawFd for TcpListener {
+    /// Converts a `RawFd` to a `TcpListener`.
+    ///
+    /// # Notes
+    ///
+    /// The caller is responsible for ensuring that the socket is in
+    /// non-blocking mode.
+    unsafe fn from_raw_fd(fd: RawFd) -> TcpListener {
+        TcpListener::from_std(FromRawFd::from_raw_fd(fd))
     }
 }
--- a/src/sys/wasi/mod.rs
+++ b/src/sys/wasi/mod.rs
@@ -1,0 +1,356 @@
+//! # Notes
+//!
+//! The current implementation is somewhat limited. The `Waker` is not
+//! implemented, as at the time of writing there is no way to support to wake-up
+//! a thread from calling `poll_oneoff`.
+//!
+//! Furthermore the (re/de)register functions also don't work while concurrently
+//! polling as both registering and polling requires a lock on the
+//! `subscriptions`.
+//!
+//! Finally `Selector::try_clone`, required by `Registry::try_clone`, doesn't
+//! work. However this could be implemented by use of an `Arc`.
+//!
+//! In summary, this only (barely) works using a single thread.
+
+use std::cmp::min;
+use std::io;
+#[cfg(all(feature = "net", debug_assertions))]
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+
+#[cfg(feature = "net")]
+use crate::{Interest, Token};
+
+cfg_net! {
+    pub(crate) mod tcp {
+        use std::io;
+        use std::net::{self, SocketAddr};
+
+        pub(crate) fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream, SocketAddr)> {
+            let (stream, addr) = listener.accept()?;
+            stream.set_nonblocking(true)?;
+            Ok((stream, addr))
+        }
+    }
+}
+
+/// Unique id for use as `SelectorId`.
+#[cfg(all(debug_assertions, feature = "net"))]
+static NEXT_ID: AtomicUsize = AtomicUsize::new(1);
+
+pub(crate) struct Selector {
+    #[cfg(all(debug_assertions, feature = "net"))]
+    id: usize,
+    /// Subscriptions (reads events) we're interested in.
+    subscriptions: Arc<Mutex<Vec<wasi::Subscription>>>,
+}
+
+impl Selector {
+    pub(crate) fn new() -> io::Result<Selector> {
+        Ok(Selector {
+            #[cfg(all(debug_assertions, feature = "net"))]
+            id: NEXT_ID.fetch_add(1, Ordering::Relaxed),
+            subscriptions: Arc::new(Mutex::new(Vec::new())),
+        })
+    }
+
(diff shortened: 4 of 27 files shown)


Thomasdezeeuw:
Have you actually tested this? Because last I remember my code didn't really work.

haraldh:
yes I did 😀

haraldh:
Corrected the `tcp_server` example build and run instructions.

src/sys/wasi/net.rs
```diff
@@ -0,0 +1,45 @@
...
+        Err(unsupported())
+    }
+
+    pub(crate) fn listen(_: &net::TcpListener, _: u32) -> io::Result<()> {
+        Err(unsupported())
+    }
+
+    pub(crate) fn set_reuseaddr(_: &net::TcpListener, _: bool) -> io::Result<()> {
+        Err(unsupported())
+    }
```
npmccallum, line 38: It is true that this is unsupported. However, I think we can also safely lie here and just return `Ok(())`. I don't see any negative consequences of this.

haraldh: sure, we can lie... but it does not gain anything for the mio internals.

haraldh:
> @haraldh I just tried the TCP server example, but it doesn't work. Have you tried running the non-wasm version? The non-wasm returns a response on each packet of bytes send, but the wasm only responds when the connection is closed. When using `nc` you see a response every time yo hit enter (which sends the line you entered), but that doesn't work for the wasm version.

You are totally right... how about now with `stream.set_nonblocking(true)` ??

```rust
    pub(crate) fn accept(
        listener: &net::TcpListener,
    ) -> io::Result<(net::TcpStream, net::SocketAddr)> {
        let (stream, addr) = listener.accept()?;
        stream.set_nonblocking(true)?;
        Ok((stream, addr))
    }
```

@Thomasdezeeuw should now work as expected.

Thomasdezeeuw (commented):
Their are some more problems with this stil.

 * Can't use read and write `Interest` in a single register call. This is a major issue.
 * `Waker` not implemented.
 * `Selector::try_clone` is broken.
 * Still needs UDP implementation.
 * Multiple threads not supported. Not a super big issue I think.

Thomasdezeeuw:
I put some review notes and change in https://github.com/Thomasdezeeuw/mio/commit/e5307791d853c1cab8c0c0c9e7617e0d198f4afc.

npmccallum:
> Their are some more problems with this stil.
> 
> * Can't use read and write `Interest` in a single register call. This is a major issue.
> * `Waker` not implemented.
> * `Selector::try_clone` is broken.
> * Still needs UDP implementation.
> * Multiple threads not supported. Not a super big issue I think.

WebAssembly has no threads. Therefore, at least some of these issues don't apply. For example, what is the point of implementing `Waker` if there is only one thread?

Thomasdezeeuw:
> > Their are some more problems with this stil.
> > 
> > * Can't use read and write `Interest` in a single register call. This is a major issue.
> > * `Waker` not implemented.
> > * `Selector::try_clone` is broken.
> > * Still needs UDP implementation.
> > * Multiple threads not supported. Not a super big issue I think.
> 
> WebAssembly has no threads. Therefore, at least some of these issues don't apply. For example, what is the point of implementing `Waker` if there is only one thread?

It's that something in the pipeline? I don't remember the details of it, but I thought it was, but last I checked was a couple of years ago.

> WASI has no UDP support. Only streams are supported at this time.

Ok.

So then we have:
 * Can't use read and write `Interest` in a single register call. This is still a major issue.
 * ~~`Waker` not implemented.~~ We'll need to remove the public type.
 * `Selector::try_clone` is broken. We could remove the implementation for now.
 * ~~Still needs UDP implementation.~~
 * Multiple threads not supported. Not a super big issue I think.

Could you `cfg` out the things that don't make sense for wasm, like the `Waker` type.

haraldh:
@Thomasdezeeuw thanks for your review. I merged your PR and added additional `cfg` according to your comments.

Also added more `cfg`, so that `cargo hack test --feature-powerset` also works with:
```
$ CARGO_TARGET_WASM32_WASI_RUNNER="wasmtime run" cargo +nightly hack test --target wasm32-wasi --feature-powerset
```

src/sys/wasi/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+//! `subscriptions`.
+//!
+//! Finally `Selector::try_clone`, required by `Registry::try_clone`, doesn't
+//! work. However this could be implemented by use of an `Arc`.
+//!
+//! In summary, this only (barely) works using a single thread.
+
+use std::cmp::min;
+use std::io;
+#[cfg(debug_assertions)]
```
Thomasdezeeuw, line 18: Nit: these `cfg`s can be merge, also in other locations.

src/sys/wasi/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+    #[cfg(feature = "net")]
+    pub(crate) fn register(
+        &self,
+        fd: wasi::Fd,
+        token: Token,
+        interests: Interest,
+    ) -> io::Result<()> {
+        let mut subscriptions = self.subscriptions.lock().unwrap();
+
+        // Cannot subscribe to read and write at the same time.
```
Thomasdezeeuw, line 127: This is still a blocking problem. It breaks Mio's design.

Thomasdezeeuw (changes requested):

src/sys/wasi/mod.rs
haraldh: That should be brought up upstream then, because that's how it currently works or better does not work.

npmccallum: @haraldh What is upstream in this case? Is it `wasmtime` or `WASI`?

npmccallum: It is possible to fix `wasmtime`. A fix for `WASI` is highly unlikely.

Thomasdezeeuw: Perhaps you could open an issue here https://github.com/WebAssembly/WASI.

haraldh: https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/witx/typenames.witx#L486-L499

haraldh: maybe push 2 subscriptions then?

npmccallum: @Thomasdezeeuw There will be no new changes WASI snapshots until modularization is complete. So we have to work with what we have. Please note that I'm the person who pushed `sock_accept()` through the committee.

Thomasdezeeuw: > maybe push 2 subscriptions then?

If I remember correctly that doesn't work, but you can try.

> @Thomasdezeeuw There will be no new changes WASI snapshots until modularization is complete. So we have to work with what we have.

Is their an ETA for that? Because if we can't call `register` with both read and write interest Mio simply won't work with WASI.

> Please note that I'm the person who pushed `sock_accept()` through the committee.

Good to know 👍

Thomasdezeeuw: To be clear if creating 2 subscriptions work we're all good, kqueue also needs separate events: https://github.com/tokio-rs/mio/blob/e8aed7baea88e844a7402a29ec8675ddc341edd9/src/sys/unix/selector/kqueue.rs#L133-L147.

npmccallum: @Thomasdezeeuw Unfortunately, there is no ETA for modularization.

Fortunately, I don't see anything in the WASI spec that makes multiple subscriptions invalid. So I propose we treat this as a `wasmtime` bug, if the bug exists at all. I suggest we make the minor change to allow simultaneous read and write subscriptions and if `wasmtime` chokes on this we file a followup PR to `wasmtime` under the assumption they are non-compliant.

npmccallum: @haraldh Any update?

haraldh: It is wasmtime... see https://github.com/bytecodealliance/wasmtime/pull/3866

haraldh: I added one commit, which made polling on read and write possible with a patched wasmtime.

src/sys/wasi/mod.rs
```diff
@@ -186,12 +187,14 @@ impl Selector {
         };
         if let Some(index) = subscriptions.iter().position(predicate) {
             subscriptions.swap_remove(index);
-            /* NOTE: can't have read and write subscriptions.
             // We might have two subscriptions (read and write).
-            if let Some(index) = subscriptions.iter().skip(index).position(predicate) {
-                subscriptions.swap_remove(index);
+            loop {
```
Thomasdezeeuw, line 191: I think this loop should be moved to be outside of the `if` statement on 188, which is the same as the contents of the loop.

Thomasdezeeuw, line 191: I think entire if-else block can also be replace with `while let Some(index) = subscriptions.iter().position(predicate) { subscriptions.swap_remove(index); }` Then we just need to check if any subscription was removed to return the `NotFound` error.

haraldh: done

haraldh: done

examples/tcp_server.rs
```diff
@@ -92,92 +90,109 @@ fn main() -> io::Result<()> {
...
+    pub fn interrupted(err: &io::Error) -> bool {
+        err.kind() == io::ErrorKind::Interrupted
+    }
 }
 
-fn interrupted(err: &io::Error) -> bool {
-    err.kind() == io::ErrorKind::Interrupted
+#[cfg(target_os = "wasi")]
+fn main() {
+    unimplemented!()
```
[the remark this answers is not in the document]
haraldh: done

src/sys/wasi/mod.rs
haraldh: done

examples/udp_server.rs
```diff
@@ -75,3 +78,8 @@ fn main() -> io::Result<()> {
         }
     }
 }
+
+#[cfg(target_os = "wasi")]
+fn main() {
+    unimplemented!()
```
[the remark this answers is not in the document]
haraldh: done

src/net/tcp/listener.rs
```diff
@@ -52,9 +55,10 @@ impl TcpListener {
     /// 2. Set the `SO_REUSEADDR` option on the socket on Unix.
     /// 3. Bind the socket to the specified address.
     /// 4. Calls `listen` on the socket to prepare it to receive new connections.
+    #[cfg(not(target_os = "wasi"))]
     pub fn bind(addr: SocketAddr) -> io::Result<TcpListener> {
         let socket = new_for_addr(addr)?;
-        #[cfg(unix)]
+        #[cfg(any(unix, target_os = "wasi"))]
```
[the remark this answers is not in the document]
haraldh: done

haraldh:
pushed more commits to be squashed in the end

examples/tcp_server.rs
```diff
@@ -1,19 +1,17 @@
...
-use mio::event::Event;
-use mio::net::{TcpListener, TcpStream};
-use mio::{Events, Interest, Poll, Registry, Token};
-use std::collections::HashMap;
-use std::io::{self, Read, Write};
-use std::str::from_utf8;
 
-// Setup some tokens to allow us to identify which event is for which socket.
-const SERVER: Token = Token(0);
+#[cfg(not(target_os = "wasi"))]
```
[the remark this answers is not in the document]
haraldh: done

examples/tcp_server.rs
```diff
@@ -13,6 +13,24 @@ const SERVER: Token = Token(0);
 // Some data we'll send over the connection.
 const DATA: &[u8] = b"Hello world!\n";
 
+#[cfg(not(windows))]
```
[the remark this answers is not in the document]
haraldh: done

src/sys/wasi/mod.rs
haraldh: wasmtime PR was merged 😀

npmccallum:
@Thomasdezeeuw Any chance we can merge this? Are we waiting for something?

examples/tcp_server.rs
```diff
@@ -1,19 +1,17 @@
 // You can run this example from the root of the mio repo:
 // cargo run --example tcp_server --features="os-poll net"
-use mio::event::Event;
```
Thomasdezeeuw, line 3: Can you revert all changes here, the changes make the example much harder to read. Please use the same structure as for the UDP example, with the extra `main`. I know it will generate a number of warnings, but that's OK.

Thomasdezeeuw (approved):
One final nit, but otherwise LGTM.

If you don't get to the final change @haraldh I'll do it tonight and merge the pr.

examples/tcp_server.rs
haraldh: yeah, will do

haraldh:
@Thomasdezeeuw squashed and left the old examples alone

haraldh:
ok, ready to be merged, except if you want to squash all commits.

Thomasdezeeuw (approved):

npmccallum:
@Thomasdezeeuw Thanks! When can we get this in a release so that we can start working on the higher level components?

hi-rustin:
I can't seem to run it in a non-wasm environment anymore. How should I run this example on macos?

Thomasdezeeuw:
Did you follow the instruction in the example? https://github.com/tokio-rs/mio/blob/e077a23a4d7fd548da1381889ae3eaa5341512e7/examples/tcp_listenfd_server.rs#L4-L5

That run for me on macOS. If it doesn't work for you can you provide some more details such as the error you're seeing.

haraldh:
Oh, yeah the non-wasi version needs some update.

Thomasdezeeuw:
> What is the purpose of our setting LISTEN_FDS here? We don't seem to be using this value.

The `$LISTEN_FDS` env var is a commonly used init systems, such as systemd, to indicate opened file descriptors (outside of standard in/out/err).

PR #1551, opened 2022-03-08
add try_io

### LINKED ISSUE #1550: make do_io() public to allow customized I/O operations
I'm trying to use WSARecvMsg() on mio::net::UDPSocket. And I find that I should call WSARecvMsg() in do_io() like UDPSocket#recv(). But I cannot do because do_io() is not public.

So, I think that do_io() should be callable by the user of mio::net::UDPSocket.
Is this ok?

Currently, I'm working in https://github.com/masa-koz/mio/blob/6dd2162e526cc74bdca183b7c1ec46565aecad83/src/net/udp.rs#L329

masa-koz:
fix #1550

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,9 +40,9 @@ net = []
 
 [dependencies]
 log = "0.4.8"
-libc = "0.2.86"
 
 [target.'cfg(unix)'.dependencies]
+libc = "0.2.86"
 
 [target.'cfg(windows)'.dependencies]
 miow   = "0.3.6"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,9 +40,9 @@ net = []
 
 [dependencies]
 log = "0.4.8"
+libc = "0.2.86"
 
 [target.'cfg(unix)'.dependencies]
-libc = "0.2.86"
 
 [target.'cfg(windows)'.dependencies]
 miow   = "0.3.6"
--- a/src/net/tcp/stream.rs
+++ b/src/net/tcp/stream.rs
@@ -224,7 +224,7 @@ impl TcpStream {
     /// let address = "127.0.0.1:8080".parse().unwrap();
     /// let stream = TcpStream::connect(address)?;
     ///
-    /// // Wait until the steram is readable...
+    /// // Wait until the stream is readable...
     ///
     /// // Read from the stream using a direct libc call, of course the
     /// // `io::Read` implementation would be easier to use.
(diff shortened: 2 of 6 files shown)


src/net/tcp/stream.rs
```diff
@@ -199,6 +199,55 @@ impl TcpStream {
...
+    ///
+    /// // Wait until the steram is readable...
+    ///
+    /// // Read from the stream using a direct libc call, of course the
+    /// // `io::Read` implementation would be easier to use.
+    /// let mut buf = [0; 512];
+    /// let n = stream.try_io(|| {
+    ///     let buf_ptr = &mut buf as *mut _ as *mut _;
+    ///     let res = unsafe { libc::recv(stream.as_raw_fd(), buf_ptr, buf.len(), 0) };
+    ///     if res != -1 {
```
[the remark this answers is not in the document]
masa-koz: Ok. I will add it.

src/net/tcp/stream.rs
```diff
@@ -199,6 +199,57 @@ impl TcpStream {
...
+    /// #
+    /// # fn main() -> Result<(), Box<dyn Error>> {
+    /// use std::io;
+    /// use std::os::unix::io::AsRawFd;
+    /// use mio::net::TcpStream;
+    ///
+    /// let address = "127.0.0.1:8080".parse().unwrap();
+    /// let stream = TcpStream::connect(address)?;
+    ///
+    /// // Wait until the steram is readable...
```
haraldh, line 227:
```suggestion
    /// // Wait until the stream is readable...
```

masa-koz: Thanks!

Cargo.toml
```diff
@@ -40,9 +40,9 @@ net = []
 
 [dependencies]
 log = "0.4.8"
+libc = "0.2.86"
```
Thomasdezeeuw, line 43: Revert this, not needed for this change. If needed you can make the example unix only.

Thomasdezeeuw (changes requested):

Cargo.toml
masa-koz: Can I add libc in [target.'cfg(windows)'.dev-dependencies]?
If not acceptable, I would like to know how to make the example unix only.

masa-koz: Thanks! I will try it.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @masa-koz.

PR #1556, opened 2022-03-18
Replace miow, winapi, and ntapi with windows-sys

Jake-Shadle:
This removes the dependency on `miow`, `winapi`, and `ntapi` in favor of just using `windows-sys` directly. A few types/functions from `miow` were copied, and added, namely the `Handle` wrapper (for `CloseHandle` dropping) and the `CompletionPort` and `CompletionStatus` types.

There was only a single function, `NtCancelIoFileEx` that was present in `ntapi` but not `windows-sys`, so I merely added the extern declaration in the one place it was used as it is not worth bringing in a dependency just for that.

Note this uses the latest `window-sys` version, but I think the version restriction can be loosened to `>=0.32` as (probably) the only actual relevant change for mio was the change of `HANDLE` from `*mut c_void` to `isize`. This is unfortunately a downside of the `windows-sys` crate, its minor version is bumped fairly frequently, but other than changes such as the one previously mentioned, most of those versions won't actually meaningfully affect mio's usage, and having an exact version requirement could mean duplicate versions of the `windows-sys` crates being used in downstream crates if they use a slightly newer version of it that is otherwise fully API compatible. Though of course the safest is to have `>=0.32, <=0.34` to avoid a minor version > 0.34 _actually_ having breaking changes for mio and downstream crates failing to compile unless they manually pin the version.

### DIFF
--- a/src/sys/windows/afd.rs
+++ b/src/sys/windows/afd.rs
@@ -14,9 +14,13 @@ use windows_sys::Win32::{
 
 const IOCTL_AFD_POLL: u32 = 0x00012024;
 
-// <https://processhacker.sourceforge.io/doc/ntioapi_8h.html#a0d4d550cad4d62d75b76961e25f6550c>
 #[link(name = "ntdll")]
 extern "system" {
+    /// See <https://processhacker.sourceforge.io/doc/ntioapi_8h.html#a0d4d550cad4d62d75b76961e25f6550c>
+    ///
+    /// This is an undocumented API and as such not part of <https://github.com/microsoft/win32metadata>
+    /// from which `windows-sys` is generated, and also unlikely to be added, so
+    /// we manually declare it here
     fn NtCancelIoFileEx(
         FileHandle: HANDLE,
         IoRequestToCancel: *mut IO_STATUS_BLOCK,
--- a/src/sys/windows/afd.rs
+++ b/src/sys/windows/afd.rs
@@ -1,17 +1,28 @@
-use ntapi::ntioapi::{IO_STATUS_BLOCK_u, IO_STATUS_BLOCK};
-use ntapi::ntioapi::{NtCancelIoFileEx, NtDeviceIoControlFile};
-use ntapi::ntrtl::RtlNtStatusToDosError;
+use std::ffi::c_void;
 use std::fmt;
 use std::fs::File;
 use std::io;
 use std::mem::size_of;
 use std::os::windows::io::AsRawHandle;
 use std::ptr::null_mut;
-use winapi::shared::ntdef::{HANDLE, LARGE_INTEGER, NTSTATUS, PVOID, ULONG};
-use winapi::shared::ntstatus::{STATUS_NOT_FOUND, STATUS_PENDING, STATUS_SUCCESS};
+use windows_sys::Win32::{
+    Foundation::{
+        RtlNtStatusToDosError, HANDLE, NTSTATUS, STATUS_NOT_FOUND, STATUS_PENDING, STATUS_SUCCESS,
+    },
+    System::WindowsProgramming::{NtDeviceIoControlFile, IO_STATUS_BLOCK, IO_STATUS_BLOCK_0},
+};
 
-const IOCTL_AFD_POLL: ULONG = 0x00012024;
+const IOCTL_AFD_POLL: u32 = 0x00012024;
 
+/// <https://processhacker.sourceforge.io/doc/ntioapi_8h.html#a0d4d550cad4d62d75b76961e25f6550c>
+#[link(name = "ntdll")]
+extern "system" {
+    fn NtCancelIoFileEx(
+        FileHandle: HANDLE,
+        IoRequestToCancel: *mut IO_STATUS_BLOCK,
+        IoStatusBlock: *mut IO_STATUS_BLOCK,
+    ) -> NTSTATUS;
+}
 /// Winsock2 AFD driver instance.
 ///
 /// All operations are unsafe due to IO_STATUS_BLOCK parameter are being used by Afd driver during STATUS_PENDING before I/O Completion Port returns its result.
@@ -24,7 +35,7 @@ pub struct Afd {
 #[derive(Debug)]
 pub struct AfdPollHandleInfo {
     pub handle: HANDLE,
-    pub events: ULONG,
+    pub events: u32,
     pub status: NTSTATUS,
 }
 
@@ -32,10 +43,10 @@ unsafe impl Send for AfdPollHandleInfo {}
 
 #[repr(C)]
 pub struct AfdPollInfo {
-    pub timeout: LARGE_INTEGER,
+    pub timeout: i64,
     // Can have only value 1.
-    pub number_of_handles: ULONG,
-    pub exclusive: ULONG,
+    pub number_of_handles: u32,
+    pub exclusive: u32,
     pub handles: [AfdPollHandleInfo; 1],
 }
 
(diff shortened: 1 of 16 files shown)


Thomasdezeeuw:
I think @Noah-Kennedy was working on this as well.

kennykerr:
Might be worth switching `miow` and `mio` directly to `windows-sys` version 0.33 as that is what the prominent [parking_lot](https://github.com/Amanieu/parking_lot/blob/master/core/Cargo.toml#L26) and [winit](https://github.com/rust-windowing/winit/blob/master/Cargo.toml#L61) crates have adopted. 

@yoshuawuyts

kennykerr:
Here's a PR update `miow` to use `windows-sys` 0.33: https://github.com/yoshuawuyts/miow/pull/54

Thomasdezeeuw:
> Might be worth switching `miow` and `mio` directly to `windows-sys` version 0.33 as that is what the prominent [parking_lot](https://github.com/Amanieu/parking_lot/blob/master/core/Cargo.toml#L26) and [winit](https://github.com/rust-windowing/winit/blob/master/Cargo.toml#L61) crates have adopted.

I think it was @Noah-Kennedy place to switch directly to use windows-sys/winapi, but I'm not 100% sure about that.

Noah-Kennedy:
I think its probably best to just use either winapi or windows-sys directly and not depend on miow.

Jake-Shadle:
I can remove the dependency on miow in this PR if that's what's wanted? I love removing dependencies.

Thomasdezeeuw:
> I can remove the dependency on miow in this PR if that's what's wanted? I love removing dependencies.

Sounds good.

src/sys/windows/afd.rs
```diff
@@ -1,17 +1,28 @@
...
+        RtlNtStatusToDosError, HANDLE, NTSTATUS, STATUS_NOT_FOUND, STATUS_PENDING, STATUS_SUCCESS,
+    },
+    System::WindowsProgramming::{NtDeviceIoControlFile, IO_STATUS_BLOCK, IO_STATUS_BLOCK_0},
+};
+
+const IOCTL_AFD_POLL: u32 = 0x00012024;
+
+// <https://processhacker.sourceforge.io/doc/ntioapi_8h.html#a0d4d550cad4d62d75b76961e25f6550c>
+#[link(name = "ntdll")]
+extern "system" {
```
Thomasdezeeuw, line 19: Should this be a part of `windows-sys`?

src/sys/windows/afd.rs
```diff
@@ -1,17 +1,28 @@
...
-use winapi::shared::ntstatus::{STATUS_NOT_FOUND, STATUS_PENDING, STATUS_SUCCESS};
-
-const IOCTL_AFD_POLL: ULONG = 0x00012024;
 
+use windows_sys::Win32::{
+    Foundation::{
+        RtlNtStatusToDosError, HANDLE, NTSTATUS, STATUS_NOT_FOUND, STATUS_PENDING, STATUS_SUCCESS,
+    },
+    System::WindowsProgramming::{NtDeviceIoControlFile, IO_STATUS_BLOCK, IO_STATUS_BLOCK_0},
+};
```
Thomasdezeeuw, line 13: Nit: Could you not use nested imports, I find them quite annoying and needlessly difficult too read.

```suggestion
use windows_sys::Win32::Foundation::{
    RtlNtStatusToDosError, HANDLE, NTSTATUS, STATUS_NOT_FOUND, STATUS_PENDING, STATUS_SUCCESS,
};
use windows_sys::Win32::Foundation::System::WindowsProgramming::{NtDeviceIoControlFile, IO_STATUS_BLOCK, IO_STATUS_BLOCK_0};
```

src/sys/windows/handle.rs
```diff
@@ -0,0 +1,30 @@
+use std::os::windows::io::RawHandle;
+use windows_sys::Win32::Foundation::{CloseHandle, HANDLE};
+
+/// Wrapper around a Windows HANDLE so that we close it upon drop in all scenarios
+#[derive(Debug)]
+pub struct Handle(HANDLE);
```
Thomasdezeeuw, line 6: Not a review comment, but future work: we can replace this with [`std::os::windows::io::OwnedHandle`](https://doc.rust-lang.org/std/os/windows/io/struct.OwnedHandle.html) (currently unstable).

src/sys/windows/net.rs
```diff
@@ -3,12 +3,16 @@ use std::mem;
...
-use winapi::shared::ws2ipdef::{SOCKADDR_IN6_LH_u, SOCKADDR_IN6_LH};
-use winapi::um::winsock2::{ioctlsocket, socket, FIONBIO, INVALID_SOCKET, SOCKET};
+use windows_sys::Win32::Networking::WinSock::{
+    ioctlsocket, socket, FIONBIO, IN6_ADDR, IN6_ADDR_0, INVALID_SOCKET, IN_ADDR, IN_ADDR_0,
+    SOCKADDR, SOCKADDR_IN, SOCKADDR_IN6, SOCKADDR_IN6_0, SOCKET,
+};
+
+// AF_INET/6 are in the Win32_NetworkManagement_IpHelper feature for some reason,
+// so just directly define them here rather than import that whole feature set
+// these constants won't ever change
```
Thomasdezeeuw, line 13: Although I get not wanting to add a feature for two constant I don't want to keep these constant in Mio, that should be done in `windows-sys`.

Thomasdezeeuw (commented):
Gave it a could look now, will do a proper review later.

One thing I noticed that you define various constants, I think those should all be moved to `windows-sys`.

src/sys/windows/net.rs
kennykerr: Feel free to suggest a more appropriate feature here: https://github.com/microsoft/win32metadata

src/sys/windows/afd.rs
kennykerr: This appears to be an undocumented API that does not appear in the Windows SDK. The `windows-sys` crate is generated from metadata for the Windows SDK. You can [request it be added here](https://github.com/microsoft/win32metadata) but given that it is undocumented, it may not be accepted.

Thomasdezeeuw (approved):
LGTM, @Noah-Kennedy can you review this as well?

Some future work:
 * Move the `SIO_BSP_*` constants, https://github.com/microsoft/win32metadata/issues/844.
 * Replace `Handle` with [`std::os::windows::io::OwnedHandle`](https://doc.rust-lang.org/std/os/windows/io/struct.OwnedHandle.html) (once stable).
 * Remove the feature for `AF_INET(6)` constants, see https://github.com/microsoft/win32metadata/issues/845.

Thomasdezeeuw:
One open question is should this be a part of v0.8.x or v0.9.x? We don't use winapi's types publicly, but the rest of the ecosystem might still be using it, having both winapi and windows-sys might become a problem.

src/sys/windows/named_pipe.rs
```diff
@@ -103,6 +113,193 @@ impl Inner {
...
+    /// `overlapped` pointers to be valid until the end of the I/O operation.
+    /// The kernel also requires that `overlapped` is unique for this I/O
+    /// operation and is not in use for any other I/O.
+    ///
+    /// To safely use this function callers must ensure that the pointers are
+    /// valid until the I/O operation is completed, typically via completion
+    /// ports and waiting to receive the completion notification on the port.
+    pub unsafe fn read_overlapped(
+        &self,
+        buf: &mut [u8],
```
Noah-Kennedy, line 188: I've never quite been a fan of how miow did this, as its kinda a footgun. Still, I can't think of anything actively harmful about taking this in as an arg.

Noah-Kennedy (approved):

Thomasdezeeuw:
Thanks @Jake-Shadle! We can do some of the future work lined out in https://github.com/tokio-rs/mio/pull/1556#pullrequestreview-920136232 in a future pr

Thomasdezeeuw:
Created https://github.com/tokio-rs/mio/issues/1558 to keep track of that future work.

PR #1564, opened 2022-04-10
Don't ignore EAGAIN in UDS connect call

### LINKED ISSUE #1560: UnixStream::connect() returns unconnected socket on Linux
This happens when the server's backlog is full.

In this scenario, on linux, `connect()` returns `EAGAIN`, which means that you have to call `connect()` again.

Note that is different from `EINPROGRESS` which means that the connection is being established in the background, and you should poll the socket to find out when it's done.

Mio doesn't do any of this; it just returns the unconnected socket: https://github.com/tokio-rs/mio/blob/fc2080cd0a663125b59f0afa2bc1c771a3fbec1e/src/sys/unix/uds/stream.rs#L16.

Thomasdezeeuw:
Per the Linux manual:
> EAGAIN For nonblocking UNIX domain sockets, the socket is
>        nonblocking, and the connection cannot be completed
>        immediately.

Previously the code incorrectly assumed that the connection was now in
progress, which is incorrect. If that were the case EINPROGRESS
would/should be returned by the OS.

Also documents that connect calls can return a WouldBlock error.
This is a reflect of the Linux manual, which unfortunately doesn't
specify how to determine when the connect call should be retried.

Fixes #1560

### DIFF
--- a/src/net/uds/stream.rs
+++ b/src/net/uds/stream.rs
@@ -15,6 +15,9 @@ pub struct UnixStream {
 
 impl UnixStream {
     /// Connects to the socket named by `path`.
+    ///
+    /// This may return a `WouldBlock` in which case the socket connection
+    /// cannot be completed immediately.
     pub fn connect<P: AsRef<Path>>(path: P) -> io::Result<UnixStream> {
         sys::uds::stream::connect(path.as_ref()).map(UnixStream::from_std)
     }
--- a/src/net/uds/stream.rs
+++ b/src/net/uds/stream.rs
@@ -17,7 +17,7 @@ impl UnixStream {
     /// Connects to the socket named by `path`.
     ///
     /// This may return a `WouldBlock` in which case the socket connection
-    /// cannot be completed immediately.
+    /// cannot be completed immediately. Usually it means the backlog is full.
     pub fn connect<P: AsRef<Path>>(path: P) -> io::Result<UnixStream> {
         sys::uds::stream::connect(path.as_ref()).map(UnixStream::from_std)
     }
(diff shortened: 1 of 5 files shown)


src/sys/unix/uds/stream.rs
```diff
@@ -13,7 +13,7 @@ pub(crate) fn connect(path: &Path) -> io::Result<net::UnixStream> {
 
     match syscall!(connect(socket, sockaddr, socklen)) {
         Ok(_) => {}
-        Err(ref e) if e.kind() == io::ErrorKind::WouldBlock => {}
+        Err(ref err) if err.raw_os_error() == Some(libc::EINPROGRESS) => {}
```
Thomasdezeeuw, line 16: I've added `EINPROGRESS` instead because I can't really tell from the macOS manual if they guarantee that `EAGAIN` is always returned (unlike e.g. Linux).

Noah-Kennedy: Unfortunately I can't tell either; that manual isn't great.

Noah-Kennedy (approved):

src/net/uds/stream.rs
```diff
@@ -15,6 +15,9 @@ pub struct UnixStream {
 
 impl UnixStream {
     /// Connects to the socket named by `path`.
+    ///
+    /// This may return a `WouldBlock` in which case the socket connection
+    /// cannot be completed immediately.
```
piscisaureus, line 20: Usually the backlog is full.

piscisaureus:
Thanks @Thomasdezeeuw !

PR #1569, opened 2022-04-27
Update to windows-sys 0.36

Thomasdezeeuw:
Removes the need for the Win32_NetworkManagement_IpHelper feature as
AF_INET and AF_INET6 are now included in the Win32_Networking_WinSock
feature.

Updates #1558

PR #1570, opened 2022-05-02
Relax windows-sys dependency version

atouchet:
Was this supposed to be set to 0.36 exactly or was that an oversight from #1569?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ log = "0.4.8"
 libc = "0.2.121"
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = "=0.36"
+version = "0.36"
 features = [
   "Win32_Storage_FileSystem",         # Enables NtCreateFile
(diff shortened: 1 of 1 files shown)


Noah-Kennedy:
@Thomasdezeeuw I don't see any reason for this to be necessary

Noah-Kennedy:
I am pretty sure this is an oversight but I'd like to check with thomas

Thomasdezeeuw (approved):
LGTM.

Thomasdezeeuw:
Thanks @atouchet.

101100:
With the very rapid release schedule of the `windows-sys` crate with apparently no regard for semantic versioning would you accept a PR to relax the version even further to either `>0.36` or `0` (which is equivalent to `0.*` as I understand it)?

Thomasdezeeuw:
@101100 I don't think we can do that as windows-sys 0.37 is allowed to break the API, which means that at some point Mio might not build any more.

PR #1572, opened 2022-05-07
Future-proof the kevent ABI

asomers:
FreeBSD 12 adds some new private fields to struct kevent, and it changes
the type of "data" from intptr_t to i64.  For now libc only binds the
11-compat ABI, but that will change some day.  Adjust mio's structure
definitions for compatibility with either ABI.

This should work on all 32 and 64 bit platforms.

### DIFF
--- a/src/sys/unix/selector/kqueue.rs
+++ b/src/sys/unix/selector/kqueue.rs
@@ -1,6 +1,6 @@
 use crate::{Interest, Token};
 use log::error;
-use std::mem::MaybeUninit;
+use std::mem::{self, MaybeUninit};
 use std::ops::{Deref, DerefMut};
 use std::os::unix::io::{AsRawFd, RawFd};
 #[cfg(debug_assertions)]
@@ -34,17 +34,6 @@ type Flags = u16;
 #[cfg(target_os = "netbsd")]
 type Flags = u32;
 
-// Type of the `data` field in the `kevent` structure.
-#[cfg(any(
-    target_os = "dragonfly",
-    target_os = "freebsd",
-    target_os = "ios",
-    target_os = "macos"
-))]
-type Data = libc::intptr_t;
-#[cfg(any(target_os = "netbsd", target_os = "openbsd"))]
-type Data = i64;
-
 // Type of the `udata` field in the `kevent` structure.
 #[cfg(not(target_os = "netbsd"))]
 type UData = *mut libc::c_void;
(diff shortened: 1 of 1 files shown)


src/sys/unix/selector/kqueue.rs
```diff
@@ -34,17 +34,6 @@ type Flags = u16;
...
 type Flags = u32;
 
-// Type of the `data` field in the `kevent` structure.
-#[cfg(any(
-    target_os = "dragonfly",
-    target_os = "freebsd",
-    target_os = "ios",
-    target_os = "macos"
-))]
-type Data = libc::intptr_t;
```
Thomasdezeeuw, line 44: DragonFlyBSD and MacOS (and likely iOS) still use `intptr_t` in the manual. I think to be safe we should stick to that as well. You can just move FreeBSD down to NetBSD and OpenBSD.

src/sys/unix/selector/kqueue.rs
```diff
@@ -60,6 +49,7 @@ macro_rules! kevent {
             fflags: 0,
             data: 0,
             udata: $data as UData,
+            ..unsafe { mem::zeroed() }
```
Thomasdezeeuw, line 52: Could you add a note that this is needed for FreeBSD?

src/sys/unix/selector/kqueue.rs
asomers: Except that FreeBSD 11 also still uses `intptr_t`.  And currently there's no way for a crate to select which libc ABI it wants to use.  So by removing that type entirely we'll just rely on type inference.

src/sys/unix/selector/kqueue.rs
asomers: Sure.  Except, looking at it again, I realize that I should've just deleted the fflags and data field initializers.

src/sys/unix/selector/kqueue.rs
Thomasdezeeuw: But you're using `as i64` in multiple places below, so it doesn't use type interference.

asomers: Yeah.  I changed the argument type of `kevent_register` to use i64 everywhere, instead of being platform-dependent.  So now there's only one place that needs type inference: line 284 (and line 61, before my second commit).  Those are the only places that actually use the libc type.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @asomers.

Issue #1561, opened 2022-04-09
Remove v2.7.x workarounds from Linux code


Thomasdezeeuw:
If https://github.com/rust-lang/rust/pull/95026 goes through.

PR #1575, opened 2022-05-20
feat: fallback to epoll_create when epoll_create1 not implement

honsunrise:
Fallback to epoll_create when epoll_create1 not implement.

### DIFF
--- a/src/sys/unix/selector/epoll.rs
+++ b/src/sys/unix/selector/epoll.rs
@@ -31,13 +31,26 @@ impl Selector {
         #[cfg(not(target_os = "android"))]
         let flag = libc::EPOLL_CLOEXEC;
 
-        syscall!(epoll_create1(flag)).map(|ep| Selector {
-            #[cfg(debug_assertions)]
-            id: NEXT_ID.fetch_add(1, Ordering::Relaxed),
-            ep,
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
@honsunrise why do you need this change?

honsunrise:
because some linux kernel not implement epoll_create1.
so tokio will panic when i run tokio on the system.

Thomasdezeeuw:
> because some linux kernel not implement epoll_create1. so tokio will panic when i run tokio on the system.

What Linux version are you using? Because `epoll_create1` was added in v2.6.27, after https://github.com/rust-lang/rust/pull/95026 Rust will only support 3.2+.

honsunrise:
v2.6.18

I did not use *-gnu toolchains, compile with the musl toolchains, it works on v2.6.18.
However, this PR is incomplete and the eventfd also needs to fallback.

Thomasdezeeuw:
Kernel v2.6.18 is not supported. The minimum we supported was RHEL 6, kernel 2.6.32-754, but that is no longer maintained by RedHat. So, I think we'll want to increase our minimum support kernel to 3.2 based on https://github.com/tokio-rs/mio/issues/1561 and https://github.com/rust-lang/rust/pull/95026.

PR #1576, opened 2022-05-24
Wasm32-wasi target support

rjzak:
I'm working on getting Tokio working on Rust, and some changes were needed for Mio.

Companion to https://github.com/tokio-rs/tokio/pull/4716

Signed-off-by: Richard Zak <richard@profian.com>

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -635,7 +635,6 @@ impl Registry {
     ///
     /// Event sources registered with this `Registry` will be registered with
     /// the original `Registry` and `Poll` instance.
-    #[cfg(not(target_os = "wasi"))]
     pub fn try_clone(&self) -> io::Result<Registry> {
         self.selector
             .try_clone()
--- a/src/sys/wasi/mod.rs
+++ b/src/sys/wasi/mod.rs
@@ -110,6 +110,13 @@ impl Selector {
         }
     }
 
+    pub(crate) fn try_clone(&self) -> io::Result<Selector> {
+        Ok(Selector {
+            id: self.id,
+            subscriptions: self.subscriptions.clone(),
+        })
+    }
+
     #[cfg(feature = "net")]
     pub(crate) fn register(
         &self,
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,6 +72,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 targets = [
   "aarch64-apple-ios",
   "aarch64-linux-android",
+  "wasm32-wasi",
   "x86_64-apple-darwin",
   "x86_64-pc-windows-msvc",
   "x86_64-unknown-dragonfly",
--- a/src/sys/shell/selector.rs
+++ b/src/sys/shell/selector.rs
@@ -11,7 +11,6 @@ pub type Events = Vec<Event>;
 pub struct Selector {}
 
 impl Selector {
-    #[cfg(not(target_os = "wasi"))]
     pub fn try_clone(&self) -> io::Result<Selector> {
         os_required!();
     }


src/poll.rs
```diff
@@ -225,7 +226,7 @@ use std::{fmt, io};
 /// [`SourceFd`]: unix/struct.SourceFd.html
 /// [`Poll::poll`]: struct.Poll.html#method.poll
 pub struct Poll {
-    registry: Registry,
+    registry: Arc<Registry>,
```
carllerche, line 229: You will probably want to move this `Arc` into the wasi implementation of `Selector`.

src/poll.rs
```diff
@@ -270,13 +271,21 @@ impl Poll {
...
-                registry: Registry { selector },
+                registry: Arc::new(Registry { selector }),
             })
         }
     }
 
     /// Create a separate `Registry` which can be used to register
     /// `event::Source`s.
+    #[cfg(target_os = "wasi")]
+    pub fn registry(&self) -> Arc<Registry> {
```
Thomasdezeeuw, line 282: We're not going to accept a different public API based on the compile target.

Thomasdezeeuw (changes requested):
@rjzak we're not going to accept this as is. `Poll` needs to have mutable access on Windows as it can only handle a single poll call at a time. Furthermore we don't want to force everyone to use an `Arc`.

Could you please explain why you need this change? "Wasm improvements" are not really clear.

rjzak:
@Thomasdezeeuw I certainly wouldn't want this PR accepted as-is, as the test application here https://github.com/rjzak/tokio-echo-test compiles to Wasm but doesn't work. I thought that by sharing the code I have so far, I could receive some feedback and pointers.

Thomasdezeeuw:
Ping @rjzak.

rjzak:
@Thomasdezeeuw I'm still working on it!

src/poll.rs
rjzak: What would that look like? I just need a `Registry` object, not a reference, and not an object that gets "moved".

rjzak:
@Thomasdezeeuw I have Axum & Tokio working on WebAssembly.
Still baffled at the moment with that to do with `Arc<Registry>`. This is where it's used in Tokio: https://github.com/rjzak/tokio/blob/wasi_wip/tokio/src/io/driver/mod.rs#L145. I could use some assistance.

Thomasdezeeuw:
> @Thomasdezeeuw I have Axum & Tokio working on WebAssembly. Still baffled at the moment with that to do with `Arc<Registry>`. This is where it's used in Tokio: https://github.com/rjzak/tokio/blob/wasi_wip/tokio/src/io/driver/mod.rs#L145. I could use some assistance.

Maybe you can put the [`Selector`](https://github.com/tokio-rs/mio/blob/c6f39d6ddba12b32793453dfe65a74104445f68d/src/sys/wasi/mod.rs#L43-L48) in an `Arc`?

rjzak:
@Thomasdezeeuw @carllerche I moved the `Arc<>` to `Registry::selector`, and all seems to be okay, except for the tests on Windows. On Windows, the `Poll::poll()` function is failing with an error stating that a mutable reference is needed for `Arc<Selector>`, however, the function `Selector::select()` doesn't take a mutable reference to itself, so this *shouldn't* be a problem, unless I'm misunderstanding the error.

Line in question: https://github.com/tokio-rs/mio/blob/6312041389bc37343d1e81151a29eb8ac68e87ea/src/poll.rs#L378
Error:

    error[E0596]: cannot borrow data in an `Arc` as mutable
       --> src\poll.rs:378:9
        |
    378 |         self.registry.selector.select(events.sys(), timeout)
        |         ^^^^^^^^^^^^^^^^^^^^^^ cannot borrow as mutable
        |
        = help: trait `DerefMut` is required to modify through a dereference, but it is not implemented for `std::sync::Arc<sys::windows::selector::Selector>`

Otherwise, formatting issues and the previously API difference between platforms has been resolved.

src/sys/wasi/mod.rs
```diff
@@ -110,6 +110,13 @@ impl Selector {
         }
     }
 
+    pub fn try_clone(&self) -> io::Result<Selector> {
```
Thomasdezeeuw, line 113: Nit: `pub(crate)`.

src/poll.rs
```diff
@@ -275,8 +275,7 @@ impl Poll {
         }
     }
 
-    /// Create a separate `Registry` which can be used to register
-    /// `event::Source`s.
+    /// Get access to the `Registry`, which can be used to register `event::Source`s.
```
Thomasdezeeuw, line 278: Nit: please revert.

Thomasdezeeuw (commented):
Two small nits, LGTM otherwise.

src/poll.rs
rjzak: Reverted

src/sys/wasi/mod.rs
rjzak: Changed

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @rjzak.

rjzak:
Thank you!

PR #1590, opened 2022-07-02
Use epoll_create() as a fallback for epoll_create1()

### LINKED ISSUE #1473: error: undefined reference to 'epoll_create1'
I saw this bug after upgrade to tokio 1.0,
but it is also possible to introduce with mio but itself.

For (current master) 58e589fddeac03453ffe60718c64e2a582c12275 :

```
mio $ cargo test --all-features --target=armv7-linux-androideabi
io/src/sys/unix/selector/epoll.rs:34: error: undefined reference to 'epoll_create1'
          clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

I use such `.cargo/config`:
```
[target.armv7-linux-androideabi]
linker = "armv7a-linux-androideabi16-clang++"
```

before that I used  mio 0.6.23 and all compiles for android just fine.

link2xt:
Fixes #1473

### DIFF
--- a/src/sys/unix/selector/epoll.rs
+++ b/src/sys/unix/selector/epoll.rs
@@ -31,7 +31,13 @@ impl Selector {
         #[cfg(not(target_os = "android"))]
         let flag = libc::EPOLL_CLOEXEC;
 
-        syscall!(syscall(libc::SYS_epoll_create1, flag))
+        // Illumos has epoll_create1() function, but no epoll syscalls.
+        #[cfg(target_os = "illumos")]
+        let ep = syscall!(epoll_create1(flag))?;
+
+        // Try epoll_create1 syscall with an epoll_create fallback.
+        #[cfg(not(target_os = "illumos"))]
+        let ep = syscall!(syscall(libc::SYS_epoll_create1, flag))
             .map(|fd| fd as libc::c_int)
             .or_else(|e| {
                 match e.raw_os_error() {
@@ -44,14 +50,15 @@ impl Selector {
                     }
                     _ => Err(e),
                 }
-            })
-            .map(|ep| Selector {
-                #[cfg(debug_assertions)]
-                id: NEXT_ID.fetch_add(1, Ordering::Relaxed),
-                ep,
-                #[cfg(debug_assertions)]
-                has_waker: AtomicBool::new(false),
-            })
+            })?;
+
+        Ok(Selector {
+            #[cfg(debug_assertions)]
+            id: NEXT_ID.fetch_add(1, Ordering::Relaxed),
+            ep,
+            #[cfg(debug_assertions)]
+            has_waker: AtomicBool::new(false),
+        })
     }
 
     pub fn try_clone(&self) -> io::Result<Selector> {
--- a/src/sys/unix/selector/epoll.rs
+++ b/src/sys/unix/selector/epoll.rs
@@ -31,12 +31,12 @@ impl Selector {
         #[cfg(not(target_os = "android"))]
         let flag = libc::EPOLL_CLOEXEC;
 
-        // Illumos has epoll_create1() function, but no epoll syscalls.
-        #[cfg(target_os = "illumos")]
+        // Illumos and Redox have epoll_create1() function, but no epoll syscalls.
+        #[cfg(any(target_os = "illumos", target_os = "redox"))]
         let ep = syscall!(epoll_create1(flag))?;
 
         // Try epoll_create1 syscall with an epoll_create fallback.
-        #[cfg(not(target_os = "illumos"))]
+        #[cfg(not(any(target_os = "illumos", target_os = "redox")))]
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw (changes requested):
@link2xt as I've said it can't impact the performance of existing code, which this does. This will not be accepted as is.

Thomasdezeeuw:
> @Thomasdezeeuw How does this impact the performance? Is epoll_create such a frequent operation that not calling fcntl saves any cycles? The crate doesn't have any benchmark, how do you measure it?

I'm not willing to revert to `epoll_create` on all platforms, I've stated this before.

Thomasdezeeuw:
> Threre is zero difference in syscall implementation, both call `do_epoll_create`. `epoll_create1` is not faster than `epoll_create`: https://elixir.bootlin.com/linux/v5.19-rc4/source/fs/eventpoll.c#L2010
> 
> The only difference is that there is an additional `fcntl` syscall to set a flag once at the beginning of event loop.

Can I just say that this is becoming really tiring...

They reason that `epoll_create1` was introduced is the same reason we use it. We don't control the entire program, so it's possible to `fork` between `epoll_create` and the next call to `fcntl`, causing the file descriptor to be leaked. We don't want this so we use `epoll_create1`.

notgull:
Is there any particular reason we can't use what the `polling` crate does for [`epoll`](https://github.com/smol-rs/polling/blob/989c515fa2ebaafecb0df933dd6a081c3571f0ec/src/epoll.rs#L28), where it tries to use a syscall to call `epoll_create1`, and if that fails with `ENOSYS`, it falls back onto the `epoll_create` implementation? That would be the best of both worlds.

Noah-Kennedy:
I think that would be fair to do. It shouldn't degrade performance or worsen anything for system which support the newer API, and would provide support for the older systems.

Thomasdezeeuw:
That should fine with the addition of some documentation that we fall back to a racy version using `epoll_create`.

notgull:
Hmm, I don't think this works. I think you need to use the `libc::syscall` function to call `epoll_create1`. Otherwise, it'll try to link to the `epoll_create1` function in `libc`, which may not exist.

src/sys/unix/selector/epoll.rs
```diff
@@ -31,7 +31,28 @@ impl Selector {
...
         let flag = libc::EPOLL_CLOEXEC;
 
-        syscall!(epoll_create1(flag)).map(|ep| Selector {
+        // Illumos and Redox have epoll_create1() function, but no epoll syscalls.
+        #[cfg(any(target_os = "illumos", target_os = "redox"))]
+        let ep = syscall!(epoll_create1(flag))?;
+
+        // Try epoll_create1 syscall with an epoll_create fallback.
+        #[cfg(not(any(target_os = "illumos", target_os = "redox")))]
+        let ep = syscall!(syscall(libc::SYS_epoll_create1, flag))
```
Thomasdezeeuw, line 40: We're not going to use raw system calls. Please add the `epoll_create1` function the libc crate instead.

Thomasdezeeuw (changes requested):
This is going into the right direction. However we're not going to use raw system calls. Also please add the the documentation to `Poll::new` to I asked for regarding the possible data race.

src/sys/unix/selector/epoll.rs
Thomasdezeeuw: > Is there any problem with using a raw system call?

It makes maintenance harder.

But I guess we can make an exception here, could you extract the system call stuff to a new `epoll_create1` function, that for platforms that support it call `libc::epoll_create1` and for others use the raw system interface?

src/sys/unix/selector/epoll.rs
```diff
@@ -133,6 +126,43 @@ impl Drop for Selector {
...
+    let flag = libc::O_CLOEXEC;
+    #[cfg(not(target_os = "android"))]
+    let flag = libc::EPOLL_CLOEXEC;
+
+    // Illumos and Redox have epoll_create1() function, but no epoll syscalls.
+    #[cfg(any(target_os = "illumos", target_os = "redox"))]
+    let ep = syscall!(epoll_create1(flag))?;
+
+    // Try epoll_create1 syscall with an epoll_create fallback.
+    #[cfg(not(any(target_os = "illumos", target_os = "redox")))]
```
Thomasdezeeuw, line 147: Could you change this to be android only with a comment referring to https://github.com/tokio-rs/mio/issues/1473 and explaining that Android API level 16 doesn't define `epoll_create1`. I really like to minimise the use of raw system calls.

Thomasdezeeuw (commented):
Almost their I think, I'm still not super happy about using raw system calls, but if we can minimise the usage to just Android I think it would be acceptable.

Thomasdezeeuw (approved):

Thomasdezeeuw:
I'm going to merge this as and will do a little cleanup and more docs into another pr.

PR #1593, opened 2022-07-17
Update MSRV to 1.61 and edition to 2021

### LINKED ISSUE #1562: Update to 2021 edition and MSRV 1.61
MSRV to support https://github.com/tokio-rs/mio/issues/1527, and 2021 because we can at that point.

TmLev:
For edition, I ran `cargo fix --edition` and updated `edition` in `Cargo.toml`.

Resolves #1562.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-edition = "2018"
+edition = "2021"
 name = "mio"
 # When releasing to crates.io:
 # - Update CHANGELOG.md.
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
We're not going to this for v0.8 yet as that would be a breaking change.

TmLev:
So how can one contribute towards v1.0 and not v0.8?

Thomasdezeeuw:
We haven't started the v1 branch yet, we support 6 month olds compilers so we'll have to wait until 1.61 is ~6 months old before we'll create v1 I think.

PR #1597, opened 2022-07-19
Add GitHub Actions

Thomasdezeeuw:


Thomasdezeeuw:
> What's the intent? To replace azure-pipelines with GH actions? We could use Cirrus instead, because all 4 OSes can run on it.

I'm fine with either, I did this in response to https://github.com/tokio-rs/mio/issues/1444#issuecomment-1189212858. Azure pipelines isn't used as much I think (based on personal experience alone).

Noah-Kennedy:
I'd prefer GitHub actions for now, as it's much easier to maintain and deal with.

Noah-Kennedy:
@Thomasdezeeuw I'll much about with this.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,105 @@
+name: CI
+on:
+  push:
+    branches: [ master ]
```
Noah-Kennedy, line 4: @Thomasdezeeuw add quotations around `master`

Noah-Kennedy:
@Thomasdezeeuw you might want to do this from a branch off mainline mio rather than a fork, I don't know if GH just let's you run newly-introduced status checks from a fork like this

Thomasdezeeuw:
> @Thomasdezeeuw you might want to do this from a branch off mainline mio rather than a fork, I don't know if GH just let's you run newly-introduced status checks from a fork like this

@Noah-Kennedy that worked, closing in favour of https://github.com/tokio-rs/mio/pull/1598.

PR #1598, opened 2022-07-19
Add GitHub Actions

Thomasdezeeuw:


Thomasdezeeuw:
@Noah-Kennedy what do you think about the checking all the targets in separate steps (like for Azure) vs in one step (like the `CheckTargets` step I've added).

Noah-Kennedy:
I like the CheckTargets step.

Noah-Kennedy:
@Thomasdezeeuw we will need to have someone with perms set the required status checks in GitHub to be the actions versions, not the azure versions.

Noah-Kennedy:
Also, you should add a job like tokio does that acts as the only needed status check, which simply depends on all other jobs to pass: https://github.com/tokio-rs/tokio/blob/master/.github/workflows/ci.yml#L32

Thomasdezeeuw:
> @Thomasdezeeuw we will need to have someone with perms set the required status checks in GitHub to be the actions versions, not the azure versions.

I got those permission.

> Also, you should add a job like tokio does that acts as the only needed status check, which simply depends on all other jobs to pass: https://github.com/tokio-rs/tokio/blob/master/.github/workflows/ci.yml#L32

You mean only make that single job required?

Noah-Kennedy:
> I got those permission.

excellent

> You mean only make that single job required?

Yes, this way we don't need you to manually add or remove status checks every time they change, we can just handle this through the normal PR process.

Thomasdezeeuw:
I'm removing the deploy docs action: https://github.com/tokio-rs/mio/blob/4532c424dc9f8f28c6ff1f21e7ea2f911f8dd98b/ci/azure-deploy-docs.yml, I don't think it's used anymore as we moved to https://docs.rs/mio.

Noah-Kennedy (approved):
LGTM!

PR #1599, opened 2022-07-24
fix minor typo

joleeee:


Thomasdezeeuw:
It's not a typo it stands for *if and only if*.

Issue #1472, opened 2021-03-10
Mio build broken on Haiku, no kpoll/kqueue


kallisti5:
```
   Compiling parking_lot v0.9.0
   Compiling hyper v0.12.35
   Compiling aho-corasick v0.7.15
   Compiling addr2line v0.14.0
   Compiling idna v0.1.5
   Compiling idna v0.2.0
   Compiling mime_guess v2.0.3
   Compiling futures-cpupool v0.1.8
   Compiling bytes v0.4.12
   Compiling mio v0.6.23
error[E0432]: unresolved imports `self::unix::Events`, `self::unix::Selector`
 --> /boot/home/.cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.6.23/src/sys/mod.rs:5:5
  |
5 |     Events,
  |     ^^^^^^ no `Events` in `sys::unix`
6 |     Io,
7 |     Selector,
  |     ^^^^^^^^ no `Selector` in `sys::unix`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0432`.
error: could not compile `mio`
```

Thomasdezeeuw:
We don't really support Haiku. Is there are support for `epoll` or `kqueue`?

kallisti5:
Nope. No support for either. We follow a pretty strict posix model

kallisti5:
*actually* it looks like mio 0.8.0 just works.  The error above is on 0.6.23

```
/Data/Code/mio> cargo build
   Compiling libc v0.2.88
   Compiling log v0.4.11
   Compiling cfg-if v0.1.10
   Compiling mio v0.8.0 (/Data/Code/mio)
    Finished dev [unoptimized + debuginfo] target(s) in 2.63s
/Data/Code/mio> cargo test
   Compiling env_logger v0.6.2
   Compiling rand v0.4.6
   Compiling mio v0.8.0 (/Data/Code/mio)
    Finished test [unoptimized + debuginfo] target(s) in 5.95s
     Running target/debug/deps/mio-753c3d872fa80196

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running target/debug/deps/aio-ddb6b9e09505d866

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running target/debug/deps/close_on_drop-6b57ddb8ed3d603e

[...]

PR #1602, opened 2022-08-09
Poll selector

Janrupf:
This PR adds support for using the Unix `poll` API as a backend.

### Why this is required
There are some operating systems out there which only support `poll` (or other API's). The main challenge here was to track the file descriptors, as the `poll` syscall requires all file descriptors at call site. This is solved by keeping a mutex around the state of the selector and thus tracking everything added/removed from it.

The main motivation for this PR is to enable support for the esp-idf and it's growing eco system, eventually allowing tokio to fully run on these devices. As the esp-idf only supports `poll`, mio (and by this extent the full feature set of tokio) is unable to run on it. Additionally this benefits OS such as Haiku (see https://github.com/tokio-rs/mio/issues/1472), which also only provide a `poll` based API.

### Does this work?
The short answer is: Yes, it should.

The long answer:
the `polling` crate already uses this, and in fact, the implementation here is largely a copy-and-paste of the one provided by `polling`. There are some notable differences whatsoever:
- mio additionally needs cloning support, this is achieved by keeping everything in an `Arc`
- mio (and tokio) used to assume edge triggering, but `poll` is level triggered (**may be a breaking change!**)

### What did change
In addition to a new a new polling backend, a feature called `force-old-poll` was added in order to allow further testing of the `poll` implementation (this feature is required to force use of the `poll` API, whenever the OS provides `kqueue` or `epoll`). 
It also now really is necessary to deregister event sources, else stale events may be fired (the general rules seems to always have been do deregister event sources, tokio's `PollEvented` for example does so as well as all mio examples). 

The mio `Waker` also gained a new method called `did_wake`, which needs to be called after a task has been woken up in order to reset the event source (and not wake the task again when polling). This requires a very small patch in tokio which I have ready.

### Effect on downstream crates
Probably none, but this is still a breaking change (see the notes about `Waker` and the edge vs. level triggered situation).

Since the `poll` implementation will virtually not be used anywhere other than new systems, no existing code should break on existing systems. Still, the API did change and does have new requirements.
I expect most downstream crates to properly deregister event sources before destruction, but this now is absolutely required to not run into weird race conditions. Additionally, every crate using mio `Waker`s now needs to call `did_wake` after being woken by the Waker.

### Notes
This needs a lot of testing and feedback from the tokio and mio maintainers. Also thanks to @Kestrer, who implemented large portions of this code for the `polling` crate.


Thomasdezeeuw:
Before starting to work on large things like this, please open an issue first explain why you need this.

Janrupf:
This started out as an experiment to see if it even was possible - turns out, it is and works pretty well. I thought I'd PR that for a review - that being said, if you still want me to open an issue I can do so (though I guess that would mostly be a copy'n paste of the PR description)

src/poll.rs
```diff
@@ -1,10 +1,26 @@
 use crate::{event, sys, Events, Interest, Token};
 use log::trace;
-#[cfg(unix)]
-use std::os::unix::io::{AsRawFd, RawFd};
 use std::time::Duration;
 use std::{fmt, io};
 
+#[cfg(all(
```
notgull, line 6: I wouldn't repeat this `cfg` block so often; I'd follow along with what the rest of `mio` does and create a macro that encapsulates the `cfg`.

src/sys/unix/selector/poll.rs
```diff
@@ -0,0 +1,518 @@
...
+        #[cfg(debug_assertions)]
+        pub fn id(&self) -> usize {
+            self.state.id
+        }
+    }
+}
+
+/// Interface to poll.
+#[derive(Debug)]
+struct SelectorState {
```
notgull, line 83: I might be remembering this wrong, but didn't the `polling` implementation also have an optional `timerfd` object to support sub-millisecond timeouts?

notgull (commented):
Overall, I'm a huge fan of this PR, since it will go a long way towards supporting several currently unsupported systems, including the most important system: the Nintendo 3DS.

I have to wonder: should we also support the `select()` libc function, for the systems that don't comply with POSIX?

Also, I'd at least have some note in there about how the implementation is partially derived from `polling`.

src/sys/unix/selector/poll.rs
Janrupf: I've copy & pasted most of this straight from `polling`, which doesn't have a `timerfd`. That being said, adding a `timerfd` is certainly possible. I however would like to not introduce it in this PR as to not further increase complexity.

For the time being you can technically add your own `timerfd` just fine to the registry and supply a `None` timeout.

Janrupf:
> I have to wonder: should we also support the `select()` libc function, for the systems that don't comply with POSIX?

I thought about this as well, my reasoning for not using `select` directly was its limitation of 1024 fd's. A `Selector` using `select` would look mostly the same as this one.

Fun fact: the platform I'm specifically targetting with this PR implements `poll` using `select` and just starts screaming when you pass in too many file descriptors (see https://github.com/espressif/esp-idf/blob/master/components/newlib/poll.c#L13). For a proper Unix system using `poll` is still better than `select`.

So if you need this _right now_, you can also just implement `poll` this way :smile:

Cargo.toml
```diff
@@ -42,6 +42,10 @@ os-ext = [
 # Enables `mio::net` module containing networking primitives.
 net = []
 
+# Forces the use of the poll system call instead of epoll on systems
+# where both are available
+force-old-poll = ["os-poll"]
```
Thomasdezeeuw, line 47: We're not going to add a feature for this or support `poll` on any OS that has something better (e.g. `epoll` or `kqueue`). Mio is complex enough as-is it maintain with the matrix OSs architectures we're not getting selectors in the mix as well.

examples/test_example.rs
```diff
@@ -0,0 +1,39 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
Thomasdezeeuw, line 1: What does this example showcase?

src/sys/unix/pipe.rs
```diff
@@ -109,22 +109,21 @@ use crate::{event, Interest, Registry, Token};
...
 /// # fn main() -> io::Result<()> {
 /// // Same setup as in the example above.
 /// let mut poll = Poll::new()?;
 /// let mut events = Events::with_capacity(8);
 ///
-/// let (mut sender, mut receiver) = pipe::new()?;
+/// let (sender, mut receiver) = pipe::new()?;
 ///
 /// poll.registry().register(&mut receiver, PIPE_RECV, Interest::READABLE)?;
-/// poll.registry().register(&mut sender, PIPE_SEND, Interest::WRITABLE)?;
```
Thomasdezeeuw, line 127: Why did you do this?

Thomasdezeeuw (commented):
Honesty I'm not really a fan of the implementation. It just doesn't fit Mio's model very well.

Also just mentioned a couple of time you copy and pasted this from the `polling` crate? If so, we can't accept that without a copyright notice. So, we'll likely simply not accept the code.

examples/test_example.rs
Janrupf: This is only temporary, will be removed once this PR is ready. I use this for easier debugging

Cargo.toml
Janrupf: This will also be removed once this PR is ready, I currently use this to develop the `poll` selector on Linux because development on Haiku and the espidf is quite hard. I still run the tests for Haiku, but this allows me to iron out bugs on a system which is easy to develop for.

TL;DR: Will be removed when PR is ready

Janrupf:
> Also just mentioned a couple of time you copy and pasted this from the `polling` crate? If so, we can't accept that without a copyright notice. So, we'll likely simply not accept the code.

Yes. I haven't yet added a copyright notice, because I'm not exactly sure how it is supposed to look. Both libraries use the MIT license (`polling` allows you to choose either Apache or MIT), but neither have a copyright header in any of their files. Do you want a proper copyright notice with a fully blown header or rather a statement, that the code has been taken from `polling`?

Thomasdezeeuw:
> Yes. I haven't yet added a copyright notice, because I'm not exactly sure how it is supposed to look. Both libraries use the MIT license (`polling` allows you to choose either Apache or MIT), but neither have a copyright header in any of their files. 

That's because all files have the same copyright, so it's not needed to add one to each file.

> Do you want a proper copyright notice with a fully blown header or rather a statement, that the code has been taken from `polling`?

No, I'm afraid it likely means this code will not be accepted.

Kestrer:
Well, as the sole contributor to `polling`’s `poll.rs` implementation I do give you permission to add this code to Mio if you want.

Thomasdezeeuw:
> Well, as the sole contributor to `polling`’s `poll.rs` implementation I do give you permission to add this code to Mio if you want.

That solves that then. 👍

Janrupf:
An update on this PR: I have implemented tracking fd's in userspace and hopefully removed all breaking API changes. The `poll` selector now uses the same technique as the Win32 implementation for edge based triggering. This in return also means, that it behaves the same way.

I found one case where mio wasn't following its own docs, according to which one needs to ensure `WouldBlock` is returned before further readiness events can be received (epoll and kqueue don't have that problem, but the Win32 implementation and poll do actually suffer from this limitation). 

I think this PR in general is ready for another code review pass - I'm still figuring out a CI solution (turns out mio went from crashing on haiku to crashing haiku, so my next attempt for CI will be Solaris [which also only has poll]).

The `force-old-poll` feature will be removed once I confirm this to be working on an OS which has only `poll`.

src/sys/unix/pipe.rs
Janrupf: Resolved. Has been reverted as edge triggering works now.

examples/test_example.rs
Janrupf: Example has been removed.

Thomasdezeeuw:
> An update on this PR: I have implemented tracking fd's in userspace and hopefully removed all breaking API changes. The `poll` selector now uses the same technique as the Win32 implementation for edge based triggering. This in return also means, that it behaves the same way.

👍 

> I found one case where mio wasn't following its own docs, according to which one needs to ensure `WouldBlock` is returned before further readiness events can be received (epoll and kqueue don't have that problem, but the Win32 implementation and poll do actually suffer from this limitation).

To be clear the docs say that:

> there is no guarantee that another readiness event will be delivered

Basically, without properly draining the readiness events Mio doesn't *guarantee* you'll get another event, but the underlying OS might return a readiness event any way. Note however that for Windows we control that behaviour and we're thinking about changing it (see https://github.com/tokio-rs/mio/issues/1611), basically following epoll and kqueue.

> I think this PR in general is ready for another code review pass - I'm still figuring out a CI solution (turns out mio went from crashing on haiku to crashing haiku, so my next attempt for CI will be Solaris [which also only has poll]).

Wait... we crash haiku? Any way any OS with just poll (and nothing better) will do. I'll try a do review pass this weekend.

> The `force-old-poll` feature will be removed once I confirm this to be working on an OS which has only `poll`.

👍

src/sys/unix/uds/mod.rs
```diff
@@ -77,25 +78,23 @@ cfg_os_poll! {
...
         // there is an error, the file descriptors are closed.
-        #[cfg(any(target_os = "ios", target_os = "macos"))]
+        #[cfg(any(target_os = "ios", target_os = "macos", target_os = "haiku"))]
         {
-            syscall!(fcntl(fds[0], libc::F_SETFL, libc::O_NONBLOCK))?;
-            syscall!(fcntl(fds[0], libc::F_SETFD, libc::FD_CLOEXEC))?;
-            syscall!(fcntl(fds[1], libc::F_SETFL, libc::O_NONBLOCK))?;
-            syscall!(fcntl(fds[1], libc::F_SETFD, libc::FD_CLOEXEC))?;
+            syscall!(fcntl(fds[0], libc::F_SETFD, libc::FD_CLOEXEC | libc::O_NONBLOCK))?;
+            syscall!(fcntl(fds[1], libc::F_SETFL, libc::FD_CLOEXEC | libc::O_NONBLOCK))?;
```
Thomasdezeeuw, line 97: This uses `F_SETFD` and `F_SETFL` (`FD` vs `FL`), two different commands. So this doesn't work.

src/sys/unix/uds/listener.rs
```diff
@@ -77,11 +79,7 @@ pub(crate) fn accept(listener: &net::UnixListener) -> io::Result<(UnixStream, So
         // Ensure the socket is closed if either of the `fcntl` calls
         // error below.
         let s = unsafe { net::UnixStream::from_raw_fd(socket) };
-        syscall!(fcntl(socket, libc::F_SETFD, libc::FD_CLOEXEC))?;
-
-        // See https://github.com/tokio-rs/mio/issues/1450
-        #[cfg(all(target_arch = "x86", target_os = "android"))]
-        syscall!(fcntl(socket, libc::F_SETFL, libc::O_NONBLOCK))?;
+        syscall!(fcntl(socket, libc::F_SETFD, libc::FD_CLOEXEC | libc::O_NONBLOCK))?;
```
Thomasdezeeuw, line 82: This uses `F_SETFD` and `F_SETFL` (`FD` vs `FL`), two different commands. So this doesn't work. See https://man7.org/linux/man-pages/man2/fcntl.2.html.

src/poll.rs
```diff
@@ -1,10 +1,26 @@
 use crate::{event, sys, Events, Interest, Token};
 use log::trace;
-#[cfg(unix)]
-use std::os::unix::io::{AsRawFd, RawFd};
 use std::time::Duration;
 use std::{fmt, io};
```
Thomasdezeeuw, line 5: Nit.
```suggestion
```

src/sys/unix/mod.rs
```diff
@@ -15,7 +15,7 @@ macro_rules! syscall {
 
 cfg_os_poll! {
     mod selector;
-    pub(crate) use self::selector::{event, Event, Events, Selector};
+    pub(crate) use self::selector::{event, Event, Events, Selector, IoSourceState};
```
Thomasdezeeuw, line 18: I think you need to put this inside of `cfg_io_source!`.

src/sys/unix/selector/io_source/stateless.rs
```diff
@@ -0,0 +1,47 @@
+use crate::{Interest, Registry, Token};
+use std::io;
+use std::os::unix::io::RawFd;
+
+pub(crate) struct IoSourceState;
+
+impl IoSourceState {
+    pub fn new() -> IoSourceState {
```
Thomasdezeeuw, line 8: Nit: also for the others.
```suggestion
    pub(crate) fn new() -> IoSourceState {
```

src/sys/unix/selector/io_source/edge_triggered.rs
```diff
@@ -0,0 +1,104 @@
...
+                .map(|()| {
+                    state.token = token;
+                    state.interests = interests;
+                }),
+            None => Err(io::ErrorKind::NotFound.into()),
+        }
+    }
+
+    pub fn deregister(&mut self, registry: &Registry, fd: RawFd) -> io::Result<()> {
+        if let Some(mut state) = self.inner.take() {
```
Thomasdezeeuw, line 96: Why are you `take`ing the state? Why not leave it in place so that we don't have to allocate when we register again? We can use the `is_registered` boolean to return `AlreadyExists`. Then we also drop the Option.

src/sys/unix/selector/io_source/edge_triggered.rs
```diff
@@ -0,0 +1,104 @@
+use crate::sys::Selector;
+use crate::{Interest, Registry, Token};
+use std::io;
+use std::os::unix::io::RawFd;
+
+struct InternalState {
+    selector: Selector,
+    token: Token,
+    interests: Interest,
```
Thomasdezeeuw, line 9: I think this can be `Option<Interest>`, where `None` means not registered and `Some` means registered, allows us to drop the `is_registered` boolean.

src/sys/unix/selector/io_source/edge_triggered.rs
```diff
@@ -0,0 +1,104 @@
...
+impl Drop for InternalState {
+    fn drop(&mut self) {
+        if self.is_registered {
+            let _ = self.selector.deregister(self.fd);
+        }
+    }
+}
+
+pub(crate) struct IoSourceState {
+    inner: Option<Box<InternalState>>,
```
Thomasdezeeuw, line 23: We can drop this `Option` if we change deregister below, see comment.

src/sys/unix/selector/io_source/edge_triggered.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+impl IoSourceState {
+    pub fn new() -> IoSourceState {
```
Thomasdezeeuw, line 27: Nit: same for below.

```suggestion
    pub(crate) fn new() -> IoSourceState {
```

src/sys/unix/selector/io_source/edge_triggered.rs
Thomasdezeeuw: For `RawFd` we can use -1, with a note saying it can be invalid if `interests` is `None`.

src/sys/unix/waker.rs
```diff
@@ -16,17 +17,20 @@ mod eventfd {
     #[derive(Debug)]
     pub struct Waker {
         fd: File,
+        selector: Selector,
```
Thomasdezeeuw, line 20: These changes are needed for Linux or Android, so please remove.

src/sys/unix/waker.rs
```diff
@@ -121,19 +145,46 @@ mod pipe {
     pub struct Waker {
         sender: File,
         receiver: File,
+        selector: Selector,
```
Thomasdezeeuw, line 148: Let's limit this to haiku, the other OSs don't need it.

src/sys/unix/waker.rs
```diff
@@ -148,26 +199,38 @@ mod pipe {
...
                     self.wake()
                 }
                 Err(ref err) if err.kind() == io::ErrorKind::Interrupted => self.wake(),
                 Err(err) => Err(err),
             }
         }
 
-        /// Empty the pipe's buffer, only need to call this if `wake` fails.
+        fn reset(&self) {
+            let _ = Self::reset_by_fd(self.receiver.as_raw_fd());
```
Thomasdezeeuw, line 211: Nit: don't use `Self`.
```suggestion
            let _ = Waker::reset_by_fd(self.receiver.as_raw_fd());
```

src/sys/unix/waker.rs
```diff
@@ -148,26 +199,38 @@ mod pipe {
...
+                    self.reset();
                     self.wake()
                 }
                 Err(ref err) if err.kind() == io::ErrorKind::Interrupted => self.wake(),
                 Err(err) => Err(err),
             }
         }
 
-        /// Empty the pipe's buffer, only need to call this if `wake` fails.
+        fn reset(&self) {
```
Thomasdezeeuw, line 210: Let's keep the original name, it's accurate, `reset`ting is not as the fd is not reset is any way.
```suggestion
        fn empty(&self) {
```

src/sys/unix/waker.rs
```diff
@@ -148,26 +199,38 @@ mod pipe {
...
+            let _ = Self::reset_by_fd(self.receiver.as_raw_fd());
+        }
+
+        /// Empty the pipe's buffer.
         /// This ignores any errors.
-        fn empty(&self) {
+        pub(crate) fn reset_by_fd(fd: RawFd) -> io::Result<()> {
```
Thomasdezeeuw, line 216: This should be unsafe as the call must ensure the `fd` is valid.

src/sys/unix/waker.rs
```diff
@@ -148,26 +199,38 @@ mod pipe {
...
+            let mut file = ManuallyDrop::new(unsafe { File::from_raw_fd(fd) });
+
             let mut buf = [0; 4096];
             loop {
-                match (&self.receiver).read(&mut buf) {
+                match file.read(&mut buf) {
                     Ok(n) if n > 0 => continue,
-                    _ => return,
+                    _ => return Ok(()),
```
Thomasdezeeuw, line 223: Why both returning an error, if we're always returning `Ok`?

src/sys/unix/waker.rs
```diff
@@ -148,26 +199,38 @@ mod pipe {
...
                 }
             }
         }
     }
+
+    impl Drop for Waker {
```
Thomasdezeeuw, line 229: Let's limit to Haiku only.

tests/tcp_stream.rs
```diff
@@ -639,14 +639,13 @@ fn tcp_reset_close_event() {
...
 
-    // Make sure we quiesce. `expect_no_events` seems to flake sometimes on mac/freebsd.
-    loop {
-        poll.poll(&mut events, Some(Duration::from_millis(100)))
-            .expect("poll failed");
-        if events.iter().count() == 0 {
-            break;
-        }
-    }
+    // Something weird is going on here. Linux keeps emitting HUP, so even though everything is
```
Thomasdezeeuw, line 642: I'm assuming this is using the `poll` interface? If so, please remove along with the `force-old-poll` feature.

Thomasdezeeuw (changes requested):
You're using `fcntl` wrong in various places. It has two different commands (we use):
* `F_SETFL`: for setting the flags status, e.g. `FD_CLOEXEC`, and
* `F_SETFD`: for setting the file descriptor flags, e.g. `O_NONBLOCK`.

waddlesplash:
> I'm still figuring out a CI solution (turns out mio went from crashing on haiku to crashing haiku

Is this with a nightly build or the beta? If you have a reliable reproducer then please open a ticket with Haiku and we will investigate (and probably fix within days.)

src/sys/unix/pipe.rs
```diff
@@ -138,6 +139,15 @@ use crate::{event, Interest, Registry, Token};
 ///             println!("Sender dropped!");
 ///             return Ok(());
 ///         },
+///         PIPE_RECV => {
+///             // Some platforms only signal a read readines event
```
gco, line 143: typo: s/readines/readiness/

src/sys/unix/pipe.rs
```diff
@@ -172,9 +182,7 @@ pub fn new() -> io::Result<(Sender, Receiver)> {
         }
 
         for fd in &fds {
-            if libc::fcntl(*fd, libc::F_SETFL, libc::O_NONBLOCK) != 0
-                || libc::fcntl(*fd, libc::F_SETFD, libc::FD_CLOEXEC) != 0
-            {
+            if libc::fcntl(*fd, libc::F_SETFL, libc::O_NONBLOCK | libc::FD_CLOEXEC) != 0 {
```
gco, line 185: As pointed out by @Thomasdezeeuw elsewhere: "This uses `F_SETFD` and `F_SETFL` (`FD` vs `FL`), two different commands. So this doesn't work."

pheki:
@Janrupf are you still working on this?

Janrupf:
Not really - the implementation itself was (and still is) working, but I have no way to proof it using a CI pipeline or something along those lines (at least not with the constraints previously set by the maintainers). There are now a bunch of merge conflicts as you can see.

I'm using an implementation of this in a production environment where it has been running on a few hundred embedded devices without any issues, but that's about as good as it gets. If you have interest on picking this up or getting it merged, I'm willing to help.

notgull:
We test it in `polling` by using an environment variable to conditionally compile it for Linux, see [here](https://github.com/smol-rs/polling/blob/master/.github/workflows/ci.yml).

Thomasdezeeuw:
> We test it in `polling` by using an environment variable to conditionally compile it for Linux, see [here](https://github.com/smol-rs/polling/blob/master/.github/workflows/ci.yml).

The problem with that is that I'm not willing to maintain this implementation on any OS that supports something "better", e.g. epoll on Linux (better might be debatable here, but let's not). Having 1 implementation per OS is enough already.

pheki:
On the following month I'll try implementing CI for one of the poll-only OSes then.

pheki:
Hey, I've managed to compile and run the tests using qemu, but it takes a lot longer to run. Even if we optimize and cache in a lot of different ways, like building our own haiku image with rust pre-installed and running `cargo vendor` before copying the library into the VM, I think it would still take at least 15 minutes to run , while also being a lot more complex. See [1](https://github.com/pheki/mio/actions/runs/5359417746/jobs/9722984818) and [2](https://github.com/pheki/mio/actions/runs/5361792963/jobs/9728108059). The first one was hanging and I cancelled, just like it happened in my local (and fast) qemu, so I suppose it's an implementation issue.

From the issue #1682, it looks like you're willing to reconsider unsing a cfg exclusively to run tests on linux? We could have a `mio_unsupported_force_poll` cfg which would switch the implementation to poll to be used just to test the poll implementation, like [here](https://github.com/smol-rs/polling/blob/7a1fd31944d8fb55acf2c209144196d259bafb89/.github/workflows/ci.yml#L56). This way we could also make this PR a lot simpler, as it would only implement poll, not full Haiku support. If there is interest on merging Haiku support in the future it can be done by someone else.

Thomasdezeeuw:
@pheki (see first part of your comment below)

> From the issue #1682, it looks like you're willing to reconsider unsing a cfg exclusively to run tests on linux? We could have a `mio_unsupported_force_poll` cfg which would switch the implementation to poll to be used just to test the poll implementation, like [here](https://github.com/smol-rs/polling/blob/7a1fd31944d8fb55acf2c209144196d259bafb89/.github/workflows/ci.yml#L56). This way we could also make this PR a lot simpler, as it would only implement poll, not full Haiku support. If there is interest on merging Haiku support in the future it can be done by someone else.

I've documented my thoughts on it in https://github.com/tokio-rs/mio/pull/1684, but yes let's continue with a `mio_unspported_force_poll_poll` flag, which we can use in the CI. But note that this means poll will **not be officially supported** for anything but (I think) Haiku.

> Hey, I've managed to compile and run the tests using qemu, but it takes a lot longer to run. Even if we optimize and cache in a lot of different ways, like building our own haiku image with rust pre-installed and running `cargo vendor` before copying the library into the VM, I think it would still take at least 15 minutes to run , while also being a lot more complex. See [1](https://github.com/pheki/mio/actions/runs/5359417746/jobs/9722984818) and [2](https://github.com/pheki/mio/actions/runs/5361792963/jobs/9728108059). The first one was hanging and I cancelled, just like it happened in my local (and fast) qemu, so I suppose it's an implementation issue.

I still like to see some kind of Haiku support in the CI if we want to officially support it, even if it's just a `cargo check` and then we test the actual implementation on Linux with the cfg flag (not ideal, but it will have to do). We can only run it for merges into the master branch so that the prs aren't blocked on slow CI.

Thank you for looking to this.

jasta:
I'm interested in throwing my hat in the ring here to try to help get this over the line (and with luck I can get some other esp-rs folks onboard!).  Before I get too far though I want to double check that I understand the issues, which AFAICT comes down to two missing requirements:

1. Automated tests covering a specific platform that is intended to (not just can) use the poll() code path introduced in this PR
2. A reliable story around ownership/maintenance for one or more of the platform in (1).

If this understanding is correct, I think the major issue we're facing right now is that the kinds of platforms that want the poll support aren't well set up for efficient, low cost automation which of course makes (1) difficult.  Separately, (2) is complicated by the fact that these platforms don't tend to be stable over long periods of time and many dozens of such platforms have come and gone while Linux persists.  This doesn't mean they don't deserve support, it just means that the overall community of people is likely going to be a tiny fraction of Linux.  Combine that with the fact that the Rust community is a small fraction of developers as a whole and in practice you can never expect to find a large group of committed maintainers.  An unfortunate situation for both tokio maintainers and the communities that would benefit from their great work.  I think we can find good solutions and move forward though, hopefully others agree :)

Let's take a closer look at (1) first.  As an example, esp-idf requires a lot of extra build support and tools (e.g. https://github.com/esp-rs/embuild) as well as a somewhat unsupported qemu fork (https://esp-rs.github.io/book/tooling/simulating/qemu.html) to run the code.  I'm confident that we _could_ get this off the ground to run in github's CI, but I doubt very much that we'd ever be able to get the performance good enough to run within a couple of minutes.  Realistically I expect this will be 10+ minutes as @pheki has discovered with Haiku as well.  Most of the embedded community chooses instead to write platform agnostic code that can be tested on the host, as is the case with this PR being able to run on Linux, then move the rest to run on real hardware that requires significant overhead and cost to maintain in a CI infrastructure.  Even companies with deep pockets do it this way just simply because it can become a real money pit to try to get $6 devices to ever be as robust and reliable as Linux rackmount servers (trust me, I tried to do this at Meta for quite some time hehe).

While solutions do exist, I do think we need to maybe reexamine how flexible the requirements really are.  For example, we are already in the business of pragmatic support for platforms like Android and iOS despite being unable to test those reliably in GitHub CI as well for much the same reasons.  I'd propose that we're relying on the fact that a good number of supported targets are "close enough" to FreeBSD, Linux, Windows, and OS X that testing on those should be sufficient to cover the rest.  The real issue with (1) then is that there are no platforms currently that actually would use the poll code path and so the tests for this code feel arbitrary and contrived to use Linux just because it technically supports poll.  I agree here (and actually I agree with everything @Thomasdezeeuw has been saying), but consider for a moment that what we're really testing here is that Linux, a POSIX-compliant platform, should work the same as any other POSIX-compliant platform, in the same way we're saying that epoll on any obscure configuration of Linux should work the same as our desktop or rackmount server Linux configurations.  Definitely not literally true, but true enough to say we've done our due diligence by relying on reliable, stable standards and doing automated testing against a well known and correct implementation.  This actually makes the case stronger for using Linux with poll instead of, say, esp-idf's newlib poll which might be buggy and not a good proxy for, say, Haiku.

For (2), I personally think we can find strong partners in the embedded community, and in particular esp-rs which is a large and growing community with committed maintainers across a number of components.  I initially assumed that the gap to get tokio support for esp32 was so large that it wouldn't be practical even technically but honestly @Janrupf 's work will likely inspire others to step up and commit to getting this work landed and maintain it going forward, myself included.

So I guess all that novel was really to say, what exactly do I need to do to get this landed?  I see the PR is missing some of the cfg changes you proposed, and a few things could use tidying per your review.  Is there anything else that's insufficient?  What more do we need to convince tokio maintainers that we've got the right long term support?  I know that for example Rust async has a big push from Espressif as evidenced by Scott Mabin's blog posts: https://mabez.dev/blog/posts/esp-rust-30-06-2023/.  Probably not a stretch to imagine I could convince the esp-rs maintainers to commit to having tokio support on the roadmap...

Janrupf:
Hi @jasta, I just want to add as a note that I'm in fact using this with tokio on the ESP32. I created a support PR on the Tokio repository some time ago, which should contain the changes neccessary to also get tokio to run this. Once mio works as expected it's only a matter of compiling Tokio with the right features.

And thanks a bunch for the long comment, I really appreciate it!

Thomasdezeeuw:
> I'm interested in throwing my hat in the ring here to try to help get this over the line (and with luck I can get some other esp-rs folks onboard!). Before I get too far though I want to double check that I understand the issues, which AFAICT comes down to two missing requirements:
> 
>     1. Automated tests covering a specific platform that is intended to (not just can) use the poll() code path introduced in this PR

Ideally the tests don't need to be modified (except for maybe adding a `cfg(target_os = "haiku")` or similar). Because that means the behaviour for the `poll(2)` implementation is the same as the on the other platforms. Since Haiku isn't supported by rustup (yet) it's ok to add a `mio_unsupported_force_poll_poll` flag to force the use of the `poll(2)` implementation so we can test on Linux on the CI. See https://github.com/tokio-rs/mio/pull/1684 for docs and https://github.com/tokio-rs/mio/pull/1685 for an example.

>     2. A reliable story around ownership/maintenance for one or more of the platform in (1).

This is pretty important, but mostly for Haiku or other OS(s) that I don't use. Basically if something breaks on Haiku I want to ask somewhere with at least the OS running to help. With the `mio_unsupported_force_poll_poll` I should be able to reproduce most issues on Linux, but not the one specific to Haiku (or other OS that uses the `poll(2)` implementation).

> If this understanding is correct, I think the major issue we're facing right now is that the kinds of platforms that want the poll support aren't well set up for efficient, low cost automation which of course makes (1) difficult. Separately, (2) is complicated by the fact that these platforms don't tend to be stable over long periods of time and many dozens of such platforms have come and gone while Linux persists. This doesn't mean they don't deserve support, it just means that the overall community of people is likely going to be a tiny fraction of Linux. Combine that with the fact that the Rust community is a small fraction of developers as a whole and in practice you can never expect to find a large group of committed maintainers. An unfortunate situation for both tokio maintainers and the communities that would benefit from their great work. I think we can find good solutions and move forward though, hopefully others agree :)

This is spot on, we all have limited time so for me (other another Mio maintainer) to spend hours trying to setup a Haiku (or other) machine just to fix an issue is quite a tough ask. That is why (2) is important, for people with Haiku already running it can be a 2 minute check to at least reproduce an issue.

> Let's take a closer look at (1) first. As an example, esp-idf requires a lot of extra build support and tools (e.g. https://github.com/esp-rs/embuild) as well as a somewhat unsupported qemu fork (https://esp-rs.github.io/book/tooling/simulating/qemu.html) to run the code. I'm confident that we _could_ get this off the ground to run in github's CI, but I doubt very much that we'd ever be able to get the performance good enough to run within a couple of minutes. Realistically I expect this will be 10+ minutes as @pheki has discovered with Haiku as well. Most of the embedded community chooses instead to write platform agnostic code that can be tested on the host, as is the case with this PR being able to run on Linux, then move the rest to run on real hardware that requires significant overhead and cost to maintain in a CI infrastructure. Even companies with deep pockets do it this way just simply because it can become a real money pit to try to get $6 devices to ever be as robust and reliable as Linux rackmount servers (trust me, I tried to do this at Meta for quite some time hehe).
>
> While solutions do exist, I do think we need to maybe reexamine how flexible the requirements really are. For example, we are already in the business of pragmatic support for platforms like Android and iOS despite being unable to test those reliably in GitHub CI as well for much the same reasons. I'd propose that we're relying on the fact that a good number of supported targets are "close enough" to FreeBSD, Linux, Windows, and OS X that testing on those should be sufficient to cover the rest. The real issue with (1) then is that there are no platforms currently that actually would use the poll code path and so the tests for this code feel arbitrary and contrived to use Linux just because it technically supports poll. I agree here (and actually I agree with everything @Thomasdezeeuw has been saying), but consider for a moment that what we're really testing here is that Linux, a POSIX-compliant platform, should work the same as any other POSIX-compliant platform, in the same way we're saying that epoll on any obscure configuration of Linux should work the same as our desktop or rackmount server Linux configurations. Definitely not literally true, but true enough to say we've done our due diligence by relying on reliable, stable standards and doing automated testing against a well known and correct implementation. This actually makes the case stronger for using Linux with poll instead of, say, esp-idf's newlib poll which might be buggy and not a good proxy for, say, Haiku.

Agreed, testing on Linux with `mio_unsupported_force_poll_poll` would be good enough. Of course this assumes that the Linux `poll(2)` implementation works exactly like the implementation found in Haikue/esp-idf/etc., which is a big if, but it's (I think) the best we can realistically do right now.

> For (2), I personally think we can find strong partners in the embedded community, and in particular esp-rs which is a large and growing community with committed maintainers across a number of components. I initially assumed that the gap to get tokio support for esp32 was so large that it wouldn't be practical even technically but honestly @Janrupf 's work will likely inspire others to step up and commit to getting this work landed and maintain it going forward, myself included.

👍 

> So I guess all that novel was really to say, what exactly do I need to do to get this landed? I see the PR is missing some of the cfg changes you proposed, and a few things could use tidying per your review. The CI checks are currently failing so obviously we gotta fix that. Is there anything else that's insufficient? What more do we need to convince tokio maintainers that we've got the right long term support? I know that for example Rust async has a big push from Espressif as evidenced by Scott Mabin's blog posts: https://mabez.dev/blog/posts/esp-rust-30-06-2023/. Probably not a stretch to imagine I could convince the esp-rs maintainers to commit to having tokio support on the roadmap...

From my end:
* Add the `mio_unsupported_force_poll_poll` flag as discussed above.
* Use the `mio_unsupported_force_poll_poll` to the `poll(2)` implementation on Linux. Maybe FreeBSD as well so we have at least two different `poll(2)` implementations we test, so that we don't depend on Linux's specific behaviours/bugs in `poll(2)`?
* No changes to the tests, ideally at least, so that we know Mio stays cross platform. Exception here for things like add `cfg(target_os = "haiku")`.
* And of course a rebase + passing CI.

I think that about it?

(note that the list might not seem to long, but I only recently agreed to the `mio_unsupported_force_poll_poll` flag, this was mainly stuck on getting Haiku or `poll(2)` implementations specific support in the CI)

jasta:
Ok I'm going to work on the technical aspects of this first:

1. Use `mio_unsupported_force_poll_poll`
2. Rebase + get CI checks passing generally
3. Add a CI configuration for the tests in GitHub Actions and make sure all that passes

Once I'm ready I'll submit a new PR and then start socializing the maintenance issues a bit with the esp-rs folks.  I'm feeling really good that we can get this over the line soon-ish!

jasta:
I noticed a couple of bugs in the poll implementation while doing this work that result directly from inconsistencies to how the Windows implementation works.  For example, when poll yields POLLHUP the edge-triggered IoSourceState doesn't work properly and then the caller just repeatedly gets spammed POLLHUP because the fd is not deregistered internally.  The Windows implementation solves this by using a Mutex and shared state between IoSourceState and the Selector, so I think I'll just further copy that behaviour.

This will fix a failing test I'm seeing locally in tcp_stream.rs:tcp_reset_close_event

jasta:
Only one issue left to fix and I could use your help @Janrupf.  Specifically I bumped into: https://github.com/tokio-rs/tokio/issues/5866 which it looks like you should've hit as well unless you were only testing recv_from and send_to APIs or anything else not covered by PollEvented.

Thomasdezeeuw:
> Only one issue left to fix and I could use your help @Janrupf. Specifically I bumped into: [tokio-rs/tokio#5866](https://github.com/tokio-rs/tokio/issues/5866) which it looks like you should've hit as well unless you were only testing recv_from and send_to APIs or anything else not covered by PollEvented.

Also see https://github.com/tokio-rs/mio/issues/1611. We could also decide that reading less then `buf.len()` bytes should trigger reregistering, matching what epoll and kqueue do already (with edge triggers as used by Mio).

waddlesplash:
FWIW, a (very basic, FDs/PROC only) implementation of `kqueue` is being worked on and should make its way into Haiku nightly builds before too long, hopefully removing the need for `poll` there.

Above, however, it was noted that some versions of this change caused Haiku to crash. Is that reproducible? Do you have a reproducer, or even just a screenshot of the crash?

Thomasdezeeuw:
> FWIW, a (very basic, FDs/PROC only) implementation of `kqueue` is being worked on and should make its way into Haiku nightly builds before too long, hopefully removing the need for `poll` there.

Nice. We still need the `poll(2)` based implementation for ESP-IDF, see https://github.com/tokio-rs/mio/pull/1687 and https://github.com/tokio-rs/mio/pull/1692.

> Above, however, it was noted that some versions of this change caused Haiku to crash. Is that reproducible? Do you have a reproducer, or even just a screenshot of the crash?

psumbera:
> > FWIW, a (very basic, FDs/PROC only) implementation of `kqueue` is being worked on and should make its way into Haiku nightly builds before too long, hopefully removing the need for `poll` there.
> 
> Nice. We still need the `poll(2)` based implementation for ESP-IDF, see #1687 and #1692.

I would also hope that Solaris would be able to use poll(2) too. See #1152.

Janrupf:
Superseded by #1687, thanks to everyone who continued this!

Thomasdezeeuw:
Thanks @Janrupf for the implementation!

PR #1603, opened 2022-08-10
Document the current Poll::poll timeout behaviour

### LINKED ISSUE #1600: poll()’s return on timeout not documented
The [documentation for the poll() method](https://docs.rs/mio/0.8.4/mio/struct.Poll.html#method.poll) doesn’t say what it returns when the timeout elapses. (It returns `Ok(())`.)

I naively expected it to return something like `Err(std::io::Error::from(std::io::ErrorKind::TimedOut))`. I see that it behaves like the `select()` and `poll()` syscalls though, so maybe I shouldn’t have been surprised.

It would be good to document this, including whether or not you can rely on the fact of the events collection being empty to indicate that it timed out.

Thomasdezeeuw:
/cc @danielparks

Closes #1600

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -349,6 +349,10 @@ impl Poll {
     /// of Mio would automatically retry the poll call if it was interrupted
     /// (if `EINTR` was returned).
     ///
+    /// Currently if the `timeout` elapses without any readiness events
+    /// triggering this will return `Ok(())`. However we're not guaranteeing
+    /// this behaviour as this depends on the OS.
+    ///
     /// # Examples
     ///
     /// A basic example -- establishing a `TcpStream` connection.


Noah-Kennedy (approved):

Thomasdezeeuw:
What do you think @danielparks?

danielparks:
Looks good. Thanks!

jkarneges:
Hmm. My interpretation of the updated documentation is that mio's poll is free to return something other than `Ok(())` if the call times out. Is that intentional? This would mean all apps would need to be very defensive and swallow all errors. Else apps risk crashing if a future version of mio decides to return `Err` on timeout.

Based on the discussion in #1600, I suspect the documentation was meant to say something like: "timeouts always return `Ok(())`, and a return value of `Ok(())` with zero events likely indicates a timeout but is not guaranteed to be a timeout". This is quite a different claim, and would not require apps to swallow all errors.

Thomasdezeeuw:
> Hmm. My interpretation of the updated documentation is that mio's poll is free to return something other than `Ok(())` if the call times out. Is that intentional? This would mean all apps would need to be very defensive and swallow all errors. Else apps risk crashing if a future version of mio decides to return `Err` on timeout.
> 
> Based on the discussion in #1600, I suspect the documentation was meant to say something like: "timeouts always return `Ok(())`, and a return value of `Ok(())` with zero events likely indicates a timeout but is not guaranteed to be a timeout". This is quite a different claim, and would not require apps to swallow all errors.

This documentation intends to document that we follow whatever the OS does, we have no special handling for timeouts. At the moment that means it return `Ok(())`. However a new implementation might return an error on timeout, we'll have consider it when it comes up.

Issue #1527, opened 2021-11-07
Use std::os::unix::net::SocketAddr instead of mio::net::SocketAddr


Thomasdezeeuw:
It's the same type. It's only implemented by us because we can't create `std::os::unix::net::SocketAddr`.

Thomasdezeeuw:
We can do this in 1.61.

Noah-Kennedy:
@Thomasdezeeuw is it confirmed to be stabilizing then?

Thomasdezeeuw:
@Noah-Kennedy it's stabilised in https://github.com/rust-lang/rust/pull/94356.

RealXuChe:
From #1605, I want to notice you that there may be a regression in the replacement: As you can see [here](https://github.com/rust-lang/rust/blob/3fdd578d72a24d4efc2fe2ad18eec3b6ba72271e/library/std/src/os/unix/net/addr.rs#L32), `std::os::unix::net::SocketAddr` doesn't support binding to a path with  `\0` in it, which means it doesn't support abstract Unix socket.

Of course, there's no reason for `std::os::unix::net::SocketAddr` to support abstract Unix socket, since abstract Unix socket is a Linux feature, and not in the Unix spec.

But mio has already supported abstract Unix socket, though it's a little buggy, like there's no conditional compilation to make it only work on Linux. 

So this replacement will possibly break the use of abstract Unix socket, I think we need to be careful on this.

(Sorry for my bad English)

Thomasdezeeuw:
I think the support is still unstable, but it's should come at some point: https://doc.rust-lang.org/std/os/unix/net/struct.SocketAddr.html#method.from_abstract_namespace. Maybe we can create a pr to stabilise it.

gustav3d:
That's a good idea indeed.

PR #1605, opened 2022-08-11
Fix path validation for abstract unix socket.

RealXuChe:
The validation of abstract unix socket is wrong, and I fixed it.
Look at the original validation [code](https://github.com/tokio-rs/mio/blob/d4cbf76f60d6472524dddd74746a861c79bf42c5/src/sys/unix/uds/mod.rs#L43):
```rust 
match (bytes.get(0), bytes.len().cmp(&sockaddr.sun_path.len())) {
    // Abstract paths don't need a null terminator
    (Some(&0), Ordering::Greater) => {
        return Err(io::Error::new(
            io::ErrorKind::InvalidInput,
            "path must be no longer than libc::sockaddr_un.sun_path",
        ));
    }
    (_, Ordering::Greater) | (_, Ordering::Equal) => {
        return Err(io::Error::new(
            io::ErrorKind::InvalidInput,
            "path must be shorter than libc::sockaddr_un.sun_path",
        ));
    }
    _ => {}
}
```
Abstract paths don't need a null terminator, so it's valid when it's no longer than the `libc::sockaddr_un.sun_path`. However, when the length of an abstract path equals to `sun_path`, `(Some(&0), Ordering::Equal)` will match on the second arm and will return an error.

I added the branch `(Some(&0), _) => {}` to accept such case.

### DIFF
--- a/src/sys/unix/uds/mod.rs
+++ b/src/sys/unix/uds/mod.rs
@@ -48,6 +48,7 @@ cfg_os_poll! {
                     "path must be no longer than libc::sockaddr_un.sun_path",
                 ));
             }
+            (Some(&0), _) => {}
             (_, Ordering::Greater) | (_, Ordering::Equal) => {
                 return Err(io::Error::new(
                     io::ErrorKind::InvalidInput,


Thomasdezeeuw (changes requested):
Can you add a test for this?

Also this is going to be replace pretty soon: https://github.com/tokio-rs/mio/issues/1527.

Thomasdezeeuw:
Closing in favour of #1749, which removes this code.

Issue #1609, opened 2022-08-16
Support UnixStream on Windows


sullivan-sean:
Windows 10 added support for unix sockets in a 2018 general Windows 10 release. Building this functionality into Mio would allow support in Tokio and other downstream libs.

This seems to be of interest in several downstream libs, example issues here:
- [Tokio](https://github.com/tokio-rs/tokio/issues/2201)
- [Tonic](https://github.com/hyperium/tonic/issues/238)

Mio's current AFD approach on Windows actually seems to be compatible with AF_UNIX sockets  (noted e.g. in this `wepoll` [issue](https://github.com/piscisaureus/wepoll/issues/3)).

sullivan-sean:
With some tinkering I've put together a first pass [here](https://github.com/sullivan-sean/mio) that passes all existing `unix_stream` and `unix_listener` tests in `mio`. 

There are a few tweaks that need to be made, e.g. right now the selector must be [re-registered between events](
https://github.com/sullivan-sean/mio/blob/9015ca296f8ff8eb39a5e25fdcaf87b973e0b973/tests/unix_stream.rs#L223-L230) which I think quirk may be related to how `user_evt` flags are not fully reset on the selector after an event:
https://github.com/tokio-rs/mio/blob/c6b5f13adf67483d927b1766faf984c1871e8abc/src/sys/windows/selector.rs#L230-L233

Another question is where the windows analog of `std::os::unix::net` should live, given that UDS is [not in `std` yet](https://github.com/rust-lang/rust/issues/56533)

For now I have included a module directly within the repo modeled after https://crates.io/crates/uds_windows using `windows-sys` instead of `winapi`. But I hope there is a better way to do this

dojyorin:
Hi😀

I want to implement an IPC mechanism in an application using Windows version of Deno.

Currently, Deno UDS is a trial support only for the Linux version, but I hope that it will eventually support the Windows version.

And I'm very interested in implementing this feature upstream of Deno.

Thanks for development😄

Noah-Kennedy:
There is a PR right now open for this. It hasn't been updated in several weeks though. Not sure if stalled or just going slowly.

keithmattix:
Looks like the PRs in question have stalled....is anyone looking at this recently?

Noah-Kennedy:
@keithmattix if you want to have a go at this please feel free to do so!

juliusl:
[...]

PR #1610, opened 2022-08-16
Adds support for UnixStream and UnixListener on Windows

sullivan-sean:
Uses `mio`s existing AFD approach for TCP/UDP to support `AF_UNIX` sockets on Windows. See https://github.com/tokio-rs/mio/issues/1609 for context


tests/unix_stream.rs
```diff
@@ -444,6 +462,14 @@ where
     expect_read!(stream.read(&mut buf), DATA1);
 
     assert!(stream.take_error().unwrap().is_none());
+    // TODO: have to re-register here to reset user_events
```
sullivan-sean, line 465: Is this expected behavior on windows given this note in the selector? If so, I can make this re-registration in the test conditional on the platform

https://github.com/tokio-rs/mio/blob/c6b5f13adf67483d927b1766faf984c1871e8abc/src/sys/windows/selector.rs#L230-L233

sullivan-sean (commented):
Some of the doc tests are currently failing because examples for `UnixStream` are unix only. Is there an easy way to write cross-platform examples for docs to avoid this?

Cargo.toml
```diff
@@ -48,13 +48,17 @@ log = "0.4.8"
 [target.'cfg(unix)'.dependencies]
 libc = "0.2.121"
 
+[target.'cfg(windows)'.dependencies]
+tempfile = "3"
```
Thomasdezeeuw, line 52: Can we do without this?

tests/unix_stream.rs
Thomasdezeeuw: The tests should work the same across platforms, so this needs to change in the implementation.

Thomasdezeeuw:
@sullivan-sean also see https://github.com/rust-lang/socket2/pull/249.

sullivan-sean:
I removed the dependency on `tempfile` and fixed the tests.

I think the current implementation for Windows actually confuses the notion of edge triggered events with oneshot notifications i.e. `EPOLLONESHOT` vs `EPOLLET` in `epoll` and does not successfully re-register for events after successful io. 

> @sullivan-sean also see [rust-lang/socket2#249](https://github.com/rust-lang/socket2/pull/249).

Would you prefer I introduce a dependency on this library for the socket operations? Looks like it would be pretty easy to use this instead of much of the socket code I've added

tests/unix_listener.rs
```diff
@@ -187,6 +190,57 @@ where
     handle.join().unwrap();
 }
 
+#[test]
+fn unix_listener_multiple_accepts() {
```
sullivan-sean, line 194: Current implementation on Windows fails this test. Subsequent `accept` calls after the first to the same listener hang because `accept` is not considered an I/O operation and won't send event when ready.

I've fixed this test by wrapping `accept` in `do_io` in `net/uds/listener.rs`

Thomasdezeeuw:
> I removed the dependency on `tempfile` and fixed the tests.

👍 

> I think the current implementation for Windows actually confuses the notion of edge triggered events with oneshot notifications i.e. `EPOLLET` vs `EPOLLONESHOT` in `epoll` and so it was not correctly re-registering for events after successful io.

That's an issue we should fix separately from adding Unix sockets for Windows.  Can you open an issue for this?

> > @sullivan-sean also see [rust-lang/socket2#249](https://github.com/rust-lang/socket2/pull/249).
> 
> Would you prefer I introduce a dependency on this library for the socket operations? Looks like it would be pretty easy to use this instead of much of the socket code I've added

No. We decided not to add any dependencies and focus on a mostly std based API, i.e. all the net types in there but non-blocking.

Cargo.toml
```diff
@@ -48,13 +48,17 @@ log = "0.4.8"
 [target.'cfg(unix)'.dependencies]
 libc = "0.2.121"
 
+[target.'cfg(windows)'.dependencies]
+rand = "0.8"
```
Darksonn, line 52: It seems like this is used only in tests?
```suggestion
[target.'cfg(windows)'.dev-dependencies]
rand = "0.8"
```

sullivan-sean: It's used to emulate `socketpair` functionality on Windows because there's unfortunately no equivalent system call. The best we can do is create a temporary file, connect through it, and then delete it. I create the file by generating a random temp file name (https://github.com/sullivan-sean/mio/blob/488254d2b509b4b1152ee118aabf8533616b5d47/src/sys/windows/uds/stdnet/net.rs#L24-L28)

sullivan-sean:
> That's an issue we should fix separately from adding Unix sockets for Windows. Can you open an issue for this?

Issue created here https://github.com/tokio-rs/mio/issues/1612 - I can make a separate PR from https://github.com/sullivan-sean/mio/commit/c725b035a06738ee9cd3e9b017cd28c8cac60e31 which contains just the fix for that, if this is indeed an issue

tests/unix_stream.rs
sullivan-sean: Given https://github.com/tokio-rs/mio/issues/1611 I think the only reason these tests are failing is that there is not a non-blocking read call after the initial read, as there is in the corresponding `tcp` tests 🤦 

I think adding a non-blocking read here could fix these tests for now, rather than changing draining behavior.

sullivan-sean:
I've removed the changes in draining behavior, which requires more discussion in https://github.com/tokio-rs/mio/issues/1611 and fixed the tests instead by adding a would block assertion in compliance with the current drain behavior on windows documented [here](https://docs.rs/mio/latest/mio/struct.Poll.html#draining-readiness).

All tests are passing and a naive windows implementation in tokio (just copy pasting the unix implementation https://github.com/sullivan-sean/tokio) seems to pass all of the unix stream + listener tests there as well

Cargo.toml
notgull: I'd personally recommend using `getrandom` instead of importing the entire `rand` crate.

sullivan-sean: Since it looks like `getrandom` just abstracts across platforms and we only need this in windows we could also just directly use `BCryptGenRandom` from `windows_sys`? (this is the same approach as [in `getrandom`](https://github.com/rust-random/getrandom/blob/0b71d50b4975b492c7ed7185d5d366403966c01b/src/windows.rs#L29))

src/sys/windows/stdnet/stream.rs
```diff
@@ -373,31 +373,25 @@ impl IntoRawSocket for UnixStream {
...
 
 fn sample_ascii_string(len: usize) -> io::Result<String> {
-    const RANGE: u32 = 26 + 26 + 10;
     const GEN_ASCII_STR_CHARSET: &[u8] = b"ABCDEFGHIJKLMNOPQRSTUVWXYZ\
             abcdefghijklmnopqrstuvwxyz\
-            0123456789";
+            0123456789-_";
     let mut result = String::with_capacity(len);
+    let mut buf = [0; 4];
     for _ in 0..len {
```
Darksonn, line 381: Is the loop still necessary? Can we make a single call to fill the entire string in one go?

sullivan-sean: removed this loop! in practice now it should only be a single call but I added a new loop to avoid overflow and make sure this function is safe even for really long strings

Thomasdezeeuw (changes requested):
I'm not sure about this any more. It adds a lot of code while normally we depend on `std`, now we implement it all ourselves. This also means we need an upgrade path for when `std` *does* support and we'll want to use that code instead.

src/lib.rs
```diff
@@ -91,6 +91,8 @@ pub mod windows {
     //! Windows only extensions.
 
     pub use crate::sys::named_pipe::NamedPipe;
+
+    pub use crate::sys::windows::std;
```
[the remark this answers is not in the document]
sullivan-sean: The `std::net` types are used in `tests/unix_listener.rs` and `test/unix_stream.rs` for unix, and also accepted as an argument in `from_std` on `mio::net::{UnixStream, UnixListener}`. To remove this, I can swap to using `mio::net` for windows tests and make `from_std` `pub(crate)` on windows. Does this sound okay?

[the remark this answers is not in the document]
sullivan-sean: Hm this doesn't quite work as nicely as I'd hoped since the `std::net` types are blocking and the tests rely on this

notgull:
I think the usual strategy is to just wait for libstd to implement it and not get too hasty with a patch we'll have to revert once libstd gains support.

Is there a tracking issue for implementing it into the standard library, by the way?

sullivan-sean:
There is an open issue since 2018 https://github.com/tokio-rs/tokio/issues/2201, and progress is unclear. It also seems like support for named pipes was added to `mio` [without std support](https://github.com/tokio-rs/tokio/issues/2201#issuecomment-1028637982)? Though to be fair maybe that added less code than this...

I imagine that the `std` implementation, whenever it arrives, would be a pretty straightforward drop-in given that the internal `UnixStream` and `UnixListener` structs in this PR are built to mirror the `std::os::unix::net` equivalents almost exactly.

If you would prefer to not add this to `mio` yet it would be really nice to expose a way to patch an event source like this into `mio`/`tokio` as an end developer (i.e. given that `PollEvented` is hidden in `tokio` it is quite hard to create a compatible source if it is not supported by `mio`/`tokio` out of the box).

Thomasdezeeuw:
> There is an open issue since 2018 [rust-lang/rust#56533](https://github.com/rust-lang/rust/issues/56533), and progress seems to have stagnated - the mentioned `min_target_api_version` [RFC ](https://github.com/rust-lang/rfcs/pull/3036) is no longer being worked on.

I think that was superseded by https://github.com/rust-lang/rfcs/pull/2523.

> It also seems like support for named pipes was added to `mio` [without std support](https://github.com/tokio-rs/tokio/issues/2201#issuecomment-1028637982)? Though to be fair maybe that added less code than this...

True, but it's only 1k lines in a single file: https://github.com/tokio-rs/mio/blob/master/src/sys/windows/named_pipe.rs. This pr is larger.

> I imagine that the `std` implementation, whenever it arrives, would be a pretty straightforward drop-in given that the internal `UnixStream` and `UnixListener` structs in this PR are built to mirror the `std::os::unix::net` equivalents almost exactly.

Hopefully, that's why I'm even considering this, but it would push back our v1 or we need to add an unstable feature.

> If you would prefer to not add this to `mio` yet it would be really nice to expose a way to patch an event source like this into `mio`/`tokio` as an end developer (i.e. given that `PollEvented` is hidden in `tokio` it is quite hard to create a compatible source if it is not supported by `mio`/`tokio` out of the box).

That's basically https://github.com/tokio-rs/mio/issues/880, but TL;DR that's not really possible.

sullivan-sean:
> I think that was superseded by [rust-lang/rfcs#2523](https://github.com/rust-lang/rfcs/pull/2523).

👍 
 
> True, but it's only 1k lines in a single file: https://github.com/tokio-rs/mio/blob/master/src/sys/windows/named_pipe.rs. This pr is larger.

Still not a single file, but after removing unused functions this PR is significantly smaller. There are also a non-trivial number of changes I had to make to [pass lint checks](https://github.com/tokio-rs/mio/runs/7941108502?check_suite_focus=true) that are unrelated to the actual content of this PR, e.g. all of the changes in `src/net/tcp/stream.rs`, `src/sys/unix/pipe.rs`, `src/sys/unix/uds/mod.rs` and a few other files.


> > I imagine that the `std` implementation, whenever it arrives, would be a pretty straightforward drop-in given that the internal `UnixStream` and `UnixListener` structs in this PR are built to mirror the `std::os::unix::net` equivalents almost exactly.
> 
> Hopefully, that's why I'm even considering this, but it would push back our v1 or we need to add an unstable feature.
> 
> > If you would prefer to not add this to `mio` yet it would be really nice to expose a way to patch an event source like this into `mio`/`tokio` as an end developer (i.e. given that `PollEvented` is hidden in `tokio` it is quite hard to create a compatible source if it is not supported by `mio`/`tokio` out of the box).
> 
> That's basically #880, but TL;DR that's not really possible.

If it doesn't make sense to add this to `mio` now, pre-`std` support for UDS on Windows, what would you advise for users who would like to use `mio`/`tokio` with UDS in a cross-platform way? Maintain a personal fork? or maybe turn this into a more up-to-date https://github.com/Azure/mio-uds-windows?

src/sys/windows/mod.rs
```diff
@@ -1,3 +1,19 @@
+/// Helper macro to execute a WinSock system call that returns an `io::Result`.
+#[allow(unused_macros)]
+macro_rules! wsa_syscall {
+    ($fn: ident ( $($arg: expr),* $(,)* ), $err_value: expr) => {{
+        let res = unsafe { windows_sys::Win32::Networking::WinSock::$fn($($arg, )*) };
+        if PartialEq::eq(&res, &$err_value) {
+            Err(io::Error::from_raw_os_error(unsafe {
```
Thomasdezeeuw, line 7: Seems unlikely, but just in case `std::io` isn't imported.
```suggestion
            Err(std::io::Error::from_raw_os_error(unsafe {
```

src/sys/windows/stdnet/mod.rs
```diff
@@ -0,0 +1,25 @@
+mod addr;
```
Thomasdezeeuw, line 1: Can you add some docs to this module that it's "port" of `std::os::unix::net` that should be added to std in the future.

tests/close_on_drop.rs
```diff
@@ -58,7 +58,7 @@ impl TestHandler {
                     AfterRead => {}
                 }
 
-                let mut buf = Vec::with_capacity(1024);
+                let mut buf = vec![0; 1024];
```
Thomasdezeeuw, line 61: Why is this needed?

tests/unix_listener.rs
```diff
@@ -1,8 +1,11 @@
-#![cfg(all(unix, feature = "os-poll", feature = "net"))]
+#![cfg(all(feature = "os-poll", feature = "net", any(unix, feature = "os-ext")))]
```
Thomasdezeeuw, line 1: Why do we need `os-ext` here? Removing `unix` enough?

```suggestion
#![cfg(all(feature = "os-poll", feature = "net"))]
```

tests/unix_listener.rs
```diff
@@ -1,8 +1,11 @@
...
 
+#[cfg(windows)]
+use mio::net;
```
Thomasdezeeuw, line 4: I think we need this because we need blocking version of the UDS types right? If so, let's put that in a separate module with that clearly documented.

tests/unix_stream.rs
```diff
@@ -77,6 +82,7 @@ fn unix_stream_connect() {
     handle.join().unwrap();
 }
 
+#[cfg(unix)]
```
Thomasdezeeuw, line 85: The code has `cfg(windows)`, but here you're making it Unix only?

tests/unix_stream.rs
```diff
@@ -445,6 +457,8 @@ where
 
     assert!(stream.take_error().unwrap().is_none());
 
+    assert_would_block(stream.read(&mut buf));
```
Thomasdezeeuw, line 460: Why is this needed? Ideally the test are unchanged as the behaviour is already dependent on in existing code.

src/sys/windows/stdnet/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+mod listener;
+mod socket;
+mod stream;
+
+pub use self::addr::SocketAddr;
+pub(crate) use self::listener::UnixListener;
+pub(crate) use self::stream::UnixStream;
+
+cfg_os_poll! {
+pub(self) use self::addr::socket_addr;
```
Thomasdezeeuw, line 11: Can you indent inside of `cfg_os_poll!`? Also elsewhere.

Thomasdezeeuw (commented):
Start of the review. I also start the CI, didn't realise it didn't start.

tests/unix_stream.rs
sullivan-sean: We need to make a call that would block to reset the interests before trying to write per https://docs.rs/mio/latest/mio/struct.Poll.html#draining-readiness. I believe right now this test does not fully follow this documentation.

tests/unix_listener.rs
sullivan-sean: oh I interpreted your comment [here](https://github.com/tokio-rs/mio/pull/1610#discussion_r951790084) to mean we don't want to expose the blocking `std::os::unix::net` Windows equivalent types outside of the crate. 

If you'd like me to expose these for use in tests, where would be the best place for them to live?

tests/unix_stream.rs
sullivan-sean: Currently I was not exposing the internal blocking UDS types for windows, so there is no `from_std` function on `UnixStream` in Windows, and this test wouldn't really make sense on windows

tests/unix_listener.rs
sullivan-sean: Ah good catch, previously I was exposing the blocking Windows UDS types under `os-ext` for Windows and so I'd added this to ensure we could access the Windows `std::net` types on Windows.

sullivan-sean:
@Thomasdezeeuw any updates on this? Is this still something I should be working on and if so, should I be exposing the Windows blocking types per https://github.com/tokio-rs/mio/pull/1610#discussion_r956163129?

Thomasdezeeuw:
> @Thomasdezeeuw any updates on this? Is this still something I should be working on and if so, should I be exposing the Windows blocking types per [#1610 (comment)](https://github.com/tokio-rs/mio/pull/1610#discussion_r956163129)?

I just haven't had the time to review this properly. Perhaps I'll have some this weekend.

src/net/uds/listener.rs
```diff
@@ -24,18 +30,26 @@ impl UnixListener {
     /// standard library in the Mio equivalent. The conversion assumes nothing
     /// about the underlying listener; it is left up to the user to set it in
     /// non-blocking mode.
+    #[cfg(unix)]
```
Thomasdezeeuw, line 33: This will need a `#[cfg_attr(docsrs, doc(cfg(linux)))]`.

src/net/uds/mod.rs
```diff
@@ -1,4 +1,6 @@
+#[cfg(unix)]
 mod datagram;
+#[cfg(unix)]
```
Thomasdezeeuw, line 3: Needs `#[cfg_attr(docsrs, doc(cfg(all(unix)))]`.

src/net/uds/listener.rs
```diff
@@ -79,18 +93,21 @@ impl fmt::Debug for UnixListener {
     }
 }
 
+#[cfg(unix)]
```
Thomasdezeeuw, line 96: Needs `#[cfg_attr(docsrs, doc(cfg(unix))]`, same for Window's socket impls.

src/net/uds/stream.rs
```diff
@@ -34,12 +40,20 @@ impl UnixStream {
     /// The Unix stream here will not have `connect` called on it, so it
     /// should already be connected via some other means (be it manually, or
     /// the standard library).
+    #[cfg(unix)]
```
Thomasdezeeuw, line 43: Needs `#[cfg_attr(docsrs, doc(cfg(unix))]`.

src/net/uds/stream.rs
```diff
@@ -144,49 +226,49 @@ impl UnixStream {
 
 impl Read for UnixStream {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        self.inner.do_io(|inner| (&*inner).read(buf))
+        self.inner.do_io(|mut inner| inner.read(buf))
```
Thomasdezeeuw, line 229: Can we revert these unrelated changes.

src/net/uds/stream.rs
```diff
@@ -220,18 +302,21 @@ impl fmt::Debug for UnixStream {
     }
 }
 
+#[cfg(unix)]
```
Thomasdezeeuw, line 305: Needs `#[cfg_attr(docsrs, doc(cfg(unix))]`, same for Windows's impls below.

src/sys/unix/pipe.rs
```diff
@@ -313,29 +313,29 @@ impl event::Source for Sender {
 
 impl Write for Sender {
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
-        self.inner.do_io(|sender| (&*sender).write(buf))
+        self.inner.do_io(|mut sender| sender.write(buf))
```
Thomasdezeeuw, line 316: Can we revert these unrelated changes.

src/sys/unix/uds/mod.rs
```diff
@@ -40,7 +40,7 @@ cfg_os_poll! {
         sockaddr.sun_family = libc::AF_UNIX as libc::sa_family_t;
 
         let bytes = path.as_os_str().as_bytes();
-        match (bytes.get(0), bytes.len().cmp(&sockaddr.sun_path.len())) {
+        match (bytes.first(), bytes.len().cmp(&sockaddr.sun_path.len())) {
```
Thomasdezeeuw, line 43: Also revert these unrelated changes please.

src/sys/windows/uds/listener.rs
```diff
@@ -0,0 +1,23 @@
...
+use crate::sys::windows::stdnet as net;
+
+pub(crate) fn bind(path: &Path) -> io::Result<net::UnixListener> {
+    let listener = net::UnixListener::bind(path)?;
+    listener.set_nonblocking(true)?;
+    Ok(listener)
+}
+
+pub(crate) fn accept(listener: &net::UnixListener) -> io::Result<(UnixStream, SocketAddr)> {
+    listener.set_nonblocking(true)?;
```
Thomasdezeeuw, line 15: Why are you calling `set_nonblocking` in a call to accept?

src/sys/windows/uds/listener.rs
```diff
@@ -0,0 +1,23 @@
...
+    listener
+        .accept()
+        .map(|(stream, addr)| (UnixStream::from_std(stream), addr))
```
Thomasdezeeuw, line 18: I think the accepted stream needs to be set to non_blocking, `from_std` doesn't do that.

src/sys/windows/stdnet/addr.rs
```diff
@@ -0,0 +1,170 @@
...
+fn path_offset(addr: &sockaddr_un) -> usize {
+    // Work with an actual instance of the type since using a null pointer is UB
+    let base = addr as *const _ as usize;
+    let path = &addr.sun_path as *const _ as usize;
+    path - base
+}
+
+cfg_os_poll! {
+    use windows_sys::Win32::Networking::WinSock::AF_UNIX;
+    pub(super) fn socket_addr(path: &Path) -> io::Result<(sockaddr_un, c_int)> {
```
Thomasdezeeuw, line 19: I've reviewed way to many of this function... up to 4 now...

src/sys/windows/stdnet/addr.rs
```diff
@@ -0,0 +1,170 @@
...
+            ));
+        }
+
+        if bytes.len() >= sockaddr.sun_path.len() {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "path must be shorter than SUN_LEN",
+            ));
+        }
+        for (dst, src) in sockaddr.sun_path.iter_mut().zip(bytes.iter()) {
```
Thomasdezeeuw, line 54: Nit: can you replace this with `copy_from_slice`?

src/sys/windows/stdnet/addr.rs
```diff
@@ -0,0 +1,170 @@
...
+            *dst = *src as u8;
+        }
+
+        let offset = path_offset(&sockaddr);
+        let mut socklen = offset + bytes.len();
+
+        match bytes.get(0) {
```
Thomasdezeeuw, line 61: Nit: can you use first?

src/sys/windows/stdnet/addr.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub(crate) fn new<F>(f: F) -> io::Result<SocketAddr>
+    where
+        F: FnOnce(*mut SOCKADDR, *mut c_int) -> io::Result<c_int>,
+    {
+        SocketAddr::init(f).map(|(_, addr)| addr)
+    }
+
+    pub(crate) fn from_parts(addr: sockaddr_un, mut len: c_int) -> SocketAddr {
+        if len == 0 {
+            // When there is a datagram from unnamed unix socket
```
Thomasdezeeuw, line 123: Don't think this applies.

src/sys/windows/stdnet/listener.rs
```diff
@@ -0,0 +1,83 @@
+use std::os::windows::io::{AsRawSocket, FromRawSocket, IntoRawSocket, RawSocket};
+use std::{fmt, io, mem};
+
+use windows_sys::Win32::Networking::WinSock::SOCKET_ERROR;
+
+use super::{socket::Socket, SocketAddr};
+
+pub(crate) struct UnixListener(Socket);
+
+impl fmt::Debug for UnixListener {
```
Thomasdezeeuw, line 10: Can move this below the impls, following the same conventions as the rest of the code base.

src/sys/windows/stdnet/listener.rs
```diff
@@ -0,0 +1,83 @@
...
+        builder.field("socket", &self.0.as_raw_socket());
+        if let Ok(addr) = self.local_addr() {
+            builder.field("local", &addr);
+        }
+        builder.finish()
+    }
+}
+
+impl UnixListener {
+    pub fn local_addr(&self) -> io::Result<SocketAddr> {
```
Thomasdezeeuw, line 22: These can all be `pub(crate)`.

src/sys/windows/stdnet/listener.rs
```diff
@@ -0,0 +1,83 @@
...
+            )
+        })
+    }
+
+    pub fn take_error(&self) -> io::Result<Option<io::Error>> {
+        self.0.take_error()
+    }
+}
+
+impl AsRawSocket for UnixListener {
```
Thomasdezeeuw, line 36: Can you also move these below the impl block below.

src/sys/windows/stdnet/listener.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+cfg_os_poll! {
+    use std::path::Path;
+
+    use super::{socket_addr, UnixStream};
+
+    impl UnixListener {
+        pub fn bind<P: AsRef<Path>>(path: P) -> io::Result<UnixListener> {
```
Thomasdezeeuw, line 62: Also `pub(crate)`.

src/sys/windows/stdnet/listener.rs
```diff
@@ -0,0 +1,83 @@
...
+            let inner = Socket::new()?;
+            let (addr, len) = socket_addr(path.as_ref())?;
+
+            wsa_syscall!(
+                bind(inner.as_raw_socket() as _, &addr as *const _ as *const _, len as _),
+                SOCKET_ERROR
+            )?;
+            wsa_syscall!(listen(inner.as_raw_socket() as _, 128), SOCKET_ERROR)?;
```
Thomasdezeeuw, line 70: Please match the backlog on Unix, 1024.

src/sys/windows/stdnet/stream.rs
```diff
@@ -0,0 +1,226 @@
+use std::io::{self, IoSlice, IoSliceMut};
```
Thomasdezeeuw, line 1: Same here as in the other file with regards to:
 * ordering of `impl` blocks
 * `pub(crate)`.

src/sys/windows/stdnet/stream.rs
```diff
@@ -0,0 +1,226 @@
...
+        pub fn pair() -> io::Result<(Self, Self)> {
+            use std::sync::{Arc, RwLock};
+            use std::thread::spawn;
+
+            let file_path = temp_path(10)?;
+            let a: Arc<RwLock<Option<io::Result<UnixStream>>>> = Arc::new(RwLock::new(None));
+            let ul = UnixListener::bind(&file_path).unwrap();
+            let server = {
+                let a = a.clone();
+                spawn(move || {
```
Thomasdezeeuw, line 165: No.
Just no.

This isn't going to happen.

src/sys/windows/stdnet/stream.rs
```diff
@@ -0,0 +1,226 @@
...
+            .map(|r| {
+                // We pick from 64=2^6 characters so we can use a simple bitshift.
+                let idx = r >> (8 - 6);
+                char::from(GEN_ASCII_STR_CHARSET[idx as usize])
+            })
+            .collect();
+        Ok(result)
+    }
+
+    fn temp_path(len: usize) -> io::Result<PathBuf> {
```
Thomasdezeeuw, line 209: Don't know about this either.

Thomasdezeeuw (changes requested):
Overall this is looking ok, not good, but ok. There are some things in this pr which aren't going to happen. For example spawning threads, that's a 100% blocker of this pr. But overall I'm willing to accept the new types for Windows, even though I prefer to have them in the standard library first so we can match the API, but I release that will take a while.

Some more questions:
 * Can remove all code changes not related to UDS. If that means the CI fails we'll fix that in another pr as this one is large enough as it is.
 * I assume `UnixDatagram`s are not supported?
 * What is path forward regarding:
   * Getting these types into the standard library?
   * Using `SocketAddr` from the standard library (see https://github.com/tokio-rs/mio/issues/1527).
 * I'm too well versed in Windows' inheritance, do we need any no inheritances after calls to accept?
 * All `cfg(...)` need a `#[cfg_attr(docsrs, doc(cfg(...))]`, I pointed out some, might have missed some others.
 * We'll probably want to wrap `SocketAddr` in a type in `mio::net` to keep the docs and methods in sync. It already has different methods and trait implementation in this pr.

I didn't have time to review the tests, but ideally very little to nothing changes with them as that means that existing code doesn't have to change to support Windows UDS.

sullivan-sean:
>  - Can remove all code changes not related to UDS. If that means the CI fails we'll fix that in another pr as this one is large enough as it is.

👍 

>  - I assume `UnixDatagram`s are not supported?

Correct, the datagram socket type is not supported by Windows implementation of Unix sockets (https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/#unsupportedunavailable)

>  - What is path forward regarding:
>    - Getting these types into the standard library?
>    - Using `SocketAddr` from the standard library (see #1527).

Looks like there's some recent movement on https://github.com/rust-lang/rust/issues/56533. I'm not familiar with how/if we could go about adding this to the standard library, and if so what the route would be. But I am happy to push for this with guidance on the approach. I imagine a `SocketAddr` type for Windows could also be added to `std` when the other uds types are.

>  - I'm too well versed in Windows' inheritance, do we need any no inheritances after calls to accept?

Seems like sockets returned from `accept` have the [same properties](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-accept#remarks) as the listener socket they are called on - so if the listener has [`WSA_FLAG_NO_HANDLE_INHERIT`](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsasocketa) set this is handled automatically. I will remove the explicit inheritance calls after accept. Then if somebody creates a listener with `from_raw_socket`, the streams returned from calling `accept` on that listener will have the same flag as the original socket, which I imagine is desired behavior.

> - We'll probably want to wrap SocketAddr in a type in mio::net to keep the docs and methods in sync. It already has different methods and trait implementation in this pr.

I'll work on adding this

> I didn't have time to review the tests, but ideally very little to nothing changes with them as that means that existing code doesn't have to change to support Windows UDS.

Thanks for the very thorough review of everything else, I know it's a big PR and appreciate your patience! I've minimized test changes as much as possible while accounting for the lack of blocking socket types in `std` for Windows.

src/sys/windows/uds/listener.rs
sullivan-sean: The stream returned by `accept` inherits properties from the listener socket, including non-blocking. Since this function is crate internal and only called on non-blocking listeners, it should always return non-blocking streams

src/sys/windows/stdnet/addr.rs
sullivan-sean: sorry 😬 will aim to not make cosmetic changes to functions mid-review going forward

Thomasdezeeuw:
> Looks like there's some recent movement on [rust-lang/rust#56533](https://github.com/rust-lang/rust/issues/56533). I'm not familiar with how/if we could go about adding this to the standard library, and if so what the route would be. But I am happy to push for this with guidance on the approach. I imagine a `SocketAddr` type for Windows could also be added to `std` when the other uds types are.

You can start with a small RFC or maybe ask it on the Rust Zulip chat. Then just implement the types in `std::os::windows` and later merge them with Unix, maybe under `std::net`. But the libs team can help with that.

> > * I'm too well versed in Windows' inheritance, do we need any no inheritances after calls to accept?
> 
> Seems like sockets returned from `accept` have the [same properties](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-accept#remarks) as the listener socket they are called on - so if the listener has [`WSA_FLAG_NO_HANDLE_INHERIT`](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsasocketa) set this is handled automatically. I will remove the explicit inheritance calls after accept. Then if somebody creates a listener with `from_raw_socket`, the streams returned from calling `accept` on that listener will have the same flag as the original socket, which I imagine is desired behavior.

👍 can you add a comment saying as much.

> > I didn't have time to review the tests, but ideally very little to nothing changes with them as that means that existing code doesn't have to change to support Windows UDS.
> 
> Thanks for the very thorough review of everything else, I know it's a big PR and appreciate your patience! I've minimized test changes as much as possible while accounting for the lack of blocking socket types in `std` for Windows.

👍

Noah-Kennedy:
Where are we with this?

sullivan-sean:
I've been away from my Windows machine and so haven't had a chance to fix the failing docs test. I will add the comment about `WSA_FLAG_NO_HANDLE_INHERIT` and try to fix the doc test tomorrow.

I posted in the rust-lang Zulip about adding Window UDS to `std` and the libs-api team seemed at least open to the idea on the condition that the failure mode on older platforms without UDS support is that "call[ing] the relevant functions...give[s] an error" which is the current behavior here for WinSock calls in general.

lumattr:
How is this one doing? Are we waiting for the rust changes of does this just need a re-review?

Thomasdezeeuw:
> How is this one doing? Are we waiting for the rust changes of does this just need a re-review?

I think this is pretty close, we need to resolve some small things and need to make sure that the test behave the same on Unix and Windows (which is harder then it sounds).

sullivan-sean:
Closed in favor of https://github.com/tokio-rs/mio/pull/1667. Thanks @KolbyML for cleaning this up :)

PR #1613, opened 2022-08-25
Support bind TcpStream to source IP address and connect to remote IP with bind_connect()

AlexZhao:
Hi, Dear mio experts

can you kindly help to check do you see is it possible to add interface can support bind source IP address when trying to create TcpStream ?

For many cases it is useful if want to clear define source IP/Port to Dest IP/Port  in fully controlled manner, so please try to see is it possible to add this support? currently it only have Unix system support, if it is required by community I can add windows support later

Br
Alex

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-.cargo
+.vscode
 Cargo.lock
-target*
-libs
+target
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw:
I'm afraid this won't be added to Mio as it follows the same API as the standard library. However you can use the socket2 crate for something like this and then convert it into `mio::net::TcpStream` for usage with Mio.

PR #1615, opened 2022-09-03
Round up when converting timeout to ms for epoll_wait()

### LINKED ISSUE #1614: poll() returns immdiately with timeouts less than one millisecond
This contradicts the documentation, which says:

>Note that the timeout will be rounded up to the system clock granularity (usually 1ms), and kernel scheduling delays mean that the blocking interval may be overrun by a small amount.

I ran into this chasing down high CPU usage from [below](https://github.com/facebookincubator/below)'s replay view, which is just a TUI for viewing existing logs and does not collect samples.  Some notes from the investigation, trimmed for brevity:

`below replay --time now` consumes enormous CPU time even though the screen is completely static.

- ~4% (4200 MHz) built from upstream git

As for the bulk of the CPU hogging, Looks like culprits are,

(snip!)

3. There are some 50,000 calls to `epoll_wait()` per second.

Very suspicious profile (event=cycles):

```
- 96.97% cursive_core::cursive_run::CursiveRunner<C>::run
   - 96.78% cursive_core::cursive_run::CursiveRunner<C>::step (inlined)
      - 73.51% cursive_core::cursive_run::CursiveRunner<C>::process_events (inlined)
         - 73.23% <cursive_buffered_backend::BufferedBackend as cursive_core::backend::Backend>::poll_event
              <cursive::backends::crossterm::Backend as cursive_core::backend::Backend>::poll_event
              crossterm::event::poll
            - crossterm::event::poll_internal (inlined)
               - 73.18% crossterm::event::read::InternalEventReader::poll
                  - 65.03% <crossterm::event::source::unix::UnixInternalEventSource as crossterm::event::source::EventSource>::try_read
                     + 57.68% mio::poll::Poll::poll
                     + 4.68% crossterm::event::timeout::PollTimeout::leftover (inlined)
                     + 1.73% crossterm::event::timeout::PollTimeout::new (inlined)
                     + 0.06% mio::event::events::Events::is_empty
```

I built crossterm, modified it's example code to poll with 1ms timeout like crossterm uses it, and threw it in gdb.  Backtrace looks like:

```
#0  0x00007ffff7e8c74a in epoll_wait () from /lib64/libc.so.6
#1  0x00005555555ca615 in mio::sys::unix::selector::epoll::Selector::select (self=0x555555668c20, events=0x555555668c30, timeout=...)
    at /home/rhaley/.cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.8.4/src/sys/unix/selector/epoll.rs:68
#2  0x00005555555cdd22 in mio::poll::Poll::poll (self=0x555555668c20, events=0x555555668c30, timeout=...)
    at /home/rhaley/.cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.8.4/src/poll.rs:400
#3  0x0000555555585be6 in crossterm::event::source::unix::{impl#1}::try_read (self=0x555555668c20, timeout=...) at src/event/source/unix.rs:83
#4  0x000055555557d6cb in crossterm::event::read::InternalEventReader::poll<crossterm::event::filter::EventFilter> (
    self=0x5555556630a0 <crossterm::event::INTERNAL_EVENT_READER+8>, timeout=..., filter=0x5555556228dc) at src/event/read.rs:64
#5  0x000055555556f7c4 in crossterm::event::poll_internal<crossterm::event::filter::EventFilter> (timeout=..., filter=0x5555556228dc) at src/event.rs:233
#6  0x0000555555578958 in crossterm::event::poll (timeout=...) at src/event.rs:168
#7  0x000055555556c58e in event_poll_read::print_events () at examples/event-poll-read.rs:25
#8  0x000055555556cb3b in event_poll_read::main () at examples/event-poll-read.rs:55
```

The main problem is that mio rounds <1ms timeouts to 0, which results in `epoll_wait()` returning immediately.  But how does a =1ms timeout become a <1ms timeout?

Because crossterm doesn't pass the timeout straight to mio.  Instead, it converts it to a `crossterm::timeout::PollTimeout`, and then passes the *leftover* time to mio.  The leftover time is calculated as the original timeout less the elapsed time since the `PollTimout` was constructed. In fact, crossterm converts back and forth between `std::time::Duration` and its own `PollTimeout` by calling `leftover()` and constructing new `PollTimeout`s several times, potentially losing a few nanos on each conversion.

Compounding the problem, crossterm polls repeatedly in a loop until the timeout is fully elapsed.  So the 1ms timeout becomes a 1ms spinwait.

Patched mio to round timeouts up instead of down.  Result:

Before, with 4% CPU usage:

```
07:10:00
@[tracepoint:syscalls:sys_enter_openat, below, 2606215, 2606215]: 8
@[tracepoint:syscalls:sys_enter_ioctl, below, 2606215, 2606215]: 8
@[tracepoint:syscalls:sys_enter_close, below, 2606215, 2606215]: 8
@[tracepoint:syscalls:sys_enter_clock_nanosleep, below, 2606215, 2606215]: 32
@[tracepoint:syscalls:sys_enter_epoll_wait, below, 2606215, 2606215]: 50275
```

After, with 0.9% CPU usage:

```
07:06:30
@[tracepoint:syscalls:sys_enter_openat, below, 2571785, 2571785]: 8
@[tracepoint:syscalls:sys_enter_ioctl, below, 2571785, 2571785]: 8
@[tracepoint:syscalls:sys_enter_close, below, 2571785, 2571785]: 8
@[tracepoint:syscalls:sys_enter_clock_nanosleep, below, 2571785, 2571785]: 31
@[tracepoint:syscalls:sys_enter_epoll_wait, below, 2571785, 2571785]: 31
```

(per-second syscall counts collected with [this bpftrace script](https://github.com/yump/bpftrace-tricks/blob/master/hits_per_second.bt.fish))

yump:
Fixes #1614

I'm a total beginner to rust, so there may be a better way to do this.  I'm also unfamiliar with async I/O, so it's entirely possible that crossterm is misusing the API.  But the `epoll_wait` manpage does say that the timeout will be rounded up.

### DIFF
--- a/src/sys/unix/selector/epoll.rs
+++ b/src/sys/unix/selector/epoll.rs
@@ -88,19 +88,17 @@ impl Selector {
 
         let timeout = timeout
             .map(|to| {
-                cmp::min(
-                    MAX_SAFE_TIMEOUT,
-                    to.as_millis()
-                        // as_millis() truncates, so round up to the nearest millisecond as the
-                        // documentation says can happen.  This avoids turning submillisecond
-                        // timeouts into immediate returns unless the caller explicitly requests
-                        // that by specifying a zero timeout.
-			+ if to.subsec_nanos() % 1_000_000 == 0 {
-			    0
-			} else {
-			    1
-			},
-                ) as libc::c_int
+                let to_ms = to.as_millis();
+                // as_millis() truncates, so round up to 1 ms as the documentation says can happen.
+                // This avoids turning submillisecond timeouts into immediate returns unless the
+                // caller explicitly requests that by specifying a zero timeout.
+                let to_ms = to_ms
+                    + if to_ms == 0 && to.subsec_nanos() != 0 {
+                        1
+                    } else {
+                        0
+                    };
+                cmp::min(MAX_SAFE_TIMEOUT, to_ms) as libc::c_int
             })
             .unwrap_or(-1);
 
--- a/src/sys/unix/selector/epoll.rs
+++ b/src/sys/unix/selector/epoll.rs
@@ -87,7 +87,18 @@ impl Selector {
         const MAX_SAFE_TIMEOUT: u128 = libc::c_int::max_value() as u128;
 
         let timeout = timeout
-            .map(|to| cmp::min(to.as_millis(), MAX_SAFE_TIMEOUT) as libc::c_int)
+            // as_millis() truncates, so round up to the nearest millisecond as the documentation
+            // says can happen.  This avoids turning submillisecond timeouts into immediate returns
+            // unless the caller explicitly requests that by specifying a zero timeout.
+            .map(|to| {
+                to.as_millis()
+                    + if to.subsec_nanos() % 1_000_000 == 0 {
+                        0
+                    } else {
+                        1
+                    }
+            })
+            .map(|to| cmp::min(to, MAX_SAFE_TIMEOUT) as libc::c_int)
             .unwrap_or(-1);
 
         events.clear();
--- a/src/sys/unix/selector/epoll.rs
+++ b/src/sys/unix/selector/epoll.rs
@@ -87,18 +87,21 @@ impl Selector {
         const MAX_SAFE_TIMEOUT: u128 = libc::c_int::max_value() as u128;
 
         let timeout = timeout
-            // as_millis() truncates, so round up to the nearest millisecond as the documentation
-            // says can happen.  This avoids turning submillisecond timeouts into immediate returns
-            // unless the caller explicitly requests that by specifying a zero timeout.
             .map(|to| {
-                to.as_millis()
-                    + if to.subsec_nanos() % 1_000_000 == 0 {
-                        0
-                    } else {
-                        1
-                    }
+                cmp::min(
+                    MAX_SAFE_TIMEOUT,
+                    to.as_millis()
+                        // as_millis() truncates, so round up to the nearest millisecond as the
+                        // documentation says can happen.  This avoids turning submillisecond
+                        // timeouts into immediate returns unless the caller explicitly requests
+                        // that by specifying a zero timeout.
+			+ if to.subsec_nanos() % 1_000_000 == 0 {
+			    0
+			} else {
+			    1
+			},
+                ) as libc::c_int
             })
-            .map(|to| cmp::min(to, MAX_SAFE_TIMEOUT) as libc::c_int)
             .unwrap_or(-1);
 
         events.clear();


src/sys/unix/selector/epoll.rs
```diff
@@ -87,7 +87,18 @@ impl Selector {
...
+            // unless the caller explicitly requests that by specifying a zero timeout.
+            .map(|to| {
+                to.as_millis()
+                    + if to.subsec_nanos() % 1_000_000 == 0 {
+                        0
+                    } else {
+                        1
+                    }
+            })
+            .map(|to| cmp::min(to, MAX_SAFE_TIMEOUT) as libc::c_int)
```
Thomasdezeeuw, line 101: We don't need the the remainder, we can just get the `subsec_nanos` directly (just a read of a field). Also let's put into a single `map` closure. Something like the following should would

```suggestion
            .map(|to| cmp::min(
                to.as_millis()
                    // as_millis() truncates, so round up to the nearest millisecond as the documentation
                    // says can happen.  This avoids turning submillisecond timeouts into immediate returns
                    // unless the caller explicitly requests that by specifying a zero timeout.
                    + if to.subsec_nanos() == 0 { 0 } else { 1 }
            , MAX_SAFE_TIMEOUT) as libc::c_int)
```

This will need some manual formatting of the code and the comment.

yump: Closures merged and re-rustfmt'ed.

The remainder is not needed for timeouts between 0 and 1 ms, but without it one extra ms would be added to any exact-integer-ms timeout.  The biggest relative error is that a 1ms timeout gets doubled to 2ms.  If the timeouts callers specify are a smooth random variable that's no problem, but humans tend to pick nice round numbers.

The docs do say overrun is possible, and I recognize the desire to avoid a divide instruction, however the syscall a few lines later is probably a lot more expensive, especially on a CPU requiring spectre mitigations.

But it's your API and I have no reason to fight about it, so the remainder can go if you still want it gone.

Thomasdezeeuw: You're right, good catch. Let's do something like the following (will need formatting again):

```rust
            .map(|to| {
                let to_ms = to.as_millis();
                // as_millis() truncates, so round up to the nearest millisecond as the
                // documentation says can happen. This avoids turning submillisecond
                // timeouts into immediate returns unless the caller explicitly requests
                // that by specifying a zero timeout.
                let to_ms = to_ms + if to_ms == 0 && to.subsec_nanos() != 0 { 1 } else { 0 };
                cmp::min(MAX_SAFE_TIMEOUT, to_ms) as libc::c_int 
            })
```

Don't need the remainder and we only do it if the timeout is less than 1 millisecond, but not zero.

Thomasdezeeuw (approved):
Any chance you want to apply the same patch to Windows? Related code: https://github.com/tokio-rs/mio/blob/c6b5f13adf67483d927b1766faf984c1871e8abc/src/sys/windows/iocp.rs#L225-L231

If not it's no problem.

yump:
> Any chance you want to apply the same patch to Windows? Related code:
> 
> https://github.com/tokio-rs/mio/blob/c6b5f13adf67483d927b1766faf984c1871e8abc/src/sys/windows/iocp.rs#L225-L231
> 
> If not it's no problem.

[Same patch applied to Windows, on another branch](https://github.com/yump/mio/tree/short_timeout_fix_windows).  I don't have any Windows machines, but `cargo test --all-features` passes under Linux.

Thomasdezeeuw:
Thanks @yump. If you want to create another pr with the Windows change our CI will test it. You can check it locally using `make check_all_target` (might want to change the `$TARGETS` to just Windows) using `cargo check`.

PR #1618, opened 2022-09-10
protocol: add feature to create MPTCP sockets

Martichou:
Hey !

I was working on this "feature" for a side project, I thought it might be nice to submit it upstream :)

-----

This commit add support for creating MPTCP sockets.

Small details:
- We can only create a MPTCP socket in a TCP context. This is why we check that we're in an SOCK_STREAM socket type and that the domain is either AF_INET(6).
- Disabled the TTL test and helper in case of MPTCP as it's not supported by the protocol.

	-> "Operation not supported"

As a more general note, MPTCP is now more widely
available since Linux Kernel >= 5.6. But it still
need to be enabled manually using:
$ sysctl net.mptcp.enabled=1

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,8 @@ os-ext = [
 # Enables `mio::net` module containing networking primitives.
 net = []
 
+mptcp = []
+
 [dependencies]
 log = "0.4.8"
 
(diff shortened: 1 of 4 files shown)


Thomasdezeeuw:
I don't think we'll support it Mio as it tries to follow the API in std lib. I'm afraid you'll have to use socket2. It may be possible to convert a socket2 socket to a `TcpStream` and use that? Otherwise you'll have to use [`SourceFd`](https://docs.rs/mio/latest/mio/unix/struct.SourceFd.html).

PR #1624, opened 2022-10-19
deps: update to windows-sys 0.42.0

poliorcetics:
`windows-sys 0.42.0` was released some time ago and should break neither MSRV nor public API
(I hope).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,11 +49,11 @@ log = "0.4.8"
 libc = "0.2.121"
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = "0.36"
+version = "0.42"
 features = [
-  "Win32_Storage_FileSystem",         # Enables NtCreateFile
   "Win32_Foundation",                 # Basic types eg HANDLE
   "Win32_Networking_WinSock",         # winsock2 types/functions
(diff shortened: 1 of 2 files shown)


poliorcetics:
Strangely running `cargo check --all-targets --target x86_64-pc-windows-msvc` did not raise the errors caught by CI, I wonder why

Thomasdezeeuw:
@poliorcetics You can use `make check_all_targets`, possibly requiring `make install_target`, or use the command it uses with the correct target set.

Thomasdezeeuw (approved):
I think this is fine, it shouldn't break anything looking at the CI. Unfortunately windows-sys doesn't seem to keep a changelog, but looking at https://github.com/microsoft/windows-rs/compare/0.36.0...0.42.0 it seems ok.

Thomasdezeeuw:
Thanks @poliorcetics

poliorcetics:
@Thomasdezeeuw is releasing a new patch version possible ? Are there other changes in `master` that would preclude that ?

Thomasdezeeuw:
See https://github.com/tokio-rs/mio/pull/1625, I think I can do that later today.

poliorcetics:
Thank you very much !

PR #1625, opened 2022-10-24
Release v0.8.5

nox:


Thomasdezeeuw:
I've opened https://github.com/tokio-rs/mio/pull/1626 to use the changelog format we use, and include a couple more items.

PR #1627, opened 2022-11-02
Add SO_REUSEADDR for UDP sockets

louisna:
The SO_REUSEADDR setsockopt option is available for the [listener TcpSocket](https://github.com/tokio-rs/mio/blob/0accf7dc22f197245e6a1aa84096262cd6f6e4d4/src/net/tcp/listener.rs#L74).

[A previous issue](https://github.com/tokio-rs/mio/issues/186) suggested adding the support for the option for UdpSocket, but it was never added.

The QUIC protocol (RFC9000) uses UDP sockets to carry QUIC packets. As QUIC is also a connection-oriented protocol, it makes sense to me to add the same support for SO_REUSEADDR for these UdpSockets.

I c/p the code from the TCP socket codebase and enabled SO_REUSEADDR by default. Of course, I am willing to modify the code if it does not suits the expectations.

Best,

Louis

### DIFF
--- a/src/sys/unix/udp.rs
+++ b/src/sys/unix/udp.rs
@@ -2,6 +2,7 @@ use crate::sys::unix::net::{new_ip_socket, socket_addr};
 
 use std::io;
 use std::mem;
+use std::mem::size_of;
 use std::net::{self, SocketAddr};
 use std::os::unix::io::{AsRawFd, FromRawFd};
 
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
Sorry, this is out of scope for Mio. You can use socket2 to setup a socket with configuration options and turn that into a Mio socket.

PR #1630, opened 2022-11-14
net/uds: add methods to connect/bind with a socket address

lucab:
This enhances `connect` and `bind` logic for Unix Domain Sockets (UDS), by adding methods which allow to directly use a socket address. This mirrors similar features which already exist in `mio` for TCP and UDP sockets.

### DIFF
--- a/tests/unix_stream.rs
+++ b/tests/unix_stream.rs
@@ -78,6 +78,55 @@ fn unix_stream_connect() {
 }
 
 #[test]
+fn unix_stream_connect_addr() {
+    let (mut poll, mut events) = init_with_poll();
+    let barrier = Arc::new(Barrier::new(2));
+    let local_addr = {
+        // Workaround through a temporary listener using the same address,
+        // as there is currently no way of directly building a `SocketAddr`.
+        let path = temp_file("unix_stream_connect_addr");
+        let listener = net::UnixListener::bind(path.clone()).unwrap();
+        let mio_listener = mio::net::UnixListener::from_std(listener);
+        let address = mio_listener.local_addr().unwrap();
+        drop(mio_listener);
+        _ = std::fs::remove_file(&path);
+        address
+    };
+    let path = temp_file("unix_stream_connect_addr");
+    let listener = net::UnixListener::bind(path).unwrap();
+    let mut stream = UnixStream::connect_addr(&local_addr).unwrap();
+
+    let barrier_clone = barrier.clone();
+    let handle = thread::spawn(move || {
+        let (stream, _) = listener.accept().unwrap();
+        barrier_clone.wait();
+        drop(stream);
+    });
+
+    poll.registry()
+        .register(
+            &mut stream,
+            TOKEN_1,
+            Interest::READABLE | Interest::WRITABLE,
+        )
+        .unwrap();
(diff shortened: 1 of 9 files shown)


Thomasdezeeuw:
I assume this is based on https://github.com/rust-lang/rust/issues/85410, which seems to be on it's way to stable.

lucab:
@Thomasdezeeuw yes, kinda. My final goal would be to use **abstract** UDS in tokio (and higher up). For that I'm really looking forward to `SocketAddr::from_abstract_namespace(&[u8])`, which is still currently moving a bit. So maybe later, I'm not in a hurry.

I started with this which only tackles the connect/bind side, as it seemed quite safe to do and I noticed that TCP/UDP already had them.

lucab:
Just asking explicitly, what are the next steps here? Is this additional API fine here in `mio` or was the comment above suggesting to directly put this in `tokio` instead?

Thomasdezeeuw:
> Just asking explicitly, what are the next steps here? Is this additional API fine here in `mio`

Yes.

> or was the comment above suggesting to directly put this in `tokio` instead?

Yes. :)

Basically, Mio follows the std lib, so I think we should wait until the API at least on the road of stabilisation in std lib. Then we match the api and aren't stuck with a different API due to a last minute name change in std lib, after we already made a release with the function.

So, if you want in Tokio today your best bet is to use socket2 directly their. But I don't decide what does and doesn't go in the Tokio api.

lucab:
Ack, thanks for the feedback. I'll leave this open here for the moment then, and revisit once stdlib stabilization is done.

I'll go check Tokio next, and see if they like a `socket2` implementation there. I can then hopefully switch that to use this logic directly from `mio`, later on.

lucab:
I got busy with many other things in life, but the good news is that in the meanwhile `bind_addr()` and `connect_addr()` got stabilized in Rust 1.70. The additional methods in this patch already use the same signatures.

I've rebased and de-conflicted this PR, it should be ready for reviewing and merging, PTAL.

tests/unix_stream.rs
```diff
@@ -77,6 +77,55 @@ fn unix_stream_connect() {
...
+        // Workaround through a temporary listener using the same address,
+        // as there is currently no way of directly building a `SocketAddr`.
+        let path = temp_file("unix_stream_connect_addr");
+        let listener = net::UnixListener::bind(path.clone()).unwrap();
+        let mio_listener = mio::net::UnixListener::from_std(listener);
+        let address = mio_listener.local_addr().unwrap();
+        drop(mio_listener);
+        _ = std::fs::remove_file(&path);
+        address
+    };
```
Thomasdezeeuw, line 94: Can't you just use `listener.local_addr()` after creating it?

Thomasdezeeuw (approved):
One question, otherwise LGTM.

tests/unix_stream.rs
lucab: Absolutely yes, you are right. I pushed a commit on top to simplify this, and rebased everything to current `master`.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @lucab

PR #1631, opened 2022-11-14
Update to windows-rs 0.43

Jasper-Bekkers:
Another version came out...


Thomasdezeeuw:
Doesn't seem to be release to crates.io yet: https://crates.io/crates/windows-sys/versions.

PR #420, opened 2016-07-02
Less truncating when casting timeout values for `Poll::poll`

tbu-:

PR #1616, opened 2022-09-03
Round nonzero <1ms timeouts to 1ms for Windows too

yump:
Fixes #1614

PR #1642, opened 2022-12-25
Again less truncating for timeout values in `Poll::poll`

tbu-:
Round up submillisecond wait times to the next millisecond on platforms where submillisecond sleep times aren't supported. This makes the behavior actually match the documentation.

See also #420, where this was implemented for the first time.

### DIFF
--- a/src/sys/unix/selector/epoll.rs
+++ b/src/sys/unix/selector/epoll.rs
@@ -88,12 +88,11 @@ impl Selector {
 
         let timeout = timeout
             .map(|to| {
-                let to_ms = to.as_millis();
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
Didn't https://github.com/tokio-rs/mio/pull/1615 and https://github.com/tokio-rs/mio/pull/1616 fix this?

tbu-:
Only for the special case of waiting between 0ms and 1ms.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @tbu-

PR #1646, opened 2023-01-30
Add a test to tcp_stream that tests if priority events work

poeschel:
This adds a test to the tcp_stream to test if priority events are working. This only works on linux. Linux sets the EPOLLPRI event type for those events.
This works by sending a tcp packet with the MSG_OOB flag set.

### DIFF
--- a/tests/tcp_stream.rs
+++ b/tests/tcp_stream.rs
@@ -793,3 +793,54 @@ fn hup_event_on_disconnect() {
         vec![ExpectEvent::new(Token(1), Interest::READABLE)],
     );
 }
+
+#[test]
+#[cfg(target_os = "linux")]
+fn pollpri() {
+    use mio::net::TcpListener;
+    use nix::sys::socket::{
+        connect, send, socket, AddressFamily, MsgFlags, SockFlag, SockType, SockaddrIn,
+    };
+
+    let (mut poll, mut events) = init_with_poll();
+    let addr = "127.0.0.1:0".parse().unwrap();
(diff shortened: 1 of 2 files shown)


tests/tcp_stream.rs
```diff
@@ -793,3 +793,56 @@ fn hup_event_on_disconnect() {
...
+    send(fd, &DATA1, MsgFlags::empty()).unwrap();
+    expect_events(
+        &mut poll,
+        &mut events,
+        vec![ExpectEvent::new(Token(1), Interest::READABLE)],
+    );
+
+    // MsgFlags::MSG_OOB sends a tcp packet which is received by linux kernel with
+    // POLLPRI event type aka Readiness::PRIORITY set.
+    send(fd, &DATA2, MsgFlags::MSG_OOB).unwrap();
```
Thomasdezeeuw, line 842: I've said the same thing about the previous example you send. This will not generate an event as you're not draining the read readiness of the stream. Remove the `send` call above on line 833.

poeschel: I updated the test as requested.

poeschel:
I updated the
`cfg(unix)` conditional compilation to
`cfg(target_os = "linux")` since I do not expect this to work on *BSDs.

poeschel:
I `cargo fmt`ed the change.

Thomasdezeeuw:
Closing in favour of https://github.com/tokio-rs/mio/pull/1647.

PR #1647, opened 2023-02-04
Add Interest::PRIORITY

### LINKED ISSUE #1645: A way to add EPOLLPRI event type to the epoll selector
While working with Video4Linux2 I am searching for a way to (a)wait `POLLPRI` events coming from the kernel.
I do see, that an `Event` can be checked if it `is_prioprity`:
https://github.com/tokio-rs/mio/blob/fa4e4b3c58af76909d047abd81d3eaca1d8c5736/src/sys/unix/selector/epoll.rs#L227-L229
But I wonder how to register at the `Selector` in a way that it receives this type of events?
Looking at https://github.com/tokio-rs/mio/blob/fa4e4b3c58af76909d047abd81d3eaca1d8c5736/src/sys/unix/selector/epoll.rs#L114-L123 it seems, that some kind of  PRIORITY `Interest` would be needed, but it is not there: https://github.com/tokio-rs/mio/blob/fa4e4b3c58af76909d047abd81d3eaca1d8c5736/src/interest.rs#L19-L34
I searched a bit through the history and found something similar was already there:
https://github.com/tokio-rs/mio/blob/434405a66c66c1e2b15a07388dd0ed4edcc7e3c8/src/event_imp.rs#L613-L622
But it is gone and I don't know why.
So my questions:
* Why is the `is_priority` function there? It is of no use at the moment, right?
* What happened to the old `Ready:PRIORITY`? Is there some reason why this got removed or was it a mistake?
* What would be a good (upstreamable) way to implement the support for `POLLPRI` type events?

Thomasdezeeuw:
To trigger Event::is_priority on Linux and Android.

In the future we might want to include EV_OOBAND for kqueue, but that seems to be Apple only (FreeBSD and OpenBSD don't implement it at least).

Closes #1645

Does this solve the problem @poeschel, @Urist-McGit?

### DIFF
--- a/src/interest.rs
+++ b/src/interest.rs
@@ -20,18 +20,9 @@ pub struct Interest(NonZeroU8);
 const READABLE: u8 = 0b0001;
 const WRITABLE: u8 = 0b0010;
 // The following are not available on all platforms.
-#[cfg_attr(
-    not(any(
-        target_os = "dragonfly",
-        target_os = "freebsd",
-        target_os = "ios",
-        target_os = "macos",
-    )),
-    allow(dead_code)
-)]
 const AIO: u8 = 0b0100;
-#[cfg_attr(not(target_os = "freebsd"), allow(dead_code))]
 const LIO: u8 = 0b1000;
+const PRIORITY: u8 = 0b10000;
 
 impl Interest {
     /// Returns a `Interest` set representing readable interests.
@@ -52,6 +43,10 @@ impl Interest {
     /// Returns a `Interest` set representing LIO completion interests.
     #[cfg(target_os = "freebsd")]
     pub const LIO: Interest = Interest(unsafe { NonZeroU8::new_unchecked(LIO) });
+
+    /// Returns a `Interest` set representing priority completion interests.
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    pub const PRIORITY: Interest = Interest(unsafe { NonZeroU8::new_unchecked(PRIORITY) });
 
     /// Add together two `Interest`.
     ///
@@ -104,14 +99,19 @@ impl Interest {
         (self.0.get() & WRITABLE) != 0
     }
 
-    /// Returns true if `Interest` contains AIO readiness
+    /// Returns true if `Interest` contains AIO readiness.
     pub const fn is_aio(self) -> bool {
         (self.0.get() & AIO) != 0
     }
 
-    /// Returns true if `Interest` contains LIO readiness
+    /// Returns true if `Interest` contains LIO readiness.
     pub const fn is_lio(self) -> bool {
         (self.0.get() & LIO) != 0
+    }
+
+    /// Returns true if `Interest` contains priority readiness.
+    pub const fn is_priority(self) -> bool {
+        (self.0.get() & PRIORITY) != 0
     }
 }
 
@@ -170,6 +170,16 @@ impl fmt::Debug for Interest {
                     write!(fmt, " | ")?
                 }
                 write!(fmt, "LIO")?;
+                one = true
+            }
+        }
+        #[cfg(any(target_os = "linux", target_os = "android"))]
+        {
+            if self.is_priority() {
+                if one {
+                    write!(fmt, " | ")?
+                }
+                write!(fmt, "PRIORITY")?;
                 one = true
             }
         }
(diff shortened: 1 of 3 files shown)


poeschel:
Since this is pretty much, what I proposed here: https://github.com/tokio-rs/mio/issues/1645#issuecomment-1408943515
I do not understand why you try to sell this as your work, but okay. Main thing is we solve the problem.

> Does this solve the problem @poeschel, @Urist-McGit?

Unfortunately no, it does not solve it fully. You missed debug-fmt-ing the priority interest. In your case debug printing a priority interest alone does panic. Have a look at my [proposal](https://github.com/poeschel/mio/commit/282edd32c7459d14cd4efc0ecafd652ab806f30f#) in src/interest.rs lines 186-192.

Thank you very much for your work regarding this issue.

Thomasdezeeuw:
> Since this is pretty much, what I proposed here: [#1645 (comment)](https://github.com/tokio-rs/mio/issues/1645#issuecomment-1408943515) I do not understand why you try to sell this as your work, but okay. Main thing is we solve the problem.

I only say the example, which had a bug I pointed out, then I got a pr (https://github.com/tokio-rs/mio/pull/1646) with the exact same bug and an unwanted dependency. So, in my free time I tried fix the problem you were having...

If you had the solution and a working fix, why not submit that in as a pr?

poeschel:
> I only saw the example, which had a bug I pointed out, then I got a pr (https://github.com/tokio-rs/mio/pull/1646) with the exact same bug and an unwanted dependency. So, in my free time I tried fix the problem you were having...

Don't get me wrong. As I said already: I am very thankful for your work. Be it:
- looking and commenting on my [issue](https://github.com/tokio-rs/mio/issues/1645#issuecomment-1408943515)
- looking and commenting on the [example](https://github.com/poeschel/mio-pollpri-test) we provided
- looking at the [PR](https://github.com/tokio-rs/mio/pull/1646) with the test demonstrating the problem
- your PR here
- your work on mio in general

I am sorry, that you did not see my [proposed fix](https://github.com/poeschel/mio/commit/282edd32c7459d14cd4efc0ecafd652ab806f30f) hidden in the comment on the [issue](https://github.com/tokio-rs/mio/issues/1645#issuecomment-1408943515). This is my fault. I thought you get notified, if you were commenting on an issue and I add a new comment.

> If you had the solution and a working fix, why not submit that in as a pr?

This is because I am really unsure. You are the holy magicans that make this work and keep the code tidy andclean and I am a clueless guy having an issue. At first I was unsure if the problem I saw is really an issue in mio or if it is just me doing something wrong. While investigating the issue, I saw in the history that there actually has been some PRIORITY stuff already, but it was removed. Maybe for a good reason? Before submitting a PR I wanted to know where the problem really is. During our discussion on the issue, it became clear to me, that there really is an issue and I started thinking about how a solution could look like. This then became the code in my proposal.

poeschel:
Could you please add something like
https://github.com/poeschel/mio/blob/282edd32c7459d14cd4efc0ecafd652ab806f30f/src/interest.rs#L186-L192
? Maybe you also need ` #[cfg(any(target_os = "linux", target_os = "android"))]` ?
 I think then your PR is fine.

Thomasdezeeuw:
> Don't get me wrong. As I said already: I am very thankful for your work. Be it:

Thank you.

> This is because I am really unsure.

That's what pr reviews are for, it's an easy place to see what you're proposing and it's a place for a discussion.

> You are the holy magicans that make this work and keep the code tidy andclean and I am a clueless guy having an issue. At first I was unsure if the problem I saw is really an issue in mio or if it is just me doing something wrong. While investigating the issue, I saw in the history that there actually has been some PRIORITY stuff already, but it was removed. Maybe for a good reason? Before submitting a PR I wanted to know where the problem really is. During our discussion on the issue, it became clear to me, that there really is an issue and I started thinking about how a solution could look like. This then became the code in my proposal.

No magicians at work, we're all writing the same code.

I'll push a commit with the formatting fix and give you co-author credit.

poeschel (approved):
The code looks good to me now.
But what happened to the Co-authored-by?

Thank you!

Darksonn (approved):

Thomasdezeeuw:
@poeschel it seems you commit with a different email address than GitHub recognises, what do you prefer I use?

poeschel:
> @poeschel it seems you commit with a different email address than GitHub recognises, what do you prefer I use?

Best is, you use `lars.poeschel@edag.com` I also added this to my github profile.
Thank you!

PR #1655, opened 2023-02-13
Use consistent cfg attributes

Thomasdezeeuw:


taiki-e:
Note that there is a proposal to deprecate target_vendor: https://github.com/rust-lang/rust/issues/100343 / https://github.com/rust-lang/lang-team/issues/102

Thomasdezeeuw:
> Note that there is a proposal to deprecate target_vendor: [rust-lang/rust#100343](https://github.com/rust-lang/rust/issues/100343) / [rust-lang/lang-team#102](https://github.com/rust-lang/lang-team/issues/102)

Well that's great... I guess we shouldn't go with this this then.

I've pushed another commit to clean up the remaining attributes which I've found while doing the previous work.

src/sys/unix/net.rs
```diff
@@ -20,14 +20,14 @@ pub(crate) fn new_socket(domain: libc::c_int, socket_type: libc::c_int) -> io::R
...
-        target_os = "openbsd"
+        target_os = "openbsd",
     ))]
     let socket_type = socket_type | libc::SOCK_NONBLOCK | libc::SOCK_CLOEXEC;
 
     let socket = syscall!(socket(domain, socket_type, 0))?;
 
     // Mimick `libstd` and set `SO_NOSIGPIPE` on apple systems.
-    #[cfg(target_vendor = "apple")]
+    #[cfg(any(target_os = "ios", target_os = "macos"))]
```
hawkw, line 30: Nit, take it or leave it: rust-lang/lang-team#102 also mentions the existence of `target_family = "apple"`, which is set in addition to `target_family = "unix"` when building for both iOS and macOS. It might be a little bit simpler to use `target_family = "apple"` here instead?

src/sys/unix/selector/kqueue.rs
```diff
@@ -21,15 +21,15 @@ type Count = libc::size_t;
...
+#[cfg(any(target_os = "ios", target_os = "macos"))]
 type Filter = i16;
 #[cfg(target_os = "netbsd")]
 type Filter = u32;
 
 // Type of the `flags` field in the `kevent` structure.
 #[cfg(any(target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd"))]
 type Flags = libc::c_ushort;
-#[cfg(any(target_os = "macos", target_os = "ios"))]
+#[cfg(any(target_os = "ios", target_os = "macos"))]
```
hawkw, line 32: similarly, would `target_family="apple"` make sense for these?

tests/util/mod.rs
```diff
@@ -249,9 +249,9 @@ pub fn set_linger_zero(socket: &TcpStream) {
         libc::setsockopt(
             socket.as_raw_fd(),
             libc::SOL_SOCKET,
-            #[cfg(target_vendor = "apple")]
+            #[cfg(any(target_os = "ios", target_os = "macos"))]
             libc::SO_LINGER_SEC,
-            #[cfg(not(target_vendor = "apple"))]
+            #[cfg(not(any(target_os = "ios", target_os = "macos")))]
```
hawkw, line 254: Nit: I believe we could use `target_family = "apple"` here?

hawkw (approved):
This looks good to me. However, I did notice while reading https://github.com/rust-lang/lang-team/issues/102 that there's also a `target_family = "apple"` cfg that's set for both iOS and macOS. Would it be a bit simpler to use that rather than testing for both `target_os` cfgs?

taiki-e:
@hawkw 

> target_family = "apple"

This has not yet been implemented.

```console
$ rustc --print cfg --target x86_64-apple-darwin | grep target_family
target_family="unix"
$ rustc -V
rustc 1.69.0-nightly (5b8f28453 2023-02-12)
```

Thomasdezeeuw:
> This looks good to me. However, I did notice while reading [rust-lang/lang-team#102](https://github.com/rust-lang/lang-team/issues/102) that there's also a `target_family = "apple"` cfg that's set for both iOS and macOS. Would it be a bit simpler to use that rather than testing for both `target_os` cfgs?

I agree I would prefer to use `target_family = "apple"`, but as @taiki-e pointed it doesn't exists yet. So for now let's make it consistent and when the rustc side of things are resolved we can rethink this.

PR #1658, opened 2023-02-25
Add tvOS and watchOS

Thomasdezeeuw:
Follow up to https://github.com/tokio-rs/mio/pull/1655.

### DIFF
--- a/src/interest.rs
+++ b/src/interest.rs
@@ -37,6 +37,8 @@ impl Interest {
         target_os = "freebsd",
         target_os = "ios",
         target_os = "macos",
+        target_os = "tvos",
+        target_os = "watchos",
     ))]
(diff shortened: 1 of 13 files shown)


notgull:
It looks like the `cfg` statements are getting pretty long. Have you considered using [`cfg_aliases`](https://crates.io/crates/cfg_aliases) to reduce the size and scope of these blocks?

Thomasdezeeuw:
> It looks like the `cfg` statements are getting pretty long. Have you considered using [`cfg_aliases`](https://crates.io/crates/cfg_aliases) to reduce the size and scope of these blocks?

There are getting, but we're not going to add dependencies for this.

Darksonn (approved):

PR #1662, opened 2023-03-23
Add waker implementation for WASI

xofyarg:
There's no standard way to do it at the moment under WASI. Instead if there are two connected socket from the host side, they can be used to emulate Linux's eventfd.


Thomasdezeeuw (changes requested):
This won't be accepted as-is it would need to work with the `Waker::new` API, I won't accept `Waker::init`.

Issue #1611, opened 2022-08-17
Reconsider supported draining behaviour


Thomasdezeeuw:
For Unix (`kqueue` and `epoll`) to drain readiness all you have to do is read/write until you get a "short" read/write, i.e. less bytes are processed than your buffer size. However Mio doesn't guarantee that another event is return until the I/O operations hits a `WouldBlock` error. This means that to strictly follow Mio's docs you'll have to issue another pointless system call (read/write/etc.).

This is because on Windows Mio could only guarantee the readiness to be drained once a `WouldBlock` error was returned. However nowadays we control that ourselves in [`IoSourceState::do_io`](https://github.com/tokio-rs/mio/blob/c6b5f13adf67483d927b1766faf984c1871e8abc/src/sys/windows/mod.rs#L87-L94).  So, we could change it to reregister when e.g. the read bytes < buffer size.

Also find out if all of this is true for `poll` (see https://github.com/tokio-rs/mio/pull/1602).

Janrupf:
> Also find out if all of this is true for poll (see https://github.com/tokio-rs/mio/pull/1602).

If I understand correctly, it is. `poll` is level-based by default anyway, so we have full control over when we want to re-arm an edge trigger. https://github.com/tokio-rs/mio/pull/1602 currently does not implement this, but since for `poll` this solves other issues as well, I'm working on implement edge-based triggers using `IoSourceState`.

Noah-Kennedy:
This can be a pretty impactful observation in certain workloads.

Thomasdezeeuw:
One thing we need to consider are the `try_io` functions on various types, e.g. [`TcpStream::try_io`](https://docs.rs/mio/latest/mio/net/struct.TcpStream.html#method.try_io), which currently don't give us access to the amount of bytes read.

Thomasdezeeuw:
I don't think we can change the behaviour any more. The public `try_io` method doesn't have access to the internal success type of the function it's wrapping, nor could it know if e.g. a read was short. Furthermore I don't this working with the `poll(2)` implementation.

Based on that I'm going to close this and keep the current behaviour.

If someone does have an idea on how to support `try_io` and the `poll(2)` we can reopen this.

utkarshgupta137:
Hi,

[...]

PR #1667, opened 2023-05-01
Adds support for UnixStream and UnixListener on Windows

KolbyML:
This PR supersedes PR https://github.com/tokio-rs/mio/pull/1610 and solves issue https://github.com/tokio-rs/mio/issues/1609

### Background Info
I have spoken with Thomasdezeeuw and his main complaints with the last PR was
> It's been a while since I've looked at the pr. But basically I don't want any (substantial) changes to the tests as that would be we would break already existing (Unix only) code that people have written using Mio. Ideally we pave over an platform differences between Unix and Windows

So I was looking into cleaning up the PR and cleaned up some messy edges. I am making this draft pull request just to show in the public where it is at and what I am working on.

### Ready for review !!!

The PR is ready review I made it so we didn't have to change the tests. The only test what is different is the try_io doc test which is because the unix one uses pair which isn't supported on Windows. But all the tests in the tests folder matches.

The only addition to the tests would be in ``tests/UnixStream.rs``
```
assert_would_block(stream.read(&mut buf));
```

This is because of https://github.com/tokio-rs/mio/issues/1611 which is known about and intentional done this way.

``sys/windows/stdnet`` is a implementation matching the standards libraries implementation of Unix UDS. So if the standard library ever adds support for Unix UDS for windows it should be a drop in replacement.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -91,6 +91,10 @@ pub mod windows {
     //! Windows only extensions.
 
     pub use crate::sys::named_pipe::NamedPipe;
+    // blocking windows uds which mimick std implementation used for tests
+    cfg_net! {
+        pub use crate::sys::windows::stdnet;
+    }
 }
 
 pub mod features {
--- a/src/net/mod.rs
+++ b/src/net/mod.rs
@@ -32,8 +32,13 @@ pub use self::tcp::{TcpListener, TcpStream};
 mod udp;
 #[cfg(not(target_os = "wasi"))]
 pub use self::udp::UdpSocket;
+#[cfg(not(target_os = "wasi"))]
+mod uds;
+#[cfg(not(target_os = "wasi"))]
+pub use self::uds::{SocketAddr, UnixListener, UnixStream};
+
+#[cfg(not(target_os = "wasi"))]
+pub(crate) use self::uds::AddressKind;
 
 #[cfg(unix)]
-mod uds;
-#[cfg(unix)]
-pub use self::uds::{SocketAddr, UnixDatagram, UnixListener, UnixStream};
+pub use self::uds::UnixDatagram;
--- a/src/net/tcp/stream.rs
+++ b/src/net/tcp/stream.rs
@@ -269,49 +269,49 @@ impl TcpStream {
 
 impl Read for TcpStream {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        self.inner.do_io(|inner| (&*inner).read(buf))
+        self.inner.do_io(|mut inner| inner.read(buf))
     }
 
     fn read_vectored(&mut self, bufs: &mut [IoSliceMut<'_>]) -> io::Result<usize> {
-        self.inner.do_io(|inner| (&*inner).read_vectored(bufs))
+        self.inner.do_io(|mut inner| inner.read_vectored(bufs))
     }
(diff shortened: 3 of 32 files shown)


leiless:
Any plan for reviewing/merging this PR?

src/lib.rs
```diff
@@ -91,6 +91,10 @@ pub mod windows {
     //! Windows only extensions.
 
     pub use crate::sys::named_pipe::NamedPipe;
+    // blocking windows uds which mimick std implementation used for tests
+    cfg_net! {
+        pub use crate::sys::windows::stdnet;
+    }
```
Thomasdezeeuw, line 97: We're not going to expose this.

Thomasdezeeuw (commented):
> > Well, this pr has some of the same issues pointed out in previous iterations of this.
> 
> Would you be able to repost the "issues pointed out in previous iterations of this PR" in this PR. To make tracking the issues what still need to be done easier to track. "previous iterations" had certain changes which you thought took the PR backwards, which is what I tried to resolve in this PR. It is currently unclear exactly what you want. As your statement is quite broad. And the "history" of what you wanted done is confusing because of the reasons I stated before.

I pointed out an obvious one, but see https://github.com/tokio-rs/mio/pull/1610 for my other ~40 comments. I'm not going to repeat myself, I simply don't have the time.

src/lib.rs
KolbyML: It has previously been stated the tests shouldn't change. To make the tests not change we need to have a blocking windows UDS implementation. Having this requirement only leaves us 2 options.

- 1: Add ``#[cfg(test)]`` to the expose so it can only be used in tests
- 2: Implement UDS support in the rust STD

Would option 1 even be allowed? or an Option?

If it isn't an option then if we want windows UDS in tokio there is a requirement for it being implemented in the STD First. Originally it was thought that we could add Windows UDS since MIO already supports Windows Named Pipes without std support.

So is the president that std support is required. That is fine, it is nice to just know that is definitly the step forward. Because the only non-std way to fulfill the no-changing-test-requirement would be option 1.

Thomasdezeeuw: > It has previously been stated the tests shouldn't change.

Yes, we need to ensure that the existing behaviour works on Windows as well.

> To make the tests not change we need to have a blocking windows UDS implementation. Having this requirement only leaves us 2 options.
> 
> * 1: Add `#[cfg(test)]` to the expose so it can only be used in tests
> * 2: Implement UDS support in the rust STD
> 
> Would option 1 even be allowed? or an Option?

The first option doesn't work as we use Rust's integration tests, which doesn't have access to another crate's `#[cfg(test)]` exposed items. The second option would be ideal, but that's currently not possible as far as I know.

> If it isn't an option then if we want windows UDS in tokio there is a requirement for it being implemented in the STD First. Originally it was thought that we could add Windows UDS since MIO already supports Windows Named Pipes without std support.
> 
> So is the precedent that std support is required. That is fine, it is nice to just know that is definitly the step forward. Because the only non-std way to fulfill the no-changing-test-requirement would be option 1.

You're overlooking a third option. Have a blocking implementation inside the test code.

Thomasdezeeuw:
I'm going to close this because it still needs some work and hasn't seen any activity recently. @KolbyML thanks for the effort!

PR #1668, opened 2023-05-02
Fix FreeBSD CI test

KolbyML:
Curl is failing on CI causing FreeBSD CI test to fail on my other PR https://github.com/tokio-rs/mio/pull/1667/checks?check_run_id=13155549454

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=270940 I found other references to the same bug

Curl 7.88.1 up fails. Curl is already installed on FreeBSD so removing the command to update it fixes this issue

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -7,8 +7,7 @@ env:
 task:
   name: FreeBSD
   setup_script:
-    - pkg install -y curl
-    - curl https://sh.rustup.rs -sSf --output rustup.sh
+    - fetch https://sh.rustup.rs -o rustup.sh
     - sh rustup.sh -y --profile minimal
   cargo_cache:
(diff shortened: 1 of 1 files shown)


asomers (approved):
LGTM.  Another option is to use `fetch`, which is built into the base system rather than from ports.  Use it like this:
```
fetch https://sh.rustup.rs -o rustup.sh
sh rustup.sh -y --profile=minimal
```

KolbyML:
> fetch https://sh.rustup.rs -o rustup.sh

I updated the PR to use your suggested solution since base system would be ideal I agree

asomers (approved):

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @KolbyML

PR #1670, opened 2023-05-06
Expose Windows handle on Registry

### LINKED ISSUE #1669: Exposing `AsRawHandle` on `Registry`
It appears that while Unix platforms expose `AsRawFd`, there is no Windows equivalent. There already is `AsRawHandle` implementation [in `CompletionPort`](https://github.com/tokio-rs/mio/blob/master/src/sys/windows/iocp.rs#L141), it just needs to be propagated up the chain.

h33p:
This PR implements `AsRawHandle` on `Registry`. I believe the implications of this are the same as having `AsRawFd` on unix, but of course I could be wrong.

Closes #1669

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -2,6 +2,8 @@ use crate::{event, sys, Events, Interest, Token};
 use log::trace;
 #[cfg(unix)]
 use std::os::unix::io::{AsRawFd, RawFd};
+#[cfg(windows)]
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw:
Per your comment here: https://github.com/tokio-rs/mio/issues/1669#issuecomment-1537369859 I think we should close this.

PR #1679, opened 2023-05-30
Add recv_with_ancillary_data for udp socket

yuguorui:
Add recv_with_ancillary_data for udp socket to facilitate IP_PKTINFO and IP_RECVORIGDSTADDR usage.

IP_RECVORIGDSTADDR is very handly for UDP transparent proxy with tproxy, and these patches are precondition for tokio to support ancillary data receiving.

`struct cmsghdr` parsing is not included in these patches on purpose, to avoid introducing addtional dependencies.

### DIFF
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -303,6 +303,13 @@ impl UdpSocket {
         self.inner.do_io(|inner| inner.recv(buf))
     }
 
+    /// Receives IP_PKTINFO ancillary message that contains a pktinfo structure that supplies some information 
+    /// about the incoming packet. This works only for datagram oriented sockets.
+    #[cfg(unix)]
+    pub fn set_pktinfo(&self, enable: bool) -> io::Result<()> {
+        self.inner.do_io(|inner| sys::udp::set_pktinfo(inner, enable))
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw:
Mio follows the API provided by the std lib, so this won't be accepted by Mio. However it's something I'm willing to consider for socket2: https://github.com/rust-lang/socket2.

yuguorui:
> Mio follows the API provided by the std lib, so this won't be accepted by Mio. However it's something I'm willing to consider for socket2: https://github.com/rust-lang/socket2.

Thanks for your reminding.
I understand that socket2 maybe a more advanced networking API, but it doesn't seem to be event driven, could socket2 be effectively driven by tokio?

Thomasdezeeuw:
> Thanks for your reminding. I understand that socket2 maybe a more advanced networking API, but it doesn't seem to be event driven, could socket2 be effectively driven by tokio?

You can use [`socket2::SockRef`](https://docs.rs/socket2/latest/socket2/struct.SockRef.html) to call any method on `socket2::Socket` on `mio::net::UdpSocket`.

Issue #1682, opened 2023-06-22
Need switch to not use eventfd on `target_os = "linux"`


haraldh:
For an SGX enclave (using gramine), I can't use the eventfd mechanism. Would be nice, if I could fallback to the pipe mechanism.

Currently I am using https://github.com/haraldh/mio/commit/5852388314d46f0b231366dec683ede148a76d4e

But an additional feature might be too much maintenance burden.

Would this help? `RUSTFLAGS='--cfg no_eventfd'` ... so we can do `#[cfg(all(not(no_eventfd), any(target_os = "linux", target_os = "android")))]`

haraldh:
Oh yeah, `RUSTFLAGS='--cfg no_eventfd'` works great... see https://github.com/tokio-rs/mio/compare/master...haraldh:mio:no_eventfd_cfg

notgull:
Is it a documented issue that `eventfd` can't be used inside of an SGX enclave, but `epoll` and pipes can?

haraldh:
@notgull at least for gramine:
https://gramine.readthedocs.io/en/stable/manifest-syntax.html#allowing-eventfd

PR #1683, opened 2023-06-23
feat: add no_eventfd cfg switch

haraldh:
This disables the use of `eventfd` on linux for the `Waker` implementation. Instead `pipe` is used as a fallback.

Because some SGX runtimes like gramine do not have a secure emulation of the `eventfd` syscall, compiling with `RUSTFLAGS='--cfg no_eventfd'` will produce a more secure SGX application.

See also:
https://gramine.readthedocs.io/en/stable/manifest-syntax.html#allowing-eventfd

Resolves: #1682


Thomasdezeeuw:
I'm going to do this myself as we likely want a similar flag for https://github.com/tokio-rs/mio/pull/1602, so I'll need to think a bit more about documentation and naming. Thanks for putting in the effort.

Issue #1632, opened 2022-11-17
Linking fails with x86_64-pc-windows-gnu toolchain


eiderdaus:
Mio defines `NtCancelIoFileEx` as follows:

```
#[link(name = "ntdll")]
extern "system" {
    /// See <https://processhacker.sourceforge.io/doc/ntioapi_8h.html#a0d4d550cad4d62d75b76961e25f6550c>
    ///
    /// This is an undocumented API and as such not part of <https://github.com/microsoft/win32metadata>
    /// from which `windows-sys` is generated, and also unlikely to be added, so
    /// we manually declare it here
    fn NtCancelIoFileEx(
        FileHandle: HANDLE,
        IoRequestToCancel: *mut IO_STATUS_BLOCK,
        IoStatusBlock: *mut IO_STATUS_BLOCK,
    ) -> NTSTATUS;
}
```

If I add `fn main(){}` and build with `cargo build` I get the following linker error:

```
error: linking with `x86_64-w64-mingw32-gcc` failed: exit code: 1
  |
[...]

PR #1688, opened 2023-07-12
Support x86_64-pc-windows-gnu target

Thomasdezeeuw:
This updates windows-sys to current master and adds the x86_64-pc-windows-gnu target.

/cc @eiderdaus

Closes https://github.com/tokio-rs/mio/issues/1632

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,8 +51,11 @@ log = { version = "0.4.8", optional = true }
 libc = "0.2.149"
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = "0.48"
+version = "0.52"
 features = [
+  "Wdk_Foundation",                   # Required for AFD.
+  "Wdk_Storage_FileSystem",           # Required for AFD.
+  "Wdk_System_IO",                    # Required for AFD.
   "Win32_Foundation",                 # Basic types eg HANDLE
   "Win32_Networking_WinSock",         # winsock2 types/functions
   "Win32_Storage_FileSystem",         # Enables NtCreateFile
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,6 +79,7 @@ targets = [
   "aarch64-linux-android",
   "wasm32-wasi",
   "x86_64-apple-darwin",
+  "x86_64-pc-windows-gnu",
   "x86_64-pc-windows-msvc",
   "x86_64-unknown-dragonfly",
   "x86_64-unknown-freebsd",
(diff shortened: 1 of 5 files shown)


Thomasdezeeuw:
Ah, windows-sys also increased their MSRV beyond 1.46... That's the second dependency that did so (log being the other).

eiderdaus:
windows-sys 1.52.0 was finally released two weeks ago so I think this can be merged (after rebasing and replacing the patched windows-sys with the new release of course). I found no issues testing this locally. Meanwhile Tokio's MSRV is 1.63 while windows-sys' is 1.56, so after merging this I think it should also be no problem to make a new mio release and use it in Tokio.

Thomasdezeeuw:
> windows-sys 1.52.0 was finally released two weeks ago so I think this can be merged (after rebasing and replacing the patched windows-sys with the new release of course). I found no issues testing this locally. Meanwhile Tokio's MSRV is 1.63 while windows-sys' is 1.56, so after merging this I think it should also be no problem to make a new mio release and use it in Tokio.

Except for Tokio TLS, which has a lower MSRV :( see https://github.com/tokio-rs/mio/pull/1725.

It's an annoying problem I know.

eiderdaus:
Not sure I understand. Do you propose to wait until Tokio 1.25 is phased out? Why not integrate this change, create a new (seemingly unnecessarily breaking) mio release 0.9 (or even 1.0, why not) and then you can still make bug fix releases for mio 0.8 to support Tokio 1.25. Just like Tokio is handling LTS itself, kind of.

Thomasdezeeuw:
> Not sure I understand. Do you propose to wait until Tokio 1.25 is phased out? Why not integrate this change, create a new (seemingly unnecessarily breaking) mio release 0.9 (or even 1.0, why not) and then you can still make bug fix releases for mio 0.8 to support Tokio 1.25. Just like Tokio is handling LTS itself, kind of.

I'm not going to bump Mio's version for MSRV, that's not a scalable solution.

Thomasdezeeuw:
We can either wait until March for the TLS version to not be supported any more, or we can decide the TLS version requires a fixed Mio version. We haven't really decided yet.

Thomasdezeeuw:
> I'm not sure what you mean by "fixed" in this context.

I'm mean Tokio TLS could use a fixed version of Mio v0.8.9, as is fixed to v0.8.9 specifically (not v0.8.x).

> I just find it weird to have mio's development locked on a MSRV that is so much lower than tokio's _own_ master branch MSRV. Certainly there must be some solution. Waiting is none.

That's the price of TLS software...

eiderdaus:
> I'm mean Tokio LTS could use a fixed version of Mio v0.8.9, as is fixed to v0.8.9 specifically (not v0.8.x).

But if you bump Mio only to 0.8.10 for further development (including MSRV change etc.), a potential bug in Mio 0.8.9 then can not be fixed in this Tokio LTS version, or? This does not make any sense to me.

> That's the price of LTS software...

No. I pointed out a reasonable solution yet you decide to ignore it. _That's_ the price of _open source_ software...

The existence of LTS versions in Tokio does not prevent further development of mio, it's only you. Just edit that Cargo.toml, set `version = "0.9.0"` and move on.

Thomasdezeeuw:
> > I'm mean Tokio LTS could use a fixed version of Mio v0.8.9, as is fixed to v0.8.9 specifically (not v0.8.x).
> 
> But if you bump Mio only to 0.8.10 for further development (including MSRV change etc.), a potential bug in Mio 0.8.9 then can not be fixed in this Tokio LTS version, or? This does not make any sense to me.
> 
> > That's the price of LTS software...
> 
> No. I pointed out a reasonable solution yet you decide to ignore it. _That's_ the price of _open source_ software...

Bump Mio to v0.9 also has a cost. It means all users must update and the versions are incompatible with one another.

Also the price of open source software is that is show some respect, which this conversation is starting to lose.

> The existence of LTS versions in Tokio does not prevent further development of mio, it's only you. Just edit that Cargo.toml, set `version = "0.9.0"` and move on.

This is really written by someone who hasn't maintained a library used by multiple people. It can be quite hard for people to update, so  "edit that Cargo.toml, set `version = "0.9.0"` and move on" can't be done by everyone.

PR #1700, opened 2023-07-28
Ignore Interrupte call in example

ssrlive:


examples/tcp_server.rs
```diff
@@ -40,7 +40,13 @@ fn main() -> io::Result<()> {
...
 
     loop {
-        poll.poll(&mut events, None)?;
+        if let Err(ref e) = poll.poll(&mut events, None) {
+            if interrupted(e) {
+                continue;
+            }
+            println!("Failed to poll for events: {}", e);
+            break;
+        }
```
Thomasdezeeuw, line 49: We probably still want to return the error.
```suggestion
        if let Err(err) = poll.poll(&mut events, None) {
            if interrupted(&err) {
                continue;
            }
            return Err(err);
        }
```

With above we don't need the added `Ok(())`

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @ssrlive

PR #1704, opened 2023-08-01
Add AIX operating system support

ecnelises:
AIX does not have `epoll` or `kqueue` functionality, so it relies on `force_poll_poll` feature.

### DIFF
--- a/src/sys/unix/waker.rs
+++ b/src/sys/unix/waker.rs
@@ -207,6 +207,7 @@ pub use self::kqueue::Waker;
     target_os = "redox",
 ))]
 mod pipe {
+    use crate::unix::pipe::new as new_pipe;
     use std::fs::File;
     use std::io::{self, Read, Write};
     use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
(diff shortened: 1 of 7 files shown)


src/sys/unix/waker.rs
```diff
@@ -222,7 +224,10 @@ mod pipe {
     impl WakerInternal {
         pub fn new() -> io::Result<WakerInternal> {
             let mut fds = [-1; 2];
+            #[cfg(not(target_os = "aix"))]
             syscall!(pipe2(fds.as_mut_ptr(), libc::O_NONBLOCK | libc::O_CLOEXEC))?;
+            #[cfg(target_os = "aix")]
+            syscall!(pipe(fds.as_mut_ptr()))?;
```
Thomasdezeeuw, line 230: You still need to set the `libc::O_NONBLOCK | libc::O_CLOEXEC` otherwise this will not work.

Thomasdezeeuw (changes requested):

src/sys/unix/waker.rs
```diff
@@ -265,6 +265,32 @@ mod pipe {
                 }
             }
         }
+
+        #[cfg(target_os = "aix")]
+        fn pipe2() -> io::Result<(RawFd, RawFd)> {
```
Thomasdezeeuw, line 270: Since this is essentially the same code as in `src/sys/unix/pipe.rs`, maybe we should reuse that code.

Thomasdezeeuw (approved):

ecnelises:
`cargo check --target x86_64-unknown-illumos --no-default-features --features os-poll` fails because no `os-ext` feature specified, and `crate::unix::pipe::new` relies on `os-ext`. Should we guard `mod pipe` with `os-ext`?

Thomasdezeeuw:
@ecnelises I've opened https://github.com/ecnelises/mio/pull/1 to solve that.

Thomasdezeeuw:
Thanks @ecnelises

PR #1705, opened 2023-08-04
Reduce allocations in poll(2) implementation

### LINKED ISSUE #1701: poll: Optimize locks and allocations in the critical path
The poll implementation in #1687 makes liberal use of locks and one critical allocation in `closed_raw_fds` in fairly hot code paths that can be optimized away.  There are some dead lock and efficiency concerns as identified here: https://github.com/tokio-rs/mio/pull/1687#discussion_r1275414504 and https://github.com/tokio-rs/mio/pull/1687#discussion_r1274693454 .

These optimizations may prove significant but can be punted until we get the code stable and proven to work well in a variety of environments.  Further, it'd be nice to have some new tests that make sure we don't regress by optimizing the code.

Thomasdezeeuw:
This changes the poll(2) implementation to reuse to `Vec`tors, so that we don't allocate in calls to `Poll::poll` in most cases. [^1] The downside to this is that `Poll` and `Registry` will contains two `Vec`s, which will increase their size. Especially for `Registry` this is annoying because they are never used. It might be worth considering exposing the same `Poll`/`Registry` split in the `sys` module to avoid this overhead for `Registry`.

This also contains a small cleanup commit to make the code match the style of the other files.

[^1]: Initially the vectors will be zero-sized and thus it will allocate in the first (couple of) calls, but at some point the capacity should be sufficient for further calls to no longer allocate.

Closes #1701

/cc @jasta this is what I meant in https://github.com/tokio-rs/mio/pull/1687#discussion_r1274680675 and https://github.com/tokio-rs/mio/pull/1687#discussion_r1274693454

### DIFF
--- a/src/sys/unix/selector/poll.rs
+++ b/src/sys/unix/selector/poll.rs
@@ -5,6 +5,7 @@
 
 use std::collections::HashMap;
 use std::fmt::{Debug, Formatter};
+use std::mem::swap;
 use std::os::unix::io::{AsRawFd, RawFd};
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, Condvar, Mutex};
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
Not doing this for now.

PR #1709, opened 2023-08-13
Run the MSRV CI check on more platforms

Thomasdezeeuw:
Since we have fairly complex dependencies now I think it's best to run the check on more platforms.


Thomasdezeeuw:
Note to self: also backport this to v0.7.

Darksonn (approved):

Thomasdezeeuw:
Not sure what is going with the linking on macOS, going to merge this and hope that it resolves itself.

Issue #1328, opened 2020-05-29
Solaris 11.4 does support SOCK_CLOEXEC and SOCK_NONBLOCK


EddaAkikusa:
https://github.com/tokio-rs/mio/blob/68805cbeb7797e0d2653e861e009cc10c05bd4f9/src/sys/unix/net.rs#L56
According to their document https://docs.oracle.com/cd/E88353_01/html/E37843/socket-3c.html
but still not sure about previous versions.

Thomasdezeeuw:
@Frozen-Tofu want to open a pr moving solaris from https://github.com/tokio-rs/mio/blob/68805cbeb7797e0d2653e861e009cc10c05bd4f9/src/sys/unix/net.rs#L56 to https://github.com/tokio-rs/mio/blob/68805cbeb7797e0d2653e861e009cc10c05bd4f9/src/sys/unix/net.rs#L26-L34?

Frozen-Tofu:
Yes, but I've to check whether Solaris 10, previous Solaris 11 and Illumos have implemented those flags.



发自我的iPhone


------------------ Original ------------------

Thomasdezeeuw:
Illumos now has a different target: #1295.

Thomasdezeeuw:
We don't support Solaris (see https://github.com/tokio-rs/mio/issues/1152), illumos should work.

PR #1715, opened 2023-09-01
Add Solaris operating system support (using flag mio_unsupported_forc…

psumbera:
…e_poll_poll)


Thomasdezeeuw:
Since solaris doesn't support any selector other than poll you don't need the `mio_unsupported_force_poll_poll` cfg.

But are problems listed in https://github.com/tokio-rs/mio/issues/1152 and https://github.com/tokio-rs/mio/issues/1328 (and https://github.com/tokio-rs/mio/pull/1528, https://github.com/tokio-rs/mio/pull/1542) resolved?

notgull:
Solaris and Illumos support a mechanism called [event ports](https://docs.oracle.com/cd/E36784_01/html/E36874/port-create-3c.html), which can be used to implement event listening in a more efficient way than `poll()`.

Thomasdezeeuw:
> Solaris and Illumos support a mechanism called [event ports](https://docs.oracle.com/cd/E36784_01/html/E36874/port-create-3c.html), which can be used to implement event listening in a more efficient way than `poll()`.

Illumos already uses epoll, so this will be Solaris specific.

psumbera:
> Since solaris doesn't support any selector other than poll you don't need the `mio_unsupported_force_poll_poll` cfg.

How is that possible? So far I had to use RUSTFLAGS="--cfg mio_unsupported_force_poll_poll". 
 
> But are problems listed in #1152 and #1328 (and #1528, #1542) resolved?

#1152 could be considered as resolved once this is merged in. Originally I was thinking about implementing it via Solaris event ports but it's not possible right now since we couln't check it using continous integration.

#1328 my commit uses SOCK_CLOEXEC and SOCK_NONBLOCK as it's for Illumos. Solaris 10 will never support Rust. It's very old and now with mininal support. It was released in 2004.

Not sure about others. I can modifiy README file and remove Solaris line. But mio_unsupported_force_poll_poll is mentined there as not really supported thing.

Thomasdezeeuw:
> > Since solaris doesn't support any selector other than poll you don't need the `mio_unsupported_force_poll_poll` cfg.
> 
> How is that possible? So far I had to use RUSTFLAGS="--cfg mio_unsupported_force_poll_poll".

By adding the correct `cfg` attributes for Solaris.

> > But are problems listed in #1152 and #1328 (and #1528, #1542) resolved?
> 
> #1152 could be considered as resolved once this is merged in. Originally I was thinking about implementing it via Solaris event ports but it's not possible right now since we couln't check it using continous integration.

:+1: 

> #1328 my commit uses SOCK_CLOEXEC and SOCK_NONBLOCK as it's for Illumos. Solaris 10 will never support Rust. It's very old and now with mininal support. It was released in 2004.

I'm not familiar with Solaris, but do newer version have support for `SOCK_CLOEXEC` and `SOCK_NONBLOCK`?

> Not sure about others. I can modifiy README file and remove Solaris line.

Let's leave it for now, I want to make sure it actually works this time.

> But mio_unsupported_force_poll_poll is mentined there as not really supported thing.

They key word in `mio_unsupported_force_poll_poll`  is *force*, it overwrites the preferred polling method on the platform to use a `poll(2)` based implementation instead. This is mainly used for testing and thus the flag itself unsupported. However  the `poll(2)` implementation is fully supported on platforms that don't have anything "better", for example ESP-IDF. We can add Solaris to this list, even though it has event ports.

psumbera:
> > > Since solaris doesn't support any selector other than poll you don't need the `mio_unsupported_force_poll_poll` cfg.
> > 
> > 
> > How is that possible? So far I had to use RUSTFLAGS="--cfg mio_unsupported_force_poll_poll".
> 
> By adding the correct `cfg` attributes for Solaris.

Sorry! I think I'm lost. What and where should I put anything? Thanks!
 
> > > But are problems listed in #1152 and #1328 (and #1528, #1542) resolved?

> > #1328 my commit uses SOCK_CLOEXEC and SOCK_NONBLOCK as it's for Illumos. Solaris 10 will never support Rust. It's very old and now with mininal support. It was released in 2004.
> 
> I'm not familiar with Solaris, but do newer version have support for `SOCK_CLOEXEC` and `SOCK_NONBLOCK`?

Yes, all version of Solaris 11.4 (first release in 2018) do support them.

> > But mio_unsupported_force_poll_poll is mentined there as not really supported thing.

Thomasdezeeuw:
> > By adding the correct `cfg` attributes for Solaris.
> 
> Sorry! I think I'm lost. What and where should I put anything? Thanks!

Try compiling without `RUSTFLAGS="--cfg mio_unsupported_force_poll_poll"` and rustc will tell you ;)

> > I'm not familiar with Solaris, but do newer version have support for `SOCK_CLOEXEC` and `SOCK_NONBLOCK`?
> 
> Yes, all versions of Solaris 11.4 (first release in 2018) do support them.

:+1:

psumbera:
> > > By adding the correct `cfg` attributes for Solaris.
> > 
> > 
> > Sorry! I think I'm lost. What and where should I put anything? Thanks!
> 
> Try compiling without `RUSTFLAGS="--cfg mio_unsupported_force_poll_poll"` and rustc will tell you ;)

No it doesn't help:

```
$ RUSTFLAGS="--cfg mio_unsupported_force_poll_poll" cargo build
   Compiling libc v0.2.147
   Compiling log v0.4.20
   Compiling mio v0.8.8 (/builds/psumbera/mio-solaris)
    Finished dev [unoptimized + debuginfo] target(s) in 6.07s
```

psumbera:
> Try enabling all features `--all-features`. And as I said don't use `RUSTFLAGS="--cfg mio_unsupported_force_poll_poll"`. So just run `cargo check --all-features` on Solaris and it should fail. But still have no issue how to define 

Sorry about with[out] RUSTFLAGS. It fails like this now. But still have no clue how to define `mio_unsupported_force_poll_poll` `cfg` atribute (ideally in `Config.toml` or `src/macros.rs`).
```
cargo check --all-features
    Checking mio v0.8.8 (/builds/psumbera/mio-solaris)
error[E0432]: unresolved imports `self::selector::event`, `self::selector::Event`, `self::selector::Events`, `self::selector::Selector`
  --> src/sys/unix/mod.rs:18:37
   |
18 |     pub(crate) use self::selector::{event, Event, Events, Selector};
   |                                     ^^^^^  ^^^^^  ^^^^^^  ^^^^^^^^ no `Selector` in `sys::unix::selector`
   |                                     |      |      |
   |                                     |      |      no `Events` in `sys::unix::selector`
   |                                     |      no `Event` in `sys::unix::selector`
   |                                     no `event` in `sys::unix::selector`
   |
   = help: consider importing this module instead:
           crate::event
   = help: consider importing this struct instead:
           crate::event::Event
   = help: consider importing this struct instead:
           crate::Events

warning: unused macro definition: `debug_detail`
  --> src/sys/mod.rs:18:18
   |
18 |     macro_rules! debug_detail {
   |                  ^^^^^^^^^^^^
   |
   = note: `#[warn(unused_macros)]` on by default

For more information about this error, try `rustc --explain E0432`.
warning: `mio` (lib) generated 1 warning
error: could not compile `mio` (lib) due to previous error; 1 warning emitted
```

Thomasdezeeuw:
Well... you need to resolve those issues to ensure that Solaris works...

psumbera:
Ok, with my new change Solaris builds without need to define `mio_unsupported_force_poll_poll`.

Thomasdezeeuw:
@psumbera can you add some CI setup? `cargo test` would be best, but `cargo check` is sufficient for now. Also see https://github.com/tokio-rs/mio/blob/master/Makefile#L2.

psumbera:
> @psumbera can you add some CI setup? `cargo test` would be best, but `cargo check` is sufficient for now.

Confused again. Where? `.github/workflows/ci.yml`? Any example for that?
`cargo test` passes for me. But there is no Solaris instance which is accesible from github for CI.

> Also see https://github.com/tokio-rs/mio/blob/master/Makefile#L2.

Solaris isn't supported by `rustup` right now. I think there is no need to extend the list.

Thomasdezeeuw:
> > @psumbera can you add some CI setup? `cargo test` would be best, but `cargo check` is sufficient for now.
> 
> Confused again. Where? `.github/workflows/ci.yml`? Any example for that? `cargo test` passes for me. But there is no Solaris instance which is accesible from github for CI.

... I mean come on... At some point you have put the time in. It's really not that complicated once you find the GitHub Actions file.

> > Also see https://github.com/tokio-rs/mio/blob/master/Makefile#L2.
> 
> Solaris isn't supported by `rustup` right now. I think there is no need to extend the list.

Do you have any other means to add Solaris support to the CI?

src/sys/unix/waker.rs
```diff
@@ -1,5 +1,5 @@
 #[cfg(all(
-    not(mio_unsupported_force_poll_poll),
+    not(any(target_os = "solaris", mio_unsupported_force_poll_poll)),
```
Thomasdezeeuw, line 2: Not needed.

src/sys/unix/waker.rs
```diff
@@ -13,7 +13,7 @@
 ))]
 mod fdbased {
     #[cfg(all(
-        not(mio_unsupported_force_waker_pipe),
+        not(any(target_os = "solaris", mio_unsupported_force_waker_pipe)),
```
Thomasdezeeuw, line 16: Not needed.

src/sys/unix/waker.rs
```diff
@@ -51,7 +52,7 @@ mod fdbased {
 }
 
 #[cfg(all(
-    not(mio_unsupported_force_poll_poll),
+    not(any(target_os = "solaris", mio_unsupported_force_poll_poll)),
```
Thomasdezeeuw, line 55: Not needed.

src/sys/unix/waker.rs
```diff
@@ -112,7 +113,7 @@ mod eventfd {
             }
         }
 
-        #[cfg(mio_unsupported_force_poll_poll)]
+        #[cfg(any(target_os = "solaris", mio_unsupported_force_poll_poll))]
```
Thomasdezeeuw, line 116: I don't think this is needed, but not 100% sure.

src/sys/unix/waker.rs
```diff
@@ -138,7 +139,7 @@ mod eventfd {
 }
 
 #[cfg(all(
-    mio_unsupported_force_poll_poll,
+    any(target_os = "solaris", mio_unsupported_force_poll_poll),
```
Thomasdezeeuw, line 142: Not needed.

src/sys/unix/waker.rs
```diff
@@ -276,7 +278,7 @@ mod pipe {
...
```
Thomasdezeeuw, line 281: Not need, you already added it below.

src/sys/unix/waker.rs
psumbera: I got following error without this change:
````
error: field `waker` is never read
  --> src/sys/unix/waker.rs:38:9
   |
37 |     pub struct Waker {
   |                ----- field in this struct
38 |         waker: WakerInternal,
   |         ^^^^^
   |
   = note: `Waker` has a derived impl for the trait `Debug`, but this is intentionally ignored during dead code analysis
  --> src/lib.rs:6:5
   |
6  |     dead_code
   |     ^^^^^^^^^

error: associated items `new` and `wake` are never used
  --> src/sys/unix/waker.rs:42:16
   |
41 |     impl Waker {
   |     ---------- associated items in this implementation
42 |         pub fn new(selector: &Selector, token: Token) -> io::Result<Waker> {
   |                ^^^
...
48 |         pub fn wake(&self) -> io::Result<()> {
   |                ^^^^
```

src/sys/unix/waker.rs
psumbera: I got following error without this change:
```
error[E0432]: unresolved imports `self::fdbased`, `self::waker::Waker`
  --> src/sys/unix/mod.rs:24:20
   |
24 |     pub(crate) use self::waker::Waker;
   |                    ^^^^^^^^^^^^^^^^^^
   |
  ::: src/sys/unix/waker.rs:66:15
   |
66 | pub use self::fdbased::Waker;
   |               ^^^^^^^ could not find `fdbased` in `self`

```

src/sys/unix/waker.rs
psumbera: I got following errir without this change:
```
error[E0432]: unresolved import `crate::sys::unix::waker::WakerInternal`
 --> src/sys/unix/selector/poll.rs:7:5
  |
7 | use crate::sys::unix::waker::WakerInternal;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `WakerInternal` in `sys::unix::waker`

error: unused import: `AsRawFd`
  --> src/sys/unix/selector/poll.rs:11:25
   |
11 | use std::os::unix::io::{AsRawFd, RawFd};
   |                         ^^^^^^^
   |
  --> src/lib.rs:5:5
   |
5  |     unused_imports,
   |     ^^^^^^^^^^^^^^
```

src/sys/unix/waker.rs
psumbera: I think it uses `pub use self::poll::Waker;` and thus cannot use `pub use self::fdbased::Waker`.

psumbera:
> > > @psumbera can you add some CI setup? `cargo test` would be best, but `cargo check` is sufficient for now.
> > 
> > 
> > Confused again. Where? `.github/workflows/ci.yml`? Any example for that? `cargo test` passes for me. But there is no Solaris instance which is accesible from github for CI.
> 
> ... I mean come on... At some point you have put the time in. It's really not that complicated once you find the GitHub Actions file.

Still confused. If you mean '.cirrus.yml'. Cirrus CI doesn't support Solaris OS.
 
> > > Also see https://github.com/tokio-rs/mio/blob/master/Makefile#L2.
> > 
> > Solaris isn't supported by `rustup` right now. I think there is no need to extend the list.
> 
> Do you have any other means to add Solaris support to the CI?

No.

psumbera:
@Thomasdezeeuw do you have any other commets or suggestions? Thank you!

Thomasdezeeuw (commented):
> @Thomasdezeeuw do you have any other commets or suggestions? Thank you!

Ideally we can setup some kind of CI, but that doesn't seem likely. 

Also some of the comments I made still need to be resolved, I see you answered them, but I think the warnings should we resolved in another way.

src/sys/unix/waker.rs
psumbera: @Thomasdezeeuw How can possibly following attribute be `false` on Solaris?
```
#[cfg(all(
    not(mio_unsupported_force_poll_poll),
    not(all(
        not(mio_unsupported_force_waker_pipe),
        any(
            target_os = "freebsd",
            target_os = "ios",
            target_os = "macos",
            target_os = "tvos",
            target_os = "watchos",
        )
    ))
))]
```

Thomasdezeeuw:
I need to clean up the `cfg` a bit soem that Solaris (and vita, see https://github.com/tokio-rs/mio/pull/1721) can use the `epoll(2)` implementation without any RUSTFLAGS.

Thomasdezeeuw:
@psumbera can you rebase this? I think after we merged https://github.com/tokio-rs/mio/pull/1721 we got a reasonable way forward. I'll like Solaris to be supported in a similar way, that is without any flags.

Thomasdezeeuw:
@psumbera did you mean to close this?

psumbera:
Ok. I wasn't able to reuse this pull request. Instead I have to create #1724.

PR #1720, opened 2023-10-15
Add support for backlog for bind

rustafarian-dev:


Thomasdezeeuw:
Mio follows the same API as the standard library, so I'm afraid this won't be accepted.

PR #1721, opened 2023-10-18
Added support for vita target

nikarh:
With these changes mio compiles and (mostly) works with `armv7-sony-vita-newlibeabihf` target with `--cfg mio_unsupported_force_poll_poll --cfg mio_unsupported_force_waker_pipe` rust flags.

`armv7-sony-vita-newlibeabihf` is a Tier 3 target with working std via newlib.

The target has no `epoll` or `kqueue`, only `poll` and `select` and async pipes, so cfg flags are required for compilation.
This is not an issue for devs using the target, since most if not all who work with it use `cargo-vita` tool for building, and it automatically adds these flags to `RUSTFLAGS`. 

- Like `espidf` it has no process API and thus no `CLOEXEC`
- `sockaddr_in` is mildly different on this target, and has an additional field 
- no `ioctl`, so `set_nonblocking` for vita follows illumos
- async pipes are implemented with `pipe2` call for this target

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,6 +133,22 @@ jobs:
       run: cargo install --debug cargo-hack
     - name: Check all targets
       run: make check_all_targets
+  CheckTier3Targets:
+    name: Check
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        target: ["armv7-sony-vita-newlibeabihf"]
+    steps:
+    - uses: actions/checkout@v3
+    - uses: dtolnay/rust-toolchain@nightly
+      with:
+        components: rust-src
+    - uses: taiki-e/install-action@cargo-hack
+    - name: Run check
+      run: cargo hack check -Z build-std=std,panic_abort --feature-powerset --target ${{ matrix.target }}
   Sanitizer:
     runs-on: ubuntu-latest
     timeout-minutes: 10
@@ -160,5 +176,6 @@ jobs:
       - Docs
       - Rustfmt
       - CheckTargets
+      - CheckTier3Targets
     steps:
       - run: exit 0
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ net = []
 log = { version = "0.4.8", optional = true }
 
 [target.'cfg(unix)'.dependencies]
-libc = "0.2.121"
+libc = "0.2.149"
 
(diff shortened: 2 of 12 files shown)


nikarh:
Hi!

> Can we add a `cargo check` to the CI here as well? I think the socket2 one can be copied.

I can, but it won't compile without these cfg flags right now, and I'm not sure they should be added to ci checks.
Should I change `#[cfg()]` in code and add vita target so that it compiles without RUSTFLAGS? Or wait for #1715? Or add CI but with the flags?

Thomasdezeeuw:
> I can, but it won't compile without these cfg flags right now, and I'm not sure they should be added to ci checks.

They were actually added to test the `poll(2)` implementation on the CI, but see below.

> Should I change `#[cfg()]` in code and add vita target so that it compiles without RUSTFLAGS? Or wait for #1715? Or add CI but with the flags?

If you can change them in this pr that would be great. I wanted to this myself in #1715, but I without rustup support for Solaris isn't quite a pain in the but to test.

nikarh:
Updated PR - added where necessary and a CI cargo check step.

Although for tokio we still need to set `mio_unsupported_force_poll_poll` - https://github.com/tokio-rs/tokio/blob/881b510a072f5acd773a10d3be0debf74113404e/tokio/src/io/poll_evented.rs#L171

Maybe this implementation detail can somehow be exposed? (not in the scope of this PR of course)

Thomasdezeeuw:
> Updated PR - added `target_os = "vita"` to cfg where necessary, and a CI cargo check step.

:+1: 

> Although for tokio we still need to set `mio_unsupported_force_poll_poll` - https://github.com/tokio-rs/tokio/blob/881b510a072f5acd773a10d3be0debf74113404e/tokio/src/io/poll_evented.rs#L171
> 
> Maybe this implementation detail can somehow be exposed? (not in the scope of this PR of course)

This is technically Tokio depending on unspecific behaviour (see https://github.com/tokio-rs/tokio/issues/5866#issuecomment-1639940869). Let's try resolve that in https://github.com/tokio-rs/tokio/issues/5866.

Thomasdezeeuw (approved):
LGTM, can't test, but the CI seems happy.

Thomasdezeeuw:
Thanks @nikarh

nikarh:
Hey, sorry to bother you @Thomasdezeeuw, would it be possible to make a release with these changes? 
This would help with tokio-rs/tokio/pull/6094

Thanks!

PR #1724, opened 2023-11-06
Add Solaris operating system support

psumbera:


psumbera:
@Thomasdezeeuw can you please look here?

Thomasdezeeuw (approved):
Can you confirm that it now works without `RUSTFLAGS`?

Also could you add Solaris to https://github.com/tokio-rs/mio/blob/e80c3b21b59b92238f4c8c331fcfc4a71508d1c5/.github/workflows/ci.yml#L155? Then we can make sure it builds.

psumbera:
I can confirm that `cargo test --features="log,os-poll,os-ext,net"` passes on Solaris 11.4.

psumbera:
> @psumbera can you fix the CI?

Not really sure... I can reproduce it on Linux. But now I don't know what to do...

Why it's failing for `armv7-sony-vita-newlibeabihf` even without my changes?

Thomasdezeeuw:
Maybe try (I don't have my regular dev setup atm, so can't test)

```patch
diff --git a/src/sys/unix/mod.rs b/src/sys/unix/mod.rs
index 7804236..28e0bea 100644
--- a/src/sys/unix/mod.rs
+++ b/src/sys/unix/mod.rs
@@ -17,7 +17,9 @@ cfg_os_poll! {
     mod selector;
     pub(crate) use self::selector::{event, Event, Events, Selector};
 
+    #[cfg(feature = "os-ext")]
     mod sourcefd;
+    #[cfg(feature = "os-ext")]
     pub use self::sourcefd::SourceFd;
 
     mod waker;
```

psumbera:
Now I'm getting (with  `--features os-poll`) following. Not sure what to do with it.
```
error: glob import doesn't reexport anything because no candidate is public enough
  --> src/sys/mod.rs:57:13
   |
57 |     pub use self::unix::*;
   |             ^^^^^^^^^^^^^
   |
```

psumbera:
> Now I'm getting (with `--features os-poll`) following. Not sure what to do with it.
> 
> ```
> error: glob import doesn't reexport anything because no candidate is public enough

I think this might be known Rust issue: https://github.com/rust-lang/rust/issues/115966 . Adding `#[warn(unused_imports)]` seems to resolve the issue for now.

Though not sure what is status of CI now?

psumbera:
> Though not sure what is status of CI now?

I see now new results. Thank you! The only failure is `MSRV (ubuntu-latest)`. Have no idea what is wrong there...

src/sys/mod.rs
```diff
@@ -54,6 +54,7 @@ cfg_os_poll! {
 #[cfg(unix)]
 cfg_os_poll! {
     mod unix;
+    #[warn(unused_imports)]
```
Thomasdezeeuw, line 57: This doesn't make sense. Did you mean to disable the lint? That would be the following, but really we shouldn't need it at all.

```suggestion
    #[allow(unused_imports)]
```

psumbera: Yes, `#[allow(unused_imports)]` does make sense. Fixed.

psumbera: @Thomasdezeeuw can you please merge it now?

Thomasdezeeuw:
Thanks @psumbera

PR #1725, opened 2023-11-17
Update `windows-sys` 0.52

kennykerr:
This is the first update to the `windows-sys` crate in 8 months.  https://github.com/microsoft/windows-rs/releases/tag/0.52.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ log = { version = "0.4.8", optional = true }
 libc = "0.2.149"
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = "0.48"
+version = "0.52"
 features = [
   "Win32_Foundation",                 # Basic types eg HANDLE
   "Win32_Networking_WinSock",         # winsock2 types/functions


Thomasdezeeuw:
@kennykerr it seems it contains breaking changes

kennykerr:
Yes, there were some WDK changes. I can fix those but I just noticed that `mio` hasn't updated MSRV so it may not be practical just yet. Much like `tokio`, the `windows-sys` crate now expects 1.56 to avoid a number of compatibility issues.

Thomasdezeeuw:
We'll have to discuss the MSRV, as we need to support Tokio TLS, which has, or had, a MSRV of 1.46

kennykerr:
I pushed the WDK changes. The `mio` crate uses some early support for the WDK and `windows-sys` 0.52 now has significantly improved WDK support.

Anyway, I'll probably just close this PR and you can revisit when you're ready to bump your MSRV.

kennykerr:
@Thomasdezeeuw just FYI the `NtCancelIoFileEx` function is provided by `windows-sys` 0.52:

https://github.com/tokio-rs/mio/blob/f6a20da1c81c2d56a78bc6f6832b9904b9215914/src/sys/windows/afd.rs#L21-L28

kayabaNerve:
Is there a location msrv is actively being discussed? I'd love to see this change be effected.

Thomasdezeeuw:
> Is there a location msrv is actively being discussed? I'd love to see this change be effected.

Not actively any more. The problem is Tokio LTS, which doesn't support windows-sys's MSRV. If I recall correctly it's supported until March 2024 after which we can update to windows-sys v0.52's MSRV.

Some related discussions:
 * https://github.com/tokio-rs/mio/pull/1688
 * https://github.com/tokio-rs/mio/issues/1632

kayabaNerve:
Hm.

This PR is purely import-based, with no logic/actual type changes.

Would a PR be accepted in the form of a build script which configures which windows-sys to pull in? 0.52 if >= Rust 1.56, else 0.48? That'd prevent de-duplication of windows-sys in users' trees while also not being a breaking change.

Thomasdezeeuw:
> Hm.
> 
> This PR is purely import-based, with no logic/actual type changes.

No, v0.48 -> v0.52 has breaking changes, see the failing CI: https://github.com/tokio-rs/mio/actions/runs/6905918873.

> Would a PR be accepted in the form of a build script which configures which windows-sys to pull in? 0.52 if >= Rust 1.56, else 0.48? That'd prevent de-duplication of windows-sys in users' trees while also not being a breaking change.

I don't really want to complicate Mio further.


I've started a thread on Discord to discuss this further: https://discord.com/channels/500028886025895936/627986131702644744/1181530156528041984.

kayabaNerve:
Got it. Thanks for all the answers.

PR #1726, opened 2023-12-04
Mentioned TcpStream::take error on TcpStream::connect documentation.

Andrepuel:
As discussed on #1565 .


src/net/tcp/stream.rs
```diff
@@ -68,12 +68,15 @@ impl TcpStream {
     ///  1. Call `TcpStream::connect`
     ///  2. Register the returned stream with at least [write interest].
     ///  3. Wait for a (writable) event.
-    ///  4. Check `TcpStream::peer_addr`. If it returns `libc::EINPROGRESS` or
+    ///  4. Check `TcpStream::take_error`. If it returns an error on the `Err`
+    ///  or in the `Ok(Some)` variants something went wrong. If it returns
+    ///  `Ok(None)`, then proceed to step 5.
```
Thomasdezeeuw, line 73: We don't have to explain the `Result` variants, we assume readers know about them.

```suggestion
    ///  4. Check `TcpStream::take_error`. If it returns an error something
    ///     went wrong. If it returns `Ok(None)`, then proceed to step 5.
```

You might need to format this.

Thomasdezeeuw (approved):
Small nit, but otherwise LGTM.

PR #1729, opened 2023-12-06
poll: Fix old comment not relevant after a refactor in the original PR.

### LINKED ISSUE #1728: WakerRegistrar mentioned in comment is not present
https://github.com/tokio-rs/mio/blame/f6a20da1c81c2d56a78bc6f6832b9904b9215914/src/sys/unix/selector/poll.rs#L267

```
                        // Remove the interest which just got triggered
                        // the IoSourceState/WakerRegistrar used with this selector will add back
                        // the interest using reregister.
                        poll_fd.events &= !poll_fd.revents;
```

when I use `mio` within https://github.com/rust-lang/rustup on Solaris it does not reregister (and `rustup` doesn't download stuff)

Looking at `IoSourceState` (mentioned in the comment as well) it does reregistration just in case of error!?

https://github.com/tokio-rs/mio/blob/master/src/sys/unix/selector/poll.rs#L663

@jasta, where is `WakerRegistrar` and how is `reregistration` supposed to work?

jasta:
Closes #1728

### DIFF
--- a/src/sys/unix/selector/poll.rs
+++ b/src/sys/unix/selector/poll.rs
@@ -263,9 +263,9 @@ impl SelectorState {
                             closed_raw_fds.push(poll_fd.fd);
                         }
 
-                        // Remove the interest which just got triggered
-                        // the IoSourceState/WakerRegistrar used with this selector will add back
-                        // the interest using reregister.
+                        // Remove the interest which just got triggered the IoSourceState's do_io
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @jasta

PR #1742, opened 2023-12-11
Add more CI targets

Thomasdezeeuw:
Adds:
 * iOS
 * tvOS
 * watchOS
 * OpenBSD
 * DragonflyBSD
 * NetBSD

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,7 +142,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: ["armv7-sony-vita-newlibeabihf", "sparcv9-sun-solaris", "x86_64-pc-solaris"]
+        target: ["aarch64-apple-ios", "aarch64-apple-tvos", "aarch64-unknown-openbsd", "arm64_32-apple-watchos", "armv7-sony-vita-newlibeabihf", "sparcv9-sun-solaris", "x86_64-pc-solaris", "x86_64-unknown-dragonfly", "x86_64-unknown-openbsd"]
     steps:
     - uses: actions/checkout@v3
     - uses: dtolnay/rust-toolchain@nightly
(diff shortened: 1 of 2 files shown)


asomers:
Within libc, the various Apple targets (ios, macos, tvos and watchos) differed from each other in exactly three symbols: `clock_settime`, `memmem`, and `task_set_info`.  So unless mio uses one of those three, it's a waste of resources to do CI for all four targets separately.  And even if mio does use all of them, it's still a waste of resources to test ios separately from tvos.

Thomasdezeeuw:
> Within libc, the various Apple targets (ios, macos, tvos and watchos) differ from each other in exactly three symbols: `clock_settime`, `memmem`, and `task_set_info`.

It's not really about the difference in the OS, more so in the fact that the cfg attribute for vendors has been soft-deprecated and now we need to use the OSs separately, e.g. see https://github.com/tokio-rs/mio/blob/master/src/sys/unix/selector/kqueue.rs#L23-L28. This is quite error prone and something I want to check.

> So unless mio uses one of those three, it's a waste of resources to do CI for all four targets separately. And even if mio does use all of them, it's still a waste of resources to test ios separately from tvos.

That's a fair point.

Maybe we can have some CI that only run when we're about to make a release or something.

Darksonn (approved):

asomers:
> > Within libc, the various Apple targets (ios, macos, tvos and watchos) differ from each other in exactly three symbols: `clock_settime`, `memmem`, and `task_set_info`.
> 
> It's not really about the difference in the OS, more so in the fact that the cfg attribute for vendors has been soft-deprecated and now we need to use the OSs separately, e.g. see https://github.com/tokio-rs/mio/blob/master/src/sys/unix/selector/kqueue.rs#L23-L28. This is quite error prone and something I want to check.

The Nix project dealt with that by defining our own config aliases.  See https://github.com/nix-rust/nix/blob/88c73492f616f15fd4135849f0a692f6a9651071/build.rs#L18 .

Thomasdezeeuw:
> The Nix project dealt with that by defining our own config aliases. See https://github.com/nix-rust/nix/blob/88c73492f616f15fd4135849f0a692f6a9651071/build.rs#L18 .

I don't really want to add dependencies and looking at the the crate itself it's quite a bit of code so I'm not sure doing it ourselves is a good idea either.

I would have been best if they could have kept the vendor attribute, or introduce something similar.

PR #1756, opened 2024-01-25
Add do_io on peek for TcpStream

### LINKED ISSUE #1755: peek blocks after read on windows
Hi! I wanted to continue the issue of https://github.com/tokio-rs/tokio/issues/3789

I used the following code as reproduction:
```rust 
use tokio::io::AsyncReadExt;

#[tokio::main]
async fn main() -> std::io::Result<()> {
    let listener = tokio::net::TcpListener::bind("0.0.0.0:2222").await?;

    loop {
        let (mut socket, _) = listener.accept().await?;

        tokio::spawn(async move {
            let mut buf = [0; 1];
            let _ = socket.read(&mut buf).await;
            let _ = socket.peek(&mut buf).await;
        });
    }
}
```

So I don't know much about Tokio internals, but I spent the last couple of hours debugging.

I found that the waker for the `Readiness` future used by peek is never called after there was one call to read. Digging deeper the `GetQueuedCompletionStatusEx` is blocking when issuing the peek after read.

I think this is because Windows doesn't know we are interested by `READABLE` because peek never registers the fact that it has the `READABLE` interest at the OS level. It does register the interest at the `ScheduledIO` level but we never get there obviously because the thread is parked.

When we issue a `read` we use the `IoSourceState` which `reregisters` our interests with the IOCP if it would block, but when we peek we directly use the STD IO. See: 
https://github.com/tokio-rs/mio/blob/08ee541bf00c2cc81f9acc5486851ca1d59c25a3/src/net/tcp/stream.rs#L210-L212

I think the fix is to use  `IoSourceState::do_io` when we peek, but I must test it.

Sytten:
Fixes #1755 
I am not sure of the implications of this change for all platforms, but it fixes the issue on windows. I believe it should be fine on unix too, but more tests are needed.

### DIFF
--- a/src/net/tcp/stream.rs
+++ b/src/net/tcp/stream.rs
@@ -208,7 +208,7 @@ impl TcpStream {
     /// Successive calls return the same data. This is accomplished by passing
     /// `MSG_PEEK` as a flag to the underlying recv system call.
     pub fn peek(&self, buf: &mut [u8]) -> io::Result<usize> {
-        self.inner.peek(buf)
+        self.inner.do_io(|inner| inner.peek(buf))
     }
 
     /// Execute an I/O operation ensuring that the socket receives more events


Sytten:
FYI its a potential security issue that cirrus ci runs without any approval, probably should restrict that...

Thomasdezeeuw (commented):
This is incorrect. Peeking shouldn't reset our readiness internally as more data can be read (when not peeking).

Thomasdezeeuw:
> FYI its a potential security issue that cirrus ci runs without any approval, probably should restrict that...

We have no secrets, so it's not really.

Sytten:
> This is incorrect. Peeking shouldn't reset our readiness internally as more data can be read (when not peeking).

I am not sure I understand what that means. A peek does currently reset internal readiness if it would would block which is fine since we need to wait for more data to come in. A read also reset internal readiness. If you don't reregister the interest with the OS when that happens (which is essentially what do_io does) then I am not sure how we expect this to work at all. I am a bit surprised that Unix currently work, seems more like an accident than planned behaviour to me but I didnt check the code much.

Thomasdezeeuw:
If a peek call returns a short read (peek) on Unix it will trigger our internal registering mechanism, but it shouldn't.

Sytten:
Does internal mean the readiness stored in ScheduledIO or the libc/iocp calls?

Each read does a reregister if it would block, so I fail to understand why that would be a bad thing for a peek to do it too. But I probably don't understand the implications of reregister, the name suggest it is idempotent.

Now moving forward if this is not the solution what is? Is it to only reregister on windows? Is it to not clear readiness on WouldBlock (that seems wrong)?

Sytten:
PR is incorrect closing

PR #1759, opened 2024-02-21
add support of HermitOS

stlankes:
[HermitOS](http://hermit-os.org/) is a library operating system. The handling of the socket streams is compatible to unix. However, other parts are not compatible to Unix. Consequently, HermitOS doesn't belong to the unix family.

To support MIO, MIO's unix is revised to support HermitOS. Currently, the implementation based on the unix system `poll`.

### DIFF
--- a/src/net/tcp/stream.rs
+++ b/src/net/tcp/stream.rs
@@ -1,6 +1,8 @@
 use std::fmt;
 use std::io::{self, IoSlice, IoSliceMut, Read, Write};
 use std::net::{self, Shutdown, SocketAddr};
+#[cfg(target_os = "hermit")]
+use std::os::hermit::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
 #[cfg(unix)]
 use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
 #[cfg(target_os = "wasi")]
@@ -87,6 +89,8 @@ impl TcpStream {
     pub fn connect(addr: SocketAddr) -> io::Result<TcpStream> {
         let socket = new_for_addr(addr)?;
         #[cfg(unix)]
+        let stream = unsafe { TcpStream::from_raw_fd(socket) };
+        #[cfg(target_os = "hermit")]
         let stream = unsafe { TcpStream::from_raw_fd(socket) };
         #[cfg(windows)]
         let stream = unsafe { TcpStream::from_raw_socket(socket as _) };
@@ -348,21 +352,21 @@ impl fmt::Debug for TcpStream {
     }
 }
 
-#[cfg(unix)]
+#[cfg(any(unix, target_os = "hermit"))]
 impl IntoRawFd for TcpStream {
     fn into_raw_fd(self) -> RawFd {
         self.inner.into_inner().into_raw_fd()
     }
 }
 
-#[cfg(unix)]
+#[cfg(any(unix, target_os = "hermit"))]
 impl AsRawFd for TcpStream {
     fn as_raw_fd(&self) -> RawFd {
         self.inner.as_raw_fd()
     }
 }
 
-#[cfg(unix)]
+#[cfg(any(unix, target_os = "hermit"))]
 impl FromRawFd for TcpStream {
     /// Converts a `RawFd` to a `TcpStream`.
     ///
--- a/src/net/tcp/stream.rs
+++ b/src/net/tcp/stream.rs
@@ -88,9 +88,7 @@ impl TcpStream {
     #[cfg(not(target_os = "wasi"))]
     pub fn connect(addr: SocketAddr) -> io::Result<TcpStream> {
         let socket = new_for_addr(addr)?;
-        #[cfg(unix)]
-        let stream = unsafe { TcpStream::from_raw_fd(socket) };
-        #[cfg(target_os = "hermit")]
+        #[cfg(any(unix, target_os = "hermit"))]
         let stream = unsafe { TcpStream::from_raw_fd(socket) };
         #[cfg(windows)]
         let stream = unsafe { TcpStream::from_raw_socket(socket as _) };
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -51,7 +51,7 @@ cfg_os_poll! {
     }
 }
 
-#[cfg(unix)]
+#[cfg(any(unix, target_os = "hermit"))]
 cfg_os_poll! {
     mod unix;
     #[allow(unused_imports)]
@@ -68,12 +68,6 @@ cfg_os_poll! {
 cfg_os_poll! {
     mod wasi;
     pub(crate) use self::wasi::*;
-}
-
-#[cfg(target_os = "hermit")]
-cfg_os_poll! {
-    mod unix;
-    pub(crate) use self::unix::*;
 }
 
 cfg_not_os_poll! {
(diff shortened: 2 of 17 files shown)


stlankes:
We have also version, which based on the branch v0.8.x. Are you interested to integrate HermitOS in the version 0.8?

Thomasdezeeuw (commented):
We're not the biggest fan of dependency. Any chance we can do this with just the libc crate?

Thomasdezeeuw:
> We have also version, which based on the branch v0.8.x. Are you interested to integrate HermitOS in the version 0.8?

We can backport later.

stlankes:
Yes, I can understand that dependencies are critical. Our OS is written completely in Rust and a classical C library is missing. `hermit-abi` is also part of [std](https://github.com/rust-lang/rust/blob/master/library/std/Cargo.toml#L50-L51). Do you think that you can support `hermit-abi`?

By the way, we used [a revised version of your examples](https://github.com/hermit-os/hermit-rs/tree/main/examples/miotcp/src) in our [CI](https://github.com/hermit-os/kernel/actions/runs/7971482269/job/21761264009).

src/sys/mod.rs
```diff
@@ -70,6 +70,12 @@ cfg_os_poll! {
     pub(crate) use self::wasi::*;
 }
 
+#[cfg(target_os = "hermit")]
+cfg_os_poll! {
+    mod unix;
+    pub(crate) use self::unix::*;
+}
```
Thomasdezeeuw, line 77: Can you merge this with `cfg(unix)` on line 54?

src/sys/unix/mod.rs
```diff
@@ -5,6 +5,31 @@
...
+                libc::errno::ENOTCONN => std::io::ErrorKind::NotConnected,
+                libc::errno::EPERM => std::io::ErrorKind::PermissionDenied,
+                libc::errno::EPIPE => std::io::ErrorKind::BrokenPipe,
+                libc::errno::ETIMEDOUT => std::io::ErrorKind::TimedOut,
+                _ => panic!("Unknown error {}", res),
+            };
+            Err(std::io::Error::from(e))
+        } else {
+            Ok(res)
+        }
```
Thomasdezeeuw, line 31: Does `std::io::Error::last_os_error()` not work on HermitOS?

src/sys/unix/mod.rs
```diff
@@ -5,6 +5,31 @@
...
+                libc::errno::ECONNRESET => std::io::ErrorKind::ConnectionReset,
+                libc::errno::EEXIST => std::io::ErrorKind::AlreadyExists,
+                libc::errno::EINTR => std::io::ErrorKind::Interrupted,
+                libc::errno::EINVAL => std::io::ErrorKind::InvalidInput,
+                libc::errno::ENOENT => std::io::ErrorKind::NotFound,
+                libc::errno::ENOTCONN => std::io::ErrorKind::NotConnected,
+                libc::errno::EPERM => std::io::ErrorKind::PermissionDenied,
+                libc::errno::EPIPE => std::io::ErrorKind::BrokenPipe,
+                libc::errno::ETIMEDOUT => std::io::ErrorKind::TimedOut,
+                _ => panic!("Unknown error {}", res),
```
Thomasdezeeuw, line 26: This is a no-go.

Thomasdezeeuw:
> Yes, I can understand that dependencies are critical. Our OS is written completely in Rust and a classical C library is missing. `hermit-abi` is also part of [std](https://github.com/rust-lang/rust/blob/master/library/std/Cargo.toml#L50-L51). Do you think that you can support `hermit-abi`?

It was pointed out to me that Hermit does support the libc crate, e.g. https://github.com/rust-lang/libc/pull/3348. If that's the case we're not accepting hermit-abi as dependency.

mkroening:
> > Yes, I can understand that dependencies are critical. Our OS is written completely in Rust and a classical C library is missing. `hermit-abi` is also part of [std](https://github.com/rust-lang/rust/blob/master/library/std/Cargo.toml#L50-L51). Do you think that you can support `hermit-abi`?
> 
> It was pointed out to me that Hermit does support the libc crate, e.g. [rust-lang/libc#3348](https://github.com/rust-lang/libc/pull/3348). If that's the case we're not accepting hermit-abi as dependency.

The only code in `libc` for Hermit are the definitions of C integer types: https://github.com/rust-lang/libc/blob/main/src/hermit/mod.rs

Thomasdezeeuw:
After a bit of discussion we decided to accept the dependency and support of HermitOS as _experimental_. If you can address the three comments I made earlier we can merge this.

Thomasdezeeuw:
Thanks for making the changes, I'm traveling right now so I can't properly review this until next week.

stlankes:
The changes aren't finished. I will "ping" you, when I solve everything.

src/sys/unix/mod.rs
stlankes: Within `std` we determined the last error in a different way and ignore `std::io::Error::last_os_error()`. In the meantime we fixed it (see rust-lang/rust#121765). Consequently, the nightly compiler should support it. Should we use `last_os_error` or should we directly call our [abi](https://github.com/hermit-os/hermit-rs/blob/main/hermit-abi/src/lib.rs#L341-L342)?

src/sys/unix/mod.rs
stlankes: Yeah, I will remove it. The solution depends on the question above.

src/sys/unix/mod.rs
Thomasdezeeuw: That's a though one. Is it common for Hermit to run on Rust stable? Or are most users on nightly? If the latter is true I think it's ok to be on nightly-only for now.

stlankes: The kernel use some nightly feature and the user-space is able to use stable. I think we can use the nightly compiler. I will fix it.

src/sys/unix/mod.rs
stlankes: Is the new solution better (see latest commit)?

src/sys/unix/mod.rs
```diff
@@ -5,6 +5,13 @@
 macro_rules! syscall {
     ($fn: ident ( $($arg: expr),* $(,)* ) ) => {{
         let res = unsafe { libc::$fn($($arg, )*) };
+        #[cfg(target_os = "hermit")]
```
Thomasdezeeuw, line 8: Since this code is now the same as on Unix, let's merge the two blocks.

I think it's fine to change `res == -1` to `res < 0` if that is needed on Hermit.

stlankes: I merged both block together.

src/net/tcp/stream.rs
```diff
@@ -88,6 +90,8 @@ impl TcpStream {
         let socket = new_for_addr(addr)?;
         #[cfg(unix)]
         let stream = unsafe { TcpStream::from_raw_fd(socket) };
+        #[cfg(target_os = "hermit")]
+        let stream = unsafe { TcpStream::from_raw_fd(socket) };
```
Thomasdezeeuw, line 94: Nit: I think we can merge this with the Unix code as it's the same API, even though it's two different traits.

Thomasdezeeuw (approved):
LGTM.

Is this ready to merge?

src/net/tcp/stream.rs
stlankes: Yeah, I oversaw it

stlankes:
I think that the current version is ready to merge

PR #1760, opened 2024-02-29
named-pipes: fix receiving IOCP events after deregister

carllerche:
There is a race condition with named pipes where it is possible to receive events for a named pipe after it has been deregistered with the selector. This is because named pipes use IOCP, and there is a path where it posts a raw event, and by the time that raw event is returned by the selector the user, the user could have deregistered the I/O resource.

The fix is to post the event while maintaining the association with the I/O resource. Then, before returning the event to the user, do a final check to ensure the I/O resource (named pipe) is still registered with the selector.

Fixes [#6369](https://github.com/tokio-rs/tokio/issues/6369)

### DIFF
--- a/src/sys/windows/named_pipe.rs
+++ b/src/sys/windows/named_pipe.rs
@@ -1001,6 +1001,11 @@ impl Io {
     fn schedule_event(&self, me: &Arc<Inner>, mut event: Event) {
         // Alter the token so that the selector will identify the IOCP event as
         // one for a named pipe. This will be reversed in `event_done`
+        //
+        // `data` for named pipes is an auto-incrementing counter. Because
+        // `data` is `u64` we do not risk losing the most-significant bit
+        // (unless a user creates 2^62 named pipes during the lifetime of the
+        // process).
         event.data <<= 1;
         event.data += 1;
 
--- a/src/sys/windows/named_pipe.rs
+++ b/src/sys/windows/named_pipe.rs
@@ -84,6 +84,7 @@ struct Inner {
     connect: Overlapped,
     read: Overlapped,
     write: Overlapped,
+    event: Overlapped,
     // END NOTE.
     handle: Handle,
     connecting: AtomicBool,
@@ -110,8 +111,14 @@ impl Inner {
 
     /// Same as [`ptr_from_conn_overlapped`] but for `Inner.write`.
     unsafe fn ptr_from_write_overlapped(ptr: *mut OVERLAPPED) -> *const Inner {
-        // `read` is after `connect: Overlapped` and `read: Overlapped`.
+        // `write` is after `connect: Overlapped` and `read: Overlapped`.
         (ptr as *mut Overlapped).wrapping_sub(2) as *const Inner
+    }
+
+    /// Same as [`ptr_from_conn_overlapped`] but for `Inner.event`.
+    unsafe fn ptr_from_event_overlapped(ptr: *mut OVERLAPPED) -> *const Inner {
+        // `event` is after `connect: Overlapped`, `read: Overlapped`, and `write: Overlapped`.
+        (ptr as *mut Overlapped).wrapping_sub(3) as *const Inner
     }
 
     /// Issue a connection request with the specified overlapped operation.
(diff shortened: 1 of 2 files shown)


atouchet:
That issue number doesn't exist. Typo?

Thomasdezeeuw:
> That issue number doesn't exist. Typo?

It's a Tokio issue, I've added a link.

src/sys/windows/named_pipe.rs
```diff
@@ -938,28 +967,48 @@ impl Io {
...
+        event.data <<= 1;
+        event.data += 1;
+
+        let completion_status =
+            event.to_completion_status_with_overlapped(me.event.as_ptr() as *mut _);
+
+        match self.cp.as_ref().unwrap().post(completion_status) {
+            Ok(_) => {
+                // Increase the ref count of `Inner` for the completion event.
+                mem::forget(me.clone());
```
Thomasdezeeuw, line 1008: Nit: we could use `Arc::increment_strong_count`, but I don't know if we use it anywhere else, if not it's ok to keep the current code.

src/sys/windows/named_pipe.rs
```diff
@@ -938,28 +967,48 @@ impl Io {
...
+                self.schedule_event(me, ev);
+            }
+        }
+    }
+
+    fn schedule_event(&self, me: &Arc<Inner>, mut event: Event) {
+        // Alter the token so that the selector will identify the IOCP event as
+        // one for a named pipe. This will be reversed in `event_done`
+        event.data <<= 1;
+        event.data += 1;
```
Thomasdezeeuw, line 1000: I don't see this `+= 1` being undone in `event_done`.

src/sys/windows/named_pipe.rs
```diff
@@ -938,28 +967,48 @@ impl Io {
...
-                let _ = self.cp.as_ref().unwrap().post(ev.to_completion_status());
+                self.schedule_event(me, ev);
+            }
+        }
+    }
+
+    fn schedule_event(&self, me: &Arc<Inner>, mut event: Event) {
+        // Alter the token so that the selector will identify the IOCP event as
+        // one for a named pipe. This will be reversed in `event_done`
+        event.data <<= 1;
```
Thomasdezeeuw, line 999: Do we know we won't lose a bit doing this shift?

Thomasdezeeuw (approved):
A chance that https://github.com/tokio-rs/mio/issues/1755 is related?

LGTM, but I've very jet lagged at the moment, so not the best reviewer. @Darksonn you want to take a look at this as well if you can?

Thomasdezeeuw:
The minimal CI seems to fail with a proper error: `STATUS_STACK_BUFFER_OVERRUN`: https://github.com/tokio-rs/mio/actions/runs/8103773681/job/22149086129?pr=1760#step:5:274.

Thomasdezeeuw:
Oh, I just noticed this target v0.8.x branch, can you change it to master? Then we can back port afterwards, that's how we usually do it.

src/sys/windows/named_pipe.rs
carllerche: This is the style in the rest of the file. It probably should be updated in one go.

src/sys/windows/named_pipe.rs
carllerche: `>>` handles this by dropping the lowest bit.

carllerche:
I rebased against master.

src/sys/windows/named_pipe.rs
carllerche: I added an extra comment. It isn't an issue as `event.data` is `u64` and is set internally by mio for named pipes using an incrementing counter.

src/sys/windows/named_pipe.rs
Thomasdezeeuw: Right, I should not be reviewing with this little sleep.

Thomasdezeeuw (approved):

daira:
For tokio users: this is marked as fixing tokio-rs/tokio#6369 but I think it should not have been: https://github.com/tokio-rs/tokio/issues/6369#issuecomment-1978292853

Thomasdezeeuw:
For Tokio users, run `cargo update -p mio` and you're good.

PR #1765, opened 2024-03-13
doc: Fix link for `SocketAddr`

ldm0:


src/sys/unix/uds/socketaddr.rs
```diff
@@ -8,10 +8,11 @@ use std::{ascii, fmt};
 ///
 /// This is implemented instead of imported from [`net::SocketAddr`] because
 /// there is no way to create a [`net::SocketAddr`]. One must be returned by
-/// [`accept`], so this is returned instead.
+/// [`UnixListener::accept`] or [`TcpListener::accept`], so this is returned instead.
```
Thomasdezeeuw, line 11: TcpListener doesn't return this, only the UDS sockets do.

Thomasdezeeuw (changes requested):
This type is being removed in https://github.com/tokio-rs/mio/pull/1749.

PR #1766, opened 2024-03-22
Adding support for QNX OS

SebastianSchildt:
First try to enable mio to be used on QNX operating system.

Is this the right way to go about this? Once this is in acceptable form, can/should it be backported to one of the v0.8x/0.7x

We are not that familiar with mio development, so please advise.
As mio/tokio users we would appreciate if released versions just work "out of the box"

@AkhilTThomas I added you to the fork, so you can write there as well

### DIFF
--- a/src/sys/unix/net.rs
+++ b/src/sys/unix/net.rs
@@ -25,6 +25,8 @@ pub(crate) fn new_socket(domain: libc::c_int, socket_type: libc::c_int) -> io::R
         target_os = "hermit",
     ))]
     let socket_type = socket_type | libc::SOCK_NONBLOCK | libc::SOCK_CLOEXEC;
+    #[cfg(target_os = "nto")]
+    let socket_type = socket_type | libc::SOCK_CLOEXEC;
 
     let socket = syscall!(socket(domain, socket_type, 0))?;
 
@@ -61,7 +63,7 @@ pub(crate) fn new_socket(domain: libc::c_int, socket_type: libc::c_int) -> io::R
             let _ = syscall!(close(socket));
             return Err(err);
         }
-        #[cfg(not(any(target_os = "espidf", target_os = "vita")))]
+        #[cfg(not(any(target_os = "espidf", target_os = "vita", target_os = "nto")))]
         if let Err(err) = syscall!(fcntl(socket, libc::F_SETFD, libc::FD_CLOEXEC)) {
             let _ = syscall!(close(socket));
             return Err(err);
--- a/src/sys/unix/net.rs
+++ b/src/sys/unix/net.rs
@@ -54,6 +54,7 @@ pub(crate) fn new_socket(domain: libc::c_int, socket_type: libc::c_int) -> io::R
         target_os = "watchos",
         target_os = "espidf",
         target_os = "vita",
+        target_os = "nto",
     ))]
     {
         if let Err(err) = syscall!(fcntl(socket, libc::F_SETFL, libc::O_NONBLOCK)) {
@@ -117,6 +118,7 @@ pub(crate) fn socket_addr(addr: &SocketAddr) -> (SocketAddrCRepr, libc::socklen_
                     target_os = "espidf",
                     target_os = "vita",
                     target_os = "hermit",
+                    target_os = "nto",
                 ))]
                 sin_len: 0,
                 #[cfg(target_os = "vita")]
@@ -148,6 +150,7 @@ pub(crate) fn socket_addr(addr: &SocketAddr) -> (SocketAddrCRepr, libc::socklen_
                     target_os = "watchos",
                     target_os = "espidf",
                     target_os = "vita",
+                    target_os = "nto",
                 ))]
                 sin6_len: 0,
                 #[cfg(target_os = "vita")]
--- a/src/sys/unix/tcp.rs
+++ b/src/sys/unix/tcp.rs
@@ -115,6 +115,7 @@ pub(crate) fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream,
                 target_os = "espidf",
                 target_os = "vita",
                 target_os = "hermit",
+                target_os = "nto",
             ))]
             syscall!(fcntl(s.as_raw_fd(), libc::F_SETFL, libc::O_NONBLOCK))?;
 
--- a/src/sys/unix/tcp.rs
+++ b/src/sys/unix/tcp.rs
@@ -95,6 +95,7 @@ pub(crate) fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream,
         target_os = "espidf",
         target_os = "vita",
         target_os = "hermit",
+        target_os = "nto",
         all(target_arch = "x86", target_os = "android"),
     ))]
     let stream = {
--- a/src/sys/unix/uds/listener.rs
+++ b/src/sys/unix/uds/listener.rs
@@ -99,6 +99,7 @@ pub(crate) fn accept(listener: &net::UnixListener) -> io::Result<(UnixStream, So
             all(target_arch = "x86", target_os = "android"),
             target_os = "espidf",
             target_os = "vita",
+            target_os = "nto",
         ))]
         syscall!(fcntl(socket, libc::F_SETFL, libc::O_NONBLOCK))?;
 
--- a/src/sys/unix/uds/listener.rs
+++ b/src/sys/unix/uds/listener.rs
@@ -53,6 +53,7 @@ pub(crate) fn accept(listener: &net::UnixListener) -> io::Result<(UnixStream, So
         target_os = "watchos",
         target_os = "espidf",
         target_os = "vita",
+        target_os = "nto",
         // Android x86's seccomp profile forbids calls to `accept4(2)`
         // See https://github.com/tokio-rs/mio/issues/1445 for details
         all(target_arch = "x86", target_os = "android"),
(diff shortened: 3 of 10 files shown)


notgull:
Do you have any documentation of the `poll` syscall exposed by QNX?

SebastianSchildt:
I am not so deep into this, so maybe @AkhilTThomas has some hands-on experience, but in terms of documentation, this is public

https://www.qnx.com/developers/docs/7.1/index.html#com.qnx.doc.neutrino.lib_ref/topic/p/poll.html

Thomasdezeeuw:
Couple of questions:
* Can we add something to the CI, `cargo test` would be nice, `cargo check` is sufficient.
* Is annoying willing to commit to maintaining this with QNX OS knowledge? Otherwise support will not be official.

Backporting can be done after this.

src/sys/unix/net.rs
```diff
@@ -54,6 +54,7 @@ pub(crate) fn new_socket(domain: libc::c_int, socket_type: libc::c_int) -> io::R
         target_os = "watchos",
         target_os = "espidf",
         target_os = "vita",
+        target_os = "nto",
```
gh-tr, line 57: QNX supports the SOCK_CLOEXEC option, just not the SOCK_NONBLOCK option when creating a socket or a socketpair. https://www.qnx.com/developers/docs/7.1/#com.qnx.doc.neutrino.lib_ref/topic/s/socket.html.  

Like the code for uds/mod.rs, you can enable line 27 and SOCK_NONBLOCK should be ignored:
  let socket_type = socket_type | libc::SOCK_NONBLOCK | libc::SOCK_CLOEXEC;
  
Below, you can ignore the call to fcntl with FD_CLOEXEC on the socket (add qnx to the ignore list with espidf and vita) as it's already set on socket creation.

src/sys/unix/uds/mod.rs
```diff
@@ -103,6 +104,7 @@ cfg_os_poll! {
             target_os = "watchos",
             target_os = "espidf",
             target_os = "vita",
+            target_os = "nto",
         ))]
         {
             syscall!(fcntl(fds[0], libc::F_SETFL, libc::O_NONBLOCK))?;
```
gh-tr, line 110: SOCK_CLOEXEC is already supported as part of the socketpair() call.  Can skip the FD_CLOEXEC calls below.

gh-tr (changes requested):
Couple of comments inline surrounding SOCK_CLOEXEC.

SebastianSchildt:
Hi, thanks for the hints. We will  definitely  check, although currently at least personally I  am blocked not having a QNX license (anymore?).  (not sure if @AkhilTThomas  is better equipped currently). Working on that, but always a major pita. 

I used the time to check whether something can be done with cargo check, but i think it is -without QNX toolchain installed - not possible. Do you know any way get "some level of testing" for neutrino target in an OSS project @flba-eb, as you have ported some other libs (like compiling maybe not even linking, and not running)? Or is this basically impossible currently?

gh-tr:
> Hi, thanks for the hints. We will definitely check, although currently at least personally I am blocked not having a QNX license (anymore?). (not sure if @AkhilTThomas is better equipped currently). Working on that, but always a major pita.
> 
> I used the time to check whether something can be done with cargo check, but i think it is -without QNX toolchain installed - not possible. Do you know any way get "some level of testing" for neutrino target in an OSS project @flba-eb, as you have ported some other libs (like compiling maybe not even linking, and not running)? Or is this basically impossible currently?

@flba-eb has access to a QNX license and runs checks locally; the same way I would go about it.  We generally submit the test results of the crates as part of the request.  I guess I'm curious why you're making changes to QNX specific code without a license.  Perhaps ping me via email and we can have a quick discussion surrounding that and perhaps get you what you need?

tests/unix_pipe.rs
```diff
@@ -124,10 +126,10 @@ fn event_when_receiver_is_dropped() {
 }
 
 #[test]
+#[cfg_attr(target_os = "nto", ignore = "Read close events are detected by NTO")]
 fn from_child_process_io() {
     // `cat` simply echo everything that we write via standard in.
     let mut child = Command::new("cat")
-        .env_clear()
```
AkhilTThomas, line 130: Why it was removed ? -->  `env_clear` clears the PATH variable and the system was unable to find `cat`. Is this necessary for the test framework?

SebastianSchildt:
> 
> @flba-eb has access to a QNX license and runs checks locally; the same way I would go about it. We generally submit the test results of the crates as part of the request. I guess I'm curious why you're making changes to QNX specific code without a license. Perhaps ping me via email and we can have a quick discussion surrounding that and perhaps get you what you need?

Ah 2 different  things: This was initially done in my company a while back, and well... as these things are it took some time being allowed to share it here. So my _current_ lack of license is  more an internal IT issue

Generally, however, as QNX support in Rust matures, I think it will be a bit of an issue, that support can break anytime. Communities like this one, most members likely have neither any contact with QNX nor do they care - which is fine, but due to the closedness it won't be easy to find any breaks eraly. I think currently it would be like this: One of the users of this (i.e. us) one day will do a cargo update, and well, figure out something broken, then come back here, trying to fix it and get it released again. In the long run, it would really be good if in the CI here upstream, there would be some way to at least compile, preferably run tests on that target.  

I think we are not there yet, just thinking forward :)

flba-eb:
> Hi, thanks for the hints. We will definitely check, although currently at least personally I am blocked not having a QNX license (anymore?). (not sure if @AkhilTThomas is better equipped currently). Working on that, but always a major pita.
> 
> I used the time to check whether something can be done with cargo check, but i think it is -without QNX toolchain installed - not possible. Do you know any way get "some level of testing" for neutrino target in an OSS project @flba-eb, as you have ported some other libs (like compiling maybe not even linking, and not running)? Or is this basically impossible currently?

As @gh-tr  already noted, I have a license I can use locally. Unfortunately `cargo check` is the best we can do right now, this is also what we're using in other crates.

tests/unix_pipe.rs
Thomasdezeeuw: I think it's fine.

src/sys/unix/uds/mod.rs
```diff
@@ -103,10 +106,11 @@ cfg_os_poll! {
...
-            #[cfg(not(any(target_os = "espidf", target_os = "vita")))]
+            #[cfg(not(any(target_os = "espidf", target_os = "vita", target_os = "nto")))]
             syscall!(fcntl(fds[0], libc::F_SETFD, libc::FD_CLOEXEC))?;
             syscall!(fcntl(fds[1], libc::F_SETFL, libc::O_NONBLOCK))?;
             #[cfg(not(any(target_os = "espidf", target_os = "vita")))]
```
Thomasdezeeuw, line 116: I think we need to add nto here as well, per the use of `SOCK_CLOEXEC`.

src/sys/unix/uds/listener.rs
```diff
@@ -78,6 +79,7 @@ pub(crate) fn accept(listener: &net::UnixListener) -> io::Result<(UnixStream, So
         target_os = "watchos",
         target_os = "espidf",
         target_os = "vita",
+        target_os = "nto",
         all(target_arch = "x86", target_os = "android")
     ))]
     let socket = syscall!(accept(
```
Thomasdezeeuw, line 85: On line 101/103 we need to add nto to ensure `O_NONBLOCK` is set.

src/sys/unix/tcp.rs
```diff
@@ -95,6 +95,7 @@ pub(crate) fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream,
         target_os = "espidf",
         target_os = "vita",
         target_os = "hermit",
+        target_os = "nto",
         all(target_arch = "x86", target_os = "android"),
     ))]
     let stream = {
```
Thomasdezeeuw, line 101: Need to add nto to the `cfg` on line 118/119, setting `O_NONBLOCK`.

tests/unix_pipe.rs
```diff
@@ -91,6 +92,7 @@ fn event_when_sender_is_dropped() {
 }
 
 #[test]
+#[cfg_attr(target_os = "nto", ignore = "Read fd close events do not trigger POLLHUP on nto target")]
```
AkhilTThomas, line 95: POLLHUP is not set by not, hence ignoring this test case

Thomasdezeeuw:
Has any progress has been made for a CI setup?

src/sys/unix/uds/mod.rs
Thomasdezeeuw: This still  needs to be done, same as we do for `fds[0]` above.

Thomasdezeeuw (approved):

AkhilTThomas:
> > (I saw the CI still fails with clippy erros, but I think those aren't on us, are they?
> 
> No, should be fixed by #1768.
> 
> As for the CI, I think we can accept without it, but then we won't be able to support it properly. I won't really be able to judge patches as I have no way to check them. Not an ideal situation, but better than no support at all.
> 
> I think we can accept this after:
> 
> * Fixing https://github.com/tokio-rs/mio/pull/1766/files#r1538833459,
> * a rebase (to fix the Clippy warnings),
> * a `cargo fmt` run (to fix the rustfmt warnings).

Done! Thank you for the review! 😄

Thomasdezeeuw:
Thanks @SebastianSchildt @AkhilTThomas!

Thomasdezeeuw:
Also see follow up pr #1800 to remove the need for any `RUSTFLAGS`.

PR #1767, opened 2024-03-22
Implemented the From trait for converting mio socket types into std

### LINKED ISSUE #1754: why is it possible to create a mio TcpStream from std TcpStream but going the other way around is unsafe?
Example:

` unsafe { std::net::TcpStream::from_raw_fd(mio::net::TcpStream::connect(...).unwrap().into_raw_fd()) }`

however this is perfectly safe

`mio::net::TcpStream::from_std( std::net::TcpStream::connect(....).unwrap())`

tglane:
Implemented the following traits for conversion between the mio socket types and the std socket types:
`From<mio::net::UdpSocket> for std::net::UdpSocket;`
`From<mio::net::TcpListener> for std::net::TcpListener;`
`From<mio::net::TcpStream> for std::net::TcpStream;`

This allows the conversion without using the unsafe keyword in the user code. Inside of the trait function we use the `unsafe` keyword to access the raw underlying socket of the mic sockets. But since the mid socket types are in a well-defined state after construction, the use of unsafe here should be safe.

Closes #1754

### DIFF
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -255,7 +255,7 @@ impl From<TcpListener> for net::TcpListener {
         // mio::net::TcpListener which ensures that we actually pass in a valid file
         // descriptor/socket
         unsafe {
-            #[cfg(any(unix, target_os = "hermit"))]
+            #[cfg(any(unix, target_os = "hermit", target_os = "wasi"))]
             {
                 net::TcpListener::from_raw_fd(listener.into_raw_fd())
             }
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -248,3 +248,21 @@ impl FromRawFd for TcpListener {
         TcpListener::from_std(FromRawFd::from_raw_fd(fd))
     }
 }
+
+impl From<TcpListener> for net::TcpListener {
+    /// Converts a `mio::net::TcpListener` into a `std::net::TcpListener`
+    fn from(listener: TcpListener) -> Self {
+        // Safety: This is safe since we are extracting the raw fd from a well-constructed
+        // mio::net::TcpListener which ensures that we actually pass in a valid file
+        // descriptor/socket
+        unsafe {
+            #[cfg(any(unix, target_os = "hermit"))]
+            let std_listener = net::TcpListener::from_raw_fd(listener.into_raw_fd());
+
+            #[cfg(windows)]
+            let std_listener = net::TcpListener::from_raw_socket(listener.into_raw_socket());
+
+            std_listener
+        }
+    }
+}
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -256,9 +256,13 @@ impl From<TcpListener> for net::TcpListener {
         // descriptor/socket
         unsafe {
             #[cfg(any(unix, target_os = "hermit"))]
-            net::TcpListener::from_raw_fd(listener.into_raw_fd())
+            {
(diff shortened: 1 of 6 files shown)


src/net/tcp/listener.rs
```diff
@@ -248,3 +248,21 @@ impl FromRawFd for TcpListener {
         TcpListener::from_std(FromRawFd::from_raw_fd(fd))
     }
 }
+
+impl From<TcpListener> for net::TcpListener {
+    /// Converts a `mio::net::TcpListener` into a `std::net::TcpListener`
```
Thomasdezeeuw, line 253: Don't need this, the trait and the types it uses are sufficient documentation.

```suggestion
```

src/net/tcp/listener.rs
```diff
@@ -248,3 +248,21 @@ impl FromRawFd for TcpListener {
...
+        // mio::net::TcpListener which ensures that we actually pass in a valid file
+        // descriptor/socket
+        unsafe {
+            #[cfg(any(unix, target_os = "hermit"))]
+            let std_listener = net::TcpListener::from_raw_fd(listener.into_raw_fd());
+
+            #[cfg(windows)]
+            let std_listener = net::TcpListener::from_raw_socket(listener.into_raw_socket());
+
+            std_listener
```
Thomasdezeeuw, line 265: Don't need to create a binding.
```suggestion
            #[cfg(any(unix, target_os = "hermit"))]
            net::TcpListener::from_raw_fd(listener.into_raw_fd())
            #[cfg(windows)]
            net::TcpListener::from_raw_socket(listener.into_raw_socket())
```

Thomasdezeeuw (commented):
Two small suggestions, if you can apply those to all implementations this LGTM.

Could you also add `From` implementations for the Unix-only UDS types in `src/net/uds`?

src/net/tcp/listener.rs
tglane: My editor was complaining about a missing semicolon in that case so I added the binding. But you are right here, it compiles just fine without it so I will apply this to all trait implementations.

tglane:
> Two small suggestions, if you can apply those to all implementations this LGTM.
> 
> Could you also add `From` implementations for the Unix-only UDS types in `src/net/uds`?

I added the trait implementation for `UnixDatagram`, `UnixStream` and `UnixListener` as well.

Thomasdezeeuw (approved):

src/net/tcp/listener.rs
```diff
@@ -248,3 +248,21 @@ impl FromRawFd for TcpListener {
...
     }
 }
+
+impl From<TcpListener> for net::TcpListener {
+    fn from(listener: TcpListener) -> Self {
+        // Safety: This is safe since we are extracting the raw fd from a well-constructed
+        // mio::net::TcpListener which ensures that we actually pass in a valid file
+        // descriptor/socket
+        unsafe {
+            #[cfg(any(unix, target_os = "hermit"))]
```
Thomasdezeeuw, line 258: This currently fails to build on WASI. Can you try the following, to see if that fixes it? You might need to add `use std::os::wasi::io::{FromRawFd, IntoRawFd};` to some files.

```suggestion
            #[cfg(any(unix, target_os = "hermit", target_os = "wasi"))]
```

tglane: Sure. With that addition the lib can be built for the target `wasm32-wasi` on my machine.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @tglane

PR #1768, opened 2024-03-26
Fix Clippy's unused-io-amount lint

Thomasdezeeuw:


Thomasdezeeuw:
So annoying that Clippy gives no warnings locally, but a couple on the CI.

Note to self: squash these commits.

Darksonn (approved):

PR #1773, opened 2024-04-08
Add Apple visionOS support

QuentinPerez:
## Motivation

Add the support of Apple VisionOS  [121419](https://github.com/rust-lang/rust/pull/121419)

## Solution
Add target_os = "visionos"


Thomasdezeeuw:
Can you add VisionOS to our CI and local testing in:
 * https://github.com/tokio-rs/mio/blob/15050b9a24ffb528e97c61aa789aa338093d6059/Makefile#L2
 * https://github.com/tokio-rs/mio/blob/15050b9a24ffb528e97c61aa789aa338093d6059/.github/workflows/ci.yml#L145

(same for the socket2 pr, but I'll add a comment there as well)

.github/workflows/ci.yml
```diff
@@ -1,9 +1,9 @@
 name: CI
 on:
   push:
-    branches: [ "master", "v0.7.x", "v0.8.x" ]
```
Thomasdezeeuw, line 4: Can you undo all these formatting changes? It's hard to see what has actually changed.

Thomasdezeeuw:
libc seems to not build (correctly) for visionOS, see CI.

QuentinPerez:
see https://github.com/rust-lang/libc/pull/3568#issuecomment-2047308703

QuentinPerez:
@Thomasdezeeuw same question; do you mind if I use `target_env = "apple"` ? Instead of changing this every time Apple releases a new device

simlay:
> libc seems to not build (correctly) for visionOS, see CI.

It should now be fix with https://github.com/rust-lang/rust/pull/124560. Though, I think CI may still fail because the default Xcode version for the macOS-latest runner is [currently Xcode 15.0.1](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md#xcode) but [visionOS was added with Xcode 15.2](https://developer.apple.com/documentation/xcode-release-notes/xcode-15_2-release-notes) which means that `xcrun --show-sdk-path --sdk xros` will fail when called from rustc. One could fix this in CI with something like `sudo xcode-select -s /Applications/Xcode_15.2.app` but that's a bit too specific and brittle for my taste.

Thomasdezeeuw:
@QuentinPerez can you rebase this after #1787 is merged to give this another try in the cI?

Thomasdezeeuw:
I've updated this in #1795 to prepare for the changes coming in #1749.

PR #1774, opened 2024-04-14
Update to windows-sys v0.52

skyfloogle:
This is just a cherry-pick of the corresponding commit from the master branch. Required for building on certain targets on newer versions of Rust (see https://github.com/microsoft/windows-rs/pull/2774).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,8 +49,11 @@ log = { version = "0.4.8", optional = true }
 libc = "0.2.149"
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = "0.48"
+version = "0.52"
 features = [
+  "Wdk_Foundation",                   # Required for AFD.
+  "Wdk_Storage_FileSystem",           # Required for AFD.
+  "Wdk_System_IO",                    # Required for AFD.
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw (approved):

Thomasdezeeuw:
Cherry-pick of commit eea9e3e0c469480e5c59c01e6c3c7e5fd88f0848.

Note to self add this to the commit when merging.

skyfloogle:
Looks like this would require a MSRV increase to v1.56, which I understand goes against mio's policy. Does this mean building these targets on Rust v1.76+ will have to wait until v1.0 is stabilized?

Thomasdezeeuw:
> Looks like this would require a MSRV increase to v1.56, which I understand goes against mio's policy. Does this mean building these targets on Rust v1.76+ will have to wait until v1.0 is stabilized?

Yes, I'm afraid so.

skyfloogle:
In that case, I'll close the pull request.
On the bright side, after further investigation, I found out this doesn't actually block me, as my project only uses `mio` as a transitive dependency which isn't actually invoked on Windows.
Apologies for the confusion, and good luck with future developments!

PR #1783, opened 2024-05-05
Update windows-sys to 0.52

westy92:
Land https://github.com/tokio-rs/mio/commit/eea9e3e0c469480e5c59c01e6c3c7e5fd88f0848 on `0.8.x` to unblock updating `windows-sys` in `tokio`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,8 +49,11 @@ log = { version = "0.4.8", optional = true }
 libc = "0.2.149"
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = "0.48"
+version = "0.52"
 features = [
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw:
Duplicate of #1725. Can't be accept in v0.8 because windows-sys bumped the MSRV to be higher than what Mio supports.

westy92:
Can we bump MSRV to 1.56 now that it's after March 2024?

https://github.com/tokio-rs/mio/pull/1725#issuecomment-1840332722

Thomasdezeeuw:
I rather move forward with Mio v1, for which I don't have enough time as it is.

westy92:
Fair enough. Thank you for your work on v1!

PR #1802, opened 2024-06-09
Fix compilation on Hermit

Thomasdezeeuw:
For some reason the `std::os::fd` module is not available on Hermit. It seems like an oversight and I've opened rust-lang/rust#126198 to track a fix.


Thomasdezeeuw:
> Should you create your own `crate::os` module to avoid duplicating this workaround everywhere?

A fix is already in the works: https://github.com/rust-lang/rust/pull/126346, so maybe we can switch back soon-ish.

PR #1806, opened 2024-06-14
Reduce MSRV to v1.70

### LINKED ISSUE #1732: Decide MSRV for v1

Thomasdezeeuw:
Released on 1 June 2023, gives us 1 year of supported Rust versions.

Closes #1732


Darksonn (approved):

Thomasdezeeuw:
The CI is hitting: https://github.com/dtolnay/rust-toolchain/issues/112. Going to see if there is a work around

PR #1808, opened 2024-06-14
Remove Solaris from unsupported list

### LINKED ISSUE #1152: Port to Solaris

Thomasdezeeuw:
We now support it via the poll(2) implementation.

It's also `cargo check`ed in our CI already.

Closes #1152

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,10 +152,8 @@ This uses the Windows AFD system to access socket readiness events.
 
 ### Unsupported
 
-* Solaris, see [issue #1152]
 * Wine, see [issue #1444]
 
-[issue #1152]: https://github.com/tokio-rs/mio/issues/1152
 [issue #1444]: https://github.com/tokio-rs/mio/issues/1444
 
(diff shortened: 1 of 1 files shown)


Darksonn (approved):

RalfJung:
Is it expected that is was removed from "unsupported" but not added to "supported"? Solaris is currently not listed at all.

Thomasdezeeuw:
> Is it expected that is was removed from "unsupported" but not added to "supported"? Solaris is currently not listed at all.

Good point, I'll add it once I make another release.

Thomasdezeeuw:
Solaris was added to the supported list in https://github.com/tokio-rs/mio/pull/1970.

PR #1813, opened 2024-06-21
More cleanups cfgs in unix module

Thomasdezeeuw:
Similar to the changes made the cfgs for the selector this commit set is for the waker type.

The waker implementation are split into multiple files, one per implementation, matching the selector split. The mess in src/sys/unix/waker.rs is cleaned up and moved to src/sys/unix/mod.rs and src/sys/unix/selector/*.rs, see individual commits for more details.


Thomasdezeeuw:
@Darksonn hopefully this is the last of the cfg cleanups :)

Darksonn (approved):

PR #1816, opened 2024-06-30
Add GNU/Hurd support

sthibaul:


tests/poll.rs
```diff
@@ -112,6 +112,7 @@ fn poll_closes_fd() {
 }
 
 #[test]
+#[cfg_attr(target_os = "hurd", ignore = "protocol not supported")]
```
Thomasdezeeuw, line 115: Does Hurd not support TCP? Can you expand on what protocol is not supported?

tests/tcp.rs
```diff
@@ -571,6 +571,7 @@ fn connect_error() {
 
         for event in &events {
             if event.token() == Token(0) {
+                #[cfg(not(target_os = "hurd"))]
```
Thomasdezeeuw, line 574: What is causing this? This behaviour deviates from the expected.

tests/unix_datagram.rs
```diff
@@ -100,6 +103,7 @@ fn unix_datagram_connect() {
     let datagram2_local = datagram2.local_addr().unwrap();
 
     datagram1
+
```
Thomasdezeeuw, line 106: Remove.

tests/regressions.rs
```diff
@@ -16,6 +16,7 @@ const ID1: Token = Token(1);
 const WAKE_TOKEN: Token = Token(10);
 
 #[test]
+#[cfg_attr(target_os = "hurd", ignore = "protocol not supported")]
```
Thomasdezeeuw, line 19: Same as above, can you expand on what protocol is not supported?

.github/workflows/ci.yml
```diff
@@ -151,6 +151,7 @@ jobs:
           - arm64_32-apple-watchos
           - armv7-sony-vita-newlibeabihf
           - i686-unknown-linux-gnu
+          - i686-unknown-hurd-gnu
```
Thomasdezeeuw, line 154: Does is support x86_64, maybe ARM as well?

Thomasdezeeuw (changes requested):

.github/workflows/ci.yml
sthibaul: x86_64 and arm64 are being supported, but that's not in upstream llvm etc. yet.

tests/poll.rs
sthibaul: TCP is supported of course :) I have to check again precisely, it's probably some particular sockopt, possibly the failure could be ignored inside the mio implementation.

sthibaul: It's the `set_read_timeout` call that gets `SO_RCVTIMEO` seen as not being implemented. I have disabled them, it then goes fine.

tests/regressions.rs
sthibaul: same as above

tests/tcp.rs
sthibaul: It's odd here. When I try a C equivalent of the test (connecting to a closed port, and expecting the resulting socket to be writable), I do see it writable. I don't know why I'm getting a non-writable event here. Actually I didn't find which part of the code sets `WRITABLE` in this case, I only saw `READABLE` in `src/sys/unix/waker/pipe.rs` for instance.

Thomasdezeeuw: > Actually I didn't find which part of the code sets `WRITABLE` in this case, I only saw `READABLE` in `src/sys/unix/waker/pipe.rs` for instance.

Mio doesn't set any events, is the OS/kernel that does that. The relevant check is in https://github.com/tokio-rs/mio/blob/1ce154579e1076260ea6934e89275912b01fae47/src/sys/unix/selector/poll.rs#L565.

sthibaul: Ok, I see, `is_writable` is overloaded between the events and the interests.

So I dug more, what happens is that since tcp fast_open is not supported yet, but we set `SOCK_NONBLOCK`, after connect returned `EINPROGRESS` the socket notices the refused connection, and thus sets an error condition on the socket. `poll` thus returns `POLLERR` instead of `POLLOUT`, which indeed does make sense, I have added that as an alternative.

Thomasdezeeuw: Can we have a comment explaining this?

sthibaul: Sure! Done so

tests/tcp.rs
```diff
@@ -571,7 +571,9 @@ fn connect_error() {
 
         for event in &events {
             if event.token() == Token(0) {
-                assert!(event.is_writable());
+		// With fastopen we would be able to write
+		// Without fastopen we would be getting the connection error
```
Thomasdezeeuw, line 575: Needs formatting:
```suggestion
		    // With fastopen we would be able to write.
		    // Without fastopen we would be getting the connection error.
```

Thomasdezeeuw (commented):
Except for the formatting LGTM.

tests/tcp.rs
sthibaul: ah, right, my editor wasn't set up to show me the difference between spaces and tabs, now fixed

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @sthibaul

Issue #1796, opened 2024-06-09
Switch to libc for Hermit


Thomasdezeeuw:
Depends on https://github.com/rust-lang/libc/pull/3724

After that we should be able to remove the dependency on the `hermit-abi` crate.

Thomasdezeeuw:
Now waiting on https://github.com/rust-lang/libc/pull/3766.

Thomasdezeeuw:
Done in #1830.

PR #1821, opened 2024-08-06
Update hermit-abi.

westy92:


Thomasdezeeuw:
We want to drop the hermit-abi dependency: https://github.com/tokio-rs/mio/issues/1796, so not updating it.

PR #1823, opened 2024-08-06
Update wasi

westy92:


Thomasdezeeuw:
This updates to wasi v0.2, which we don't support.

PR #1826, opened 2024-08-09
Copy illumos waker pipe work around to eventfd

Thomasdezeeuw:
It seems the same behaviour we seen for pipe waker implementation is also true for eventfd. Work around this by resetting the waker before writing to it.

### DIFF
--- a/src/sys/unix/waker/eventfd.rs
+++ b/src/sys/unix/waker/eventfd.rs
@@ -47,7 +47,7 @@ impl Waker {
         // generated.
         // See https://www.illumos.org/issues/16700.
         #[cfg(target_os = "illumos")]
-        self.reset();
+        self.reset()?;
 
         let buf: [u8; 8] = 1u64.to_ne_bytes();
         match (&self.fd).write(&buf) {
--- a/src/sys/unix/waker/eventfd.rs
+++ b/src/sys/unix/waker/eventfd.rs
@@ -42,6 +42,13 @@ impl Waker {
 
     #[allow(clippy::unused_io_amount)] // Don't care about partial writes.
     pub(crate) fn wake(&self) -> io::Result<()> {
+        // The epoll emulation on some illumos systems currently requires
+        // the eventfd to be read before an edge-triggered read event is
+        // generated.
+        // See https://www.illumos.org/issues/16700.
+        #[cfg(target_os = "illumos")]
+        self.reset();
+
         let buf: [u8; 8] = 1u64.to_ne_bytes();
         match (&self.fd).write(&buf) {
             Ok(_) => Ok(()),
--- a/src/sys/unix/waker/pipe.rs
+++ b/src/sys/unix/waker/pipe.rs
@@ -40,7 +40,7 @@ impl Waker {
         // The epoll emulation on some illumos systems currently requires
         // the pipe buffer to be completely empty for an edge-triggered
         // wakeup on the pipe read side.
-        // See https://www.illumos.org/issues/16700.
+        // See https://www.illumos.org/issues/13436.
         #[cfg(target_os = "illumos")]
         self.empty();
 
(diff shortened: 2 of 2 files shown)


src/sys/unix/waker/eventfd.rs
```diff
@@ -42,6 +42,13 @@ impl Waker {
 
     #[allow(clippy::unused_io_amount)] // Don't care about partial writes.
     pub(crate) fn wake(&self) -> io::Result<()> {
+        // The epoll emulation on some illumos systems currently requires
+        // the eventfd to be read before an edge-triggered read event is
+        // generated.
+        // See https://www.illumos.org/issues/16700.
+        #[cfg(target_os = "illumos")]
+        self.reset();
```
sunshowers, line 50: Since `self.reset()` returns an error, I think this should have a `?` at the end.

src/sys/unix/waker/pipe.rs
```diff
@@ -40,6 +40,7 @@ impl Waker {
         // The epoll emulation on some illumos systems currently requires
         // the pipe buffer to be completely empty for an edge-triggered
         // wakeup on the pipe read side.
+        // See https://www.illumos.org/issues/16700.
```
sunshowers, line 43: Thanks for adding this link! The bugs are related but actually distinct. The correct issue here is https://www.illumos.org/issues/13436.

sunshowers (commented):
Thank you for your time working on this issue! While I continue to believe that switching back to pipes is the least risky way to move forward, I have a couple of comments if you choose to go down this route.

Thomasdezeeuw:
The CI will be fixed tomorrow, see https://github.com/tokio-rs/mio/pull/1827#issuecomment-2278533810.

Darksonn (approved):
Assuming the Illumos folks will test this to sanity check it, LGTM. We may want https://github.com/tokio-rs/tokio/issues/6763 in mio as well.

Thomasdezeeuw:
> Assuming the Illumos folks will test this to sanity check it, LGTM. We may want [tokio-rs/tokio#6763](https://github.com/tokio-rs/tokio/issues/6763) in mio as well.

I would like better CI support for illumos (or any OS that we currently don't test),

Thomasdezeeuw:
CI should be fixed by #1828, so merging this.

jclulow:
> Assuming the Illumos folks will test this to sanity check it, LGTM. We may want https://github.com/tokio-rs/tokio/issues/6763 in mio as well.

For the record:

- This is expressly the opposite of what we asked for, and has been integrated over our strong objections
- A cursory inspection and passing similarity is not sufficient at all to determine that this workaround and the existing workaround are in any way related
- No amount of testing we could possibly perform in the next day with a few unit tests would produce the same risk profile as merely reverting the original regression

It's unfortunate that the members of the @tokio-rs/illumos maintenance group were ignored and bypassed here twice: once when the original change was made without notice or testing during a larger refactoring, then again here.  We are strongly motivated to help as this software is critical for us and for others; please let us do so!

Thomasdezeeuw:
@jclulow I didn't want to do this, but if we're adding to the record...

I really, really disliked communicating with you and @sunshowers. Every single comment the both of you made in #1824 read like a marketing blob. We're engineers, please be brief in your communication. No need to highlight every single first sentence, nor write paragraphs when the information can be conveyed in a single sentence.

I preferred to determine the cause of the bug, rather then ignore it. This didn't see to be too urgent for either of you. I understand that you rely on Tokio, and thus Mio, but it's fine to stick to a previous version for a couple of days longer, giving us time to fix the bug properly.

Once I did have time to look at the issue it took me whole of 2 minutes to figure out it was most likely the same issue as the one we already hit in the pipe waker implementation. Giving me another indication that neither of you really bother to look at the problem at hand (but I could be wrong here and this was simply not communicated to me).

Doubling down with comments such as https://github.com/tokio-rs/mio/pull/1824#issuecomment-2277413008 and https://github.com/tokio-rs/mio/pull/1826#issuecomment-2282190812 is really not appropriated by me. I didn't have to help, I could just say illumos is no longer supported. But I didn't and I won't. I always try to help out OSs that I personally don't run, but I *do* need help from the people who do run them. I failed that get from either of you.

In the future I hope situations like this are resolved differently.

PR #1827, opened 2024-08-09
Add rust-src component to sanitizer CI

Thomasdezeeuw:


Thomasdezeeuw:
Ah, this is caused by https://github.com/rust-lang/rust/issues/128808, fixed in https://github.com/rust-lang/cargo/pull/14370, which should be released tomorrow.

PR #1828, opened 2024-08-10
Fix Hermit build

Thomasdezeeuw:
The Hermit build would fail if --no-default-features --features os-ext were passed as flags. This is because the cfg_io_source! macro assumed that IoSource was needed for os-ext, which is true for Unix due to the usage of the Unix pipe, but not true for Hermit (it doesn't support Unix pipes).


Darksonn (approved):

Thomasdezeeuw:
Thanks @Darksonn

PR #1830, opened 2024-08-19
remove dependency to hermit-abi

stlankes:
- use directly libc as platform interface

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,10 +47,10 @@ net = []
 log = { version = "0.4.8", optional = true }
 
 [target.'cfg(unix)'.dependencies]
-libc = "0.2.149"
+libc = "0.2.159"
 
 [target.'cfg(target_os = "hermit")'.dependencies]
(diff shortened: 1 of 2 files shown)


stlankes:
Ok, we have to wait for the missing error numbers. see rust-lang/libc#3858

stlankes:
A new version of libc is published and the dependency to hermit-abi can be removed. But I don't understand that `CI / Sanitizer (leak)` failed. Does someone have an idea? I just use a new version of libc.

Thomasdezeeuw (approved):

Thomasdezeeuw:
@stlankes the CI can be ignored it's a sanitizer issue, Mio is not to blame.

Thomasdezeeuw:
Thanks @stlankes

stlankes:
@Thomasdezeeuw Do you know, when you publish a new release? I would prefer to remove this dependency from the latest release.

Thomasdezeeuw:
I'll prepare a release now. I don't really have a schedule for them.

PR #1833, opened 2024-10-02
AIX use force poll poll Selector by default

mustartt:
On AIX, we are having trouble implementing `pollset` for Selector. We would like to opt-in to the `mio_unsupported_force_poll_poll`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,6 +152,7 @@ jobs:
           - armv7-sony-vita-newlibeabihf
           - i686-unknown-linux-gnu
           - i686-unknown-hurd-gnu
+          - powerpc64-ibm-aix
           - riscv32imc-esp-espidf
           - sparcv9-sun-solaris
           - wasm32-wasi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,7 +152,7 @@ jobs:
           - armv7-sony-vita-newlibeabihf
           - i686-unknown-linux-gnu
           - i686-unknown-hurd-gnu
-          - powerpc64-ibm-aix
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw:
Can you add CI support for this? See https://github.com/tokio-rs/mio/blob/d8d68ac6373a197af435a80d61d71963678fff97/.github/workflows/ci.yml#L128-L135

Thomasdezeeuw:
Seems like the random crate doesn't compile?

```
  error[E0425]: cannot find function `fill_bytes` in module `sys`
    --> /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/random.rs:66:14
     |
  66 |         sys::fill_bytes(bytes)
     |              ^^^^^^^^^^ not found in `sys`
  
  error[E0425]: cannot find function `fill_bytes` in this scope
    --> /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sys/random/mod.rs:92:5
     |
  92 |     fill_bytes(&mut buf);
     |     ^^^^^^^^^^ not found in this scope
  
```

mustartt:
It seems like we don't have the full stdlib support yet in upstream Rust. It compiles and tests fine downstream.

Would it be okay to leave aix out of the CI for now? 

Also, I'm seeing a `CI / Sanitizer (leak)` in the checks.

mustartt:
Thanks for the info. I removed it from CI and created a new issue mentioning it https://github.com/tokio-rs/mio/issues/1834.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @mustartt

PR #1836, opened 2024-10-09
add `wasm32-wasip2` support

dicej:
This implementation currently uses a mix of POSIX-style APIs (provided by `wasi-libc` via the `libc` crate) and WASIp2-native APIs (provided by the `wasi` crate).

Alternatively, we could implement `Selector` using only POSIX APIs, e.g. `poll(2)`.  However, that would add an extra layer of abstraction to support and debug, as well as make it impossible to support polling `wasi:io/poll/pollable` objects which cannot be represented as POSIX file descriptors (e.g. timer events, DNS queries, HTTP requests, etc.).

Another approach would be to use _only_ the WASIp2 APIs and bypass `wasi-libc` entirely.  However, that would break interoperability with both Rust `std` and e.g. C libraries which expect to work with file descriptors.

Since `wasi-libc` does not yet provide a public API for converting between file descriptors and WASIp2 resource handles, we currently use a non-public API (see the `netc` module below) to do so.  Once
https://github.com/WebAssembly/wasi-libc/issues/542 is addressed, we'll be able to switch to a public API.

I've tested this end-to-end using https://github.com/dicej/wasi-sockets-tests, which includes smoke tests for `mio`, `tokio`, `tokio-postgres`, etc.


dicej:
I just realized this PR doesn't include support for accepting incoming connections -- only initiating outgoing ones.  I forgot about the former since I don't personally have an urgent need to support it, but it could be added as a follow-up PR.  To be clear: WASIp2 is fully capable of handling that case.

src/sys/wasi/p2.rs
```diff
@@ -0,0 +1,662 @@
...
+            let index = usize::try_from(index).unwrap();
+            if timeout.is_none() || index != pollables.len() - 1 {
+                let (_, fd, socket, subscription, interests) = &states[index];
+
+                let mut push_event = || {
+                    events.push(Event {
+                        token: subscription.token,
+                        interests: *interests,
+                    })
+                };
```
raskyld, line 247: Just out of curiosity, if there any reason to do that in a closure vs. doing it inline at L276, L289 and L297?

dicej: No particular reason other than avoiding the repetition.

raskyld: Alright! Was not sure if it was DRY or there was a mechanism I am not aware of, thanks for clarification 🙏

raskyld:
Also, I have a last clarifying question: reading `wasi-libc` I assumed that all `Pollable` are abstracted as a Socket but this socket is a "fake" one in the sense that we only use it to convey the "readiness" of the underlying `Pollable` irrespectively of whether it is actually obtained through `wasi:sockets`.

But your comment:

> I just realized this PR doesn't include support for accepting incoming connections -- only initiating outgoing ones. I forgot about the former since I don't personally have an urgent need to support it, but it could be added as a follow-up PR. To be clear: WASIp2 is fully capable of handling that case.

Actually speak of connections, so I am not sure my previous assumption was correct.

Anyway, great job 💪

dicej:
> Also, I have a last clarifying question: reading wasi-libc I assumed that all Pollable are abstracted as a Socket but this socket is a "fake" one in the sense that we only use it to convey the "readiness" of the underlying Pollable irrespectively of whether it is actually obtained through wasi:sockets.

WASIp2 represents things like TCP and UDP sockets as resources (e.g. [tcp-socket](https://github.com/WebAssembly/wasi-sockets/blob/main/wit/tcp.wit#L46)) which have methods for binding, listening, connecting, etc.  Once a socket is connected, you can get its `input-stream` and `output-stream`, which are also resources.  And each of those can be used to obtain a `pollable` representing read and write readiness, respectively.

Consequently, `wasi-libc` must keep track of up to six resource handles for each socket: 

- a `tcp-socket` or `udp-socket`, depending on the socket type
- a `pollable` representing readiness of any in-progress bind, listen, connect, or accept operation
- (if connected) an `input-stream`, an `output-stream`, and one `pollable` each for reading and writing

So you can think of a `wasi-libc` socket file descriptor as uniquely identifying a bundle of resource handles, the number and type of which depends on the state that socket is in.

Adding support for binding, listening and accepting WASIp2 sockets in `mio` would amount to adding `match` cases for e.g. `tcp_socket_state_tag_t::TCP_SOCKET_STATE_UNBOUND`, `tcp_socket_state_tag_t::TCP_SOCKET_STATE_BOUND`, and `tcp_socket_state_tag_t::TCP_SOCKET_STATE_LISTENING` and using the `socket_pollable` field of `tcp_socket_t` to await transitions to the next state.

Does that help?  Happy to go into more detail if desired.  Also, the [wasi-sockets docs](https://github.com/WebAssembly/wasi-sockets/blob/main/imports.md) are quite through if you haven't perused them yet.

dicej:
BTW, this PR doesn't include UDP support either, again because that hasn't been a priority for me.  Shouldn't be hard to add as a follow-up PR.

raskyld:
Thanks a lot for the time you took answering me 🙏 
Your answer is super clear!

My question was more related to how `pollable` obtained from types not related to `sockets` are usable in `mio` context, for example, the streams you mentioned can also be obtained through the `wasi:http` world: https://github.com/WebAssembly/wasi-http/blob/main/wit/types.wit#L510

In this case, I suspect you still have the `streams` and the `pollable` but not the sockets resource handles.
Could you still use those resource handles to `Poll` them in `mio`? IIUC, the philosophy of the crate is to build upon **any** event source independently of the platform primitive under of that but I may misunderstand it!

dicej:
> My question was more related to how `pollable` obtained from types not related to `sockets` are usable in `mio` context, for example, the streams you mentioned can also be obtained through the `wasi:http` world: https://github.com/WebAssembly/wasi-http/blob/main/wit/types.wit#L510

Oh right, great question.  Yeah, I think we'd need to add a new, WASIp2-only API for registering `pollable`s that have no corresponding file descriptors (and likewise for `tokio`, presumably).  This PR clearly doesn't include such a thing, but it's something we could add in another PR.

Another approach would be to add an API to `wasi-libc` that accepts an arbitrary `pollable` and allocates a file descriptor for it.

@badeend and @sunfishcode might have thoughts about this.

raskyld:
Alright thanks for the answer! Reading https://github.com/WebAssembly/wasi-libc/issues/542 it seems to me that the question of allocating fd to pollable was among the initial options, I wonder though how would you implement stuff like `fstat` for those fd 🤔 but let's not pollute your MR!

To get back to the main subject,
I just remembered I wondered why you use atomic types and mutex in the PR since `wasip2` is single-threaded?
Is it because of the thread proposal ?

dicej:
> To get back to the main subject, I just remembered I wondered why you use atomic types and mutex in the PR since `wasip2` is single-threaded? Is it because of the thread proposal ?

Honestly, I just copied that from the existing WASIp1 implementation.  I wrote this code almost a year ago and only came back to it yesterday, so it's not super fresh in my mind, but that part at least came straight from the existing code.  I'm guessing the original code either needed it to make the compiler happy (e.g. make `Selector` `Send` and `Sync`) or for future-proofing.  `Arc` and `Mutex` are roughly equivalent to `Rc` and `RefCell` on single-threaded Wasm, anyway, so there shouldn't be a performance penalty.

Cargo.toml
```diff
@@ -65,8 +65,13 @@ features = [
   "Win32_System_WindowsProgramming",  # General future used for various types/funcs
 ]
 
-[target.'cfg(target_os = "wasi")'.dependencies]
+[target.'cfg(all(target_os = "wasi", not(target_env = "p2")))'.dependencies]
 wasi = "0.11.0"
+
+[target.'cfg(all(target_os = "wasi", target_env = "p2"))'.dependencies]
+wasi = "0.13.3"
```
Darksonn, line 72: Is this really the correct way to go about this? Only 0.11 supports p1?

src/io_source.rs
```diff
@@ -175,7 +175,7 @@ where
     }
 }
 
-#[cfg(target_os = "wasi")]
+#[cfg(all(target_os = "wasi", not(target_env = "p2")))]
```
Darksonn, line 178: Can we do this?
```suggestion
#[cfg(all(target_os = "wasi", target_env = "p1"))]
```

src/net/tcp/stream.rs
```diff
@@ -87,10 +87,10 @@ impl TcpStream {
     /// entries in the routing cache.
     ///
     /// [write interest]: Interest::WRITABLE
-    #[cfg(not(target_os = "wasi"))]
+    #[cfg(any(not(target_os = "wasi"), target_env = "p2"))]
```
Darksonn, line 90: Can you use `not(all(...))` here?

Cargo.toml
raskyld: Looking at https://docs.rs/wasi/latest/wasi/
I would say yes, the last mention I see of `p1` is on `0.11`

raskyld:
After a good night of sleep, I realised the name of the PR should probably mention that it only adds support for `wasi:sockets` so it's clear that we will need follow-up PRs for other `pollable` and for initiating connections.

WDYT?

src/io_source.rs
dicej: The reason I did it that way is because `not(target_env = "p2")` is backwards compatible with `rustc` versions prior to the introduction of the `wasm32-wasip2` target (i.e. all stable `rustc` versions as of this writing).  See [here](https://bytecodealliance.zulipchat.com/#narrow/stream/235408-rust-toolchain/topic/.5Bpost.5D.20Changes.20to.20Rust's.20WASI.20targets/near/468251805) for further discussion.  It _does_ mean we'll need to make changes when `p3` becomes available, though.  Happy to do whatever you think is best here.

src/net/tcp/stream.rs
dicej: If you mean `not(all(target_os = "wasi", target_env = "p1"))` then the same rationale I gave above for not using `target_env = "p1"` applies.  Again, though. happy to change it if we're not concerned about backwards `rustc` compatibility.

dicej:
> After a good night of sleep, I realised the name of the PR should probably mention that it only adds support for established `wasi:sockets` so it's clear that we will need follow-up PRs for other `pollable` and for initiating connections.
> 
> WDYT?

Makes sense; I'll add some TODO comments to the code and to the commit message.

src/sys/wasi/p2.rs
```diff
@@ -0,0 +1,675 @@
...
+        })
+    }
+
+    pub(crate) fn select(&self, events: &mut Events, timeout: Option<Duration>) -> io::Result<()> {
+        events.clear();
+
+        let mut subscriptions = self.subscriptions.lock().unwrap();
+
+        let mut states = Vec::new();
+        for (fd, subscription) in subscriptions.deref() {
```
Thomasdezeeuw, line 159: I don't think looping of all active subscriptions is a great idea. Nor is the allocation for `states`.

Why switch from a `Vec` to a `HashMap`?

Thomasdezeeuw (changes requested):
I don't really have much time to review this.

Can we split this up in multiple prs as this seems to do multiple things.
1. A minimal pr that adds support for v2, no adding of v2 functionality.
2. The selector rewrite (not sure why this is needed)
3. Any v2 additional we make, such as support for more API

dicej:
> Can we split this up in multiple prs as this seems to do multiple things.
> 
>     1. A minimal pr that adds support for v2, no adding of v2 functionality.
> 
>     2. The selector rewrite (not sure why this is needed)
> 
>     3. Any v2 additional we make, such as support for more API

Items 1 and 2 are intertwined.  None of the WASIp1 APIs are available in WASIp2, so a new selector implementation is needed to target the new APIs.  There _is_ an adapter which implements (most of) the WASIp1 APIs in terms of their WASIp2 counterparts, but wasi-libc doesn't use the adapter for sockets -- it uses the WASIp2 APIs directly in order to access the much broader socket support that WASIp2 provides.

As I mentioned above, we _do_ have a few different options for a WASIp2 selector implementation (use POSIX `poll(2)`, use the WASIp2 APIs directly, or use a mix of POSIX file descriptors and direct WASIp2 APIs for maximum compatibility), but reusing the WASIp1 implementation is not one of those options.

Totally agreed that we can leave item 3 in your list for a later PR, though.

src/sys/wasi/p2.rs
dicej: I'm happy to optimize by reducing the number of allocations if desired.  Would you prefer that I update this PR or leave that for a follow-up PR?

> Why switch from a Vec to a HashMap?

I was aiming for O(1) lookups by file descriptor in the `IoSourceState` implementation.

banocean: > I'm happy to optimize by reducing the number of allocations if desired. Would you prefer that I update this PR or leave that for a follow-up PR?
> 
> > Why switch from a Vec to a HashMap?
> 
> I was aiming for O(1) lookups by file descriptor in the `IoSourceState` implementation.

It might be slower when taking into account hashing tho :thinking:

dicej: > It might be slower when taking into account hashing tho 🤔

Possibly; I can switch back to using a `Vec` if preferred.

jeffparsons:
Just to clarify, is this currently "waiting-on-review", to use the rust-lang parlance? Or, conversely, is further review currently blocked on any action from the author?

Thomasdezeeuw:
@jeffparsons this is waiting on an action on https://github.com/tokio-rs/mio/pull/1836#pullrequestreview-2367086886. We currently don't have the capacity to review a pr of this size, so it needs to be split up.

jeffparsons:
> @jeffparsons this is waiting on an action on [#1836 (review)](https://github.com/tokio-rs/mio/pull/1836#pullrequestreview-2367086886). We currently don't have the capacity to review a pr of this size, so it needs to be split up.

Thanks, @Thomasdezeeuw. I wonder if that was perhaps not clear to @dicej because [his later comment](https://github.com/tokio-rs/mio/pull/1836#issuecomment-2411877350) suggests there's no natural decomposition of what's in this PR, which gave me the impression that he was waiting for feedback on his explanation. 

Just floating this idea: would it be possible/acceptable to both author and reviewer to split this into at least: 

- just common plumbing (e.g. selector) without any actual socket APIs 
- outgoing socket APIs 

Would that make review more manageable? Are there other axes this could be reasonably split on?

I'm just an observer, so sorry if this isn't helpful. Just trying to see if there's anything that could get this out of limbo. 

💖

dicej:
> @jeffparsons this is waiting on an action on [#1836 (review)](https://github.com/tokio-rs/mio/pull/1836#pullrequestreview-2367086886). We currently don't have the capacity to review a pr of this size, so it needs to be split up.

Along the lines of what @jeffparsons suggested, I can definitely split this up if we're not concerned about delivering useful functionality to `wasm32-wasip2` users right off the bat.  For example, I could start with a minimal scaffolding for the new selector that compiles but with only `todo!()` in the function bodies, then reintroduce the rest of the implementation incrementally until we have something that's actually useful.  How does that sound?

jeffparsons:
@Thomasdezeeuw Does the proposal in https://github.com/tokio-rs/mio/pull/1836#issuecomment-2441708065 sound okay to you?

fbjork:
What’s the path forward here? Submitting smaller PRs to make reviewing easier? Would love to see this land!

pimeys:
Could somebody tell me what is needed to get this one merged? I could try to finish it up, but I'm not sure why the current PR is not good enough and how would a possible PR from me be something that could be merged to main.

<rant>We'd like to start using the async ecosystem in wasi, and there are runtimes such as wstd that already works. But for some reason the whole rust async ecosystem depends on Tokio... I wonder how back in the days we could've abstracted the runtimes so we wouldn't be in this situation now...</rant>

dicej:
@pimeys @fbjork I'll try to summarize the situation as I understand it (see the earlier comments for details):

@Thomasdezeeuw feels this is not the right approach and would prefer a smaller, easier to review PR (i.e. not just breaking this one up into smaller pieces, but smaller overall).  To address that, [I've proposed](https://github.com/tokio-rs/mio/pull/1836#issuecomment-2474524724) that we treat WASIp2 as "just another POSIX system" and use the generic POSIX `poll(2)`-based implementation in `src/sys/unix/selector/poll.rs` instead of a WASI-specific implementation, leaving the details of WASIp2 interop to wasi-libc.  I expect that would require minimal changes to `mio`, amounting mostly to tweaking some `#[cfg(...)]` annotations here and there.  Whether such a PR would be accepted is something only a maintainer can answer, however.

pimeys:
> @pimeys @fbjork I'll try to summarize the situation as I understand it (see the earlier comments for details):
> 
> @Thomasdezeeuw feels this is not the right approach and would prefer a smaller, easier to review PR (i.e. not just breaking this one up into smaller pieces, but smaller overall). To address that, [I've proposed](https://github.com/tokio-rs/mio/pull/1836#issuecomment-2474524724) that we treat WASIp2 as "just another POSIX system" and use the generic POSIX `poll(2)`-based implementation in `src/sys/unix/selector/poll.rs` instead of a WASI-specific implementation, leaving the details of WASIp2 interop to wasi-libc. I expect that would require minimal changes to `mio`, amounting mostly to tweaking some `#[cfg(...)]` annotations here and there. Whether such a PR would be accepted is something only a maintainer can answer, however.

Right. So instead of implementing our own module, we just use the unix module? It should just work if we have the corresponding wit imported to wasmtime?

Btw, is there a way to run the tests in mio with wasm32-wasip2 target?

dicej:
> Right. So instead of implementing our own module, we just use the unix module?

Correct.

> It should just work if we have the corresponding wit imported to wasmtime?

I'm not sure what you mean by that.  When you tell Cargo to target `wasm32-wasip2`, it will statically link in wasi-libc, which will reference the WASIp2 interface definitions (e.g. `wasi-clocks`, `wasi-filesystem`, `wasi-sockets`, etc.).  The resulting component will contain everything `wasmtime run` needs to know to run it -- no additional WIT files required.

> Btw, is there a way to run the tests in mio with wasm32-wasip2 target?

I haven't tried that, but it should be doable by using `wasmtime run` as the test runner.  I've written some end-to-end tests [here](https://github.com/dicej/wasi-sockets-tests) that exercise `tokio`, `tokio-postgres`, etc.

pimeys:
@dicej how should the wakers be created in wasip2? There's some abstractions in the `wasi` crate e.g. for [pollables](https://docs.rs/wasi/0.14.1+wasi-0.2.3/wasi/io/poll/struct.Pollable.html) that could be used here. But how does this fit in the context of mio? Surely none of the eventfd, kqueue nor pipe wakers can be used here, we need something for wasi that can be used together with the pollables?

dicej:
> @dicej how should the wakers be created in wasip2? There's some abstractions in the `wasi` crate e.g. for [pollables](https://docs.rs/wasi/0.14.1+wasi-0.2.3/wasi/io/poll/struct.Pollable.html) that could be used here. But how does this fit in the context of mio? Surely none of the eventfd, kqueue nor pipe wakers can be used here, we need something for wasi that can be used together with the pollables?

The idea [presented here](https://github.com/WebAssembly/wasi-libc/issues/542) is to add an API to `wasi-libc` to allow pollables to be converted to file descriptors, making them usable with `mio`, `poll(2)`, etc.  We'd need to implement that to support polling arbitrary pollables via `mio`, but I think that should be straightforward.

pimeys:
> d need to implement that to support polling arbitrary pollables via `mio`, but I think that should be straightforward.

Interesting. But we can already work without these wakers. Sorry for asking a bit noob questions, but does that mean our waker is woken immediately and we consume quite a bit more CPU until we get proper wakers for mio?

dicej:
> Interesting. But we can already work without these wakers. Sorry for asking a bit noob questions, but does that mean our waker is woken immediately and we consume quite a bit more CPU until we get proper wakers for mio?

Sorry, I guess I don't understand the question.  `poll(2)` will block until at least one of the file descriptors you pass it is ready (or a timeout occurs).  There's no need to busy wait and waste CPU or anything like that.

What I'm proposing is that `mio` uses `poll(2)` for WASIp2, which should be only slightly less efficient (due to an extra layer of abstraction and the necessity of converting between file descriptors and pollables) than calling the native WASIp2 `wasi:io/poll#poll` function directly.  I expect the overhead will be negligible.

Can you clarify what you mean by "our waker is woken immediately"?  It should only be woken when the pollable is ready.

pimeys:
> > Interesting. But we can already work without these wakers. Sorry for asking a bit noob questions, but does that mean our waker is woken immediately and we consume quite a bit more CPU until we get proper wakers for mio?
> 
> Sorry, I guess I don't understand the question. `poll(2)` will block until at least one of the file descriptors you pass it is ready (or a timeout occurs). There's no need to busy wait and waste CPU or anything like that.
> 
> What I'm proposing is that `mio` uses `poll(2)` for WASIp2, which should be only slightly less efficient (due to an extra layer of abstraction and the necessity of converting between file descriptors and pollables) than calling the native WASIp2 `wasi:io/poll#poll` function directly. I expect the overhead will be negligible.
> 
> Can you clarify what you mean by "our waker is woken immediately"? It should only be woken when the pollable is ready.

I mean if we do not implement or use any waker in mio. If we cannot use the file descriptor waker yet from the libc-wasi, should we expect some issues? How is it currently working in your branch, I don't see you enabling any of the mio's waker implementations.

dicej:
> I mean if we do not implement or use any waker in mio. If we cannot use the file descriptor waker yet from the libc-wasi, should we expect some issues? How is it currently working in your branch, I don't see you enabling any of the mio's waker implementations.

Are you talking about https://docs.rs/mio/1.0.2/mio/struct.Waker.html?  That's only relevant for multithreaded systems, and neither WASIp1 nor WASIp2 support multithreading.

dicej:
> > I mean if we do not implement or use any waker in mio. If we cannot use the file descriptor waker yet from the libc-wasi, should we expect some issues? How is it currently working in your branch, I don't see you enabling any of the mio's waker implementations.
> 
> Are you talking about https://docs.rs/mio/1.0.2/mio/struct.Waker.html? That's only relevant for multithreaded systems, and neither WASIp1 nor WASIp2 support multithreading.

Ah, now I understand what you're getting at.  I see now that [sys/unix/selector/poll.rs](https://github.com/tokio-rs/mio/blob/master/src/sys/unix/selector/poll.rs) uses `Waker` internally and the existing implementations in `sys/unix/waker` aren't suitable.  Sorry, I misunderstood your original question completely.

Yeah, if `poll.rs` fundamentally assumes multithreading then it won't be a good fit.  In that case, we can still aim for a simpler implementation than this current PR has by using `poll(2)`, but it would look different from `poll.rs`.

dicej:
> The only missing link is the missing `eventfd` syscall, meaning we cannot use a file descriptor for waking up the poll.

Even if we had that, of what use would it be?  The only purpose of `Waker` is to allow a thread to wake another thread which is blocked on a `Selector::select` call, but since WASI is single threaded that's not possible anyway.  So I think we'd either need to modify `selector/poll.rs` to (conditionally) support single-threaded operation or pursue the second option you listed.

pimeys:
Today;s progress: cfg-gated the waker out from the poll.rs impl. Got myself to an endless loop sucking all CPU when running the postgres example. Probably going to continue a bit on Monday, but also thinking would it make sense to just clean the code a bit from this branch and try to get it approved?

It would be very cool to unlock the whole tokio ecosystem for wasi components...

dicej:
> Got myself to an endless loop sucking all CPU when running the postgres example.

Yeah, an early version of this patch had a similar issue due to an edge-triggering vs. level-triggering mismatch.  That's what motivated [this code](https://github.com/tokio-rs/mio/blob/fae1ac4baa174d3fd596b54a8a72e7429acd5077/src/sys/wasi/p2.rs#L306-L310).

pimeys:
Eventually I used this PR from you @dicej to implement our proof of concept. My take is to wait a bit and implement the needed IO in host if needed. I wrote down the current issues in a gist, if somebody's interested:

https://gist.github.com/pimeys/856846ff8718247d79f1557ed82d80f7

Snowiiii:
What is missing here ?

StaticallyTypedAnxiety:
Was wondering about this as well. Is it still on the radar? Might need it for a future project

dicej:
> > Was wondering about this as well. Is it still on the radar? Might need it for a future project
> 
> @dicej 👀

I'm happy to answer any questions about the code in this PR, but questions about if, when, and under what conditions WASIp2 support could be merged into Mio should be directed to the Mio maintainers.

banocean:
> > > Was wondering about this as well. Is it still on the radar? Might need it for a future project
> > 
> > 
> > @dicej 👀
> 
> I'm happy to answer any questions about the code in this PR, but questions about if, when, and under what conditions WASIp2 support could be merged into Mio should be directed to the Mio maintainers.

Due to amount of time this PR has been sitting unattended I through there is something missing in it, sorry

banocean:
Also there seems to be some conflicts and failing ci

dicej:
> Also there seems to be some conflicts and failing ci

I went ahead and rebased both this and the corresponding Tokio PR, and also updated https://github.com/dicej/wasi-sockets-tests to use the rebased versions.

src/io_source.rs
Thomasdezeeuw: We could increase our MSRV to 1.82 (released November, 2024) so we can use @Darksonn's suggestion.

Thomasdezeeuw (commented):
> That's an artifact of how WASIp2's [poll](https://github.com/WebAssembly/wasi-io/blob/3191bc6738d99a022c9233fed2025147bd3a3faa/wit/poll.wit#L46) function is defined, along with the underlying [ABI](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md), which determines how the host allocates and returns a `list<u32>` from a function. Specifically, the host calls a `cabi_realloc` function exported by the guest to do the allocation.

I find it quite strange to have allocation per system call (runtime call?). Wouldn't those allocation hurt performance?

> It's possible to optimize that by pre-allocating an conservatively-sized buffer before calling `poll`, storing a pointer to it in a global, and then returning that pointer when the host calls `cabi_realloc`. Then we could reuse that buffer across `Selector::select` calls, eliminating any per-call allocation overhead. The downside is that we'd have to write that code by hand rather than use the auto-generated bindings from the `wasi` crate, but maybe it's worth it in this case?

So we would be intercepting the host's call to allocate memory? That seems way too complicated, let alone intrusive for a library in my opinion.

> > Looking at wasi-p2 to me it seems we can change `poll_oneoff` to `poll` and it should mostly work (though I would expect certain edge cases).
> 
> The main difference between the WASIp1 and WASIp2 APIs is that the latter doesn't have any types corresponding to p1's [subscription](https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#-subscription-record) and [eventtype](https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#-eventtype-variant) types, etc. In p1, those types allowed the caller to tell `poll_oneoff` which events were of interest (in the case of `subcription`) and `poll_oneoff` to tell the caller which events actually occurred (in the case of `eventtype`). In p2, we must create and pass a separate `pollable` handle to `poll` for each event of interest.
> 
> In the case of p2, the `pollable` type is opaque, so we need to remember which socket (or other poll-able thing) and operation (e.g. read, write, or accept) that `pollable` represents, which accounts for some of the differences between `p1.rs` and `p2.rs`. The other differences are related to the fact that p2 simply supports more things (e.g. making outgoing connections) than p1 did.

The more I read about the `poll` function from preview2 the more confusing I find it. 

> Meanwhile, WASIp3 is expected to arrive early this year, and later this year we'll [have multithreading, too](https://github.com/WebAssembly/wasi-sdk/issues/589). At that point, we really could use `src/sys/unix/selector/poll.rs` as-is, i.e. treat `wasm32-wasip3` as just another POSIX OS. So the fallback plan if we can't reach consensus on this PR is to skip `wasm32-wasip2` and focus on p3 instead.

Maybe we should wait for preview 3 then and don't support preview 2. I don't how realistic that is though as I have no idea how fast adoption of new previews usually is.

Thomasdezeeuw:
I agree with @Darksonn on not having a readiness API, Mio (version 1) will not work with a completion based API. I'm experimenting with completion based APIs in A10 (https://github.com/Thomasdezeeuw/a10), mainly focussing on io_uring. But for it to work (without too many footguns) we need to expose an entirely different API.

So for Mio to support WASI preview 3 / 1.0 Mio would need a readiness API similar to `poll(2)` on POSIX (or something like kqueue or epoll).

Thomasdezeeuw:
@Darksonn already mentioned most of what I wanted to say, but I'll add one more point. If no one with wasm specific knowledge is able to help maintain the code then I can't accept a 1k line change, I simply don't have the capacity to maintain it.

I think the `poll(2)` based approach would be best as well. If we need to add some wasm specific code that's fine. We do something similar for illumos, which uses `epoll(2)` but it's implemented slightly differently than on Linux. In such a case I still consider it the "same" implementation.

dicej:
I've got the `mio` test suite running against my new implementation (minus the tests which require `thread::spawn` and/or UDP broadcast, neither of which are available for WASIp2).  A few of them were failing due to what turned out to be a bug in `wasi-libc`, which I've fixed locally but still need to write a C test before posting a PR to that repo.  I'll do that tomorrow and also get the `tokio` test suite running in case that shakes out any other `wasi-libc` bugs.

Once that's all done, it will take some time to get a new `wasi-sdk` released and pull that into the Rust release train, so in parallel with that I'll post new `mio` and `tokio` PRs with the broken-by-wasi-libc tests temporarily `#[ignore]`d.  Later, when the wasi-libc fix(es) land in a Rust stable release we can un-`#[ignore]` them.

dicej:
Apologies for the delay; it ended up being a busy week.  Here's the new PR: https://github.com/tokio-rs/mio/pull/1931.  Please direct further discussion there.

Issue #1803, opened 2024-06-09
Use std::os::fd on Hermit


Thomasdezeeuw:
Pr #1802 switched most code from `std::os::fd` to use `std::os::hermit::io` because of https://github.com/rust-lang/rust/issues/126198. Revert that back.

Depends on https://github.com/rust-lang/rust/issues/126198 being fixed.

Thomasdezeeuw:
This fix is in Rust 1.81: https://github.com/rust-lang/rust/commit/cd4c0f06daddd6f3b1f18ede64584f1505f1bc9b.

Thomasdezeeuw:
Reopen #1841 once we bump our MSRV past 1.81.

PR #1841, opened 2024-10-25
fix: use std::os::fd on HermitOS

joeydewaal:
Issue https://github.com/rust-lang/rust/issues/126198 was fixed and added to 1.81.

fixes: #1803


Thomasdezeeuw:
Our MSVR is 1.70, so this will have to wait until we bump it.

Thomasdezeeuw:
Since that bump will be a long time from now, I'm closing this. Thank you @joeydewaal, we'll reopen this once we bump the MSRV.

PR #1853, opened 2024-12-28
Process polling

igankevich:
This PR adds process polling capabilities via `Process` type that implements `Source`.
* On Linux the polling is done via [`pidfd`](https://man7.org/linux/man-pages/man2/pidfd_open.2.html).
* On MacOS/BSD it is done via `EVFILT_PROC` filter of [`kqueue`](https://man.freebsd.org/cgi/man.cgi?query=kqueue&sektion=2). 
* On Windows it is done via [job objects](https://learn.microsoft.com/en-us/windows/win32/api/jobapi2/nf-jobapi2-assignprocesstojobobject).

To implement `Process` I made the following changes.
* No changes to `epoll`-based and `poll`-based selectors: `pidfd` is like any other file descriptor.
* Minor changes to `kqueue`-based selector and related `is_*` functions to support new `EVFILT_PROC` filter and prevent code repetition.
* Multiple changes to Windows selector. This selector now supports registering arbitrary handles to be polled. Handle's address is used as the token (`lpCompletionKey`), and each selector maintains a mapping between the handles and the associated data `HandleInfo`. For processes the associated data consists of user-provided token. Currently only processes are registered as handles, but everything else can be re-implemented using handles to simplify the code (e.g. `Arc<CompletionPort>` will no longer be needed in `NamedPipe` and it will be possible to register the same pipe in several selectors). Please let me know if this is something you want to be done in this PR.
* `tests/process.rs` test.
* Changes to `tests/util` module to support large number of events.

Besides process polling this PR includes fixes suggested by the latest Clippy version.

*EDIT*

The motivation behind non-blocking process polling is to simplify projects that use `mio` to collect output from child processes (e.g. test harnesses, virtual networks). Currently such projects either use a separate thread to wait for child processes via `wait` syscall (that waits for _any_ process to exit) or `poll` with small timeout and check whether the processes exited after each `poll`. The former needs mutexes to synchhorinze the waiting thread with the poller, and the latter is as efficient as active polling. With proper process polling such projects would use the same thread to monitor process status, this would greatly simplify their code bases.

*EDIT 2*

I tested the changes on Linux, MacOS, FreeBSD and Windows using the same commands as in CI and latest Rust version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,10 +40,10 @@ os-ext = [
   "windows-sys/Win32_System_Pipes",
   "windows-sys/Win32_Security",
 ]
+# Enables OS process polling.
+os-proc = ["os-poll"]
 # Enables `mio::net` module containing networking primitives.
 net = []
-# Enables OS process polling.
-process = []
 
 [dependencies]
 log = { version = "0.4.8", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,12 @@ os-ext = [
   "windows-sys/Win32_Security",
 ]
 # Enables OS process polling.
-os-proc = ["os-poll"]
+os-proc = [
+  "os-poll",
+  "windows-sys/Win32_Security",
+  "windows-sys/Win32_System_JobObjects",          
+  "windows-sys/Win32_System_SystemServices",
+]
 # Enables `mio::net` module containing networking primitives.
 net = []
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,8 @@ os-ext = [
 ]
 # Enables `mio::net` module containing networking primitives.
 net = []
+# Enables OS process polling.
+process = []
 
 [dependencies]
 log = { version = "0.4.8", optional = true }
--- a/Makefile
+++ b/Makefile
@@ -19,9 +19,11 @@ test_all: check_all_targets
 # compile correctly.
 test_sanitizer:
 	@if [ -z $${SAN+x} ]; then echo "Required '\$$SAN' variable is not set" 1>&2; exit 1; fi
-	RUSTFLAGS="-Z sanitizer=$$SAN -Z sanitizer-memory-track-origins" \
-	RUSTDOCFLAGS="-Z sanitizer=$$SAN -Z sanitizer-memory-track-origins" \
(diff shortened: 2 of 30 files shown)


igankevich:
Fixed (hopefully) the errors reported in CI. Had to use `-Zexport-executable-symbols` to make leak sanitizer work in CI (https://github.com/rust-lang/rust/issues/111073).

igankevich:
Thanks for approving the pipeline. Hopefully I fixed the flaky test :)

It seems that leak sanitizer doesn't work with panicking tests, so I disabled them when compiled under this sanitizer. The error that we get here might be related to a [custom allocator](https://groups.google.com/g/thread-sanitizer/c/nmhRglhlohI). I'm not sure `mio` has something to do with this error. I couldn't find any Rust-related issue with the same error.

Thomasdezeeuw:
@igankevich is it possible for this to live outside of Mio? I've implemented process polling before (but not for Windows) and Tokio also has an implementation. At this point I'm not really interested in maintaining more code, so I'm only willing to accept it if it can't live outside of Mio and has you (or someone else) as maintainer for it.

igankevich:
> @igankevich is it possible for this to live outside of Mio? I've implemented process polling before (but not for Windows) and Tokio also has an implementation. At this point I'm not really interested in maintaining more code, so I'm only willing to accept it if it can't live outside of Mio and has you (or someone else) as maintainer for it.

Thanks for your reply Thomas!

For better or worse the project where I originally needed that functionality is fast-moving and has already switched to `tokio` from `mio`.

From what I know it's technically difficult to have portable process polling functionality outside (on top of) `mio` because different operating systems implement it differently: on Linux it's possible because the process is just a file descriptor, but this isn't the case on MacOS and Windows. The workaround is to do active polling, i.e. check periodically that the process is alive with [`wait(2)`](https://man7.org/linux/man-pages/man2/wait.2.html).

I thought that this PR would be easy, but Windows API complexity inflated it to over 1k lines. I'm not really an expert in Windows, so I'm not sure about maintaining.

Thomasdezeeuw:
> I thought that this PR would be easy, but Windows API complexity inflated it to over 1k lines. I'm not really an expert in Windows, so I'm not sure about maintaining.

We have plenty of Window specific issues as it is and no maintainer for it. So if you can't help maintain this then I'm afraid I can't accept it at this time.

Issue #1834, opened 2024-10-03
Add AIX back to CI after full stdlib support


mustartt:
Currently Rust on AIX is still under development and is missing some upstream stdlib implementations causing failed CI tests. Once `powerpc64-ibm-aix` becomes a tier 2 target, CI can be enabled.

Thomasdezeeuw:
This is specifically about uncommentating this line: https://github.com/tokio-rs/mio/pull/1833/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR155.

PR #1867, opened 2025-03-11
[AIX] fix incorrectly set nonblocking socket

mustartt:
During the initial port, setting `O_NONBLOCK` was missed in some places for AIX causing unexpected hangs in the initial port. So we are adding them back now. 

Also re-enable ci as per https://github.com/tokio-rs/mio/issues/1834 we have fully implemented all missing items for `std` on AIX.


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @mustartt

PR #1869, opened 2025-04-11
chore: use `std::os::fd` for `HermitOS`

hulxv:
Given that [rust-lang#126198](https://github.com/rust-lang/rust/issues/126198) was finally resolved, `HermitOS` should use `std::os::fd`. This PR complete this to-do.


Thomasdezeeuw:
Our MSRV is still 1.70, so we still can't do this. Same as in https://github.com/tokio-rs/mio/pull/1841.

Thomasdezeeuw:
Closing this, we can reopen once we bump our MSRV.

PR #1871, opened 2025-04-14
add cygwin support

Berrysoft:
This PR adds cygwin support. It uses the POSIX poll as backend. I tried my best to fix the tests. Some of them are ignored reasonably, while some of them are not - related to the `UnixListener` and `UnixStream`. I don't understand them yet.

### DIFF
--- a/tests/unix_listener.rs
+++ b/tests/unix_listener.rs
@@ -89,6 +89,7 @@ fn unix_listener_register() {
 }
 
 #[test]
+#[cfg_attr(target_os = "cygwin", ignore)]
 fn unix_listener_reregister() {
     let (mut poll, mut events) = init_with_poll();
     let barrier = Arc::new(Barrier::new(2));
@@ -116,6 +117,7 @@ fn unix_listener_reregister() {
 }
 
 #[test]
+#[cfg_attr(target_os = "cygwin", ignore)]
 fn unix_listener_deregister() {
     let (mut poll, mut events) = init_with_poll();
     let barrier = Arc::new(Barrier::new(2));
--- a/tests/unix_stream.rs
+++ b/tests/unix_stream.rs
@@ -37,12 +37,14 @@ fn unix_stream_send_and_sync() {
     target_os = "hurd",
     ignore = "getting pathname isn't supported on GNU/Hurd"
 )]
+#[cfg_attr(target_os = "cygwin", ignore)]
 fn unix_stream_smoke() {
     #[allow(clippy::redundant_closure)]
     smoke_test(|path| UnixStream::connect(path), "unix_stream_smoke");
 }
 
 #[test]
+#[cfg_attr(target_os = "cygwin", ignore)]
 fn unix_stream_connect() {
     let (mut poll, mut events) = init_with_poll();
     let barrier = Arc::new(Barrier::new(2));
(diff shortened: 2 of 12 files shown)


tests/unix_listener.rs
```diff
@@ -89,6 +89,7 @@ fn unix_listener_register() {
 }
 
 #[test]
+#[cfg_attr(target_os = "cygwin", ignore)]
```
Thomasdezeeuw, line 92: Why are these ignored?

tests/unix_stream.rs
```diff
@@ -37,12 +37,14 @@ fn unix_stream_send_and_sync() {
     target_os = "hurd",
     ignore = "getting pathname isn't supported on GNU/Hurd"
 )]
+#[cfg_attr(target_os = "cygwin", ignore)]
```
Thomasdezeeuw, line 40: Why are these ignored?

Thomasdezeeuw (commented):
Mostly LGTM. For the ignored tests can you add a reason why they are ignored.

tests/unix_listener.rs
Berrysoft: The `connect` method always blocks because there's no `accept`. Thus, the spawned thread never waits for the barrier, making the testing thread block.

tests/unix_stream.rs
Berrysoft: Some of them are ignored mistakenly.

I leave two of them ignored. The `accept` method needs `connect` to handshake, but the non-blocking `connect` doesn't handshake. It postpones this process to a `poll`. The ignored tests doesn't poll the client socket, so the spawned thread always blocks.

tests/unix_listener.rs
```diff
@@ -89,6 +89,7 @@ fn unix_listener_register() {
 }
 
 #[test]
+#[cfg_attr(target_os = "cygwin", ignore = "Cygwin blocks on connect")]
```
Thomasdezeeuw, line 92: We need to fix this, otherwise Mio won't work properly.

tests/unix_stream.rs
```diff
@@ -183,6 +183,7 @@ fn unix_stream_pair() {
     target_os = "hurd",
     ignore = "getting pathname isn't supported on GNU/Hurd"
 )]
+#[cfg_attr(target_os = "cygwin", ignore = "nonblocking connect doesn't handshake")]
```
Thomasdezeeuw, line 186: Not sure what this means.

tests/unix_listener.rs
Berrysoft: Welcome to join the discussion, then:)

https://inbox.sourceware.org/cygwin/Z_UERXFI1g-1v3p2@calimero.vinschen.de/T/#t

TL;DR: They provide a workaround, which sacrifices `PEERCRED`; and they don't think it's a bug.

Berrysoft: Well I think it's cygwin's problem, and mio doesn't have to try to fix it. Cygwin target is only tier-3, which means anyone can practically break it without any doubt. This PR just wants to make it compile and "nearly" work on cygwin (and it really works well except unix sockets). If there is any person complaining about the bug or misbehavior of mio on cygwin, just assign it to me:) There won't be many people concern about this target, though.

Berrysoft:
CI is failing but it's not my problem. The master branch also fails.

Berrysoft:
I proposed a fix for CI in #1883 .

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @Berrysoft

PR #1875, opened 2025-05-05
Make `Poll` a `Source` itself

evyatark2:
Both epoll and kqueue support listening for other epoll/kqueue file descriptors.

They become readable when an event in the inner epoll/kqueue is ready.

We can expose this feature by making `Poll` an `event::Source`.

I gated this feature behind `os-ext` as this was required to use `IoSource`.

Additionally, thanks to Rust's borrow checker, there can be no cycles in the listening graph i.e. this will not compile:
```rust

poll.registry().register(&mut poll, Token(0), Interest::READABLE).unwrap();
```

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -23,6 +23,9 @@ use std::{fmt, io};
 
 use crate::{event, sys, Events, Interest, Token};
 
+#[cfg(all(unix, feature = "os-ext"))]
+use crate::io_source::IoSource;
+
 /// Polls for readiness events on all registered values.
 ///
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw (changes requested):
Since this is only supported by epoll and kqueue I'm sure we want to do this.

This is also already possible by using [`SourceFd`](https://docs.rs/mio/1.0.2/mio/unix/struct.SourceFd.html) and the [`AsRawFd`](https://docs.rs/mio/1.0.2/mio/struct.Poll.html#impl-AsRawFd-for-Poll) implementation for `Poll`.

PR #1878, opened 2025-05-19
made 'Events' an alias for 'Vec<Event>'

### LINKED ISSUE #1877: Events is no longer needed
i opened this issue earlier but i learned some new info just after posting it so i panicked and closed the issue, but 
that turned out to be an overreaction and this post needed very little editing to be made correct

in the meantime i've actually implemented this already, please have a look at the PR (it's very small) & discuss any 
implementation related isssues oer' there

`Events` used to be responsible for abstracting over the differences between windows and posix events, but the
responsibility has since been moved to `Event` yet `Events` continues to exist as (for all intents and purposes)an 
opaque wrapper around `Vec<Event>` that provides a very limited subset of Vec's capabilities for no safety or 
correctness reason

defining `Events` as an alias for `Vec<Event>` & removing the entirety of events.rs would have multiple benefits: 
- we would shed the responsibility of providing a bespoke storage API for events, Vec<Event> works fine 
- the above point applies even more-so to docs than code, the current docs are very well-written but they shouldn't 
exist, and any useful example code in them could be recontextualized somewhere else
- users could use all the useful methods provided by Vec, making downstream application code more sane
- internal use of Vec<Event> would be made more clear by the reduction in type-shuffling noise
- instead of storing events as `sys::Event` (an alias for usize(sic, this depends on the target)) we can store them with 
safer types and simply downcast as appropriate (this is only necessary at the places where we used to call `Events.sys()`, and it's actually just a single place so i inlined it for now in my PR, it can be factored out again if the need arises 

the current situation is that `Events` is a wrapper around `Vec<sys::Event>` but `sys::Event` isn't exposed through
the public API & has an identical in-memory repr as `Event`

Carlyle-Foster:
this closes #1877


Carlyle-Foster:
this ain't gonna work

PR #1879, opened 2025-05-19
added 'Event.as_vec()' & 'Events.as_vec_mut()'

Carlyle-Foster:
these provide a useful "escape hatch" for any methods of `Vec` that we've yet to create wrapper methods for, i also implemented `From<&sys::Events>` & `From<&mut sys::Events>` for `&Vec<sys::Events>` & `&mut Vec<sys::Events>` respectively, so whereas before we had to implement methods of `Vec` both on `Events` & `sys::events` now we can just implement them on `Events`, i plan to eventually implement all the methods of `Vec` for `Events`

### DIFF
--- a/src/event/event.rs
+++ b/src/event/event.rs
@@ -182,15 +182,6 @@ impl Event {
     pub fn is_lio(&self) -> bool {
         sys::event::is_lio(&self.inner)
     }
-
-    /// Create a reference to an `Event` from a platform specific event.
-    pub(crate) fn from_sys_event_ref(sys_event: &sys::Event) -> &Event {
-        unsafe {
-            // This is safe because the memory layout of `Event` is
-            // the same as `sys::Event` due to the `repr(transparent)` attribute.
-            &*(sys_event as *const sys::Event as *const Event)
-        }
-    }
 }
 
 /// When the [alternate] flag is enabled this will print platform specific
--- a/src/event/events.rs
+++ b/src/event/events.rs
@@ -185,8 +185,7 @@ impl Events {
     /// # }
     /// ```
     pub fn clear(&mut self) {
-        let vec: &mut Vec<Event> = self.into();
-        vec.clear();
+        self.inner.clear();
     }
 
     /// Returns a shared reference to the underlying Vec
--- a/src/event/events.rs
+++ b/src/event/events.rs
@@ -2,6 +2,7 @@ use crate::event::Event;
 use crate::sys;
 
 use std::fmt;
+use std::mem::transmute;
 
 /// A collection of readiness events.
 ///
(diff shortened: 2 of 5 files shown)


Carlyle-Foster:
i know that the `shell/selector.rs` shouldn't have implementations, but for some reason the doctests call methods of `sys::Event` without `os_poll`?

Thomasdezeeuw (commented):
We don't want to expose the `Events` is a `Vec` in case we ever want to change that.

What is the motivation for this change?

Carlyle-Foster:
because i think the utility of having the full set of `Vec` methods on `Events` far outweighs the flexibility of hiding the implementation, `Events` serves such a specific and limited job that i think the chance that `Vec` will ever be insufficient is basically nil

Carlyle-Foster:
all the methods of `Vec` can sometimes be useful, but for one particular example when you u `Events` in a struct and then try to iterate over those events, the iterator locks up the struct by storing a shared ref inside itself but if u try to fix the problem by using `mem::take()` you can't because it doesn't implement default like `Vec` does

Carlyle-Foster:
but really, in time you'll want the full set of list operations for anything that is, conceptually, a list

Carlyle-Foster:
not to mention that it would remove the need to document `Events` because the operations it supported would be obvious, the only reason it needs docs rn is because the subset of `Vec`'s methods it provides is completely arbitrary

Thomasdezeeuw:
> in general, u can't keep the iterator around for very long because it keeps a locks things in place

Where do you get this from? `Events` doesn't use a lock.

Carlyle-Foster:
i wan referring to the effect of keeping a shared reference around, it locks the thing it's referring to in place so u can't mutate it

Carlyle-Foster:
i probably shouldn't use such overloaded terminology, but i find it helpful to think of references as compile-time locks

Carlyle-Foster:
`get()` & `get_mut()` are by far the most important missing methods, and since we already expose a non-consuming iterator worse comes to worse we can implement it on top of that

Carlyle-Foster:
why wouldn't the use-cases for `Vec.get()` & `Vec.get_mut()` not apply to `Events`? just because it doesn't commit to exposing every method of `Vec` doesn't mean it isn't basically just a `Vec` with a limited API to leave room for implementation changes

Thomasdezeeuw:
@Carlyle-Foster if you don't have a concrete use case for exposing additional API we're not going to support it. Since you haven't shared a use case despite me asking multiple times I'm going to close this prs.

Carlyle-Foster:
as to your question, my specific use case is that i'm writing a libray that has a `Server` struct with a `serve()` method that goes into an event-loop and returns to the calling application code to provide hooks on events. the `Server` is responsible for preserving the state between calls, so it stores both an `Events` and a map of `ID` to `Client` & that's where my troubles atart, i need to iterate over the events and potentally drop clients from the map while doing so

PR #1880, opened 2025-05-19
refactored 'Events' & 'sys::Events' related code

Carlyle-Foster:
before this change implementing a method for `Events` (in practice always a wrapper around a mthod of `Vec`) required also implementing the same wrapper for the implementations of `sys::Events` in the kqueue and IOCP based targets, where `sys::Events` isn't an alias for `Vec<Event>`

this change implements `From<&sys::Events>` & `From<&mut sys::Events>` for `Vec` with both of those targets,

### DIFF
--- a/src/sys/unix/selector/kqueue.rs
+++ b/src/sys/unix/selector/kqueue.rs
@@ -364,7 +364,7 @@ impl DerefMut for Events {
 
 impl fmt::Debug for Events {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "Events: TODO: actually print them")
+        write!(f, "Events(Vec<libc::kevent>)")
     }
 }
 


Carlyle-Foster:
> I don't want to force the layout of the `Events`. There might be implementations that don't want to use a `Vec`, that's the reason why it's implemented in the `sys` module and a cross-OS layer is provided on top of that.

this isn't a public-facing commitment though, it can be reverted without a breaking change if we ever have to support an implementation that can't have it's `sys::Events` contain an underlying `Vec<Event>`

Carlyle-Foster:
..and not all the methods of `Event` have to be delegated to `Vec` with this approach,  rn `Event.with_capacity()` & `Events.clear()` are still delegated to `sys::Events` rather than `Vec` because IOCP needs custom implementations that handle the completion ports

src/event/events.rs
```diff
@@ -228,3 +231,20 @@ impl fmt::Debug for Events {
...
+        unsafe { transmute(inner) }
+    }
+}
+
+impl<'a> From<&'a mut Events> for &'a mut Vec<Event> {
+    fn from(value: &'a mut Events) -> Self {
+        let inner: &mut Vec<sys::Event> = (&mut value.inner).into();
+        unsafe { transmute(inner) }
+    }
+}
```
Thomasdezeeuw, line 250: We don't want to support this.

Thomasdezeeuw (changes requested):

Thomasdezeeuw:
This is not a direct I want to take `Events` into.

Carlyle-Foster:
my mistake, i forgot that the trait impl would be public, i've fixed it now by using a private method instead

PR #1887, opened 2025-06-28
Fix various lints

Thomasdezeeuw:


Darksonn (approved):

Thomasdezeeuw:
This is already an improvement so merging this and when I have some time I'll fix the other issues.

PR #1883, opened 2025-06-05
Fix CI error by cfg some code out

Berrysoft:
Try to fix CI.

PR #1895, opened 2025-08-21
fix: peek reregister after would block

discord9:
CC #1755 https://github.com/tokio-rs/tokio/issues/3789

as title, use a fix that on windows `peek` should reregister after ~either success or~ would block, so that it's behavior would be the same as on linux(see ~`peek_readiness`~`peek_ok` and `peek_would_block`).

This should fix windows peek problem, @Darksonn might want to take a look.

The problem it seems is that in `src\sys\windows\selector.rs` `SockState::feed_event` mio have to simulate Edge-triggered behavior and clean readable interest after send a readable interest, but since peek is also a kind of special read, it also cancelled interest to readable event, hence cause the blocking peek bug on windows because interest is lost so blocking is forever

__edited__: found out that under linux with `cfg(mio_unsupported_force_poll_poll)` peek would also have similiar problem so just write `do_io` on all platform instead
# Alternatives
only do `do_io` on `#[cfg(any(windows, mio_unsupported_force_poll_poll))]` which also seem to pass all tests on all platform/cfg combination

### DIFF
--- a/tests/tcp_stream.rs
+++ b/tests/tcp_stream.rs
@@ -899,11 +899,11 @@ fn peek_ok() {
 
     assert_eq!(stream2.write(&[0]).unwrap(), 1);
     // peek multiple times until we get a byte
-    loop{
+    loop {
         let res = stream1.peek(&mut buf);
-        match res{
+        match res {
             Ok(1) => break,
-            Err(err) if err.kind() == io::ErrorKind::WouldBlock =>{continue}
+            Err(err) if err.kind() == io::ErrorKind::WouldBlock => continue,
             _ => panic!("Unexpected error: {:?}", res),
         }
     }
--- a/tests/tcp_stream.rs
+++ b/tests/tcp_stream.rs
@@ -945,3 +945,82 @@ fn peek_readiness() {
     );
     assert_eq!(stream1.peek(&mut buf).unwrap(), 0);
 }
+
+#[test]
+fn peek_ok(){
+    let mut buf = [0; 2];
+    let (mut poll, mut events) = init_with_poll();
+
+    let listener = net::TcpListener::bind(any_local_address()).unwrap();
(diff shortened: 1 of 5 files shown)


Darksonn:
Reregister is an expensive operation and I don't think that's the solution we want.

discord9:
> Reregister is an expensive operation and I don't think that's the solution we want.

How about handle peek event specially in `feed_event` on windows somehow? Also the CI test failed strangely, I don't understand why, the [TestPoll](https://github.com/tokio-rs/mio/actions/runs/17116372497/job/48557753295?pr=1895#logs) failed with not found expected event)
but the [Test (ubuntu-latest)](https://github.com/tokio-rs/mio/actions/runs/17116372497/job/48557753207?pr=1895#logs) successed, I don't understand why, also I don't think my patch affect behavior on linux platform so even more confused

discord9:
the falling CI seems related to #1883 and not caused by this patch

Thomasdezeeuw (approved):

tests/tcp_stream.rs
```diff
@@ -876,3 +876,131 @@ fn send_oob_data<S: AsRawFd>(stream: &S, data: &[u8]) -> io::Result<usize> {
...
+        vec![ExpectEvent::new(ID1, Readiness::READABLE)],
+    );
+
+    assert_eq!(stream1.read(&mut buf).unwrap(), 1);
+    assert_would_block(stream1.peek(&mut buf));
+
+    assert_eq!(stream2.write(&[0, 1, 2, 3]).unwrap(), 4);
+
+    // this panic with no event on windows if not re-register after would block peek
+    // becuase mio simulate edge-triggered behavior in `SockState::feed_event` and need reregister
```
Thomasdezeeuw, line 959:
```suggestion
```

Thomasdezeeuw:
Thanks @discord9

PR #1901, opened 2025-09-10
Update `windows-sys` and dev-dependencies

a1phyr:
- `windows-sys` to 0.61 (Requires MSRV 1.71)
- dev-dependencies: `env_logger` to 0.11 and `rand` to 0.9

It would be nice to move from the old `wasi` version to the new `wasip1` crate but this requires MSRV 1.82, so I'm not sure this is acceptable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
       with:
         # NOTE: full version (including .0) to work around
         # <https://github.com/dtolnay/rust-toolchain/issues/112>.
-        toolchain: 1.70.0
+        toolchain: 1.71.0
     - name: Check
       # We only run check allowing us to use newer features in tests.
       run: cargo check --all-features
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 edition = "2021"
-rust-version = "1.70"
+rust-version = "1.71"
 name = "mio"
 # When releasing to crates.io:
 # - Update CHANGELOG.md.
@@ -53,7 +53,7 @@ libc = "0.2.159"
 libc = "0.2.159"
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = "0.60"
+version = "0.61"
 features = [
   "Wdk_Foundation",                   # Required for AFD.
   "Wdk_Storage_FileSystem",           # Required for AFD.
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -82,7 +82,7 @@ jobs:
       with:
         # NOTE: full version (including .0) to work around
         # <https://github.com/dtolnay/rust-toolchain/issues/112>.
-        toolchain: 1.70.0
+        toolchain: 1.71.0
```
Thomasdezeeuw, line 85: We're not going to update our MSRV for these dev-dependency updates.

.github/workflows/ci.yml
```diff
@@ -196,7 +196,10 @@ jobs:
         sanitizer: [address, leak, memory, thread]
     steps:
     - uses: actions/checkout@v4
-    - uses: dtolnay/rust-toolchain@nightly
+    # FIXME: revert to using the latest nightly when https://github.com/rust-lang/rust/issues/146465 is fixed.
+    - uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: nightly-2025-09-05
```
Thomasdezeeuw, line 202: Please revert this, we'll just ignore the failure. I don't want to use an outdated nightly version that we'll never update.

Thomasdezeeuw (changes requested):

.github/workflows/ci.yml
Thomasdezeeuw: Well, then we can't update 🤷

Thomasdezeeuw: I don't want to bump MSRV without good reason. And windows-sys is not a good reason imo, they release way too often with breaking changes and bump the MSRV too often.

a1phyr:
I updated the PR to allow both 0.60 and 0.61 versions for `windows-sys`, so the last version can be used while not breaking MSRV.

`env_logger` was only updated to 0.10.

atouchet:
Tokio is updating their MSRV to 1.71: https://github.com/tokio-rs/tokio/pull/7658. Does it make sense to do the same for Mio now?

Thomasdezeeuw (commented):
We've discussed this and decided to accept the bump in MSRV to 1.71, to unlock the update of window-sys. This does mean a lot of changes are needed:
* Bump to 1.1.0
* rust-version & MSRV to 1.71
* windows-sys -> 0.61.

Are you willing to make these changes @a1phyr ? Otherwise I can do them.

Issue #1819, opened 2024-07-27
Windows named pipes hits unreachable code during read operation


akahan:
The error occurs at this location:
https://github.com/tokio-rs/mio/blob/4a5114e518b982f49ce093be6d0d2a2ab86472d1/src/sys/windows/named_pipe.rs#L871

In `io.read` comes `State::Err` with: `Err The pipe has been ended.  (os error 109)`
This appears on both Windows 11 and Windows 10.

It reproduced almost stably when closing the laptop lid with running service and client, apparently after switching to standby mode.

If I handle this state and simply exit the `read_done` method without calling `unreachable` (and assign the same error state to `io.read` again. Perhaps it's better not to do this?), everything continues to work, seemingly normal. Any comments guys?

My environment works like this:
The Windows service creates a named pipe as an administrator (both `PipeMode::Bytes` and `PipeMode::Message` modes were tested).  After which the client connects to it and sends a message 2 times per second and reads the response.

Thomasdezeeuw:
Maybe related to #1660?

@akahan what version of Mio are you running?

akahan:
last version 1.0.1

akahan:
Even if it's a similar bug https://github.com/tokio-rs/mio/issues/1660.  In any case, it is not fixed.

akahan:
I can not understand.  If `io.read | io.write state` can be `Err`, why is it not checked here, but only `Pending` is checked, and the rest is supposedly unreachable?

akahan:
Here https://github.com/akahan/mio/tree/v1.0.1-dev is the code that I used to temporarily solve this problem.

sergioabreu-g:
I'm getting constant crashes in my application due to this, could akahan's fix be applied as it is? I've read the code but it's quite convoluted, I don't see why the `Error` state should be deemed unreachable at that point.

sergioabreu-g:
> @sergioabreu-g a pull request for this would be welcome.

I can make a PR with that change but I don't know what the consequences might be. I don't have enough knowledge of the codebase to understand why that error is there or how it should be handled.

tripplet:
I'm also observing this issue.
The scenario is the same to what @akahan describes:
A windows service with a named pipe where a client connects to, sends a some bytes and waits for some response bytes from the service before closing the pipe again.
In my case I can not reliably reproduce the error it only happens occasionally. 

[...]

PR #1903, opened 2025-10-10
fix: improve state transitions of Windows named pipes

thomaseizinger:
Disclaimer: I am by no means a Windows expert. The changes presented here are built from reading through the code and trying to reason about the possible state transitions. I've tried to explain in each commit, why I think the resulting change is safe to make.

I'll apply this to our codebase and see if it fixes the crash reports that we are getting from users. In the meantime, feedback is most welcome.

Resolves: #1819


thomaseizinger:
> I actually merged this now so it can be included in the next release: #1909 😅

Fair enough! :)

PR #1915, opened 2025-11-10
Fix/improve comments for `LISTEN_BACKLOG_SIZE`

opara-tor:
The backlog size logic that was copied from the standard library was for unix sockets, not inet sockets. Mio's backlog sizes differ greatly from the standard library for inet sockets (128 or smaller). The comments here were confusing/misleading, so I've tried to clarify them.

### DIFF
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -81,9 +81,9 @@ cfg_not_os_poll! {
     }
 }
 
-/// Define the `listen` backlog parameters as in the standard library. This
-/// helps avoid hardcoded unsynchronized values and allows better control of
-/// default values depending on the target.
+/// Define the `listen` backlog parameters. This helps avoid hardcoded
+/// unsynchronized values and allows better control of default values depending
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
Reading https://github.com/rust-lang/rust/blob/4f808ba6bf9f1c8dde30d009e73386d984491587/library/std/src/sys/net/connection/socket.rs#L538-L541 now, these values might be wrong. Might have to revert #1896 or split TCP from Unix values.

Thomasdezeeuw:
Location in the standard lib:
* Unix: https://github.com/rust-lang/rust/blob/5dbf4069dc98bbbca98dd600a65f50c258fbfd56/library/std/src/os/unix/net/listener.rs#L72
* TCP: https://github.com/rust-lang/rust/blob/5dbf4069dc98bbbca98dd600a65f50c258fbfd56/library/std/src/sys/net/connection/socket/mod.rs#L563-L575

PR #1921, opened 2025-11-29
Handle ERROR_MORE_DATA for named pipes

toksdotdev:
This MR addresses an issue where a panic occurs while attempting to read messages larger than the buffer size for Windows named pipes (this is hardcoded to 4096 bytes). 

```
panicked at C:\Users\toks\mio\src\sys\windows\named_pipe.rs:900:17:
assertion `left == right` failed
  left: 4096
 right: 0
```

### Root Cause

Per [MS docs](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-readfile#pipes):
> If a named pipe is being read in message mode and the next message is longer than the nNumberOfBytesToRead parameter specifies, ReadFile returns FALSE and [GetLastError](https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-getlasterror) returns ERROR_MORE_DATA. The **remainder of the message can be read** by a subsequent call to the ReadFile or [PeekNamedPipe](https://learn.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-peeknamedpipe) function.

Furthermore, if the provided buffer to the overlapped read operation has free capacity, **data would be copied into it**. However, since no special handling is given to `ERROR_MORE_DATA`, the [`bytes transfered == 0`](https://github.com/toksdotdev/mio/blob/9b2705ec1ad6e686a217c57801a4dd7ca09abcd2/src/sys/windows/named_pipe.rs#L926) assertion would be executed, causing the above panic (where 4096 is the number of bytes read despite an `ERROR_MORE_DATA` error).

### Solution

The core idea is to gracefully handle `ERROR_MORE_DATA`, and subsequently transition the named pipe's internal state back to a `State::Ok(buf, 0)`, where `buf` is any unread data; then repeat this until the last set of bytes of the message is copied, and `ERROR_MORE_DATA` is no longer returned. 

For added context (given this part of the code is poorly documented): the repeat until no `ERROR_MORE_DATA` is achieved when an overlapped read is scheduled via [`schedule_read(..)`](https://github.com/toksdotdev/mio/blob/9b2705ec1ad6e686a217c57801a4dd7ca09abcd2/src/sys/windows/named_pipe.rs#L715), and on an I/O completion port event, [`read_done(...)`](https://github.com/toksdotdev/mio/blob/9b2705ec1ad6e686a217c57801a4dd7ca09abcd2/src/sys/windows/named_pipe.rs#L884) being called to transition to an "end" state in preparation for the next [`Read::read(..)`](https://github.com/toksdotdev/mio/blob/eac80addc85229d7c58bdbe38e467ca717cc997b/src/sys/windows/named_pipe.rs#L527) call.

I've also added relevant tests to verify the new behaviour.

## Related Issues
- #1772 
- https://github.com/tokio-rs/tokio/issues/6460

## Other Observations

I noticed https://github.com/tokio-rs/tokio/issues/5307 is also used to track this issue. However, I believe they're two distinct problems. This handles the case where more data needs to be read, but can't because we panic mid-way, while the other is more concerned about tweaking the buffer size of the named pipe (which imo is up to the implementer).

BTW, one could achieve similar fix by leveraging a larger buffer (https://github.com/tokio-rs/tokio/issues/5307) since there'd be enough capacity for the entire message to fit in. But that has its own tradeoffs primarily due to the large buffer size.

### DIFF
--- a/tests/win_named_pipe.rs
+++ b/tests/win_named_pipe.rs
@@ -10,6 +10,9 @@ use mio::windows::NamedPipe;
 use mio::{Events, Interest, Poll, Token};
 use windows_sys::Win32::{Foundation::ERROR_NO_DATA, Storage::FileSystem::FILE_FLAG_OVERLAPPED};
 
+mod util;
+use util::{expect_events, ExpectEvent};
+
 fn _assert_kinds() {
     fn _assert_send<T: Send>() {}
     fn _assert_sync<T: Sync>() {}
@@ -369,15 +372,11 @@ fn read_message_larger_than_internal_buffer() {
     let expected_msg = vec![0x5u8; 8192];
     assert_eq!(t!(client.write(&expected_msg)), 8192);
 
-    loop {
-        t!(poll.poll(&mut events, Some(Duration::from_secs(5))));
-        let events_vec = events.iter().collect::<Vec<_>>();
-        if let Some(event) = events_vec.iter().find(|e| e.token() == Token(0)) {
-            if event.is_readable() {
-                break;
-            }
-        }
-    }
+    expect_events(
+        &mut poll,
+        &mut events,
+        vec![ExpectEvent::new(Token(0), Interest::READABLE)],
+    );
 
     let mut buf = [0u8; 4000];
     let mut actual_msg = Vec::new();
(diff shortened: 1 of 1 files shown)


tests/win_named_pipe.rs
```diff
@@ -347,3 +347,109 @@ fn reregister_deregister_different_poll() {
...
+        Interest::READABLE | Interest::WRITABLE,
+    ));
+    let mut events = Events::with_capacity(128);
+    t!(poll.poll(&mut events, None));
+
+    // Send message larger than the IPC kernel buffer (4096 bytes)
+    let expected_msg = vec![0x5u8; 8192];
+    assert_eq!(t!(client.write(&expected_msg)), 8192);
+
+    loop {
```
Thomasdezeeuw: Nit: this can use `expect_events`.

tests/win_named_pipe.rs
```diff
@@ -347,3 +347,109 @@ fn reregister_deregister_different_poll() {
...
+        Interest::READABLE | Interest::WRITABLE,
+    ));
+
+    let mut events = Events::with_capacity(128);
+    t!(poll.poll(&mut events, None));
+
+    let expected_msg = vec![1u8; 10000];
+    assert_eq!(t!(client.write(&expected_msg)), 10000);
+
+    loop {
```
Thomasdezeeuw: Same as above, can use `expect_events`.

toksdotdev:
thanks for taking a look. i've addressed the feedbacks.

PR #1926, opened 2025-12-04
Add FreeBSD 14 and 15 to Cirrus CI

Thomasdezeeuw:


Darksonn (approved):

Thomasdezeeuw:
Thanks for the info @asomers. Since we deal with a lot of targets that "randomly" fail for various reasons out side of Mio I think it's ok to keep it.

PR #1927, opened 2025-12-11
A NtCreateWaitCompletionPacket and NtAssociateWaitCompletionPacket realisation to poll CreateWaitableTimer and CreateEvent handles

eesekaj:
Hello,

Since this request doesn't break the API, I decided not to open an issue. Perhaps the implementation of this feature is not yet completely sensible, so I do not insist on its adoption.

I am working on project which uses Mio crate (from private repo but anyway). I was tidying up the "zoo" of IO subsystems of different OSes (with timers and stuff) and ran into the problem with this crate that it's impossible to poll the timer/events if handles are not files. In order to make it fully compatible with Windows, I developed some experimental code which adds a feature to 'poll' the timers and events.

A **SourceEventHndl** in source_hndl.rs is acting as a wrapper for any type which holds a 'handle' to object which was created by: **CreateEvent** or **CreateWaitableTimer**.

Windows-sys crate is missing some calls to NTDLL, so ffi.rs covers this. Tests pass successfully on Windows 10, but I don't have an opportunity to check if it runs on Windows 11 because, usually I am not writing a code for Windows. (And I don't have access to Win11 machine)

The code can be enabled using feature **os-extended**.

I tried my best not to touch the base code and integrate it as extension. For more details please check attached comments and test file **win_event.rs** and example **windows_event.rs**

Problems: 

1. There is at least one problem that I haven't solved yet: WaitCompletionPacket identification, which is carried out not by the descriptor, but by the token.
2. It is not possible to tell if handle is from Event* API. So if a wrong handle is passed then program probably would crash.
3. Every time an event triggers it should be re-associated. It is ok to de-associate **CompletionPacket** many times, but in reverse it returns the error which is general and masking it create problems.

<details>
  <summary>Example:</summary>

```rust
#[derive(Debug)]
pub struct PrimitiveTimer
{
    hndl_timer: OwnedHandle
}

impl AsHandle for PrimitiveTimer
{
    fn as_handle(&self) -> std::os::windows::prelude::BorrowedHandle<'_> 
    {
        return self.hndl_timer.as_handle();
    }
}

impl PrimitiveTimer
{
    fn new(name: &str) -> PrimitiveTimer
    {
        let mut label_cstr: Vec<u16> = name.encode_utf16().collect();
        label_cstr.push(0);

        let hndl_timer = 
            unsafe
            { 
                HANDLE(
                    CreateWaitableTimerExW(
                        null(),  
                        label_cstr.as_ptr(),
                        0,
                        EVENT_ALL_ACCESS
                    )
                )
                .try_into_owned()
                .unwrap()
            };

        return Self{ hndl_timer: hndl_timer};
    }

    fn arm_relative(&self, timeout: i64) 
    {
        let time: i64 = timeout / 100;
        unsafe
        {
            SetWaitableTimer(
                self.hndl_timer.as_raw_handle(), 
                &time as *const i64,
                0,
                None,
                null(),
                false.into()
            )
        };
    }
}

#[test]
fn test_event_win()
{
   // timer 1
    let mut se_hndl_timer = 
        SourceEventHndl::new(PrimitiveTimer::new("timer_1")).unwrap();

  //  ....

    // poll

    // registering
    poll.registry().register(&mut se_hndl_timer, Token(1), Interest::READABLE).unwrap();
}
```
</details>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ include       = [
 default = ["log"]
 
 # Enables the Event* hanle registration.
-os-extended = ["os-poll", "os-ext", "net"]
+os-extended = ["os-ext"]
 # Enables the `Poll` and `Registry` types.
 os-poll = []
 # Enables additional OS specific extensions, e.g. Unix `pipe(2)`.
@@ -113,7 +113,7 @@ required-features = ["os-poll", "net"]
 
 [[example]]
 name = "windows_event"
-required-features = ["os-ext"]
+required-features = ["os-extended", "net"]
 
 [lints.rust]
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(mio_unsupported_force_poll_poll)', 'cfg(mio_unsupported_force_waker_pipe)'] }
--- a/src/sys/windows/event.rs
+++ b/src/sys/windows/event.rs
@@ -39,17 +39,8 @@ impl Event {
             data: status.token() as u64,
         }
     }
-
-    /// Used by `source_hndl.rs`
-    #[cfg(all(target_os = "windows", feature = "os-extended"))]
-    pub(super) fn from_completion_status_event(status: &CompletionStatus) -> Event {
-        Event {
-            flags: 0,
-            data: status.bytes_transferred() as u64,
-        }
-    }
     
-    #[cfg(feature = "os-ext")]
+    #[cfg(feature = "os-poll")]
     pub(super) 
     fn to_completion_status_with_overlapped(
         &self,
--- a/src/sys/windows/iocp.rs
+++ b/src/sys/windows/iocp.rs
@@ -1,6 +1,6 @@
 //! Bindings to IOCP, I/O Completion Ports
 
-use crate::Token;
+
 use crate::sys::Event;
 use crate::sys::windows::tokens::TokenDefault;
 use crate::sys::windows::tokens::TokenSelector;
@@ -196,7 +196,8 @@ impl CompletionStatus {
     /// This function is useful when creating a status to send to a port with
     /// the `post` method. The parameters are opaquely passed through and not
     /// interpreted by the system at all.
-    fn new<S>(bytes: u32, user_token: Token, token_selector: S, overlapped: *mut Overlapped) -> Self 
+    pub(crate) 
+    fn new<S>(bytes: u32, token_selector: S, overlapped: *mut Overlapped) -> Self 
     where 
         S: TokenSelector
     {
@@ -206,29 +207,12 @@ impl CompletionStatus {
                 dwNumberOfBytesTransferred: bytes,
                 lpCompletionKey: token_selector.get_token().0,
                 lpOverlapped: overlapped as *mut _,
-                Internal: user_token.0,
+                Internal: 0,
             }
         )
     }
 
-    /*
     pub(super) 
-    fn from_event_overlapped<S>(ev: &Event, internal_token: S, overlapped: *mut Overlapped) -> Self 
-    where 
-        S: TokenSelector
-    {
-        CompletionStatus(
-            OVERLAPPED_ENTRY 
-            {
-                dwNumberOfBytesTransferred: ev.flags,
-                lpCompletionKey: internal_token.get_token().0,
-                lpOverlapped: overlapped as *mut _,
-                Internal: ev.data as usize,
-            }
-        )
-    }*/
-
-     pub(super) 
     fn from_event_overlapped(ev: &Event, overlapped: *mut Overlapped) -> Self 
     {
         CompletionStatus(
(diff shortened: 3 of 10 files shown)


eesekaj:
I decided to adapt some changes from closed source MIO crate (i.e from the organisation's private repository) to your OSS crate, but I had to change some things so as not to conflict with the interested parties  and avoid making changes to the public API.

I'd like to know what you think about these changes. And how likely is it that they will be accepted, or are those changes unsuitable?

**All changes are related to Windows only!**

Changes: 

1. Internal Token generator in  `/tokens.rs`. Produces "mapped" tokens for every IO source i.e NamedPipe, Event, Waker, Afd.
2. `./source_hndl.rs` was re-implemented. The `SourceHndl` is acting as `SourceFd`. And `SourceEventHndl` is a safer realization of the former. `SourceEventHndl` is no longer using a locked list and a callback is now used. `SourceFd` stores the data on the shared list.
3. `./seletor.rs` all code from previous PR was removed and the `feed_events()` function was re-implemented to determine the event source by mapped token.

Now, a user **provided token** and **internal token** are separated. The former is used for identification of event outside of the crate. The latter is used for internal use only. The selector's event feeder receives the **internal token** and checks to which IO/event (i.e NamedPipe, Event, Waker etc. a facility) is it mapped. Then it routes a received data for parsing to this facility. A facility is holding the user provided token in `_inner_` structures and emits the **Event** (to user) replacing an **internal token** with **user token**. In some cases i.e in _NamedPipe_, when it is required to reschedule the event, the **internal token** is stored is indicated in the **Event** in order for **selector's event feeder** to determine where to send this rescheduled event.

<details>

<summary>Token Flow diagram</summary>

```text
+--------------+                                                             
|              |                                                             
|  AFD token   ----------+                                                   
|              |         |                                                   
+--------------+         |                                                   
                         | +--------------------+                            
+--------------+         | |     OVRLAPPED_ENTRY|                            
|              |         | |                    |                            
|  Pipe token  ---------->-> lpCompletionKey    |                            
|              |    +----->|                    |                            
+--------------+    |  +--->                    |                            
                    |  |   |                    |                            
+--------------+    |  |   +---------|----------+                            
|              |    |  |             |                                       
|  Event token -----+  |             |                                       
|              |       |             |                                       
+--------------+       |    +--------v---------+                             
                       |    |                  |                             
+--------------+       |    |  IoCompletionPort<----------------------------+
|              |       |    |                  |                            |
|  Waker token --------+    +--------|---------+                            |
|              |                     |                                      |
+--------------+                     |                                      |
                            +--------v---------+                            |
                            |                  |                            |
          +------------------  event_feed()    |              Rescheduled   |
          |                 |                  |              event         |
          |                 +----------|-------+                            |
          |Internal token              |                                    |
          |                            |                                    |
          |                            |                                    |
   +------v----------+               +-v----------------+                   |
   |    WAKER        |               |    Pipe          |       ...         |
   | from_overlapped |               |  event_*()       |                   |
   |                 |               |                  |                   |
   +------|----------+               +|----------------|+                   |
          | ▯                         |                |                    |
          |           User token      |                | Internal token     |
          |                           |                |                    |
   +------v----------+          +-----v---------+     +v---------------+    |
   | Event{}         |          | Event{}       |     |▯Event{}        |    |
   |                 |          |               |     |                |----+
   |     OUT         |          |    OUT        |     |  RESCHEDULE    |     
   +-----------------+          +---------------+     +----------------+     
                                                                             
```
</details>

eesekaj:
This is a new feature which allows to poll handles created by [CreateWaitableTimerExW](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createwaitabletimerexw) and [CreateEventA](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createeventa) which are not a **file** type. 

For example **CreateWaitableTimerExW** can be used with [WaitMultipleObject](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitformultipleobjects) call or **IOCP** via [NtCreateWaitCompletionPacket](https://learn.microsoft.com/en-us/windows/win32/devnotes/ntcreatewaitcompletionpacket) and [NtAssociateWaitCompletionPacket](https://learn.microsoft.com/en-us/windows/win32/devnotes/ntassociatewaitcompletionpacket) to IOCP instance.

Another thing is a token for event identification. In order to introduce new feature, I decided to amend the token generation system and substitute it with a "mapped" internal token for internal events and resolve it to user defined token only when the event is emitted to the "poll" caller. (this is temporary solution) Because, I don't want to represent struct as "C" and count the offset using magic numbers instead (at least) using "offset_of()".  There is a better solution for that, but cannot copy it from closed source mio fork. I need to re-implement it somehow. I don't have time for that.

I need that feature to avoid  calling `WaitMultipleObject` on separate thread for timers and events. This allows to poll it in the same instance.

eesekaj:
In the recent commit I have fixed some issues with tests and errors.  But I didn't format the code and will do it later if it will be approved.

PR #1945, opened 2026-04-05
Add ping_pong example: demonstration of mio's core concept

siddharthteli12:


Darksonn:
This example is wrong, unfortunately. You cannot just invoke IO ops once in response to readiness events - you must invoke them until readiness is cleared by a `WouldBlock` error. You also can't just call `write_all` in the `read` branch, as it may fail with `WouldBlock` and you do not handle it.

PR #1950, opened 2026-04-16
fix: Handle State::None in named pipe read_done and write_done

netaneld122:
_Disclaimer - this is a pure AI slop. I'm mostly interested in not crashing production code that hits this edge case._

Fixes a panic observed in production (reported upstream in tokio-rs/mio#1819 as `State::None` variant).

Replace `unreachable!()` with a graceful return (but still panic in debug builds) when IOCP delivers a completion event for a named pipe whose I/O state has already been consumed (e.g. after deregistration or an unexpected pipe closure race).

The `Arc` refcount leaked by `mem::forget` during `schedule_read` / `schedule_write` is properly reclaimed by `Arc::from_raw` at the top of each callback, so returning early is safe.

### DIFF
--- a/src/sys/windows/named_pipe.rs
+++ b/src/sys/windows/named_pipe.rs
@@ -906,7 +906,16 @@ fn read_done(status: &OVERLAPPED_ENTRY, events: Option<&mut Vec<Event>>) {
             io.notify_readable(&me, events);
             return;
         }
-        State::None => unreachable!(),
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
I don't think this actually solves anything. Somehow the state machine for `NamedPipe` is broken, this only hides that, possibly creating more issues by continuing on.

PR #1951, opened 2026-04-28
fixes for horizonOS/n3ds

kore-signet:
Fixes support for HorizonOS/Nintendo 3DS, in similar style to current support for WASI & ESP-IDF.

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -16,9 +16,9 @@
     )),
 ))]
 use std::os::fd::{AsFd, AsRawFd, BorrowedFd, RawFd};
-#[cfg(all(debug_assertions, not(target_os = "wasi")))]
+#[cfg(all(debug_assertions, not(any(target_os = "wasi", target_os = "horizon"))))]
 use std::sync::atomic::{AtomicBool, Ordering};
-#[cfg(all(debug_assertions, not(target_os = "wasi")))]
+#[cfg(all(debug_assertions, not(any(target_os = "wasi", target_os = "horizon"))))]
 use std::sync::Arc;
 use std::time::Duration;
 use std::{fmt, io};
@@ -274,7 +274,7 @@ pub struct Poll {
 pub struct Registry {
     selector: sys::Selector,
     /// Whether this selector currently has an associated waker.
-    #[cfg(all(debug_assertions, not(target_os = "wasi")))]
+    #[cfg(all(debug_assertions, not(any(target_os = "wasi", target_os = "horizon"))))]
     has_waker: Arc<AtomicBool>,
 }
 
@@ -324,7 +324,7 @@ impl Poll {
             sys::Selector::new().map(|selector| Poll {
                 registry: Registry {
                     selector,
-                    #[cfg(all(debug_assertions, not(target_os = "wasi")))]
+                    #[cfg(all(debug_assertions, not(any(target_os = "wasi", target_os = "horizon"))))]
                     has_waker: Arc::new(AtomicBool::new(false)),
                 },
             })
@@ -717,14 +717,14 @@ impl Registry {
     pub fn try_clone(&self) -> io::Result<Registry> {
         self.selector.try_clone().map(|selector| Registry {
             selector,
-            #[cfg(all(debug_assertions, not(target_os = "wasi")))]
+            #[cfg(not(any(target_os = "wasi", target_os = "horizon")))]
             has_waker: Arc::clone(&self.has_waker),
         })
     }
 
     /// Internal check to ensure only a single `Waker` is active per [`Poll`]
     /// instance.
-    #[cfg(all(debug_assertions, not(target_os = "wasi")))]
+    #[cfg(all(debug_assertions, not(any(target_os = "wasi", target_os = "horizon"))))]
     pub(crate) fn register_waker(&self) {
         assert!(
             !self.has_waker.swap(true, Ordering::AcqRel),
@@ -733,7 +733,7 @@ impl Registry {
     }
 
     /// Get access to the `sys::Selector`.
-    #[cfg(any(not(target_os = "wasi"), feature = "net"))]
+    #[cfg(any(not(any(target_os = "wasi", target_os = "horizon")), feature = "net"))]
     pub(crate) fn selector(&self) -> &sys::Selector {
         &self.selector
     }
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -734,6 +734,7 @@ impl Registry {
 
     /// Get access to the `sys::Selector`.
     #[cfg(any(not(target_os = "wasi"), feature = "net"))]
+    #[allow(dead_code)]
     pub(crate) fn selector(&self) -> &sys::Selector {
         &self.selector
     }
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -717,7 +717,7 @@ impl Registry {
     pub fn try_clone(&self) -> io::Result<Registry> {
         self.selector.try_clone().map(|selector| Registry {
             selector,
-            #[cfg(not(any(target_os = "wasi", target_os = "horizon")))]
+            #[cfg(all(debug_assertions, not(any(target_os = "wasi", target_os = "horizon"))))]
             has_waker: Arc::clone(&self.has_waker),
         })
     }
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -12,6 +12,7 @@
         target_os = "solaris",
         target_os = "vita",
         target_os = "cygwin",
+        target_os = "horizon"
     )),
 ))]
 use std::os::fd::{AsFd, AsRawFd, BorrowedFd, RawFd};
@@ -455,6 +456,7 @@ impl Poll {
         target_os = "solaris",
         target_os = "vita",
         target_os = "cygwin",
+        target_os = "horizon"
     )),
 ))]
 impl AsRawFd for Poll {
@@ -757,6 +759,7 @@ impl fmt::Debug for Registry {
         target_os = "solaris",
         target_os = "vita",
         target_os = "cygwin",
+        target_os = "horizon"
     )),
 ))]
 impl AsFd for Registry {
@@ -779,6 +782,7 @@ impl AsFd for Registry {
         target_os = "solaris",
         target_os = "vita",
         target_os = "cygwin",
+        target_os = "horizon"
     )),
 ))]
 impl AsRawFd for Registry {
@@ -800,6 +804,7 @@ cfg_os_poll! {
             target_os = "solaris",
             target_os = "vita",
             target_os = "cygwin",
+            target_os = "horizon"
         )),
     ))]
     #[test]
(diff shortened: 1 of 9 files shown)


src/sys/unix/selector/poll.rs
```diff
@@ -490,34 +490,47 @@ impl RegistrationRecord {
...
 
 #[cfg(not(target_os = "wasi"))]
-const POLLWRBAND: libc::c_short = libc::POLLWRBAND;
+const POLLWRBAND: PollFlagInt = libc::POLLWRBAND;
 #[cfg(target_os = "wasi")]
-const POLLWRBAND: libc::c_short = 0;
+const POLLWRBAND: PollFlagInt = 0;
 
-const READ_EVENTS: libc::c_short = libc::POLLIN | POLLRDHUP;
+#[cfg(not(target_os = "wasi"))]
```
Thomasdezeeuw: This was incorrectly added.

src/sys/unix/selector/poll.rs
```diff
@@ -490,34 +490,47 @@ impl RegistrationRecord {
...
+const READ_EVENTS: PollFlagInt = libc::POLLIN | POLLRDHUP;
 
-const WRITE_EVENTS: libc::c_short = libc::POLLOUT;
+#[cfg(not(target_os = "wasi"))]
+const WRITE_EVENTS: PollFlagInt = 0x08;
```
Thomasdezeeuw: I don't want to maintain the value of flags, it must use a constant from the libc crate.

src/sys/unix/selector/poll.rs
```diff
@@ -578,15 +596,18 @@ fn poll(fds: &mut [PollFd], timeout: Option<Duration>) -> io::Result<usize> {
 #[derive(Debug, Clone)]
 pub struct Event {
     token: Token,
+    #[cfg(not(target_os = "horizon"))]
     events: libc::c_short,
+    #[cfg(target_os = "horizon")]
+    events: libc::c_int
```
Thomasdezeeuw: Can you make this a type, it's quite confusing to define the field twice.

src/sys/unix/selector/poll.rs
```diff
@@ -578,15 +596,18 @@ fn poll(fds: &mut [PollFd], timeout: Option<Duration>) -> io::Result<usize> {
...
 }
 
 pub type Events = Vec<Event>;
 
 pub mod event {
     use std::fmt;
 
-    use crate::sys::Event;
+    use crate::sys::{Event, unix::selector::PollFlagInt};
```
Thomasdezeeuw: Let's use `super::PollFlagInt` instead, like we do with the constants.

src/sys/unix/uds/listener.rs
```diff
@@ -18,23 +18,35 @@ pub(crate) fn bind_addr(address: &SocketAddr) -> io::Result<net::UnixListener> {
...
+        target_os = "linux",
+        target_os = "freebsd",
+        target_os = "openbsd",
+        target_vendor = "apple",
+        target_os = "windows",
+        target_os = "redox",
+        target_os = "espidf",
+        target_os = "horizon"
+    )))]
+    let backlog = libc::SOMAXCONNS;
```
Thomasdezeeuw: Can you revert the unrelated changes here.

Thomasdezeeuw (changes requested):
This needs some more changes than the socket2 pr.

src/sys/unix/selector/poll.rs
kore-signet: That's very fair! Currently, the libc definition of POLLOUT for HorizonOS is incorrectly assigned as POLLERR. I'll submit a PR over there, too, so that it can stay a libc constant here.

src/sys/unix/selector/poll.rs
kore-signet: Oops, good catch :)

src/sys/unix/selector/poll.rs
kore-signet: This is also incorrect on most platforms, I think - since it should just have been using the self-defined POLLOUT constant anyways. Changed to libc in upcoming commit.

src/sys/unix/uds/listener.rs
kore-signet: Unfortunately they're load-bearing: HorizonOS' libc defs do not even have a symbol for libc::SOMAXCONNS, so a #[cfg] is necessary instead of an if cfg!(). If there'd be a cleaner way to format this while not relying on a defined libc::SOMAXCONNS, I can add change it over, though.

kore-signet:
Latest commit fixes *most* of the issues, though the libc::SOMAXCONNS is left standing. I've also added the 3DS/V6K target to the CI, and configured out waker support on horizonos (ala single-threaded WASI), since none of the current waker mechanisms (eventfd/kqueue/pipe) are supported.

src/poll.rs
```diff
@@ -715,14 +717,14 @@ impl Registry {
     pub fn try_clone(&self) -> io::Result<Registry> {
         self.selector.try_clone().map(|selector| Registry {
             selector,
-            #[cfg(all(debug_assertions, not(target_os = "wasi")))]
+            #[cfg(not(any(target_os = "wasi", target_os = "horizon")))]
```
Thomasdezeeuw: This is missing `debug_assertions`.

src/sys/unix/selector/poll.rs
```diff
@@ -559,13 +565,20 @@ fn poll(fds: &mut [PollFd], timeout: Option<Duration>) -> io::Result<usize> {
...
+        }
+        
 
         let res = syscall!(poll(
             fds.as_mut_ptr() as *mut libc::pollfd,
             fds.len() as libc::nfds_t,
             timeout,
         ));
 
+
```
Thomasdezeeuw: Nit: can you remove this.

Also shouldn't rustfmt catch this?

src/sys/unix/selector/poll.rs
```diff
@@ -490,34 +490,40 @@ impl RegistrationRecord {
...
+const POLLWRBAND: PollFlagInt = 0;
 
-const READ_EVENTS: libc::c_short = libc::POLLIN | POLLRDHUP;
+const READ_EVENTS: PollFlagInt = libc::POLLIN | POLLRDHUP;
 
-const WRITE_EVENTS: libc::c_short = libc::POLLOUT;
+const WRITE_EVENTS: PollFlagInt = libc::POLLOUT;
 
-const PRIORITY_EVENTS: libc::c_short = POLLPRI;
+#[cfg(not(target_os = "wasi"))]
```
Thomasdezeeuw: I think this `cfg` attribute is wrong, it's not there before the patch.

src/sys/unix/uds/listener.rs
```diff
@@ -18,23 +18,35 @@ pub(crate) fn bind_addr(address: &SocketAddr) -> io::Result<net::UnixListener> {
     syscall!(bind(fd, sockaddr, addrlen))?;
     // Use the same backlog value as the standard libary.
     // <https://github.com/rust-lang/rust/blob/0028f344ce9f64766259577c998a1959ca1f6a0b/library/std/src/os/unix/net/listener.rs#L75-L106>
-    let backlog = if cfg!(any(
+
```
Thomasdezeeuw: Nit: can we remove all the empty lines between the backlog variables?

Thomasdezeeuw (approved):
Two more cfg attributes need changes, but otherwise LGTM.

src/sys/unix/selector/poll.rs
kore-signet: Oop, must have copy pasted the attr by mistake lol. Fixed!

src/sys/unix/selector/poll.rs
kore-signet: Done! You'd think it would, but seemingly not.

src/poll.rs
kore-signet: Fixed :)

kore-signet:
Hopefully all fixed now! :)

src/sys/unix/uds/listener.rs
Thomasdezeeuw: This has a typo causing CI failures:
```suggestion
    let backlog = libc::SOMAXCONN;
```

kore-signet: Ah, heck! Fixed :)

Thomasdezeeuw:
@kore-signet the build is failing, so is rustfmt.

kore-signet:
Yeah, sorry for the back and forth >_< No idea what's up with my copy of rustfmt - it passes checks locally. Going to try to reproduce both failures locally and then fix them, will lyk!

kore-signet:
Should finally be fixed! Rustfmt seems to be behaving differently than wrapped cargo fmt for some reason on my machine o.o As for the build issue, it happened with a specific feature set (lacking net but with os-poll/os-ext) due to a miswritten #[cfg] on the Registry impl, which is now fixed.

Thomasdezeeuw:
@kore-signet still has a build issue. You can test all combinations of features using the following:
```
cargo check -Z build-std=std,panic_abort --target armv6k-nintendo-3ds --all-features
cargo check -Z build-std=std,panic_abort --target armv6k-nintendo-3ds --no-default-features
cargo check -Z build-std=std,panic_abort --target armv6k-nintendo-3ds --no-default-features --features default
cargo check -Z build-std=std,panic_abort --target armv6k-nintendo-3ds --no-default-features --features net
cargo check -Z build-std=std,panic_abort --target armv6k-nintendo-3ds --no-default-features --features default,net
cargo check -Z build-std=std,panic_abort --target armv6k-nintendo-3ds --no-default-features --features os-ext
cargo check -Z build-std=std,panic_abort --target armv6k-nintendo-3ds --no-default-features --features default,os-ext
cargo check -Z build-std=std,panic_abort --target armv6k-nintendo-3ds --no-default-features --features net,os-ext
cargo check -Z build-std=std,panic_abort --target armv6k-nintendo-3ds --no-default-features --features default,net,os-ext
cargo check -Z build-std=std,panic_abort --target armv6k-nintendo-3ds --no-default-features --features os-poll
cargo check -Z build-std=std,panic_abort --target armv6k-nintendo-3ds --no-default-features --features default,os-poll
cargo check -Z build-std=std,panic_abort --target armv6k-nintendo-3ds --no-default-features --features net,os-poll
cargo check -Z build-std=std,panic_abort --target armv6k-nintendo-3ds --no-default-features --features default,net,os-poll
```

kore-signet:
Hi! Sorry for the delay on this. Turns out that it was a dead-code lint on Registry::selector in poll.rs, and on Selector::register in sys/unix/selector/poll.rs. I can imagine allowing #[dead_code] here isn't ideal per-se, but it does get it building, and seems to be done a few other places in the codebase already. I know this isn't the best compared to tracking it down "more properly", but I hope that's alright for this instance!

Thomasdezeeuw:
@kore-signet can we gate the `dead_code` attribute to horizon? I.e. `#[cfg_attr(target_os = "horizon", allow(dead_code))]`.

kore-signet:
Done :D

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @kore-signet

PR #1954, opened 2026-05-18
Fix race condition in NamedPipe::connect

zdule:
There is a race condition in NamePipe::connect between the asynchronous completion of `connect_overlapped` and bumping the reference count of `self.inner`. If `connect_overlapped` completes before we bump the reference count, then reference count of `self.inner` will temporarily hit zero and `self.inner` will be dropped. Note that even though `Inner` is dropped, and it's memory is returned to the allocator, the object is still used by `NamedPipe`. After `NamedPipe` is itself dropped, the reference count for `Inner` drops to zero again, resulting in `drop` being called for the second time.

The fix is to bump the reference count of `self.inner` before calling `connect_overlapped`. Then, if `connect_overlapped` does not finish immediately we forget the clone of `Arc<Inner>`.

I usually hit this issue in my module within 2 million connects. I don't hit the issue with the proposed fix. Adding a `thread::sleep` immediately before `forget` makes the issue occur almost immediately. I don't have a minimal repro that doesn't use my module. If I had to write a unit test for this I would make a wrapper of `ConnectNamedPipe` that converts synchronous completions into asynchronous ones by manually posting an event to the  IOCP.

Other operations (read / write) use a similar pattern, but unlike connect, both issuing the operation (`schedule_read` / `schedule_write`) and completions (`read_done` / `write_done`) are executed under a mutex (`self.io`).

### DIFF
--- a/src/sys/windows/named_pipe.rs
+++ b/src/sys/windows/named_pipe.rs
@@ -423,6 +423,12 @@ impl NamedPipe {
             return Err(would_block());
         }
 
+        // If `connect_overlapped` does not complete immediately then we need
+        // to pass a reference to `self.inner` to the completion handler.
+        // We need to increase the reference count here because `connect_done`
+        // could be called as soon as we call `connect_overlapped`.
+        let inner_for_completion = self.inner.clone();
+
         // Now that we've flagged ourselves in the connecting state, issue the
         // connection attempt. Afterwards interpret the return value and set
         // internal state accordingly.
@@ -448,7 +454,7 @@ impl NamedPipe {
             // `connect_done` function will "reify" this forgotten pointer to
             // drop the refcount on the other side.
             Ok(false) => {
-                mem::forget(self.inner.clone());
+                mem::forget(inner_for_completion);
                 Err(would_block())
             }
 


src/sys/windows/named_pipe.rs
```diff
@@ -423,6 +423,12 @@ impl NamedPipe {
             return Err(would_block());
         }
 
+        // If `connect_overlapped` does not complete immediately then we need
+        // to pass a reference to `self.inner` to the completion handler.
+        // We need to increase the reference count here because `connect_done`
+        // could be called as soon as we call `connect_overlapped`.
+        let inner_for_completion = self.inner.clone();
```
Thomasdezeeuw, line 430: Instead of cloning and forgetting should we use `Arc::increment_strong_count` and `Arc::decrement_strong_count`? Those were probably not available when the code was originally written, but we should be able to use them now.

zdule: Hi Thomas, thanks for the comment.

With `Arc::increment_strong_count` it would look like this [commit](https://github.com/zdule/mio/commit/300a2636ce8e9e666fb7b5db5c0e2f1030eee83a).

Note that we have call `decrement_strong_count` in two cases where `connect_overlapped` completed synchronously. 

I prefer the current code with `forget` because:
1. It ties the `clone` with the `forget` by explicitly referencing the same variable (`inner_for_completion`).
2. Forget is used in the branch that is related to something unusual -- passing the ownership of a reference to the completion function, instead of having `decrement_strong_count` in all other branches.

I'll respect whatever you decide.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @zdule

PR #1960, opened 2026-06-15
Make connect_error test use a deterministic closed port

### LINKED ISSUE #959: connect_error test takes too long
In recent CI runs on FreeBSD, the `test_tcp::connect_error` test takes > 60s.  However, on my personal machine it's < 1s, so I don't know what the problem is.
https://cirrus-ci.com/task/5165242334576640

ameyypawar:
## Problem

`connect_error` connects to a hardcoded "probably unused" port (`127.0.0.1:58381`). On a host behind a firewall configured to DROP packets to closed endpoints, the SYN is silently dropped, so the non-blocking connect never receives a RST and `poll` blocks indefinitely — the ~60s CI timeouts reported in #959. On a normal host it returns `ECONNREFUSED` in <1s.

## Fix

Bind a listener to an OS-assigned ephemeral loopback port, record its address, then drop the listener so the port is **guaranteed closed**, and connect to that. A closed loopback port returns a fast, deterministic `ECONNREFUSED` regardless of any inbound firewall policy, so the connect can no longer hang. This uses the same bind → `local_addr` → drop idiom already used throughout `tests/tcp.rs`.

The existing poll-based coverage is preserved: mio's `connect` is non-blocking, so the refusal is normally delivered as a writable/error event through `poll`, exercising the same `is_writable() || is_error()` / `is_write_closed()` / `take_error()` checks as before.

There is a small window in which the freed ephemeral port could be reassigned to a concurrent listener (tests run in parallel) between the drop and the connect. If that happens the connection succeeds and there is no error to observe, which would otherwise fail the refusal-specific assertions. The test now detects this (no pending socket error) and returns early instead, matching the existing "couldn't get the coverage we want, so skip" behavior already in the synchronous-refusal branch.

## Notes

- Test-only change; no library behavior changes.
- The original 60s hang only manifests behind a DROP firewall and isn't reproducible on a normal loopback, so this removes the failure mode by construction rather than by reproducing it. The `TestFreeBSD` CI job already exercises this test on the originally-affected target.

Closes #959

### DIFF
--- a/tests/tcp.rs
+++ b/tests/tcp.rs
@@ -566,8 +566,17 @@ fn connection_reset_by_peer() {
 fn connect_error() {
     let (mut poll, mut events) = init_with_poll();
 
-    // Pick a "random" port that shouldn't be in use.
-    let mut stream = match TcpStream::connect("127.0.0.1:58381".parse().unwrap()) {
+    // Bind a listener to an OS-assigned ephemeral port on loopback, record its
+    // address, then drop it so the port is guaranteed closed. Connecting to a
+    // closed loopback port yields a fast, deterministic ECONNREFUSED, unlike a
+    // hardcoded "probably unused" port which can hang for ~60s behind a DROP
+    // firewall: <https://github.com/tokio-rs/mio/issues/959>.
+    let addr = {
+        let listener = net::TcpListener::bind("127.0.0.1:0").unwrap();
+        listener.local_addr().unwrap()
+    };
+
+    let mut stream = match TcpStream::connect(addr) {
         Ok(l) => l,
         Err(ref e) if e.kind() == io::ErrorKind::ConnectionRefused => {
             // Connection failed synchronously.  This is not a bug, but it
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw (commented):
I don't see how this changes anything as we still have an unused address. Now it's just recently used...

ameyypawar:
Rather than chase the unused-address behaviour, I could make the failure deterministic the way `connection_reset_by_peer` does — connect to a real listener and reset it from the server side (`set_linger_zero` + drop). That gives a local RST instead of relying on the environment to refuse an unused port. Tradeoff: it would test connection-error-on-reset rather than refused-connect specifically. Would you be open to that, or do you want to keep the refused-connect semantics here?

Thomasdezeeuw:
> Rather than chase the unused-address behaviour, I could make the failure deterministic the way `connection_reset_by_peer` does — connect to a real listener and reset it from the server side (`set_linger_zero` + drop).

That would be testing something different, so no.

Thomasdezeeuw:
I'm going to close this as it doesn't solve the problem described in #959.

PR #1962, opened 2026-06-18
Add Solaris event ports selector

psumbera:
Use Solaris event ports as the native os-poll backend instead of the poll(2)-based fallback. Keep the existing poll backend available behind mio_unsupported_force_poll_poll for comparison and unsupported builds.

Add a regression test for readiness after WouldBlock and add Solaris VM coverage to CI.

### DIFF
--- a/src/sys/unix/selector/event_ports.rs
+++ b/src/sys/unix/selector/event_ports.rs
@@ -24,14 +24,27 @@ pub struct Selector {
     id: usize,
     port: OwnedFd,
     state: Arc<SelectorState>,
-    raw_events: EventBuffer,
+    events: EventBuffer,
 }
 
 #[derive(Debug)]
-struct EventBuffer(Vec<libc::port_event>);
+struct EventBuffer {
+    port_events: Vec<libc::port_event>,
+    poll_fds: Vec<libc::pollfd>,
+}
 
-// SAFETY: `EventBuffer` is only scratch storage for `port_getn` results and is
-// only accessed through `Selector::select`, which requires `&mut self`.
+impl EventBuffer {
+    fn new() -> EventBuffer {
+        EventBuffer {
+            port_events: Vec::new(),
+            poll_fds: Vec::new(),
+        }
+    }
+}
+
+// SAFETY: `EventBuffer` is only scratch storage for `port_getn` and `poll`
+// results and is only accessed through `Selector::select`, which requires
+// `&mut self`.
 unsafe impl Send for EventBuffer {}
 unsafe impl Sync for EventBuffer {}
 
@@ -60,7 +73,7 @@ impl Selector {
             #[cfg(debug_assertions)]
             id: NEXT_ID.fetch_add(1, Ordering::Relaxed),
             port,
-            raw_events: EventBuffer(Vec::new()),
+            events: EventBuffer::new(),
             state: Arc::new(SelectorState {
                 fds: Mutex::new(HashMap::new()),
             }),
@@ -73,14 +86,14 @@ impl Selector {
             #[cfg(debug_assertions)]
             id: self.id,
             port,
-            raw_events: EventBuffer(Vec::new()),
+            events: EventBuffer::new(),
             state: Arc::clone(&self.state),
         })
     }
 
     pub fn select(&mut self, events: &mut Events, timeout: Option<Duration>) -> io::Result<()> {
         self.state
-            .select(self.port.as_raw_fd(), events, timeout, &mut self.raw_events)
+            .select(self.port.as_raw_fd(), events, timeout, &mut self.events)
     }
 
     cfg_io_source! {
@@ -130,7 +143,7 @@ impl SelectorState {
         port: RawFd,
         events: &mut Events,
         timeout: Option<Duration>,
-        raw_events: &mut EventBuffer,
+        event_buffer: &mut EventBuffer,
     ) -> io::Result<()> {
         events.clear();
 
@@ -138,11 +151,11 @@ impl SelectorState {
             return Ok(());
         }
 
-        raw_events.0.clear();
-        if raw_events.0.capacity() < events.capacity() {
-            raw_events
-                .0
-                .reserve(events.capacity() - raw_events.0.capacity());
+        event_buffer.port_events.clear();
+        if event_buffer.port_events.capacity() < events.capacity() {
+            event_buffer
+                .port_events
+                .reserve(events.capacity() - event_buffer.port_events.capacity());
         }
 
         let start = Instant::now();
@@ -174,13 +187,13 @@ impl SelectorState {
                 .map(|timeout| timeout as *mut _)
                 .unwrap_or(std::ptr::null_mut());
             let mut nget = 1;
-            raw_events.0.clear();
+            event_buffer.port_events.clear();
 
             trace!("waiting for Solaris event port events");
             let res = syscall!(port_getn(
                 port,
-                raw_events.0.as_mut_ptr(),
-                raw_events.0.capacity() as libc::c_uint,
+                event_buffer.port_events.as_mut_ptr(),
+                event_buffer.port_events.capacity() as libc::c_uint,
                 &mut nget,
                 timeout,
             ));
@@ -200,8 +213,8 @@ impl SelectorState {
 
             trace!("Solaris event port returned {nget} events");
             // SAFETY: `port_getn` initialized exactly the first `nget` entries.
-            unsafe { raw_events.0.set_len(nget as usize) };
-            self.process_events(port, raw_events.0.as_slice(), events)?;
+            unsafe { event_buffer.port_events.set_len(nget as usize) };
+            self.process_events(port, event_buffer, events)?;
             if !events.is_empty() {
                 return Ok(());
             }
@@ -211,12 +224,41 @@ impl SelectorState {
     fn process_events(
         &self,
         port: RawFd,
-        raw_events: &[libc::port_event],
+        event_buffer: &mut EventBuffer,
         events: &mut Events,
     ) -> io::Result<()> {
         let mut fds = self.fds.lock().unwrap();
+        let raw_events = event_buffer.port_events.as_slice();
+        let poll_fds = &mut event_buffer.poll_fds;
 
         events.reserve(raw_events.len());
+        poll_fds.clear();
+        poll_fds.reserve(raw_events.len());
+        for event in raw_events {
+            if event.portev_source != libc::PORT_SOURCE_FD as libc::c_ushort {
+                continue;
+            }
+
+            let fd = event.portev_object as RawFd;
+            let Some(data) = fds.get(&fd) else {
+                continue;
+            };
+            poll_fds.push(libc::pollfd {
+                fd,
+                events: interests_to_port(data.interests) as libc::c_short,
+                revents: 0,
+            });
+        }
+
+        if !poll_fds.is_empty() {
+            syscall!(poll(
+                poll_fds.as_mut_ptr(),
+                poll_fds.len() as libc::nfds_t,
+                0
+            ))?;
+        }
+
+        let mut poll_fds = poll_fds.iter();
         for event in raw_events {
             if event.portev_source == libc::PORT_SOURCE_USER as libc::c_ushort {
                 push_event(
@@ -234,7 +276,15 @@ impl SelectorState {
                     continue;
                 };
                 let current_interest_events = interests_to_port(data.interests);
-                let Some(poll_events) = poll_fd_readiness(fd, data.interests)? else {
+                let poll_events = poll_fds.next().and_then(|poll_fd| {
+                    debug_assert_eq!(poll_fd.fd, fd);
+                    if poll_fd.revents == 0 {
+                        None
+                    } else {
+                        Some(EventMask::from(poll_fd.revents))
+                    }
+                });
+                let Some(poll_events) = poll_events else {
                     if event.portev_events & ERROR_EVENTS == 0 {
                         let generation = data.generation.wrapping_add(1).max(1);
                         associate(port, fd, data.interests, generation)?;
@@ -479,20 +529,6 @@ fn push_event(
             portev_object: object,
             portev_user: usize::from(token) as *mut libc::c_void,
         }));
-    }
-}
-
-fn poll_fd_readiness(fd: RawFd, interests: Interest) -> io::Result<Option<EventMask>> {
-    let mut poll_fd = libc::pollfd {
-        fd,
-        events: interests_to_port(interests) as libc::c_short,
-        revents: 0,
-    };
-
-    if syscall!(poll(&mut poll_fd, 1, 0))? == 0 {
-        Ok(None)
-    } else {
-        Ok(Some(EventMask::from(poll_fd.revents)))
     }
 }
 
--- a/src/sys/unix/selector/event_ports.rs
+++ b/src/sys/unix/selector/event_ports.rs
@@ -14,10 +14,8 @@ static NEXT_ID: AtomicUsize = AtomicUsize::new(1);
 
 type EventMask = libc::c_int;
 
-const POLLRDHUP: EventMask = 0;
 const READ_EVENTS: EventMask = libc::POLLIN as EventMask;
 const WRITE_EVENTS: EventMask = libc::POLLOUT as EventMask;
-const PRIORITY_EVENTS: EventMask = libc::POLLPRI as EventMask;
 const ERROR_EVENTS: EventMask = libc::POLLERR as EventMask | libc::POLLHUP as EventMask;
 
 #[derive(Debug)]
@@ -541,10 +539,6 @@ fn interests_to_port(interests: Interest) -> EventMask {
         events |= WRITE_EVENTS;
     }
 
-    if interests.is_priority() {
-        events |= PRIORITY_EVENTS;
-    }
-
     events
 }
 
@@ -621,7 +615,7 @@ pub mod event {
     use crate::sys::Event;
     use crate::Token;
 
-    use super::{EventMask, POLLRDHUP};
+    use super::EventMask;
 
     pub fn token(event: &Event) -> Token {
         Token(event.0.portev_user as usize)
@@ -642,7 +636,6 @@ pub mod event {
 
     pub fn is_read_closed(event: &Event) -> bool {
         (event.0.portev_events & libc::POLLHUP as EventMask) != 0
-            || (event.0.portev_events & POLLRDHUP) != 0
     }
 
     pub fn is_write_closed(event: &Event) -> bool {
--- a/src/sys/unix/selector/event_ports.rs
+++ b/src/sys/unix/selector/event_ports.rs
@@ -1,0 +1,829 @@
+use std::collections::HashMap;
+use std::ops::{Deref, DerefMut};
+use std::os::fd::{AsFd, AsRawFd, BorrowedFd, FromRawFd, OwnedFd, RawFd};
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
+use std::{cmp, io};
+
+use crate::{Interest, Token};
+
+/// Unique id for use as `SelectorId`.
+#[cfg(debug_assertions)]
+static NEXT_ID: AtomicUsize = AtomicUsize::new(1);
+
+type EventMask = libc::c_int;
+
+const POLLRDHUP: EventMask = 0;
+const READ_EVENTS: EventMask = libc::POLLIN as EventMask;
+const WRITE_EVENTS: EventMask = libc::POLLOUT as EventMask;
+const PRIORITY_EVENTS: EventMask = libc::POLLPRI as EventMask;
+const ERROR_EVENTS: EventMask = libc::POLLERR as EventMask | libc::POLLHUP as EventMask;
+
+#[derive(Debug)]
+pub struct Selector {
+    #[cfg(debug_assertions)]
+    id: usize,
+    port: OwnedFd,
+    state: Arc<SelectorState>,
+    raw_events: EventBuffer,
+}
+
+#[derive(Debug)]
+struct EventBuffer(Vec<libc::port_event>);
+
+// SAFETY: `EventBuffer` is only scratch storage for `port_getn` results and is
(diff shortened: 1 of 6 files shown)


Thomasdezeeuw:
Before I review this, are you @psumbera or someone else able to provide support to maintain this?

psumbera:
> Before I review this, are you @psumbera or someone else able to provide support to maintain this?

Yes, I can help maintain and support this.

Solaris is now testable in CI via the Solaris CBE GitHub Actions VM, so regressions should be caught by the normal test workflow.

The main motivation is Tokio compatibility: with the current poll-based backend, Tokio's `io_async_fd` tests fail or hang on Solaris. With the event ports backend, those tests pass, and Mio`s own test suite passes as well.

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,905 @@
...
+
+        f.debug_struct("port_event")
+            .field("token", &event.token)
+            .field("events", &EventsDetails(event.events))
+            .finish()
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct Waker {
```
Thomasdezeeuw: Can we move this to `sys/unix/waker/event_ports.rs`? This makes the code similar to the existing code, making it easier to follow. We do something similar for the kqueue based waker.

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,905 @@
...
+
+const POLLRDHUP: EventMask = 0;
+const READ_EVENTS: EventMask = libc::POLLIN as EventMask;
+const WRITE_EVENTS: EventMask = libc::POLLOUT as EventMask;
+const PRIORITY_EVENTS: EventMask = libc::POLLPRI as EventMask;
+const ERROR_EVENTS: EventMask = libc::POLLERR as EventMask | libc::POLLHUP as EventMask;
+
+#[repr(C)]
+#[derive(Clone, Copy)]
+struct PortEvent {
```
Thomasdezeeuw: Can we add this to libc? I don't want to maintain these kind of definitions.

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,905 @@
...
+    ) -> io::Result<()> {
+        events.clear();
+
+        if events.capacity() == 0 {
+            return Ok(());
+        }
+
+        let start = Instant::now();
+        let mut attempted = false;
+        let mut raw_events = vec![empty_port_event(); events.capacity()];
```
Thomasdezeeuw: I don't think we have to initialise/zero this vector since it's passed to `port_getn`. From the docs at https://docs.oracle.com/cd/E88353_01/html/E37843/port-getn-3c.html (emphasis mine):
>  The list argument is an array of **uninitialized** port_event_t structures that is filled in by the system [...]

So we can get away with `Vec::with_capacity(events.capacity())`.

Thomasdezeeuw: Also can we do something about the allocation on each poll? Usually the `events` passed to `select` is the same vector for each call and there is only one call at a time (through `Poll::poll` requiring a mutable reference). I think `Selector::select` and `SelectorState::select` can both be `&mut self` and then `Selector` can have a preallocated & reuse buffer for the events.

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,905 @@
...
+            });
+            let timeout = timeout_to_timespec(current_timeout);
+            let timeout = timeout
+                .as_ref()
+                .map(|timeout| timeout as *const _)
+                .unwrap_or(ptr::null());
+            let mut nget = 1;
+
+            trace!("waiting for Solaris event port events");
+            let res = unsafe {
```
Thomasdezeeuw: We can use the `syscall!` macro here for error handling.

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,905 @@
...
+            if err.raw_os_error() == Some(libc::ENOENT) {
+                Ok(())
+            } else {
+                Err(err)
+            }
+        })
+}
+}
+
+fn syscall_result(res: libc::c_int) -> io::Result<libc::c_int> {
```
Thomasdezeeuw: This can be replaced with the `syscall!` macro.

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,905 @@
...
+                    timeout,
+                )
+            };
+            attempted = true;
+
+            if res < 0 {
+                let err = io::Error::last_os_error();
+                match err.raw_os_error() {
+                    Some(libc::ETIME) => {
+                        self.poll_registered_fds(events, poll_fallback_on_timeout)?;
```
Thomasdezeeuw: Can we set `nget = 0` and continue below?

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,905 @@
...
+                        return Ok(());
+                    }
+                    Some(libc::EINTR) => continue,
```
Thomasdezeeuw: Mio doesn't handle `EINTR`, it must be returned. See the docs on `Poll::poll`.

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,905 @@
...
+                    },
+                    None => None,
+                }
+            } else {
+                timeout
+            };
+            let poll_fallback_on_timeout = current_timeout.is_none() && self.has_fallback_fds();
+            let current_timeout = current_timeout.or_else(|| {
+                if poll_fallback_on_timeout {
+                    Some(Duration::from_millis(100))
```
Thomasdezeeuw: This code is only entered if `self.has_fallback_fds()` is true and we have no timeout. In that case why are we waiting 100ms? Can we pass a zero (but not null) timeout to only get the events that are ready to be processed instead?

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,905 @@
...
+            if nget == 0 {
+                self.poll_registered_fds(events, poll_fallback_on_timeout)?;
+                return Ok(());
+            }
+
+            trace!(
+                "Solaris event port returned {:?}",
+                &raw_events[..nget as usize]
+            );
+            self.convert_events(port, &raw_events[..nget as usize], events)?;
```
Thomasdezeeuw: Why do we need to convert the events? Can we use the same design we have to kqueue and epoll and use a wrapper around the raw event type?

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,905 @@
...
+            if !events.is_empty() {
+                self.drain_events(port, &mut raw_events, events)?;
```
Thomasdezeeuw: I don't understand why we need to call `port_getn` again here.

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,905 @@
...
+
+        if data.associated {
+            dissociate(port, fd)?;
+        }
+        Ok(())
+    }
+    }
+
+    fn wake(&self, port: RawFd, token: Token) -> io::Result<()> {
+        self.pending_wake_token.lock().unwrap().replace(token);
```
Thomasdezeeuw: Why do we need the `pending_wake_token` if we send the token in the event below? Do multiple messages/tokens send via `port_send` get coalesced?

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,905 @@
...
+
+        if self.wake_queued.swap(true, Ordering::AcqRel) {
+            return Ok(());
+        }
+
+        let result = syscall_result(unsafe {
+            port_send(
+                port,
+                libc::POLLIN as libc::c_int,
```
Thomasdezeeuw: Can you add a note saying that the value is controlled by us (Solaris doesn't do anything with it) and we use `POLLIN` to trigger the readability on the event to match other implementations?

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,905 @@
...
+            dissociate(port, fd)?;
+        }
+        Ok(())
+    }
+    }
+
+    fn wake(&self, port: RawFd, token: Token) -> io::Result<()> {
+        self.pending_wake_token.lock().unwrap().replace(token);
+
+        if self.wake_queued.swap(true, Ordering::AcqRel) {
```
Thomasdezeeuw: We don't have this kind of check for the other implementations, they always create a new event in the OS queue (single thread implementation excluded), so let's match that here.

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,905 @@
...
+        fd: RawFd,
+        token: Token,
+        interests: Interest,
+    ) -> io::Result<Arc<RegistrationRecord>> {
+        let mut fds = self.fds.lock().unwrap();
+        if fds.contains_key(&fd) {
+            return Err(io::Error::new(
+                io::ErrorKind::AlreadyExists,
+                "I/O source already registered this `Registry` \
+                (an old file descriptor might have been closed without deregistration)",
```
Thomasdezeeuw: Can we do without this check in any way? According to https://docs.oracle.com/cd/E88353_01/html/E37843/port-associate-3c.html#REFMAN3Aport-associate-3c: 

> The association is also removed if the port gets closed or when port_dissociate() is called.

So I would expect closing the fd to be enough to deregister it from Mio. I know the `poll(2)` based implementation doesn't do this, but the others does and most code using Mio expect it to work like that. I rather not make another exception, only if it's really necessary.

Thomasdezeeuw (commented):
I took a look at this implementation and I have a bunch of questions. I will need to take a another look and I think will need some more work.

Overall the implementation, to me with little knowledge of the Solaris port stuff that backs this, does some odd things, but doesn't document why (not in the comments or in the commit history as it's only 1 commit). I would really like to see some of it documented so that it's easier to understand.

Reading https://docs.oracle.com/cd/E88353_01/html/E37843/port-associate-3c.html#REFMAN3Aport-associate-3c, it seems that `PORT_SOURCE_FD` is a one-shot mechanism for fd events. However, the kernel does keep track of the fds we're interested in. I think `IoSourceState` does the right thing and reregister once `WouldBlock` is hit. However, I don't get why we need the `fds` field on `SelectorState`, and it's not documented either.

It's also not entirely clear to me why we need `convert_events`.

src/sys/unix/selector/event_ports.rs
psumbera: Done. I moved the event ports waker into src/sys/unix/waker/event_ports.rs, matching the existing Unix waker layout. The selector now only uses the shared Waker type instead of defining it inline.

src/sys/unix/selector/event_ports.rs
psumbera: Done. I added the missing event port definitions to libc and updated Mio to use libc::port_event instead of carrying a local PortEvent definition. This keeps the  FFI types in libc rather than in Mio.

See: https://github.com/rust-lang/libc/pull/5203

src/sys/unix/selector/event_ports.rs
psumbera: Done. The event buffer now uses spare capacity and passes uninitialized port_event_t storage to port_getn, so we no longer zero-initialize the vector before the syscall.

psumbera: Done. Selector::select now takes &mut self and keeps a reusable event buffer on Selector, so each poll reuses the same allocation instead of allocating a fresh vector.

src/sys/unix/selector/event_ports.rs
psumbera: Done. The port_getn call now goes through the existing syscall! macro for error handling.

src/sys/unix/selector/event_ports.rs
psumbera: Done. I removed the local syscall_result helper and switched those calls to Mio's existing syscall! macro.

src/sys/unix/selector/event_ports.rs
psumbera: Done. ETIME now sets nget = 0 and falls through to the shared post-port_getn path instead of returning early from that branch.

src/sys/unix/selector/event_ports.rs
psumbera: Done. EINTR is no longer swallowed/retried in the event ports selector; it is returned to the caller like the other Mio poll backends.

src/sys/unix/selector/event_ports.rs
psumbera: Done. The fallback path now uses a zero, non-null timeout instead of waiting 100ms, so it only drains currently ready event port events before checking fallback fds.

src/sys/unix/selector/event_ports.rs
psumbera: Done. The selector now stores raw port_event_t values in an Event wrapper, matching the epoll/kqueue style. The separate conversion step was removed; Mio events are decoded from the wrapped raw event when iterating.

src/sys/unix/selector/event_ports.rs
psumbera: Done. That extra port_getn drain was removed. The selector now processes only the events returned by the main port_getn call.

src/sys/unix/selector/event_ports.rs
psumbera: Done. I removed pending_wake_token. port_send queues the user event with the token in portev_user, so the waker now just sends that token directly and reads it back from the delivered event.

src/sys/unix/selector/event_ports.rs
psumbera: Done. I added a note explaining that the event value is controlled by Mio for user events, and that POLLIN is used so waker events map to readable readiness like the other backends.

src/sys/unix/selector/event_ports.rs
psumbera: Done. I removed the wake_queued coalescing check. Each call to wake now sends a new user event to the event port, matching the other multi-threaded waker implementations.

src/sys/unix/selector/event_ports.rs
psumbera: Done. I removed that pre-check. Registration now relies on port_associate/port_dissociate and replaces the local fd entry after a successful association, so closing an fd can behave like the other selector backends instead of adding a poll-style exception.

psumbera:
Agreed, the first version did not explain the event port-specific state well enough.

`fds` is needed because `PORT_SOURCE_FD` is one-shot: after an fd event is delivered, the fd is no longer fully associated with the port. The kernel tracks the active association, but Mio still needs to keep the user-visible state across those one-shot deliveries: current token, interests, generation for stale event suppression, whether the fd is currently associated, and whether delivered interests need fallback polling until `IoSourceState` reregisters after `WouldBlock`.

I also removed `convert_events`. Events are now stored as a wrapper around `libc::port_event`, following the epoll/kqueue style.

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,829 @@
...
+        (event.0.portev_events & libc::POLLOUT as EventMask) != 0
+    }
+
+    pub fn is_error(event: &Event) -> bool {
+        (event.0.portev_events & libc::POLLERR as EventMask) != 0
+    }
+
+    pub fn is_read_closed(event: &Event) -> bool {
+        (event.0.portev_events & libc::POLLHUP as EventMask) != 0
+            || (event.0.portev_events & POLLRDHUP) != 0
```
Thomasdezeeuw: `POLLRDHUP` is zero, so this check does `portev_events & 0 != 0`, which is always false.

src/sys/unix/selector/event_ports.rs
```diff
@@ -0,0 +1,829 @@
...
+    if interests.is_readable() {
+        events |= READ_EVENTS;
+    }
+
+    if interests.is_writable() {
+        events |= WRITE_EVENTS;
+    }
+
+    if interests.is_priority() {
+        events |= PRIORITY_EVENTS;
```
Thomasdezeeuw: `Interest::PRIORITY` is Linux and Android only. Should that be enable for Solaris as well? If so `interests_from_event` would would need to be adjusted as well I think.

Thomasdezeeuw (commented):
Thank you for the changes. Could you not rebase your commits? It makes reviewing a lot more difficult as GitHub doesn't show the difference at all.

> `fds` is needed because `PORT_SOURCE_FD` is one-shot: after an fd event is delivered, the fd is no longer fully associated with the port. The kernel tracks the active association, but Mio still needs to keep the user-visible state across those one-shot deliveries: current token, interests, generation for stale event suppression, whether the fd is currently associated, and whether delivered interests need fallback polling until `IoSourceState` reregisters after `WouldBlock`.

Can we track that in `InternalState` of `IoSource`? That already stores a lot of that information and does so without a global lock.

> I also removed `convert_events`. Events are now stored as a wrapper around `libc::port_event`, following the epoll/kqueue style.

👍

I still have to properly review `select` and `process_events` functions. Could you explain why `poll_fd_readiness` is needed *per* event, because that greatly increases the number of system calls we make.

src/sys/unix/selector/event_ports.rs
psumbera: Done in a follow-up commit. I removed the zero `POLLRDHUP` placeholder and simplified `is_read_closed` to check `POLLHUP` only on Solaris.

src/sys/unix/selector/event_ports.rs
psumbera: Done in the same follow-up commit. I did not enable `Interest::PRIORITY` for Solaris in this PR, since the public interest is currently Linux/Android-only. I removed the dead priority-interest registration path from the event ports selector instead.

psumbera:
> Thank you for the changes. Could you not rebase your commits? It makes reviewing a lot more difficult as GitHub doesn't show the difference at all.

Sure, I'll avoid rebasing from now on and keep review fixes as follow-up commits.
 
> Can we track that in `InternalState` of `IoSource`? That already stores a lot of that information and does so without a global lock.

Not completely. `IoSourceState::InternalState` only exists for Mio-owned `IoSource<T>`. The selector also supports `SourceFd` and custom `event::Source` implementations that call `Registry::register` directly, so `Poll::poll` still needs selector-owned state for those fds.

Also, when processing `port_getn` results, the selector only has the delivered fd event. It does not have access to the original `IoSource`, so it needs enough state to map the fd back to the token/interests, suppress stale events, know whether the fd is currently associated, and track whether fallback polling is needed until `IoSourceState::do_io` sees WouldBlock.

So some state overlaps with `IoSourceState`, but moving it there would not cover all registration paths and would not remove the need for selector-side state.

> I still have to properly review `select` and `process_events` functions. Could you explain why `poll_fd_readiness` is needed _per_ event, because that greatly increases the number of system calls we make.

For `poll_fd_readiness`: event ports tell us that an fd caused a one-shot event, but by the time Mio processes that event the fd has already been disassociated and the readiness may no longer match `portev_events`. The extra zero-timeout `poll` is used to get the current fd readiness before deciding what to return to the user and what to re-associate.

It is doing three things:

- avoids returning stale readiness if the condition disappeared before `Poll::poll` processed it;
- determines which interests are currently ready, so delivered interests can stay in fallback polling until `IoSourceState::do_io` sees `WouldBlock`;
 - lets us re-associate only the remaining interests after the one-shot event delivery.

So yes, it adds a syscall per delivered fd event, but it is there to preserve Mio's readiness semantics on top of event ports one-shot fd model. If there is a simpler way to do that without the extra `poll(0)`, I'm happy to adjust.

Thomasdezeeuw (approved):
Sorry for the delay, Europe was melting 🫠 

> > Can we track that in `InternalState` of `IoSource`? That already stores a lot of that information and does so without a global lock.
> 
> Not completely. `IoSourceState::InternalState` only exists for Mio-owned `IoSource<T>`. The selector also supports `SourceFd` and custom `event::Source` implementations that call `Registry::register` directly, so `Poll::poll` still needs selector-owned state for those fds.
> 
> Also, when processing `port_getn` results, the selector only has the delivered fd event. It does not have access to the original `IoSource`, so it needs enough state to map the fd back to the token/interests, suppress stale events, know whether the fd is currently associated, and track whether fallback polling is needed until `IoSourceState::do_io` sees WouldBlock.
> 
> So some state overlaps with `IoSourceState`, but moving it there would not cover all registration paths and would not remove the need for selector-side state.

That makes sense 👍 

> > I still have to properly review `select` and `process_events` functions. Could you explain why `poll_fd_readiness` is needed _per_ event, because that greatly increases the number of system calls we make.
> 
> For `poll_fd_readiness`: event ports tell us that an fd caused a one-shot event, but by the time Mio processes that event the fd has already been disassociated and the readiness may no longer match `portev_events`. The extra zero-timeout `poll` is used to get the current fd readiness before deciding what to return to the user and what to re-associate.
> 
> It is doing three things:
> 
> * avoids returning stale readiness if the condition disappeared before `Poll::poll` processed it;
> * determines which interests are currently ready, so delivered interests can stay in fallback polling until `IoSourceState::do_io` sees `WouldBlock`;
> * lets us re-associate only the remaining interests after the one-shot event delivery.
> 
> So yes, it adds a syscall per delivered fd event, but it is there to preserve Mio's readiness semantics on top of event ports one-shot fd model. If there is a simpler way to do that without the extra `poll(0)`, I'm happy to adjust.

Would it be possible to change `poll_fd_readiness` to poll all fd in one go? Similar to `poll_registered_fds`.

Other than the above question, this LGTM.

psumbera:
> Sorry for the delay, Europe was melting 🫠

I know :-)

> Would it be possible to change `poll_fd_readiness` to poll all fd in one go? Similar to `poll_registered_fds`.

Yes, that makes sense. I changed this to batch the readiness check for all fd events returned by one port_getn call. process_events now builds one pollfd array for the live fd events in the batch and does a single zero-timeout poll,   then uses the corresponding revents while processing each event.

That keeps the existing stale-event/re-association/fallback behavior, but reduces the extra readiness check from one poll syscall per fd event to one poll syscall per port_getn batch.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @psumbera

PR #1637, opened 2022-11-30
Test x86_64-pc-windows-gnu with Wine on CI

taiki-e:
This adds test for x86_64-pc-windows-gnu with Wine 7.13 on CI.

~~This also documents that Wine 7.13+ is supported.~~ EDIT: see https://github.com/tokio-rs/mio/pull/1637#issuecomment-1428318706

cc https://github.com/tokio-rs/mio/issues/1444

PR #1965, opened 2026-06-30
Test x86_64-pc-windows-gnu with Wine on CI, fix some tests in Wine

tbu-:
Revival of #1637, thanks to @taiki-e.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -284,7 +284,7 @@ jobs:
     - uses: taiki-e/setup-cross-toolchain-action@v1
       with:
         target: x86_64-pc-windows-gnu
-        runner: wine@7.13
+        runner: wine@8.18
     - name: Tests
       run: cargo test --all-features --target x86_64-pc-windows-gnu
     - name: Tests release build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -275,6 +275,21 @@ jobs:
           . $HOME/.cargo/env
           cargo test --all-features
 
+  Wine:
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@stable
+    # Install linker, libraries, and test runner (Wine) for x86_64-pc-windows-gnu.
+    - uses: taiki-e/setup-cross-toolchain-action@v1
+      with:
+        target: x86_64-pc-windows-gnu
+        runner: wine@7.13
+    - name: Tests
+      run: cargo test --all-features --target x86_64-pc-windows-gnu
+    - name: Tests release build
+      run: cargo test --release --all-features --target x86_64-pc-windows-gnu
+
   # Single job required to merge the pr.
   Passed:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -280,11 +280,20 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - uses: dtolnay/rust-toolchain@stable
-    # Install linker, libraries, and test runner (Wine) for x86_64-pc-windows-gnu.
-    - uses: taiki-e/setup-cross-toolchain-action@v1
-      with:
-        target: x86_64-pc-windows-gnu
-        runner: wine@8.18
+    - name: Install wine
+      run: |
+        # https://gitlab.winehq.org/wine/wine/-/wikis/Debian-Ubuntu
+        wget -O - https://dl.winehq.org/wine-builds/winehq.key | sudo gpg --dearmor -o /etc/apt/keyrings/winehq-archive.key -
+        sudo dpkg --add-architecture i386
+        sudo wget -NP /etc/apt/sources.list.d/ https://dl.winehq.org/wine-builds/ubuntu/dists/jammy/winehq-jammy.sources
+        sudo apt-get update
+        sudo apt-get install -y winehq-stable wine-binfmt
+        echo -1 | sudo tee /proc/sys/fs/binfmt_misc/cli # deregister mono handler for PE files
+    - name: Install linker and libraries
+      run: |
+        sudo apt-get install -y gcc-mingw-w64-x86-64
+    - name: Install target
+      run: rustup target add x86_64-pc-windows-gnu
     - name: Tests
       run: cargo test --all-features --target x86_64-pc-windows-gnu
     - name: Tests release build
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -267,5 +280,6 @@ jobs:
       - Rustfmt
       - CheckTargets
       - TestFreeBSD
+      - Wine
```
Thomasdezeeuw: Since Wine is not officially supported let's remove it from here so it doesn't block the CI.

.github/workflows/ci.yml
```diff
@@ -252,7 +252,20 @@ jobs:
...
+  Wine:
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@stable
+    # Install linker, libraries, and test runner (Wine) for x86_64-pc-windows-gnu.
+    - uses: taiki-e/setup-cross-toolchain-action@v1
+      with:
+        target: x86_64-pc-windows-gnu
+        runner: wine@8.18
```
Thomasdezeeuw: According to the website the latest stable 11.0, should we use a newer version?

.github/workflows/ci.yml
```diff
@@ -252,7 +252,20 @@ jobs:
...
           . $HOME/.cargo/env
           cargo test --target ${{ matrix.target }} --all-features
-
+  Wine:
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@stable
+    # Install linker, libraries, and test runner (Wine) for x86_64-pc-windows-gnu.
+    - uses: taiki-e/setup-cross-toolchain-action@v1
```
Thomasdezeeuw: Can we do it without this? As this contains a 1200 line script I would have to review.

.github/workflows/ci.yml
tbu-: Done.

.github/workflows/ci.yml
tbu-: Your call. I used the smallest version that included all the necessary fixes for mio. I thought it would be useful to use the lowest version so one could later decide on a supported version, if desired.

.github/workflows/ci.yml
tbu-: I can try looking into replacing this, if you want.

Do you have to review the 1200 line script though? It's authored by someone whose actions are already used in mio's workflows (`taiki-e/install-action@cargo-hack`), and it ultimately can't do much wrong — I don't see any secrets or permissions being passed to this workflow. The worst I can imagine is that it could maliciously make the (Wine!) tests fail or pass.

What is the concern you'd be looking for when reviewing this 1200 line script or my rewritten form of it?

tbu-:
Thanks for the incredibly quick review by the way!

.github/workflows/ci.yml
Thomasdezeeuw: It's still here?

.github/workflows/ci.yml
Thomasdezeeuw: The way I see it is that we really only need the `wine` command, while this action provides a lot more. I don't want a lot of dependencies for Mio, even the VM actions I'm not a fan of.

Installing Wine via `apt-get install wine`, building with `--target windows -Z build-std` and running under Wine seems a lot easier to me.

.github/workflows/ci.yml
Thomasdezeeuw: I'm not sure what Wine's supported version policy is, but I'm fine with bumping it.

.github/workflows/ci.yml
tbu-: Now really done… Sorry, forgot to force-push.

.github/workflows/ci.yml
tbu-: I'm installing the version declared as stable from wine developers now, via their official instructions: https://gitlab.winehq.org/wine/wine/-/wikis/Debian-Ubuntu. It seems to be 11.0 currently.

.github/workflows/ci.yml
tbu-: I'm now manually installing wine, and making sure that the binfmt_misc handler is correctly configured to start wine for PE files.

tbu-:
I'm now manually installing wine. I have rebased the PR onto latest master because it had merge conflicts. Should I squash the first, forth and fifth commit? :)

Sorry for the delay.

tests/tcp_stream.rs
```diff
@@ -444,8 +444,17 @@ fn shutdown_both() {
...
+    // from Rust 1.97 onwards (https://github.com/rust-lang/rust/pull/156063)
+    // and to `io::Error::Uncategorized` before).
+    //
+    // Both errors seem reasonable: we have closed the writing side, it's
+    // reasonable to receive an error for writing to the writing side, but the
+    // other party has also closed the socket, so it's reasonable to get an
+    // error because the other side went away.
     #[cfg(windows)]
-    assert_eq!(err.kind(), io::ErrorKind::ConnectionAborted);
+    assert!(err.raw_os_error() == Some(10053) || err.raw_os_error() == Some(10058));
```
Thomasdezeeuw: Nit: can we replace these with the relevant constants? `windows_sys::Win32::Networking::WinSock::WSAECONNABORTED` and `windows_sys::Win32::Networking::WinSock::WSAESHUTDOWN`.

.github/workflows/ci.yml
```diff
@@ -275,6 +275,30 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@stable
+    - name: Install wine
+      run: |
+        # https://gitlab.winehq.org/wine/wine/-/wikis/Debian-Ubuntu
+        wget -O - https://dl.winehq.org/wine-builds/winehq.key | sudo gpg --dearmor -o /etc/apt/keyrings/winehq-archive.key -
+        sudo dpkg --add-architecture i386
+        sudo wget -NP /etc/apt/sources.list.d/ https://dl.winehq.org/wine-builds/debian/dists/bookworm/winehq-bookworm.sources
```
Thomasdezeeuw: Shouldn't this use Jammy, to match the runner (Ubuntu 22.04)?
```suggestion
        sudo wget -NP /etc/apt/sources.list.d/ https://dl.winehq.org/wine-builds/ubuntu/dists/jammy/winehq-jammy.sources
```

Based on https://gitlab.winehq.org/wine/wine/-/wikis/Debian-Ubuntu#jammy.

Thomasdezeeuw (approved):
Two small comments, but overall the LGTM.

.github/workflows/ci.yml
tbu-: You're right. Fixed.

tests/tcp_stream.rs
tbu-: Done.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @tbu-

PR #1966, opened 2026-07-07
Support Nuttx target os

wllenyj:
Enable pipe2/accept4/eventfd/poll capabilities.

Use `poll` as the selector instead of `epoll`, because the `epoll_event` structure is defined differently from the linux_like. Therefore, additional support may be needed; for now, let's get tokio working.

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -4,6 +4,7 @@
     not(any(
         target_os = "aix",
         target_os = "espidf",
+        target_os = "nuttx",
         target_os = "fuchsia",
         target_os = "haiku",
         target_os = "hermit",
@@ -448,6 +449,7 @@ impl Poll {
     not(any(
         target_os = "aix",
         target_os = "espidf",
+        target_os = "nuttx",
         target_os = "fuchsia",
         target_os = "haiku",
         target_os = "hermit",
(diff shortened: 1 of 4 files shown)


Thomasdezeeuw:
I'm looking at https://github.com/apache/nuttx/blob/e280ecf73297594e2e29eebbcbe27cb328128c39/include/sys/epoll.h#L100-L114 and https://docs.rs/libc/latest/libc/struct.epoll_event.html, the only difference I see is that `data` is an union instead of a `u64`, I think we paper over that pretty quickly. I think going with epoll is the best approach here, I want to include https://github.com/tokio-rs/mio/pull/1962 in the next release any way (which might take a while to merge).

wllenyj:
@Thomasdezeeuw Thanks for your work, it is very nice.
Supporting `epoll` is definitely the final solution. It may open other pr to progress `epoll` support ? 
Currently the nuttx rust app is not compiled, when using the tokio net feature. Therefore, I want to release a minimal, runnable version as soon as possible.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Alright, I merged this as. Thanks @wllenyj

PR #1969, opened 2026-07-11
Add wasm32-unknown-emscripten target support

### LINKED ISSUE #642: Unable to build when targeting wasm32-unknown-emscripten platform
I am currently unable to build mio when targeting wasm32-unknown-emscripten platform.

I am using current Rust nightly.

I receive following errors:

```
error[E0432]: unresolved import `self::unix::Events`
 --> ~/.cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.6.10/src/sys/mod.rs:5:5
  |
5 |     Events,
  |     ^^^^^^ no `Events` in `sys::unix`

error[E0432]: unresolved import `self::unix::Selector`
 --> ~/.cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.6.10/src/sys/mod.rs:7:5
  |
7 |     Selector,
  |     ^^^^^^^^ no `Selector` in `sys::unix`
```

I have not delved too much into WASM/Rust yet, so I am not sure if this should even work or if this is the right place to report this.

guybedford:
Resolves #642.

Adds `wasm32-unknown-emscripten` as a target for mio, with a CI job running the suite under Node. Opening as a draft for now since it depends on unreleased patchsets for Emscripten and Rust's libc.

This lays the groundwork for Tokio support for Emscripten. Initially I did not plan to PR Mio, but in discussion with @Darksonn and @Noah-Kennedy it was suggested to use Mio if possible, and this has worked out well in my opinion.

With https://github.com/emscripten-core/emscripten/pull/27207 landed, Emscripten now exposes a real epoll, so the existing Linux epoll selector can reused rather than adding a new backend.

### Patch Sets

**The only remaining dependency needed to land this is libc@0.2.190, which for now is pinned to its 0.2 branch.**

This PR required the following upstream patches, now all merged:

- [x] https://github.com/rust-lang/socket2/pull/660 - Socket2 Emscripten support (landed, pending release)
- [x] emscripten-core/emscripten#27207 — The epoll APIs for Emscripten
- [x] https://github.com/rust-lang/libc/pull/5427 libc epoll definitions
- [x] emscripten-core/emscripten#27306 — NODERAWSOCKETS: socket fstat and SO_TYPE/SO_LINGER/TCP_MAXSEG options
- [x] rust-lang/libc#5256 — emscripten: add in6_pktinfo struct (merged)
- [x] Rust nightly patches in stable

### Test status

The tests run under `-pthread -sPROXY_TO_PTHREAD` in Emscripten.

In order to support the blocking accept, it was necessary to refactor the test suite to support non-blocking sockets via a `util::accept` / `util::read` abstraction.

Tests run directly on the Emscripten Node.js build with `NODERAWSOCKETS` and `NODERAWFS` to provide a transparent runner without further harness configuration being necessary.

* 147 passed, 0 failed
* 3 ignored: `tcp_stream::raw_fd` (Emscripten `getsockname` reports `0.0.0.0` as the local address until a non-blocking connect completes; port is stable since emscripten-core/emscripten#27566, address fix pending).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,6 +101,44 @@ jobs:
       run: rustup target add wasm32-wasip2
     - name: Tests
       run: cargo test --target wasm32-wasip2 --all-features
+  TestEmscripten:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    # TEMPORARY: this job builds against the guybedford/emscripten `cf` fork,
+    # which carries the epoll callback, AF_UNIX pathname sockets, and multicast
+    # getsockopt patches this branch depends on. Once those land upstream the
+    # fork checkout can be dropped and this can move to a released emsdk.
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@nightly
+      with:
+        components: rust-src
+    # JSPI is the return-to-host bridge; it needs a Node with JSPI on by
+    # default (V8 13.x+). Recorded here so the cargo runner uses it rather than
+    # the Node bundled with emsdk.
+    - uses: actions/setup-node@v4
+      with:
+        node-version: 26
+    - name: Record JSPI-capable node
+      run: echo "JSPI_NODE=$(which node)" >> "$GITHUB_ENV"
+    - name: Install emsdk (LLVM + binaryen toolchain)
+      run: |
+        git clone --depth 1 https://github.com/emscripten-core/emsdk.git
+        ./emsdk/emsdk install latest
+        ./emsdk/emsdk activate latest
+    - name: Clone guybedford/emscripten (cf) fork over the emsdk tree
+      run: |
+        git clone --depth 1 --branch cf https://github.com/guybedford/emscripten.git em-fork
+        python3 em-fork/bootstrap.py
+        rm -rf ./emsdk/upstream/emscripten
+        ln -s "$(pwd)/em-fork" ./emsdk/upstream/emscripten
+    - name: Tests
+      run: |
+        source ./emsdk/emsdk_env.sh
+        export EM_CONFIG="$EMSDK/.emscripten"
+        CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUNNER="$JSPI_NODE" \
+        RUSTFLAGS="-A linker_messages -C target-feature=+atomics,+bulk-memory,+mutable-globals -C link-arg=-pthread -C link-arg=-sPROXY_TO_PTHREAD -C link-arg=-sNODERAWSOCKETS -C link-arg=-sNODERAWFS -C link-arg=-sALLOW_MEMORY_GROWTH=1 -C link-arg=-sEXIT_RUNTIME=1 -C link-arg=-sJSPI" \
+        cargo +nightly test --target wasm32-unknown-emscripten -Zbuild-std --all-features --tests --examples --lib
   Nightly:
     runs-on: ubuntu-latest
     timeout-minutes: 10
@@ -190,6 +228,7 @@ jobs:
           #- powerpc64-ibm-aix
           - riscv32imc-esp-espidf
           - sparcv9-sun-solaris
+          - wasm32-unknown-emscripten
           - wasm32-wasip1
           - x86_64-apple-darwin
           - x86_64-apple-ios
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ net = []
 log = { version = "0.4.8", optional = true }
 
 [target.'cfg(any(unix, target_os = "hermit", target_os = "wasi"))'.dependencies]
-libc = "0.2.183"
+libc = { git = "https://github.com/guybedford/libc", branch = "emscripten" }
 
 [target.'cfg(windows)'.dependencies.windows-sys]
 version = "0.61"
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ Currently supported platforms:
 * iOS
 * macOS
 * Solaris
+* Emscripten
 
 Mio can handle interfacing with each of the event systems of the aforementioned
 platforms. The details of their implementation are further discussed in the
--- a/README.md
+++ b/README.md
@@ -137,7 +137,6 @@ Currently supported platforms:
 * iOS
 * macOS
 * Solaris
-* Emscripten
 
 Mio can handle interfacing with each of the event systems of the aforementioned
 platforms. The details of their implementation are further discussed in the
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,7 +44,7 @@
     target_family = "wasm",
     not(any(target_os = "wasi", target_os = "emscripten"))
 ))]
-compile_error!("This wasm target is unsupported by mio. If using Tokio, disable the net feature.");
+compile_error!("This wasm target is unsupported by mio (only the WASI and Emscripten wasm targets are supported). If using Tokio, disable the net feature.");
 
 // macros used internally
 #[macro_use]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,7 +40,10 @@
 //!
 //! The available features are described in the [`features`] module.
 
-#[cfg(all(target_family = "wasm", not(target_os = "wasi")))]
+#[cfg(all(
+    target_family = "wasm",
+    not(any(target_os = "wasi", target_os = "emscripten"))
+))]
 compile_error!("This wasm target is unsupported by mio. If using Tokio, disable the net feature.");
 
 // macros used internally
--- a/src/sys/unix/tcp.rs
+++ b/src/sys/unix/tcp.rs
@@ -65,6 +65,7 @@ pub(crate) fn accept(listener: &net::TcpListener) -> io::Result<(net::TcpStream,
         // See https://github.com/tokio-rs/mio/issues/1445 for details
         all(not(target_arch="x86"), target_os = "android"),
         target_os = "dragonfly",
+        target_os = "emscripten",
         target_os = "freebsd",
         target_os = "fuchsia",
         target_os = "hurd",
--- a/tests/tcp_stream.rs
+++ b/tests/tcp_stream.rs
@@ -464,6 +464,10 @@ fn shutdown_both() {
 }
 
 #[cfg(unix)]
+#[cfg_attr(
+    target_os = "emscripten",
+    ignore = "local_addr not preserved across from_raw_fd on emscripten"
+)]
 #[test]
 fn raw_fd() {
     init();
@@ -636,6 +640,7 @@ fn tcp_shutdown_client_read_close_event() {
 #[cfg_attr(
     any(
         target_os = "android",
+        target_os = "emscripten",
         target_os = "hurd",
         target_os = "illumos",
         target_os = "solaris",
--- a/tests/tcp_stream.rs
+++ b/tests/tcp_stream.rs
@@ -466,7 +466,7 @@ fn shutdown_both() {
 #[cfg(unix)]
 #[cfg_attr(
     target_os = "emscripten",
-    ignore = "local_addr not preserved across from_raw_fd on emscripten"
+    ignore = "REMOVE BEFORE LANDING - pending Node.js 26.6.0 in runner"
 )]
 #[test]
 fn raw_fd() {
--- a/tests/udp_socket.rs
+++ b/tests/udp_socket.rs
@@ -407,6 +407,10 @@ fn smoke_test_connected_udp_socket(mut socket1: UdpSocket, mut socket2: UdpSocke
     assert!(socket2.take_error().unwrap().is_none());
 }
 
+#[cfg_attr(
+    target_os = "emscripten",
+    ignore = "libuv does not re-associate a connected UDP socket on reconnect"
+)]
 #[test]
 fn reconnect_udp_socket_sending() {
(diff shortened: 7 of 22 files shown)


src/poll.rs
```diff
@@ -444,6 +444,40 @@ impl Poll {
...
+    /// Create a `Poll` whose epoll set carries a persistent, non-blocking
+    /// readiness callback for the lifetime of the `Poll`. Instead of blocking in
+    /// [`Poll::poll`], the emscripten runtime delivers up to `capacity` ready
+    /// events to `callback` on a fresh host tick whenever the set makes progress
+    /// (needs no JSPI/ASYNCIFY). The callback is unregistered when the `Poll` is
+    /// dropped.
+    ///
+    /// The `callback` may freely re-enter this `Poll`/`Registry` (e.g. register
+    /// or deregister sources): it holds no internal borrow while running.
+    pub fn new_with_callback<F>(capacity: usize, callback: F) -> io::Result<Poll>
```
Thomasdezeeuw: I won't accept this API change, it goes against how the API works on non-empscripten.

Darksonn: The code already copies the events into the `Events` buffer, so instead of passing them to a callback, I don't see why you can't just keep them around in said buffer until the next call to `poll` and return them to the caller then. That wouldn't need any user-facing API change.

guybedford:
Yeah that could work - let me see if I can come up with a way for the
callback mechanism to only be on the Tokio side instead.

Darksonn:
Why do you want the callback mechanism on the Tokio side rather than just storing them in this buffer until the next time Tokio calls `poll`? I'm not sure we can safely invoke wakers from the callback anyway.

src/poll.rs
guybedford: I've removed these callback code paths entirely, so that we are only building on the standard epoll APIs here now.

@Darksonn let's continue to discuss this on the Tokio side with regards to the distinction between the JS threads, JS Promise Integration (JSPI) and single-threaded cooperative event loop runtime behaviours.

README.md
```diff
@@ -135,6 +135,7 @@ Currently supported platforms:
 * iOS
 * macOS
 * Solaris
+* Emscripten
```
Darksonn: Hmm, WASI isn't listed here?

Darksonn (commented):
It looks like you were able to reduce this down to merely adjusting cfgs, which is great, thanks!

README.md
guybedford: Yeah I also found that odd - I can add it here, or remove Emscripten, let me know which suits.

Thomasdezeeuw: It's because WASI support is experimental, e.g. even today a pr was opened for it: #1977.

Also @guybedford should I review this? It's still marked as draft.

guybedford: I've removed Emscripten from the readme for now while experimental like for WASI.

@Thomasdezeeuw thanks for checking in - yes this is ready for review. The PR just can't land until all pre-requisite patches have been released as listed in the PR description. It is complete in the mean time though.

src/lib.rs
```diff
@@ -40,7 +40,10 @@
 //!
 //! The available features are described in the [`features`] module.
 
-#[cfg(all(target_family = "wasm", not(target_os = "wasi")))]
+#[cfg(all(
+    target_family = "wasm",
+    not(any(target_os = "wasi", target_os = "emscripten"))
+))]
 compile_error!("This wasm target is unsupported by mio. If using Tokio, disable the net feature.");
```
Thomasdezeeuw: Do we want to mention emscripten in this error message, or is it clear enough as it is?

tests/tcp_stream.rs
```diff
@@ -464,6 +464,10 @@ fn shutdown_both() {
 }
 
 #[cfg(unix)]
+#[cfg_attr(
+    target_os = "emscripten",
+    ignore = "local_addr not preserved across from_raw_fd on emscripten"
```
Thomasdezeeuw: Can you elaborate on this? How/why does the local address change?

tests/udp_socket.rs
```diff
@@ -407,6 +407,10 @@ fn smoke_test_connected_udp_socket(mut socket1: UdpSocket, mut socket2: UdpSocke
     assert!(socket2.take_error().unwrap().is_none());
 }
 
+#[cfg_attr(
+    target_os = "emscripten",
+    ignore = "libuv does not re-associate a connected UDP socket on reconnect"
```
Thomasdezeeuw: How is this related to libuv?

Thomasdezeeuw (commented):
I'm not sure what to do about the missing public api (UnixStream::pair and UnixDatagram). Removing them, as this API does, is one option, we can also include them, but always return an unsupported error.

What do you think @Darksonn?

Darksonn:
My suggestion would be to remove the unsupported APIs. IMO it's no different from the fact that `UnixDatagram` also does not exist on the windows target.

tests/udp_socket.rs
guybedford: The test runner here is Node.js. See https://github.com/libuv/libuv/pull/5193 for the upstream PR (which it seems will likely not be approved).

tests/tcp_stream.rs
guybedford: This was actually a Node.js bug, which I've fixed - and will be passing by 26.6.0 releasing next. I expect we can remove this guard entirely and use that version in the runner by the time this is ready to land.

guybedford:
> My suggestion would be to remove the unsupported APIs. IMO it's no different from the fact that UnixDatagram also does not exist on the windows target.

I've gone ahead and implemented this.

src/lib.rs
guybedford: Updated the message to mention WASI and Emscripten.

guybedford:
To give a progress update here - Emscripten landed support for epoll yesterday in https://github.com/emscripten-core/emscripten/pull/27207 🥳

This PR is now just awaiting two more PRs expected to land this week (libc and Emscripten as per the description checklist), at which point it'll be ready for final review.

In the mean time, the code remains feature complete, so reviews continue to be very much welcome.

Darksonn:
Looks like CI is failing

guybedford:
@Darksonn thanks, I've pushed up a change to fix it.

guybedford:
This PR is now ready for review, and no longer draft. All upstream patchsets have now landed.

We are just waiting for the libc@0.2.190 release further.

In the mean time, reviews are very welcome. //cc @Darksonn

src/sys/unix/tcp.rs
Darksonn, line 84: Do you also need to set the nonblocking flag via fcntl here? What about the `pipe2` syscall?

.github/workflows/ci.yml
```diff
@@ -101,6 +101,24 @@ jobs:
...
+        components: rust-src
+    - uses: mymindstorm/setup-emsdk@v14
+      with:
+        version: 6.0.9
+    # Sockets are Node raw sockets; the test peers run on pthreads under
+    # PROXY_TO_PTHREAD so their blocking waits can park the worker.
+    - name: Tests
+      run: |
+        CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUNNER=node \
+        RUSTFLAGS="-A linker_messages -C target-feature=+atomics,+bulk-memory,+mutable-globals -C link-arg=-pthread -C link-arg=-sPROXY_TO_PTHREAD -C link-arg=-sNODERAWSOCKETS -C link-arg=-sNODERAWFS -C link-arg=-sALLOW_MEMORY_GROWTH=1 -C link-arg=-sEXIT_RUNTIME=1" \
```
Darksonn: What is being silenced by `-Alinker_messages` here?

tests/util/mod.rs
```diff
@@ -208,6 +208,79 @@ pub fn assert_would_block<T>(result: io::Result<T>) {
...
+#[cfg(not(target_os = "emscripten"))]
+pub fn read<S: io::Read>(stream: &mut S, buf: &mut [u8]) -> io::Result<usize> {
+    stream.read(buf)
+}
+
+// Emscripten sockets never block: a call that would block returns `EAGAIN`
+// and the caller is expected to wait for readiness itself. `poll(2)` does
+// block on a pthread, which is where these peers run.
+#[cfg(target_os = "emscripten")]
+pub fn accept<L: Accept + AsRawFd>(listener: &L) -> io::Result<L::Conn> {
```
Darksonn, line 251: Aren't we using JSPI precisely to make blocking syscalls such as this one possible without special code such as this?

.github/workflows/ci.yml
guybedford: Two things:

1. `-Wpthreads-mem-growth`, which warns to use `-sGROWABLE_ARRAYBUFFERS=2` since shared array buffer `memory.grow` causes existing data views to be refreshed.
2. `-Wdeprecated` for `-sWASM_BIGINT` - Emscripten marked this as deprecated now that it is on by default, and rustc passes this option by default.

I've dropped the allow memory growth to rather set a large enough memory upfront, and added a specific silencing for (2) instead.

src/sys/unix/tcp.rs
guybedford: The Emscripten `NODERAWSOCKETS` implementation initially followed the websockets proxy approach - which treats all sockets as non-blocking already.

To better align with Linux semantics I've updated this to use plain `accept` with `fctl` instead of `accept4`. While it costs an extra call, it keeps the VM semantically correct.

tests/util/mod.rs
guybedford: This Mio implementation doesn't use JSPI, as that requires extra ceremony per the Tokio PR. Instead pthreads uses a JS worker to support blocking calls which gets things to mostly _just work_, and avoiding the need for the blocking JSPI Emscripten PR.

Darksonn: Still, if you're using pthreads, why can't we just use a blocking call here instead of looping on a non-blocking call?

Thomasdezeeuw: I'm also not a fan of all this special code for testing. I would prefer if we can do without it.

Cargo.toml
```diff
@@ -47,7 +47,7 @@ net = []
 log = { version = "0.4.8", optional = true }
 
 [target.'cfg(any(unix, target_os = "hermit", target_os = "wasi"))'.dependencies]
-libc = "0.2.183"
+libc = { git = "https://github.com/rust-lang/libc", branch = "libc-0.2" }
```
Thomasdezeeuw, line 50: Blocking: this needs a released version (I assume this is just a matter of time).

.github/workflows/ci.yml
```diff
@@ -101,6 +101,25 @@ jobs:
...
       run: cargo test --target wasm32-wasip2 --all-features
+  TestEmscripten:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+    - uses: actions/checkout@v7
+    - uses: dtolnay/rust-toolchain@nightly
+      with:
+        components: rust-src
+    - uses: mymindstorm/setup-emsdk@v14
```
Thomasdezeeuw, line 112: Seems like the repo changed to https://github.com/emscripten-core/setup-emsdk.

Thomasdezeeuw: It also updated to v15.

Thomasdezeeuw (changes requested):

.github/workflows/ci.yml
Darksonn: If so, can you submit a PR to update it on Tokio too?

tests/util/mod.rs
guybedford: I had an Emscripten PR to do just that open for two months here - https://github.com/emscripten-core/emscripten/pull/27342.

I just closed it to use this approach instead, because there was resistance to supporting a dual path sync/async function upstream. Accept behaves sync without jspi/pthreads, and async under jspi/pthreads, but Emscripten doesn't have a way to switch on those modes, which is what my PR added. Otherwise you need two variants of every blocking function in JS - a sync version and an async version. All difficult sells for Emscripten.

So my test refactor here was done to ensure this PR can land sooner (as soon as the next libc release).

Ideally we could just not use blocking sockets in the test suite here and everything would work out. I haven't really never needed support for blocking accept yet in test applications.

guybedford: I've posted a new attempt towards this in https://github.com/emscripten-core/emscripten/pull/27724.

Depending on progress, we can decide if we want this PR to block on that or not.

I do think ideally we should land this since it only affects the tests, and do the refactoring as a follow-on, now that it has tracking PR again.

Darksonn: I think if emscripten's goal is to get C programs to work verbatim without changes, you'll need these blocking calls to work. I would be ok with doing it *temporarily* if we can fix it in future emscripten, but it's Thomas's call.

Cargo.toml
guybedford: Yes, we are just waiting on 0.2.190.

tests/util/mod.rs
guybedford: Certianly, we can see how progress goes with that - but I would strong suggest that once libc@0.2.190 is released, we shouldn't hold back the patchset on it if it is still pending.

Thomasdezeeuw: I'm ok with merge the change to src code, but not the tests. As-is it makes too many changes that really shouldn't be necessary imo.

Darksonn: We can disable the relevant tests on emscripten for now, and add them back once it's supported.

PR #1970, opened 2026-07-13
Release v1.2.2

Thomasdezeeuw:
Note this add NuttX to the supported list of OS in the readme as it still need (released) changes in other crates (libc, etc.). We ideally also want it to use the epoll based implementation rather than poll.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 1.2.2
+
+* Add support for Solaris using event ports
+  (https://github.com/tokio-rs/mio/pull/1962).
+* Add experiment support for NuttX using `poll(2)`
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # 1.2.1
```
Darksonn:
```suggestion
# 1.2.2
```
But maybe it should be 1.3.0?

Darksonn (approved):

CHANGELOG.md
Thomasdezeeuw: Not much changed for OS other than Solaris and NuttX, nothing in fact. So I don't think it's worth a 1.3.

PR #1973, opened 2026-07-14
Extend documentation about implict source deregistrations

### LINKED ISSUE #1972: Necessity of deregistering source before dropping it
Hi,

I'm wondering whether it's required for a source to be deregistered before it's being dropped. 
From the docs I wasn't able to tell, and after looking at the `IoSource` implementation and the kqueue and epoll man pages I'd say it should be fine.
For the selector used on Windows I can't tell. 

However, due to the missing documentation[^1] I'm a bit uncertain and wanted to ask whether that's actually intended behavior.

Whatever the outcome of this discussion is, I'm thinking about opening a PR which documents the behavior on https://docs.rs/mio/latest/mio/struct.Registry.html#method.deregister.

[^1]: I've looked at https://docs.rs/mio/latest/mio/struct.Registry.html#method.deregister

WhySoBad:
This pull request extends the documentation of the `Registry::deregister` method to include a note about what happens when the last file descriptor (or file handle) of a registered event source is closed.
The purpose of this addition is to tell "mio is fine with not deregistering a source as long as it doesn't break the underlying selector". I've intentionally didn't add a list of selectors for which this behavior applies as it might not be exhaustive.

Closes #1972

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -670,6 +670,10 @@ impl Registry {
     /// deregistered; however, it must be passed back to the **same** `Poll`
     /// instance, otherwise the behavior is unspecified.
     ///
+    /// After the last file descriptor (or file handle) for a registered event source
+    /// is closed, it will be automatically deregistered if the underlying selector
+    /// provides this behavior.
+    ///
     /// # Examples
     ///
     #[cfg_attr(all(feature = "os-poll", feature = "net"), doc = "```")]


src/poll.rs
```diff
@@ -670,6 +670,10 @@ impl Registry {
     /// deregistered; however, it must be passed back to the **same** `Poll`
     /// instance, otherwise the behavior is unspecified.
     ///
+    /// After the last file descriptor (or file handle) for a registered event source
+    /// is closed, it will be automatically deregistered if the underlying selector
+    /// provides this behavior.
```
Thomasdezeeuw, line 675: I'm not sure about this. It's so noncommittal that I'm not convinced it adds anything to be honest.

WhySoBad: Fair enough.
I just thought it might be something worth knowing when implementing your own async runtime on top of mio.

WhySoBad:
I'll close this now. If others run into the same question in the future, it might be worth merging this.

PR #1974, opened 2026-07-16
Mark Wine as supported

tbu-:
mio is tested on Wine since #1965.

Additionally, #1444 hasn't been relevant for Wine support for a while.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ Currently supported platforms:
 * NetBSD
 * OpenBSD
 * Windows
+* Wine
 * iOS
 * macOS
 * Solaris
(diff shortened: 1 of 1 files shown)


tbu-:
I don't want to make decisions on behalf of mio. If running CI is considered enough for "supported" status, I'll also happily move it to the supported section.

Thomasdezeeuw (commented):
Since we now check it in the CI let's move it to the supported section, we can remove the unsupported section entirely.

tbu-:
Done.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @tbu-

PR #1977, opened 2026-07-20
enable various tests for WASI and remove obsolete `cfg` guards

dicej:
A number of tests were temporarily disabled for WASI while `wasi-libc` and `wasmtime` patches made their way into releases.  Now that Rust is using the latest `wasi-sdk` (and thus `wasi-libc`) release, we can finally enable them and cut down on the `cfg` noise.

- remove `if cfg!(target_os = "wasi") { return Ok(()) }` from `UdpSocket` doctest in `src/net/udp.rs`
- use `fcntl` instead of `ioctl` for WASI in `src/sys/unix/tcp.rs`
- enable temporarily-disabled-on-WASI tests in `tests/registering.rs`, `tests/tcp.rs`, and `tests/udp_socket.rs`
- ungate `assert_socket_non_blocking` for WASI in `tests/util/mod.rs`, `tests/tcp_listener.rs`, `tests/tcp_stream.rs`, and `tests/udp_socket.rs`

Closes #1932

### DIFF
--- a/tests/tcp_listener.rs
+++ b/tests/tcp_listener.rs
@@ -11,14 +11,14 @@ use std::thread;
 
 mod util;
 use util::{
-    any_local_address, any_local_ipv6_address, assert_send, assert_sync, assert_would_block,
-    expect_events, expect_no_events, init, init_with_poll, ExpectEvent,
+    any_local_address, any_local_ipv6_address, assert_send, assert_socket_non_blocking,
+    assert_sync, assert_would_block, expect_events, expect_no_events, init, init_with_poll,
+    ExpectEvent,
 };
 
-// Close-on-exec doesn't apply to WASI; non-blocking does, but `wasi-libc`
-// doesn't support it yet (https://github.com/WebAssembly/wasi-libc/pull/742).
+// Close-on-exec doesn't apply to WASI
 #[cfg(not(target_os = "wasi"))]
-use util::{assert_socket_close_on_exec, assert_socket_non_blocking};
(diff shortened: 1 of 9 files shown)


tests/tcp_listener.rs
```diff
@@ -11,14 +11,14 @@ use std::thread;
...
+    assert_sync, assert_would_block, expect_events, expect_no_events, init, init_with_poll,
+    ExpectEvent,
 };
 
-// Close-on-exec doesn't apply to WASI; non-blocking does, but `wasi-libc`
-// doesn't support it yet (https://github.com/WebAssembly/wasi-libc/pull/742).
+// Close-on-exec doesn't apply to WASI
 #[cfg(not(target_os = "wasi"))]
-use util::{assert_socket_close_on_exec, assert_socket_non_blocking};
+use util::assert_socket_close_on_exec;
```
Thomasdezeeuw: I wonder if we should make `assert_socket_close_on_exec` a noop on Wasi (and other OS that don't have threads, but that's out of scope for this pr), so have a single `cfg` in the function instead of multiple for each call site.

dicej: SGTM; I'll make that change.

Thomasdezeeuw (approved):
LGTM.

Do we want to add WASI to the supported OS list in the readme?

Thomasdezeeuw:
Thanks @dicej

dicej:
> Do we want to add WASI to the supported OS list in the readme?

Sure, I'll open a PR to do that.

PR #1978, opened 2026-07-21
ci: add Miri to CI

WhySoBad:
Hi,

This pull request adds a CI job to run the test suite, as well as the doc comments on Miri.
With the addition of the TCP socket and `poll` shims, Miri now supports a large enough portion of mio (and its test suite) that I thought it might be worth adding a CI job.

Also, in Miri we currently can't nicely test tokio with the `poll` selector as an integration test. We thought that it might be an option to add another job which runs the mio test suite with the `mio_unsupported_force_poll_poll` flag on a Linux target (even though that would be testing behavior this is not officially supported). What do you think of this?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,9 @@ env:
   RUST_BACKTRACE: full
   CI: true
   WASMTIME_VERSION: 47.0.0
+  # Pin a specific miri version
+  rust_miri_nightly: nightly-2026-05-30
+  MIO_MIRI_FEATURES: log,os-ext,os-poll,net
 
 permissions:
   contents: read #  to fetch code (actions/checkout)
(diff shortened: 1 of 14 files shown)


.github/workflows/ci.yml
```diff
@@ -138,6 +141,40 @@ jobs:
...
+      - uses: actions/checkout@v7
+      - name: Install Rust ${{ env.rust_miri_nightly }}
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: ${{ env.rust_miri_nightly }}
+          components: miri
+      - uses: Swatinem/rust-cache@v2
+      - name: miri-doc-test
+        run: |
+          cargo miri test --doc --features $MIO_MIRI_FEATURES --no-fail-fast
```
Thomasdezeeuw: Why is this a separate job? Can we merge it with the other job?

.github/workflows/ci.yml
```diff
@@ -138,6 +141,40 @@ jobs:
...
+          toolchain: ${{ env.rust_miri_nightly }}
+          components: miri
+      - name: Install cargo-nextest
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-nextest
+      - uses: Swatinem/rust-cache@v2
+      - name: miri
+        run: |
+          cargo miri nextest run --features $MIO_MIRI_FEATURES --test '*' --no-fail-fast
```
Thomasdezeeuw: Why are we using `nextest`? Can we use the normal test?

.github/workflows/ci.yml
```diff
@@ -138,6 +141,40 @@ jobs:
...
       #cargo fmt --all -- --check
       run: find src tests examples -type f -iname "*.rs" | xargs rustfmt --check
+  MiriTest:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - name: Install Rust ${{ env.rust_miri_nightly }}
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: ${{ env.rust_miri_nightly }}
```
Thomasdezeeuw, line 150: Do we need nightly? Can we run on stable?

.github/workflows/ci.yml
Thomasdezeeuw: Can use `--all-features` instead of a list?

.github/workflows/ci.yml
WhySoBad: Nightly is needed as Miri is only shipped with the nightly toolchain.

.github/workflows/ci.yml
WhySoBad: I've now changed it to use the normal test. I thought the parallelism of nexttest might be nice because Miri is quite slow.

src/net/udp.rs
```diff
@@ -31,8 +31,8 @@ use crate::{event, sys, Interest, Registry, Token};
 ///
 /// # Examples
 ///
-#[cfg_attr(feature = "os-poll", doc = "```")]
-#[cfg_attr(not(feature = "os-poll"), doc = "```ignore")]
+#[cfg_attr(all(feature = "os-poll", not(miri)), doc = "```")]
+#[cfg_attr(any(not(feature = "os-poll"), miri), doc = "```ignore")] // Miri doesn't support UDP sockets.
```
Darksonn: There is no reason to be overly clever here. Let's just use `x` and `not(x)` as the two conditions.
```suggestion
#[cfg_attr(all(feature = "os-poll", not(miri)), doc = "```")]
#[cfg_attr(not(all(feature = "os-poll", not(miri))), doc = "```ignore")] // Miri doesn't support UDP sockets.
```

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @WhySoBad

WhySoBad:
> Also, in Miri we currently can't nicely test tokio with the `poll` selector as an integration test. We thought that it might be an option to add another job which runs the mio test suite with the `mio_unsupported_force_poll_poll` flag on a Linux target (even though that would be testing behavior this is not officially supported). What do you think of this?

Just wanted to ask whether you saw this question in the PR description, and if not, what you think of this.

Thomasdezeeuw:
> Just wanted to ask whether you saw this question in the PR description, and if not, what you think of this.

Sure we can add another step for this.

PR #1982, opened 2026-07-27
ci: run tests with Miri using forced poll selector

WhySoBad:
As mentioned in https://github.com/tokio-rs/mio/pull/1978, this PR adds an additional step to the Miri CI job which runs the test suite with Miri using the forced `poll` selector.


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @WhySoBad

PR #1987, opened 2026-07-30
Avoid Waker setup for kqueue selector

Thomasdezeeuw:
For the Waker implementation based on the kqueue selector we used to setup a listener for EVFILT_USER events before send the events on each wake-up.

However, this is not actually needed if we use EV_ONESHOT.

### DIFF
--- a/src/sys/unix/waker/kqueue.rs
+++ b/src/sys/unix/waker/kqueue.rs
@@ -18,7 +18,6 @@ pub(crate) struct Waker {
 impl Waker {
     pub(crate) fn new(selector: &Selector, token: Token) -> io::Result<Waker> {
         let selector = selector.try_clone()?;
-        selector.setup_waker(token)?;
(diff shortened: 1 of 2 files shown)


src/sys/unix/waker/kqueue.rs
Darksonn, line 10: these docs are now wrong, setup no longer needed

Darksonn (approved):

src/sys/unix/waker/kqueue.rs
Thomasdezeeuw: You're right, I ended up removing them in a follow up pr: #1988.

Thomasdezeeuw:
I think I merged this too quickly and the FreeBSD failures seems legit, going to look into it.

Thomasdezeeuw:
Partially reverting this in #1993.

PR #1988, opened 2026-07-30
Merge kqueue and event_ports Waker implementations

Thomasdezeeuw:
Now that the kqueue implementation no longer uses any setup, the two implementations are the same.


src/sys/unix/mod.rs
```diff
@@ -79,15 +79,12 @@ cfg_os_poll! {
...
+            target_os = "solaris",
             target_os = "tvos",
             target_os = "visionos",
             target_os = "watchos",
         )
-    ), path = "waker/kqueue.rs")]
-    #[cfg_attr(all(
-        not(mio_unsupported_force_poll_poll),
-        target_os = "solaris",
-    ), path = "waker/event_ports.rs")]
```
Darksonn, line 90: What about `--cfg mio_unsupported_force_waker_pipe` on solaris?

Darksonn (approved):

src/sys/unix/mod.rs
Thomasdezeeuw: That doesn't compile on master at the moment, so I'm going to assume it's not used and not worry about it. We can fix it up if we get complaints.

PR #1991, opened 2026-07-30
Reuse IoSourceState between poll and event_port implementations

Thomasdezeeuw:
The two implementations were exactly the same, so reuse them between the two selector implementations.

### DIFF
--- a/src/sys/unix/selector/registered_io_source.rs
+++ b/src/sys/unix/selector/registered_io_source.rs
@@ -1,0 +1,151 @@
+//! Both `poll(2)` and `event_port(2)` need to hold per-fd states.
+
+use std::sync::atomic::{AtomicBool, Ordering};
+
+cfg_io_source! {
+use std::io;
+#[cfg(not(target_os = "hermit"))]
+use std::os::fd::RawFd;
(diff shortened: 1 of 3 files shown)


src/sys/unix/selector/registered_io_source.rs
```diff
@@ -0,0 +1,151 @@
...
+use std::sync::atomic::{AtomicBool, Ordering};
+
+cfg_io_source! {
+#[cfg(not(target_os = "hermit"))]
+use std::os::fd::RawFd;
+// TODO: once <https://github.com/rust-lang/rust/issues/126198> is fixed this
+// can use `std::os::fd` and be merged with the above.
+use std::io;
+#[cfg(target_os = "hermit")]
+use std::os::hermit::io::RawFd;
```
Darksonn, line 12: This `use std::io;` line goes in the way of the comment and the thing its commenting on.

Darksonn (approved):

src/sys/unix/selector/registered_io_source.rs
Thomasdezeeuw: Good catch, fixed.

PR #1993, opened 2026-07-31
Fix Waker for FreeBSD and switch DragonflyBSD, NetBSD and OpenBSD to EVFILT_USER based waker

Thomasdezeeuw:
See the individual commits.


Thomasdezeeuw:
Final one for today @Darksonn :) thanks for all the reviews

Darksonn (approved):

PR #1994, opened 2026-07-31
event: report remaining, not total, from Iter::size_hint and count

### LINKED ISSUE #1990: `event::Iter::size_hint` and `count` report the total, not the remaining
## Summary

`Iter::size_hint` returns the total number of events in the container rather than the number
of events remaining, so after `k` calls to `next()` it over-reports by `k`. `Iter::count` has
the same shape: it ignores the current position and returns the total.

`Iterator::size_hint` is specified as "the bounds on the remaining length of the iterator", and
`Iterator::count` as "the number of iterations", i.e. the count of items the iterator will
still yield.

No unsoundness — see Impact.

## The code

`src/event/events.rs:203-224` (1.2.1):

```rust
impl<'a> Iterator for Iter<'a> {
    type Item = &'a Event;

    fn next(&mut self) -> Option<Self::Item> {
        let ret = self
            .inner
            .inner
            .get(self.pos)
            .map(Event::from_sys_event_ref);
        self.pos += 1;
        ret
    }

    fn size_hint(&self) -> (usize, Option<usize>) {
        let size = self.inner.inner.len();
        (size, Some(size))
    }

    fn count(self) -> usize {
        self.inner.inner.len()
    }
}
```

`self.pos` is the read cursor, advanced by `next()`. Neither `size_hint` nor `count` consults it.

## Reproduction

```rust
let mut it = events.iter();          // suppose 4 events are present
let _ = it.next();
let _ = it.next();
assert_eq!(it.size_hint(), (2, Some(2)));  // fails: reports (4, Some(4))
assert_eq!(it.count(), 2);                 // fails: returns 4
```

## Impact

Low, and bounded by a trait that is deliberately absent.

- **No unsoundness.** `Iter` implements neither `ExactSizeIterator` nor `TrustedLen`, so no
  caller is entitled to treat the hint as authoritative for unchecked capacity or unchecked
  indexing. Adapters use it only as an allocation hint.
- The observable effects are an over-allocation in `collect()` / `extend()` when the iterator
  has been partially consumed, and a wrong answer from `count()` in the same situation.
- Callers who iterate a fresh `Iter` to exhaustion — which is the documented and overwhelmingly
  common pattern, `for event in &events` — see correct behaviour, because at position zero the
  total and the remaining count coincide.

## Suggested fix

```rust
    fn size_hint(&self) -> (usize, Option<usize>) {
        let size = self.inner.inner.len().saturating_sub(self.pos);
        (size, Some(size))
    }

    fn count(self) -> usize {
        self.inner.inner.len().saturating_sub(self.pos)
    }
```

`saturating_sub` rather than `-` is required, not stylistic: `next()` increments `self.pos`
even on the call that returns `None` (`events.rs:212`), so `pos` can exceed `len()` once the
iterator is exhausted and a plain subtraction would underflow.

Since the corrected `size_hint` would then be exact, `impl ExactSizeIterator for Iter<'_>` becomes
available if upstream wants it — but that is a separate, additive decision and is not required
to fix the contract violation.

## How this was found

Line-by-line read of mio 1.2.1 during a `cargo vet` certification of warden's dependency graph,
reading `src/event/events.rs` against the `Iterator` trait contract.

Not verified by execution: the reproduction above is derived from the source and has not been
run against a live `Poll`.

## Upstream context

Checked against the default branch before filing: `src/event/events.rs:216-222` is unchanged --
`size_hint` returns `self.inner.inner.len()` and `count` returns the same, neither consulting
`self.pos`.

3x3xX3N0N:
Fixes #1990.

`size_hint` and `count` returned the container's total rather than the number of items remaining, so after `k` calls to `next()` they over-reported by `k`.

`saturating_sub` rather than `-` because `next()` increments `pos` even on the call that returns `None`, so `pos` can exceed `len()` once the iterator is exhausted.

### DIFF
--- a/src/event/events.rs
+++ b/src/event/events.rs
@@ -214,12 +214,12 @@ impl<'a> Iterator for Iter<'a> {
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        let size = self.inner.inner.len();
+        let size = self.inner.inner.len().saturating_sub(self.pos);
         (size, Some(size))
     }
 
     fn count(self) -> usize {
-        self.inner.inner.len()
+        self.inner.inner.len().saturating_sub(self.pos)
     }
 }
 
--- a/tests/events.rs
+++ b/tests/events.rs
@@ -44,6 +44,39 @@ fn events_all() {
 }
 
 #[test]
+fn iter_size_hint_and_count_report_remaining() {
+    let (mut poll, mut events) = init_with_poll();
+
+    let waker = Waker::new(poll.registry(), WAKE_TOKEN).unwrap();
+
+    waker.wake().expect("unable to wake");
+    poll.poll(&mut events, Some(Duration::from_millis(100)))
+        .unwrap();
+
+    let total = events.iter().count();
+    assert!(total > 0, "expected at least one event");
+
+    // A fresh iterator reports everything it is going to yield.
+    let mut iter = events.iter();
+    assert_eq!(iter.size_hint(), (total, Some(total)));
+
+    // Consuming an element leaves one fewer, rather than still reporting the
+    // total.
+    iter.next().expect("expected an event");
+    let remaining = total - 1;
+    assert_eq!(iter.size_hint(), (remaining, Some(remaining)));
+    assert_eq!(iter.count(), remaining);
+
+    // `next` advances the cursor past the end once exhausted, so this also
+    // covers the cursor exceeding the length.
+    let mut iter = events.iter();
+    while iter.next().is_some() {}
+    assert!(iter.next().is_none());
+    assert_eq!(iter.size_hint(), (0, Some(0)));
+    assert_eq!(iter.count(), 0);
+}
+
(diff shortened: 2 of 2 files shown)


Thomasdezeeuw (commented):
Could you add a test to ensure we don't regress?

3x3xX3N0N:
Added in tests/events.rs — covers a partially consumed iterator and an exhausted one, the latter being the case that needs the saturating subtraction. It fails on master and passes with the fix.

3x3xX3N0N:
Rebased onto current master. The FreeBSD failures were the base, not this change — the branch predated b2504b4 ("Add back Waker setup for FreeBSD"), so every Waker-driven test failed there, including the pre-existing events_all. The two CheckTargets failures (nintendo-3ds, esp-espidf) also reproduce on current master.

3x3xX3N0N:
Closing this PR; withdrawing the contribution. Apologies for the noise.

Issue #1046, opened 2019-07-22
Determine how to do deal with WSAStartup on Windows


Thomasdezeeuw:
We first saw the issue in #1029, where the following error appeared: `Os { code: 10093, kind: Other, message: "Either the application has not called WSAStartup, or WSAStartup failed." }`.

We can't initialise it ourselves as standard library has an assertion to check if the return value is ok: https://github.com/rust-lang/rust/blob/00859e3e653973120006aaf3227823062dde1ba7/src/libstd/sys/windows/net.rs#L38.

Thomasdezeeuw:
We could do the same as in the standard library add an `init` function that somehow triggers the standard library's init function. For the tests in #1029 I added a call to `TcpListener::bind` that is dropped immediately: https://github.com/tokio-rs/mio/blob/dae8291aed0fa0c5aa7b449c630712cb2c441da7/tests/util/mod.rs#L137

However that is not really a pretty solution.

carllerche:
How exactly did you trigger it? What fn did you call that returned the error?

carllerche:
I see you are calling socket directly. Would it be possible to create all sockets by going through `std` types?

Thomasdezeeuw:
Maybe, but then we would be then we would miss out on the optimisation to set `NONBLOCK` and `CLOEXEC` in the call to `socket`, see https://github.com/tokio-rs/mio/blob/dae8291aed0fa0c5aa7b449c630712cb2c441da7/src/sys/unix/mod.rs#L76

When we create a socket via the standard library we need to set non blocking after, incurring another system call.

carllerche:
@Thomasdezeeuw I only mean do this on windows. Do those optimizations apply to windows? I know `CLOEXEC` wouldn't (since there is no fork). I'm not sure about the other one.

Thomasdezeeuw:
For `TcpStream::connect` we need to set the socket to non-blocking mode before connecting otherwise the connect system call would block.

carllerche:
@Thomasdezeeuw This is how `net2` does it: https://github.com/rust-lang-nursery/net2-rs/blob/master/src/sys/windows/mod.rs#L55-L64

Probably the best way...

andrewtj:
If an `init()` like function is going to be added, can it be made public and perhaps also be a no-op on other platforms? I ask as I'm currently doing a similar throw away bind behind a `Once` in one of my crates.

Thomasdezeeuw:
[...]

PR #1996, opened 2026-08-03
Call `WSAStartup` ourselves instead of relying on implicit init

tbu-:
Previously, `UdpSocket::bind("127.0.0.1:0")` was called, creating and destroying a socket and assuming the standard library would initialize (and not deinitialize) the socket interface.

#1046 claimed that `WSAStartup` can only be called once by each process. I believe that this is wrong and has never been the case. See e.g. this question on StackOverflow from 2008: ["What is the cost of WSAStartup and WSACleanup"](https://stackoverflow.com/q/396484) which mentions calling `WSAStartup` multiple times and quotes MSDN documentation for it.

### DIFF
--- a/src/sys/windows/net.rs
+++ b/src/sys/windows/net.rs
@@ -4,19 +4,24 @@ use std::net::SocketAddr;
 use std::sync::Once;
 
 use windows_sys::Win32::Networking::WinSock::{
-    closesocket, ioctlsocket, WSASocketW, AF_INET, AF_INET6, FIONBIO, IN6_ADDR, IN6_ADDR_0,
-    INVALID_SOCKET, IN_ADDR, IN_ADDR_0, SOCKADDR, SOCKADDR_IN, SOCKADDR_IN6, SOCKADDR_IN6_0,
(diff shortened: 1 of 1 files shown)


tbu-:
(Unrelated to Wine, I just found the code to be a bit weird.)

Thomasdezeeuw:
I'm 100% certain that at some point Windows and/or Rust's standard library did not like that Mio called `WSAStartup`. We have similar code in socket2: https://github.com/rust-lang/socket2/blob/242670a575c37cab6f74e778949417ba1470c8ba/src/sys/windows.rs#L283-L288. I'm not sure if something has changed, but I'm not sure how much time I want to spend on this to be honest.

tbu-:
> I'm 100% certain that at some point Windows and/or Rust's standard library did not like that Mio called `WSAStartup`.

Perhaps that was just a mistake in testing.

> We have similar code in socket2: https://github.com/rust-lang/socket2/blob/242670a575c37cab6f74e778949417ba1470c8ba/src/sys/windows.rs#L283-L288.

The socket2 code just shows it's another way to call WSAStartup (maybe socket2 didn't want to use WSAStartup directly)?

> I'm not sure how much time I want to spend on this to be honest.

Sure, I think you're wrong, but the current code works too, even if it's a bit weird.

PR #2000, opened 2026-08-08
Allow CI failures in tier 2 jobs

Thomasdezeeuw:
A lot of these OS fail to build the standard library at various points (see the issues list). For those cases we don't want to fail the CI any more.


Thomasdezeeuw:
Still marks the CI as failing :( Might just have to remove the OS from the matrix.

PR #2001, opened 2026-08-09
windows: only re-arm the direction that blocked

### LINKED ISSUE #1963: Windows: re-arm on a read WouldBlock produces a spurious writable event
On the Windows AFD selector, a socket registered `READABLE | WRITABLE` emits a spurious writable event on every read.

After a read returns `WouldBlock`, `IoSource::do_io` reregisters the socket's full stored interest; `set_event` resets `user_evts`, re-adding `POLL_SEND`, and since the socket is writable the next AFD poll completes immediately with `POLL_SEND`. So each read attempt produces a `WRITABLE` event nothing asked for.

**Impact.** This livelocks a `tokio` reader driven from a non-runtime thread — the spurious writable bumps the readiness tick and invalidates the reader's `clear_readiness`, so `poll_read` spins. Verified repro (Hyper): tokio-rs/tokio#8054.

**Fix direction.** Re-arm only the direction that blocked ("only reregister read", from the tokio-rs/tokio#8054 discussion). The wrinkle: `do_io` is direction-blind — `Read`/`Write` call `do_io(f)` with an opaque closure — so narrowing the re-arm touches the public `do_io`/`try_io` surface, not just the AFD `SockState`. Design discussion on tokio-rs/tokio#8054.

ameyypawar:
A socket registered `READABLE | WRITABLE` gets a writable event on every blocked read: `do_io` reregisters the full stored interest, `set_event` re-adds `POLL_SEND`, and the AFD poll completes with it immediately. Downstream this livelocks a tokio reader (tokio-rs/tokio#8054).

`do_io_with` re-arms only the direction the caller operated on, through a new `rearm` that adds to the events already being waited for rather than replacing them — narrowing one direction must not drop readiness still wanted in the other. Applied to `TcpStream`, `TcpListener` and `UdpSocket`, the types that reach this path on Windows. `try_io` keeps using `do_io`, since a caller-supplied closure has no known direction. No public API change.

Verified on Windows CI: the added test fails on `master` and passes with this change.

The tests also fail on `poll(2)`, `event_ports(2)` and WASI, which re-arm the same way; they are skipped there and left for a follow-up.

Fixes #1963

### DIFF
--- a/src/io_source.rs
+++ b/src/io_source.rs
@@ -69,6 +69,39 @@ impl<T> IoSource<T> {
         self.state.do_io(f, &self.inner)
     }
 
+    /// Same as [`do_io`], for an operation known to only involve `interest`.
+    ///
+    /// Selectors that have to re-arm a source after a blocked operation re-arm
+    /// only `interest`, instead of everything the source is registered for.
+    /// Re-arming a direction the caller did not block on can raise readiness
+    /// nothing asked for: on Windows a blocked read would otherwise re-request
+    /// write readiness, which the AFD poll completes immediately for a writable
+    /// socket.
+    ///
+    /// Only use this when the direction is actually known. An operation driven
+    /// by a caller-supplied closure, such as `try_io`, must keep using
+    /// [`do_io`].
+    ///
+    /// [`do_io`]: IoSource::do_io
+    #[cfg(feature = "net")]
+    pub(crate) fn do_io_with<F, R>(&self, interest: Interest, f: F) -> io::Result<R>
+    where
+        F: FnOnce(&T) -> io::Result<R>,
+    {
+        #[cfg(windows)]
+        {
+            self.state.do_io_with(interest, f, &self.inner)
+        }
+
+        // No other selector re-arms in a way that can raise readiness for the
+        // direction the caller did not block on, so this is exactly `do_io`.
+        #[cfg(not(windows))]
+        {
+            let _ = interest;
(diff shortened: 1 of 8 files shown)


anuraaga:
Hello - I am also working on a runtime using mio for TCP and QUIC. I need the exact rearm fix here, but for QUIC also need to be able to apply it to `WSARecvMsg`. #1551 added `try_io` exactly for this, but when needing it with a single direction rearm, I end up needing i.e. `try_io_with`, this PR with the public API exposed for use, basically being symmetric to #1551. In practice, I guess this would supercede most if not all usage of `try_io`.

I have a patch prepared to this branch, https://github.com/ameyypawar/mio/compare/fix/1963-narrow-rearm...anuraaga:pr-2001-try-io-interest?expand=1. Would it make sense to send a PR to it to incorporate into this PR to have both the automatic usage and the public try_io API support directional rearm? Or should I followup after this one is merged? Or is the idea not acceptable? Let me know any thoughts.

Thanks for the help.

ameyypawar:
Thanks — the patch is clean and looks like the right shape.

My preference is to keep this PR to just the behaviour fix; it hasn't been reviewed yet and `try_io_with` is public API that deserves its own discussion. But that's the maintainers' call, not mine — so rather than decide it: is `try_io_with` wanted, and fold in or land separately?

The QUIC/`WSARecvMsg` case argues well for it.

PR #2002, opened 2026-08-11
Fix stale read/write/connect state leaking across named pipe reuse on Windows

### LINKED ISSUE #1983: Windows named pipe: stale read state leaks across DisconnectNamedPipe/ConnectNamedPipe instance reuse
## Summary

On Windows, reusing a named pipe instance through the documented
`disconnect()` + `connect()` cycle corrupts the next connection's byte stream:
the first read on the recycled instance returns stale completion state instead
of the new client's bytes. Frame-oriented protocols on top of tokio's
`NamedPipeServer` (length-prefixed JSON, etc.) break deterministically from the
second connection onwards.

The identical reuse sequence written against the raw Win32 API works
correctly, so this appears to be mio's I/O state machine rather than Windows
behavior.

## Root cause

In `src/sys/windows/named_pipe.rs`, `impl Read for &NamedPipe`:

```rust
// Looks like an in-flight read hit an error, return that here while
// we schedule a new one.
State::Err(e) => {
    Inner::schedule_read(&self.inner, &mut state, None);
    if e.raw_os_error() == Some(ERROR_BROKEN_PIPE as i32) {
        Ok(0)
    } else {
        Err(e)
    }
}
```

When a client disconnects, the in-flight overlapped read completes with
`ERROR_BROKEN_PIPE`. Reporting EOF (`Ok(0)`) is correct, but mio also
immediately schedules **another overlapped read on the same instance**.

The caller (tokio's `NamedPipeServer`) then runs the standard
`DisconnectNamedPipe` + `ConnectNamedPipe` reuse cycle. The stale read that
mio scheduled after EOF outlives the connection it belonged to:

- its completion bytes (or its `Ok(0)` EOF result) are handed to the **next**
  connection's first `read()`, and/or
- the pending overlapped `ReadFile` swallows bytes written by the **next**
  client before the caller's own read is issued.

Either way, the recycled connection's byte stream is shifted/corrupted.

## Minimal reproduction

```rust
use tokio::io::{AsyncReadExt, AsyncWriteExt};
use tokio::net::windows::named_pipe::{ClientOptions, PipeMode, ServerOptions};

const PIPE_NAME: &str = r"\\.\pipe\mio-reuse-repro";

#[tokio::main]
async fn main() -> std::io::Result<()> {
    let server_task = tokio::spawn(async move {
        let mut instance = ServerOptions::new()
            .pipe_mode(PipeMode::Byte)
            .create(PIPE_NAME)
            .expect("create pipe");
        let mut buf = vec![0u8; 64];
        for round in 0..2u32 {
            instance.connect().await.expect("connect");
            let n = instance.read(&mut buf).await.expect("read message");
            println!("round {round}: message read {n} bytes: {:?}",
                String::from_utf8_lossy(&buf[..n]));
            // Read again; completes with EOF once the client drops.
            let eof = instance.read(&mut buf).await.expect("read EOF");
            println!("round {round}: second read -> {eof} (EOF expected)");
            instance.disconnect().expect("disconnect");
        }
    });

    tokio::time::sleep(std::time::Duration::from_millis(200)).await;
    for round in 0..2u32 {
        let mut client = ClientOptions::new().open(PIPE_NAME).expect("client open");
        let msg = format!("hello-from-round-{round}");
        client.write_all(msg.as_bytes()).await.expect("client write");
        client.flush().await.expect("client flush");
        drop(client);
        tokio::time::sleep(std::time::Duration::from_millis(200)).await;
    }
    server_task.await.expect("server join");
    println!("done");
    Ok(())
}
```

Cargo.toml: `tokio = { version = "1", features = ["full"] }`

### Actual output (deterministic, every run)

```text
round 0: message read 18 bytes: "hello-from-round-0"
round 0: second read -> 0 (EOF expected)
round 1: message read 0 bytes: ""                  <-- should be the 18-byte message
round 1: second read -> 18 (EOF expected)          <-- the round-1 message arrives one read late
done
```

The round-1 connection's first `read()` returns the **stale EOF result** from
the previous connection, and the new client's bytes surface one read later,
shifted out of their frame.

### Expected output

```text
round 0: message read 18 bytes: "hello-from-round-0"
round 0: second read -> 0 (EOF expected)
round 1: message read 18 bytes: "hello-from-round-1"
round 1: second read -> 0 (EOF expected)
```

## Control experiment: raw Win32 API works

The same reuse sequence with synchronous Win32 calls — `CreateNamedPipeW`,
then per connection `ConnectNamedPipe` -> `ReadFile` (message) -> `ReadFile`
(EOF) -> `DisconnectNamedPipe`, with the client writing via `CreateFileW` +
`WriteFile` — produces the expected output for both rounds. So
`DisconnectNamedPipe` + `ConnectNamedPipe` instance reuse itself is sound;
the corruption comes from mio's post-EOF `schedule_read`.

## Real-world impact

Found in [Trwork](https://github.com/ulchemist/Trwork), a daemon that serves
length-prefixed JSON-RPC over a Windows named pipe and recycles instances in a
`connect -> serve -> disconnect` loop. The first 32 connections (one per
pre-created instance) work; every subsequent connection fails authentication
because the stale read shifts the framing and the server rejects the torn
frame. Long-running daemons exhaust all fresh instances and become
permanently unreachable until restarted.

## Environment

- mio 1.2.2 (via tokio 1.53.1 `NamedPipeServer`)
- Windows 11 10.0.26200 (x86_64-pc-windows-msvc, Rust 1.97.1)

## Possible fix directions

- On `disconnect()`, cancel in-flight reads (`CancelIoEx`) and drain queued
  completion state before calling `DisconnectNamedPipe`.
- Alternatively, skip the post-EOF `schedule_read` for named pipes (a
  disconnected instance cannot yield more bytes), or clear read state on
  `connect()`.

Happy to test any patch on the reproduction above.

Dantsz:
Fixes #1983.

Read and write completions now carry the generation they were issued under (inspired by the Solaris Event Port implementation), so results from an in-flight operation that outlived the connection are discarded instead of leaking to the next client. `disconnect` bumps that generation and cancels or clears the pending read/write state.

`connect` tracks its own generation so an aborted `ConnectNamedPipe` can no longer park a stale error for the next successful connection.

Regression tests added:
- `issue_1983` — two connect/read/disconnect rounds on one instance(from issue)
- `issue_1983_2` — read still pending at disconnect
- `issue_1983_3` — completed but unread data buffered at disconnect
- `issue_1983_4` — write in flight when the client goes away
- `issue_1983_5` — connect aborted by disconnect


GlenDC:
Hi @Dantsz thank you for this effort. I am going to review your PR this week. Is there anything that is not in the PR description or code that I should be aware of. E.g. concerns you have, things you encountered but didn't handle, things you are less certain of, questions you might have, how you consider the state of this PR (e.g. do you deem it "ready for merge")? Normally I would ask you over a coffee, but for now it will have to be in this cold github comment area :)

Dantsz:
Hi @GlenDC , I think it's ready, can't think of anything I'd change now.

I did try to find a solution based on the Windows API but couldn't come up with anything useful, I tried to limit the fix to the named pipes module.

Issue #1984, opened 2026-07-30
Enter Unreachable code in Windows named pipe


maxmicheluttiexelio:
# Reached Unreachable code in Windows Named Pipe

I am developing a cross-platform daemon to handle multiple TTYACM devices in parallel. 
In particular, I am leveraging [serial2_tokio](https://crates.io/crates/serial2-tokio) for serial communication and a mixture of `tokio::sync::mpsc::channel` and `CancellationToken` (from [tokio_util](https://crates.io/crates/tokio-util)) for communication between threads. 
While everything works on both Linux and MacOS, on Windows I enter unreachable code at line `969` of [mio/src/sys/windows/named_pipe.rs](https://github.com/tokio-rs/mio/blob/master/src/sys/windows/named_pipe.rs). 
This happens inconsistently whenever I unplug multiple devices at once (in my tests, the minimum amount that triggered the issue is 3), becoming more consistent as the amount of devices I unplug together increases. I am unsure if the crash happens due to a serial port communication issue or a inter-thread communication issue, but I think the former possibility is much more likely.
I have been trying to produce a toy example that results in the crash, but I have not been successful yet. I willl update this issue as soon as I am able to do it.

```
2026-07-30T10:26:58Z [ERROR] (panic) thread 'tokio-rt-worker' panicked at 'internal error: entered unreachable code': /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/mio-1.2.2/src/sys/windows/named_pipe.rs:969
   0: backtrace::capture::Backtrace::create
   1: backtrace::capture::Backtrace::new
   2: log_panics::Config::install_panic_hook::{{closure}}
   3: myApp::main::{{closure}}::{{closure}}
   4: std::panicking::panic_with_hook
   5: std::panicking::panic_handler::{closure#0}
   6: std::sys::backtrace::__rust_end_short_backtrace::<std::panicking::panic_handler::{closure#0}, !>
   7: __rustc::rust_begin_unwind
   8: core::panicking::panic_fmt
   9: core::panicking::panic
  10: mio::sys::windows::named_pipe::write_done
  11: mio::sys::windows::selector::SelectorInner::select2
  12: mio::poll::Poll::poll
  13: tokio::runtime::io::driver::Driver::turn
  14: tokio::runtime::time::Driver::park_internal
  15: tokio::runtime::scheduler::multi_thread::worker::Context::park_internal
[...]

PR #2004, opened 2026-08-19
Don't treat submitted Windows named pipe I/O as failed to submit

netaneld122:
Fixes the root cause behind #1819-style `unreachable!()` panics, and a use-after-free, in Windows named pipes.

## The bug

Every submitted overlapped operation reserves one strong reference of `Inner` (`mem::forget` in `schedule_read` / `maybe_schedule_write`), and the completion callback unconditionally redeems one (`Arc::from_raw` + drop). The schedulers treat `Err` from `read_overlapped` / `write_overlapped` as "failed synchronously, no completion packet coming" and skip the reservation.

But once `ReadFile`/`WriteFile` has returned `ERROR_IO_PENDING` a completion packet is guaranteed no matter how the operation ends - and the `GetOverlappedResult(bWait = FALSE)` peek that follows can still observe the operation's failure and return `Err`: one packet in flight, zero reservations. When that packet is dispatched:

* **read path** - `schedule_read` parked `State::Err`; `read_done` (since the #1819 fix) restores it and releases the reference - one release too many. `Inner` is freed while `NamedPipe` still points at it, and the next legitimate drop double-frees (`STATUS_HEAP_CORRUPTION` at `Arc<Inner>::drop_slow`).
* **write path** - the `Err` propagates out of `Write::write`, `io.write` stays `State::None`, and `write_done` hits `unreachable!()`.

The window is tiny but the outcome is guaranteed: per the [I/O Completion Ports docs](https://learn.microsoft.com/en-us/windows/win32/fileio/i-o-completion-ports), the status block "will not be updated until the packet is removed from the completion port", so the peek can only see the failure after the selector thread has already dequeued the packet - meaning the callback is already dispatched and waiting on the `io` mutex. Hitting it requires the peer to fail the operation right after submission (e.g. a disconnect completing a pending read with `ERROR_BROKEN_PIPE`) and the submitting thread to lose the CPU before the peek - preemption or suspend/resume. #1819 matches exactly: `read_done` found `State::Err` (os error 109), reproducibly around laptop sleep/wake.

## The fix

Once the operation has been submitted, never return `Err` - report it as in flight (`Ok(None)`) so the caller reserves the reference; the callback picks up the real result via `GetOverlappedResult` and the error is still delivered through readiness as `State::Err`. This is identical to the overwhelmingly common interleaving where the peek returns `ERROR_IO_INCOMPLETE`. Synchronous submission failures (no packet) still return `Err`.

## Reproduction

The dequeue must land between the submission and the peek, which cannot be forced from outside, so the second commit adds a `#[cfg(test)]`-only hook at exactly that seam, plus two regression tests: each submits a real read/write on a connected pipe, and the hook drops the client and dequeues the resulting error completion before the peek runs. The tests were written first and verified failing on the unfixed code; the fix commit is independently green, so the history bisects.

On Windows: `cargo test --all-features submitted_` - passes on this branch; cherry-pick the test commit alone onto master and both fail:

```
a submitted read was treated as failed-to-submit, leaving its completion packet without a reserved reference

a submitted write must be reported as accepted; its failure arrives via the completion: Os { code: 109, kind: BrokenPipe, ... }
```

<details>
<summary>Standalone demonstration of the deferred-status behavior</summary>

1. Create a named pipe server (`FILE_FLAG_OVERLAPPED`), associate it with a completion port, connect a client.
2. `ReadFile` on the server with no data buffered -> `FALSE`, `ERROR_IO_PENDING`.
3. Close the client handle -> the pending read completes with `ERROR_BROKEN_PIPE`.
4. `GetOverlappedResult(bWait = FALSE)` keeps returning `ERROR_IO_INCOMPLETE`, indefinitely - the failure is not yet visible.
5. `GetQueuedCompletionStatus` dequeues the packet (operation error 109).
6. The same `GetOverlappedResult(bWait = FALSE)` **now** returns `ERROR_BROKEN_PIPE`.

Steps 4 vs 6 show the deferral: the peek can only see the error after the packet has been dequeued - i.e. only after the selector has already dispatched the callback that will consume a reservation.
</details>

### DIFF
--- a/src/sys/windows/named_pipe.rs
+++ b/src/sys/windows/named_pipe.rs
@@ -218,6 +218,9 @@ impl Inner {
             }
         }
 
+        #[cfg(test)]
+        tests::run_submit_hook();
+
         let mut bytes = 0;
         let res = GetOverlappedResult(self.handle.raw(), overlapped, &mut bytes, 0);
         if res == 0 {
@@ -279,6 +282,9 @@ impl Inner {
                 return Err(err);
             }
         }
+
+        #[cfg(test)]
+        tests::run_submit_hook();
 
         let mut bytes = 0;
         let res = GetOverlappedResult(self.handle.raw(), overlapped, &mut bytes, 0);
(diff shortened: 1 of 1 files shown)


Timiz0r:
For what it's worth, this is actually a bug we started hitting very rarely (maybe 1 crash a day across tens of thousands of nodes), and this fix looks good to me! Though I'm not the biggest Windows IO expert.

Also taking a look at the other named pipe PRs and a couple other crashes we've been hitting! Since the Windows IO experts are, well, gone, it's the least I can do! (Side-note: it's at least being discussed internally, though I'm not too privy on the details)

ralish:
This appears to be the bug causing the Firefox crash documented in [Bug #2065991](https://bugzilla.mozilla.org/show_bug.cgi?id=2065991). I don't know the why, but in the right circumstances, it triggers very easily and repeatedly.

maxmicheluttiexelio:
I have tested this PR with a cargo patch, and I can confirm that this also solves issue [#1984](https://github.com/tokio-rs/mio/issues/1984)!

Timiz0r (approved):

src/sys/windows/named_pipe.rs
```diff
@@ -218,15 +218,20 @@ impl Inner {
...
         let mut bytes = 0;
         let res = GetOverlappedResult(self.handle.raw(), overlapped, &mut bytes, 0);
         if res == 0 {
-            let err = io::Error::last_os_error();
-            if err.raw_os_error() == Some(ERROR_IO_INCOMPLETE as i32) {
-                Ok(None)
-            } else {
-                Err(err)
-            }
+            // The operation was submitted (`ReadFile` returned success or
```
GlenDC: After ERROR_IO_PENDING, the GetOverlappedResult check can't change anything: whatever it returns, a packet is coming and we must return Ok(None), no?

If so, could we return perhaps Ok(None) directly in this branch, and only call GetOverlappedResult when ReadFile returned TRUE?

(See for example remarks in https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-readfile)

GlenDC: If we agree on that we could perhaps return early here, and also keep our doc comment focussed on the contract, something like:

```rust
let err = io::Error::last_os_error();
if err.raw_os_error() != Some(ERROR_IO_PENDING as i32) {
  return Err(err);
}

// Submitted: a completion packet will be queued
// so this must not return an error.
return Ok(None);
```

This _might_ also explain why #2011 needed a second thread?

src/sys/windows/named_pipe.rs
GlenDC: > "Otherwise if an error occurred it is returned"

This is as far as I can read no longer accurate. Please document the contract callers depend on, something like:

```
/// An error means the operation was not submitted
/// and no completion packet will be queued.
```

(if you agree, than this also has to be fixed in the docstring of `read_overlapped`)

src/sys/windows/named_pipe.rs
```diff
@@ -278,15 +283,16 @@ impl Inner {
...
         let mut bytes = 0;
         let res = GetOverlappedResult(self.handle.raw(), overlapped, &mut bytes, 0);
         if res == 0 {
-            let err = io::Error::last_os_error();
-            if err.raw_os_error() == Some(ERROR_IO_INCOMPLETE as i32) {
-                Ok(None)
-            } else {
-                Err(err)
-            }
+            // See `read_overlapped`: the operation was submitted, so its
```
GlenDC: same change as what I requested for the branch in `read_overlapped`

src/sys/windows/named_pipe.rs
```diff
@@ -218,15 +218,20 @@ impl Inner {
             }
```
GlenDC: After ERROR_IO_PENDING, the GetOverlappedResult check can't change anything: whatever it returns, a packet is coming and we must return Ok(None), no?

If so, could we return perhaps Ok(None) directly in this branch, and only call GetOverlappedResult when ReadFile returned TRUE?

(See for example remarks in https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-readfile)

GlenDC:
```suggestion
            }
            
            // Submitted: a completion packet will be queued
            // so this must not return an error.
            return Ok(None);
```

GlenDC: This might also explain why #2011 needed a second thread?

src/sys/windows/named_pipe.rs
GlenDC: this comment can be removed fully I think

src/sys/windows/named_pipe.rs
```diff
@@ -218,15 +218,20 @@ impl Inner {
...
         }
 
+        #[cfg(test)]
+        tests::run_submit_hook();
```
GlenDC: these 2 lines would move above the `return Ok(None)` asked earlier, if we agree with that change

src/sys/windows/named_pipe.rs
```diff
@@ -1109,3 +1115,197 @@ impl BufferPool {
...
+    //! Every submitted overlapped operation must reserve the strong reference
+    //! that its completion callback unconditionally consumes - even when the
+    //! operation fails while it is being submitted.
+
+    use super::*;
+    use std::cell::RefCell;
+    use std::fs::{File, OpenOptions};
+    use std::time::Duration;
+
+    // A buggy completion callback can over-release at most one strong
```
GlenDC: Each test has exactly one packet in flight, so a single padding reference is enough to turn a regression into an assertion failure instead of a use-after-free. Could DISPATCH_BATCH padding be 1, and while we are at it have the comment be reduced to only stating that reason?

src/sys/windows/named_pipe.rs
```diff
@@ -1109,3 +1115,197 @@ impl BufferPool {
...
+            hook();
+        }
+    }
+
+    /// A server pipe with a connected client, registered to a completion port
+    /// that the test drains manually, playing the selector's role.
+    struct TestPipe {
+        pipe: NamedPipe,
+        cp: Arc<CompletionPort>,
+        client: RefCell<Option<File>>,
```
GlenDC: `client: RefCell<Option<File>>` is only needed because the methods take `&self`. With an exclusive reference it can be a plain `Option<File>`.

At that point the one-line helpers `read_is_pending`, `read_failed` and `write_failed` could be inlined as `matches!` at the call sites.

src/sys/windows/named_pipe.rs
```diff
@@ -1109,3 +1115,197 @@ impl BufferPool {
...
+        }
+
+        fn write_failed(&self) -> bool {
+            matches!(self.pipe.inner.io.lock().unwrap().write, State::Err(_))
+        }
+
+        /// Dispatches queued packets the way `Selector::feed_events` does.
+        fn deliver_completions(&self) -> usize {
+            let mut statuses = [CompletionStatus::zero(); DISPATCH_BATCH];
+            let mut events = Vec::new();
```
GlenDC: No too important for your PR, but it might allow me, if my plans go through, to preserve this test for a future rewrite...

`deliver_completions` collects the events and drops them. Could the tests instead assert that a readable (or writable) event for `Token(1)` was produced? That's what users actually see. It would also let these tests outlive the State enum they check today.

src/sys/windows/named_pipe.rs
```diff
@@ -1109,3 +1115,197 @@ impl BufferPool {
...
+        // surfaces the write's failure.
+        assert_eq!(pipe.deliver_completions(), 1);
+        assert_eq!(
+            pipe.reserved_refs(),
+            0,
+            "the completion must redeem exactly the reserved reference"
+        );
+        assert!(pipe.write_failed(), "the write's failure must still be delivered");
+    }
+}
```
GlenDC: Could you add an additional test for the opposite case? A failure before submission must reserve nothing.

GlenDC: Additionally I would also add a test for #2011.

It looks like this same bug, reached through Tokio: the peer disconnects right after a read is submitted, and the I/O driver thread (parked in poll) dequeues the completion before the peek. Your hook should be enough to force that deterministically, but with:

* a real `Poll` running on another thread, instead of draining the port by hand
* the read going through the public `Read` path

The test should fail on master and pass with this fix: the reference count ends back at baseline, and the disconnect still surfaces as EOF. With that in place we can actually say this PR fixes #2011, instead of only relying on patched-crate reports.

GlenDC (changes requested):
Thanks for getting this started! Your analysis checks out. Traced it by hand (must say I was a bit rusty 😓 so took me a while :().

Fits #1984, #1819, Firefox report and likely #2011.

Do you have time this week to get this PR wrapped up @netaneld122 , or are you ok with me finishing of your PR by merging your branch into a branch I would create myself with a follow-up commit (preserving your contribution in the form of your original commit). Or how do you want to proceed? Asking as there might be some more things I discover as I work through it myself, but I neither want to "take it away from you" as you clearly put work in here and might want to finish it yourself which I would respect and in that case I will await your work to review it again afterwards.

PR #2008, opened 2026-09-06
Fix unix socket accept for named peers on Darwin

bquorning:
AI disclaimer: This PR (code _and_ description) was written by AI (Claude Code), on my request. I ran into failing tests from some Ruby/Rails code that ships logs to the [Vector](https://github.com/vectordotdev/vector) agent → Tokio → Mio – Vector logged `Failed to accept socket. error=paths must not contain interior null bytes`. Claude traced the failure and suggested this fix.

## Summary

- On Darwin, `accept(2)` reports the length of the entire `sockaddr_un` structure, so a named peer's path in `sun_path` is followed by null padding up to the full length.
- `uds::listener::accept` removed at most one trailing null byte (sufficient on Linux, where the reported length covers only the path and its null terminator), leaving interior null bytes that made `SocketAddr::from_pathname` fail with `paths must not contain interior null bytes` — failing `accept()` itself for any peer bound to a named path.
- Truncate the path at the first null byte instead, which handles both layouts.

## Impact

Any client that binds its unix socket to a named path before connecting has its connections rejected on macOS. We hit this with [Vector](https://vector.dev): its unix socket sources report the peer's bound path as the event's `host`, so applications logging to Vector bind their client sockets to named paths to tell connections apart — and on macOS every such connection is dropped with `Failed to accept socket. error=paths must not contain interior null bytes`. Reproducible with a plain `tokio::net::UnixListener` (tokio delegates to mio's `accept`), so this affects all mio-based unix socket servers on Darwin.

## Test

- Added `unix_listener_accept_named_peer`, which connects from a socket bound to a named path. Neither std nor mio expose bind-before-connect for `UnixStream` (the same limitation noted in #1817), so the test uses `libc` directly, added as a unix-only dev-dependency.
- Verified the test fails on master on macOS 15 and passes with this fix; it passes on Linux with and without the fix.
- Full test suite (`--features os-poll,net`), `cargo fmt --check`, and `clippy` pass.

### DIFF
--- a/tests/unix_listener.rs
+++ b/tests/unix_listener.rs
@@ -2,6 +2,7 @@
 
 use mio::net::UnixListener;
 use mio::{Interest, Token};
+use socket2::{Domain, SockAddr, Socket, Type};
 use std::io::{self, Read};
 use std::os::unix::net;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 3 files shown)


tests/unix_listener.rs
```diff
@@ -76,6 +76,109 @@ fn unix_listener_local_addr() {
...
 
+/// Connect to the unix socket at `connect_path` from a socket that is itself
+/// bound to `bind_path`. Mio (and std) don't expose bind-before-connect for
+/// `UnixStream`, so this uses libc directly.
+#[cfg(unix)]
+fn connect_from_bound_socket(bind_path: &Path, connect_path: &Path) -> net::UnixStream {
+    use std::os::unix::ffi::OsStrExt;
+    use std::os::unix::io::FromRawFd;
+
+    fn pack_sockaddr_un(path: &Path) -> libc::sockaddr_un {
```
Thomasdezeeuw: Let's do this without using libc

bquorning: I changed the implementation to use `socket2` instead. If you like like this approach, I’ll be happy to squash the commits.

